PR #173, opened 2015-01-07
Updated arch-setup to accomidate change in makepkg, and removed spacenavd

jgkamat:
Also moved gcc arm libs from yaourt to arch repositories via pacman.

This fully removes spacenavd and comments out yaourt, as they cause build failures right now.

### DIFF
--- a/util/arch-packages-aur.txt
+++ b/util/arch-packages-aur.txt
@@ -1,8 +1,5 @@
 # This is a list of packages that we need for Arch Linux that
 # we'll install from the AUR using yaourt
 
-# gcc xcompile toolchain for mbed
-gcc-arm-none-eabi-bin
-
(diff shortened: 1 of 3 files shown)


barulicm:
Could we make spacenav into an optional item during build setup? Many projects (like opencv) have cmake options that let you specify things like --no-[module name here] or --use-[module name here].
We could add a --USE-SPACENAV for people who want to include it, and leave it default off for everyone else who won't need/want it.

justbuchanan:
Yeah we could do that, although it might get sort of messy with #defines in several places.  Is it causing issues for you on Ubuntu or did you just not want the package?

I was thinking about removing the `spacenavd` dependency since that's the driver, but without a lot of #defines, you'll still have to install `libspnav`, which is the client library for the driver.

PR #189, opened 2015-01-19
merge console for robot2015

guyfleeman:
The initial work for console over serial for mbed is done.


justbuchanan:
Good work on this!  Having a console up and running is a solid step for getting our firmware moving along.

A few suggestions after going through it briefly:
* In commands.cpp, you've got a `NUM_COMMANDS` variable to track how many command entries we have registered.  I think it would be better to remove this, then make our registered list of commands a std::vector, so you can just query it for its length.  This makes it easier to add a new command since it just has to be added to the list and you don't have to worry about updating the count.
* Similar to the above comment, it would probably be better to pass arguments to commands in the form of a std::vector<std::string> rather than having a count and an array.
* This is just a minor thing, but in your headers, you're using include guards, which is good, but the newer way to do this is to just put "#pragma once" at the top.  They essentially do the same thing, but the second way is a bit cleaner and slightly faster since it allows the compiler to ignore the rest of the file rather than having to scan down to the closing "#endif"

Overall looks really good.  If you get these things changed, I think it'll be good to merge.  And we can review at tomorrow night's meeting.

justbuchanan:
Cool, thanks for getting those changes made.  The only other improvement I can think of would be to pass the arguments to the command functions by reference rather than by copy as they are now.  I'm not sure how familiar you are with C++ references, but they're essentially pointers, but with syntax as if you were just passing a regular copy.

In commands.hpp, you can just change
`void (*handler)(std::vector<std::string> args);`
to
`void (*handler)(const std::vector<std::string> &args);`

Then change the functions themselves to match.  This will just reduce the overhead a bit since it will eliminate the copying of string vectors that it's doing currently.

PR #194, opened 2015-01-25
Albertrefactor path

### LINKED ISSUE #167: Generalize the Path class and add some specialized subclasses
In order to support more types of paths, we should make the [`Path`](https://github.com/RoboJackets/robocup-software/blob/master/soccer/planning/Path.hpp) class a fairly simple interface that the motion controller will rely on, then add some subclasses for particular path types.

I think the only methods that `Path` needs are:
* `bool evaluate(float t, const Point2d &posOut, const Point2d &velOut) const`
* `float duration() const`
* `bool hit(const Geometry2d::CompositeShape &shape) const` - detect collisions with obstacles

Specific subclasses could include:
* `BezierPath` - this will handle the functionality that the current `Path` class contains
* `CirclePath` - makes the robot go in a circle at a specified velocity
* `PiecewisePath` - a path created by stitching together subpaths
* `LatencyTestPath` - a path used to measure the latency in the system, as described in #166.

ashaw596:
Refactor Path into an abstract super class
Resolves #167


justbuchanan:
Looks good overall.  The main thing I'd change is how the BezierPath is implemented.  Right now it looks like the path planner interpolates the bezier curves over a fixed dt and stores the results in a vector.  It'd be more compact to have the BezierPath just store the control points, then interpolate when values are needed.

justbuchanan:
I'm working with bezier paths though with the rrt stuff though, so I don't mind making this change at some point.

ashaw596:
Yeah, that would't be a difficult change. I didn't do it before since I was trying to make it fit into the old Path class when I was making it. Its not really a BezierPath really. It just a vector path I guess or a velocity, points vectors path.

justbuchanan:
Yeah that's what I figured.  Want to just rename it to InterpolatedPath or something, then I'll make an actual BezierPath when I get the new path planning stuff done.

ashaw596:
Probably a good idea. I'm worst with naming variables and files :).

justbuchanan:
Haha no worries.  I think InterpolatedPath works as a name for this unless you've got something better.  Then we can use BezierPath for one that's actually represented by control points.

justbuchanan:
Also, since the path can come from the path-planner or a MotionConstraints object and gets passed around, it'd probably be good to use std::shared_ptr to hold a reference to it to make sure we don't have any memory management issues.

soccer/planning/InterpolatedPath.hpp
```diff
@@ -0,0 +1,147 @@
...
+
+			// Returns a new path starting from a given point
+			void startFrom(const Geometry2d::Point& pt, Planning::InterpolatedPath& result) const;
+
+			//Returns the destination of this path (the last point in the points array)
+			virtual boost::optional<Geometry2d::Point> destination() const;
+
+			// Returns true if the path never touches an obstacle or additionally, when exitObstacles is true, if the path
+			// starts out in an obstacle but leaves and never re-enters any obstacle.
+			virtual bool hit(const Geometry2d::CompositeShape &shape, unsigned int start = 0) const;
```
justbuchanan: The naming seems a little off here... shouldn't it return true if it finds a collision and false if it's collision-free?

ashaw596: Yeah. That's not right.... looks like it was written in 2009 lol. I'll fix
it. Also TODO:  We should also change collision detection to not use
exceptions at some point.

ashaw596:
Ok. Fixed that documentation.

soccer/MotionConstraints.cpp
```diff
@@ -6,16 +6,19 @@ REGISTER_CONFIGURABLE(MotionConstraints);
...
 ConfigDouble *MotionConstraints::_max_speed;
-ConfigDouble * MotionConstraints::_max_angle_speed;
+ConfigDouble *MotionConstraints::_max_angle_speed;
+ConfigDouble *MotionConstraints::_max_replan_distance;
 
 void MotionConstraints::createConfiguration(Configuration *cfg) {
     _max_acceleration   = new ConfigDouble(cfg, "MotionConstraints/Max Acceleration", .5);
     _max_speed          = new ConfigDouble(cfg, "MotionConstraints/Max Velocity", 2.0);
     _max_angle_speed      = new ConfigDouble(cfg, "MotionConstraints/Max Angle Speed", 80); //  degrees / second
+    _max_replan_distance      = new ConfigDouble(cfg, "MotionConstraints/Max Replan Distance", 0.6); //  degrees / second
```
justbuchanan: Could this be better named min_replan_distance or replan_threshold?

soccer/MotionConstraints.hpp
```diff
@@ -66,6 +66,7 @@ struct MotionConstraints {
 
     /// The speed we should be going when we reach the end of the path
     float endSpeed = 0;
+    float maxReplanDistance;
```
justbuchanan: Is it safe to delete this variable?

soccer/Robot.cpp
```diff
@@ -518,26 +503,14 @@ std::shared_ptr<Geometry2d::Shape> OurRobot::createBallObstacle() const {
 
 #pragma mark Motion
 
-void OurRobot::setPath(Planning::Path path) {
+void OurRobot::setPath(Planning::Path *path) {
 	_didSetPathThisIteration = true;
 
+	delete _path;
```
justbuchanan: We should check first if _pat is != nullptr.

soccer/planning/RRTPlanner.hpp
```diff
@@ -38,63 +38,62 @@ namespace Planning
...
+
 		/** best goal point */
 		Geometry2d::Point _bestGoal;
-		
+
 		///best planned path
 		///this is a fixed step path
-		Planning::Path _bestPath;
-		
+		Planning::InterpolatedPath _bestPath;
```
justbuchanan: I like that you've changed the Planner's run() method to return a new path rather than changing an existing one in place.  Is there much use in keeping a copy of the path in `_bestPath` here?

PR #199, opened 2015-02-03
Add ability to graph two data functions on one graph.

ashaw596:


justbuchanan:
Looks good, I like it!

One thought I had on the UI is that the name "Chart Menu" is a little ambiguous.  How tough would it be to change it from having:
* Chart
* Chart Menu
    * A
    * B

to something like:
* Chart
    * New Chart
    * ------------------
    * Add to "A"
    * Add to "B"

ashaw596:
Ok should be good to go.

PR #200, opened 2015-02-04
Merge build and radio updates into master.

guyfleeman:


util/robot2015-prog.sh
```diff
@@ -54,9 +54,9 @@ done
 # restart mbed code
 MBED_SERIAL_PATH="$(ls /dev/ | grep ttyACM | sed 's\.*\/dev/&\g')"
 
-for i in $MBED_SERIAL_PATH; do
-    echo Attempting reboot on $i ...
-    sudo sh -c "echo -ne 'reboot\r' > $i"
-done
+#for i in $MBED_SERIAL_PATH; do
```
justbuchanan: I see this got commented out, did it not work when you ran it?

justbuchanan:
@jjones646, what's the plan for the `SerialLink` and `USBLink` classes in here?  I remember you mentioning them before, but I wasn't sure exactly where they fit in and if we plan on using them.

firmware/robot2015/cpu/utils/FirmwareCheck/FirmwareCheck.hpp
```diff
@@ -0,0 +1,9 @@
+#ifndef FIRMWARE_CHECK_H
+#define FIRMWARE_CHECK_H
+
+#include "mbed.h"
+#include <string>
+
+extern void firmware_version(std::string&);
```
justbuchanan: Can we add a quick comment on here just to clarify that this checks the MBED's hardcore builtin firmware and not the version of the RoboCup firmware?  I think at some point, we'll bake the git repo's SHA checksum into the our firmware, which will be how we track our own versioning.

guyfleeman:
@justbuchanan the immediate write to serial caused a race condition in some old code I was testing, the issue is no longer present to my knowledge. I forgot to uncomment the lines. Sorry about that.

justbuchanan:
No worries, sounds good to me.  :+1:

jjones646:
@justbuchanan, the original intent was to have base station's usb connection using the same protocol type that the radio uses. That way, the base station code would be reduced as it would only require linking the radio and usb interface's `send` methods to the other's 'receive' callback. Then pairing this up with the ability for soccer to control a robot connected via usb/serial without writing new code (making a robot's connection type independent to how it handles the information).

Saying all that, there's more important things to worry about, and it's probably best to scrap that idea for now. I think there should still be a way to physically connect a robot to a computer and test the motors/solenoid in the form of actually operating them however.

justbuchanan:
Ok gotcha.  Well as long as you have those files on your MBED repo, let's remove them from this PR, then we can bring them back in later if we decide to use them.

I think we'll end up having console commands that we can use to drive the wheels and such over USB.

PR #202, opened 2015-02-04
Changed code blocks to non-numbered ones so you can copy paste them

jgkamat:
For @ashaw596


jgkamat:
I tried out .sh, but I didn't see any difference. At any rate, it keeps the line numbers.

PR #206, opened 2015-02-05
Fix Ubuntu Ninja Package Name

ashaw596:


ashaw596:
Changed the package name to ninja-build

Issue #142, opened 2014-11-05
Rotate radio protocol velocities


justbuchanan:
Our radio protocol cmd_x and cmd_y velocities are backwards from what would seem to make sense...

See here for info: https://github.com/RoboJackets/robocup-software/blob/master/doc/radio-protocol-2011.txt#L23

When we write the new firmware for the MBED-based control board, we'll have two sets of firmware that are completely separate, but need to speak the same radio protocol.  Now would be a good time to make sure our protocol makes more sense.

ashaw596:
Is this moot with the new firmware and stuff? Or do we still want to d othis for backwards compatibility?

justbuchanan:
Yeah we can close this.

PR #207, opened 2015-02-06
Really pretty Graphs

ashaw596:
Add body velocity to Robot protobuff so we can compare it with the command velocity.

Yay Pretty Graphs showing how bad our motion control is. (Even if it is just simulated)
Image of Screenshot

### DIFF
--- a/common/protobuf/LogFrame.proto
+++ b/common/protobuf/LogFrame.proto
@@ -87,7 +87,7 @@ message LogFrame
 	message Robot
 	{
 		required Point pos = 1;
-		required Point vel = 7;
+		required Point world_vel = 7;
 		optional Point body_vel= 16;
 		required int32 shell = 2;
 		required float angle = 3;
--- a/common/protobuf/LogFrame.proto
+++ b/common/protobuf/LogFrame.proto
@@ -88,6 +88,7 @@ message LogFrame
 	{
 		required Point pos = 1;
 		required Point vel = 7;
+		optional Point body_vel= 16;
 		required int32 shell = 2;
 		required float angle = 3;
 		optional BallSenseStatus ball_sense_status = 4;
--- a/soccer/Processor.cpp
+++ b/soccer/Processor.cpp
@@ -449,7 +449,7 @@ void Processor::run()
 			_state.logFrame->add_debug_layers(str.toStdString());
 		}
 		
-		// Our robots
+		// Add our robots data to the LogFram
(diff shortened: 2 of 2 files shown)


justbuchanan:
Cool, I like the idea of being able to graph command velocity and actual velocity side-by-side!

I think it's a little confusing though to have both `body_vel` and `vel` as properties in the tree since they're the same thing, just in different coordinate frames.  I think it would be a little cleaner if we just changed the command velocities to be sent in the same reference frame that soccer uses (issue #142).  Thoughts?

ashaw596:
I don't think its *that* confusing, but hmmm.
Sending vel would be in line with that discussion we had about telling the robot what rotation we want it to be instead of just sending the w_velocities we want. We would be able to better utilize the gyroscope for compensating for turning while moving. However, doing that much sin and cos stuff on a processor without a floating point unit. idk. Depends on how fast we can do those computations. It would probably work and would give us a bit more accuracy. We'd still have to rework the protocals as we said before to support sending the current robot rotation and the rotation we want it to be in.

ashaw596:
About the tree, I would still like to have both so we can see if our robots have weird x vs y vs diagonal movement properties. I think its very useful for debugging. I'd also like to add a seperate "target velocity" and "target position" as apart from the velocity we send to the robot so that we can test putting that into a control loop. I think if we change the vel to word_vel that would clear up any confusion.

justbuchanan:
Cool, I think renaming vel to world_vel resolves any confusion.

And you're right, I misspoke about the reference frames for radio protocol velocities.  I still think we should rotate them so if the robot is going forward down the field, it's body and world velocities match up.  Sent velocities would still be in the frame of the robot, but that frame would be rotated 90degrees from what it is currently.  That's a separate issue though, so we can leave that for later.

ashaw596:
I do think that sending it in the global reference frame would be interesting in the future tho. Depending on how good the gyroscope and sensors turn out to be.

justbuchanan:
I agree.  I think if we could tell the robot where it is and give it the current section of the path that it's on, we could get tighter controls since they'd be on-board.  It's  probably not a job for this year, but definitely something to look into.

On another note, right now the protobuf tree for RadioTx has a vel_x and a vel_y, while the protobuf for command velocity is represented as a Point2d.  They're essentially the same thing, but would making the RadioTx encapsulate the command velocity in a point make it any easier for graphing?

PR #208, opened 2015-02-11
Add Velocity Controller and Fix I term in PID code

ashaw596:


justbuchanan:
Were you able to find the article/paper on doing a position and velocity PID on top of each other?  I'm a little sketchy on this double PID thing...

PhilipBale:
You should probably study for 2200

justbuchanan:
Bro, I've got 2200 on lock down :godmode:

common/Pid.cpp
```diff
@@ -35,16 +35,19 @@ Pid::~Pid()
...
 {
-	if (w > 0)
+	if (w != _windup) 
 	{
-		if (w != _windup) {
-			_windup = w;
-			_oldErr = new float[_windup];
-			memset(_oldErr, 0, sizeof(float)*_windup);
+		if (_oldErr) 
+			delete[] _oldErr;
```
justbuchanan: dangling pointer

justbuchanan:
@ashaw596, do you want to refactor this so that it's just the PID fixes?  I like those changes, but don't think the velocity PID thing is of much use, especially since that will be wiped out by the new motion control stuff.

Issue #231, opened 2015-02-22
Sometimes MBED doesn't start after 'make robot2015-prog'


guyfleeman:
Sometimes after a prog command the mbed does not start properly, led's 1 and 2 do not flicker. I suspect this has to do with setting interrupt priorities.

jgkamat:
I can confirm this, it seems to be much more prevalent on my Debian system than my arch system. I don't know why though...

This is the command the script runs to restart the mbed, it only works when there is firmware on the machine already.

```
sudo bash -c "echo -ne '\rreboot\r' > /dev/mbed"

```

To get a better idea of the problem, run screen on the mbed while running this command.

guyfleeman:
@jgkamat do you remember any specific conditions that could have led to the failures, specifically usb ops that could generate interrupts? I've tried adding delay between the firmware copy and the reboot command using ```sleep 1.0``` which seemed to improve the situation but didn't fix it all together.

jgkamat:
I remember that when the issue happened, if I screened into the mbed, (or cat /dev/mbed), I would get a ton of garbage output. It may have something to do with buffering?

guyfleeman:
@jgkamat Ah, now that you mention it that makes a lot of sense. It seems that the stdio buffer isn't always flushed prior to the to reboot. I wonder if this could cause complications with the copy? All USB traffic is routed through stdio I think.

When I screened into it it just failed? Did you get the garbage on arch? Maybe I should do my initial debugging there.

jgkamat:
@guyfleeman I got the garbage on debian, let me see if I can reproduce it now...

jgkamat:
[Here is the output of the garbage](http://pastebin.com/qxziHUDZ)... I printed out an ls before, for reference. I can get it to reproduce quite reliably actually (ive been working on my arch machine much more often)... Do you want me to bring my Debian machine over tomorrow, so you can take a look?

guyfleeman:
@jgkamat yeah whatever you think will help us find a solution. Maybe we can go through this tomorrow? I'm aware of the problem occurring when the mbed is plugged in, the OS queries some information and if you screen in too soon the reply is garbled and the OS can't communicate with it properly.

jgkamat:
@guyfleeman thats actually probably it! On my arch system, that is minimal (as there's no auto-mounting and such), but my debian system is running KDE, which does a ton of background work.

jgkamat:
[...]

PR #233, opened 2015-02-25
Mbed travis

jgkamat:
Made the firmware compile for our travis builds! Hopefully this pull request will pass...

I also added a temp fix for #231

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ before_install:
   - export CC=gcc-4.7
   - export CXX=g++-4.7
 
-script: make && make test-cpp
+script: make && make test-cpp && make robot2015
 
 notifications:
   email: false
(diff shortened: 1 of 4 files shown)


jgkamat:
Unrelated, but i found [this page](http://docs.travis-ci.com/user/caching/) about travis ci caching, I'm trying to make it work with ccache now.

justbuchanan:
Cool, let me know what you find out about caching - it'd definitely be nice to have faster CI builds.  The article you linked to said the caching features are only available to private users and their container-based builds, which don't allow sudo, so I don't know if it'll work for us.

jgkamat:
So another article [I found](http://blog.travis-ci.com/2014-12-17-faster-builds-with-container-based-infrastructure/) says that they currently reccomend uploading precompiled binaries instead of using apt on container based systems, which is probably impractical for our purposes... They are looking into eventually providing sudo support in containers, and enabling apt caching by default on all builds though..

PR #236, opened 2015-03-01
Adds Basic IO Expander Support

jpfeltracco:
Adds the ability to toggle individual pins from the console, test all pins of the expander, and refer to pins using the same basic setup as normal MBED pins by doing this:

```c++
IOExpanderDigitalOut myPin(expander, IOExpanderPinA0);
myPin = 1; // Sets myPin to ON
```

This relies on the MCP23017 library found [here](http://developer.mbed.org/users/Nurbol/code/MCP23017/shortlog) which hasn't been added in this pull request, it will need to be added before this is pulled.


justbuchanan:
@guyfleeman, what's the best way to add the mbed library for this to CMake?

guyfleeman:
@justbuchanan @jpfeltracco it's already there. Under firmware/robot2015/cpu open arm_toolchain.cmake and change set(BUILD_MCP23017 false) to true.

jpfeltracco:
Changed it to true and it's finding the #include "MCP23017.h", but when compiling it produces a bunch of undefined reference errors. Any ideas @guyfleeman ?

guyfleeman:
@jpfeltracco are the reference errors internal or external to the MCP23017.h file?

justbuchanan:
The error message I saw was a linker error, so it was finding the header just fine, it just wasn't building and linking the MCP23017 c++ file.  I think we need to change the BUILD command for the ExternalProject from empty to something else and then make sure the lib gets linked against the robot2015 binary.

guyfleeman:
@justbuchanan @jpfeltracco I've been trying to get the driver to compile externally for a few days and I'm running into the same problem I had with the mbed library, but since the driver doesn't have a make file (or a python build system) I'm struggling to get it to compile.

That being said, I have a fix ready to be merged into @jpfeltracco  master that will download the driver to a folder in the drivers folder, a lot like we are doing with the CC1101. I've also modified the gitignore appropriately. If this sounds okay, let me know and I'll open the pull request with @jpfeltracco. If not, I'll keep digging for solutions.

Still feeling under the weather, didn't want to get anyone else sick.

guyfleeman:
@justbuchanan I've submitted a pull request to @jpfeltracco with the fix. Once he merges, or has me do additional fixing, it should be good to go into master.

justbuchanan:
Sounds good.  How does the mbed python build suite work?  I thought it was setup so you could give it a path to the source for a library and it could build it.  If that's the case, I think we can just make a list of libraries from the mbed site that we want and we can loop through to build and link them all.

justbuchanan:
It'd be best to clone all library dependencies into somewhere under the `build` directory if possible.

guyfleeman:
@justbuchanan Last I checked the mbed python script only supported official repos, but I'll look into the integration.

guyfleeman:
@justbuchanan everything's in build now. The linking might be a bit weird, but it works well.

justbuchanan:
@jpfeltracco, any luck getting this to work with @guyfleeman's changes?

PR #237, opened 2015-03-01
Adds qt5-svg package to arch setup packages.

jpfeltracco:
This fixes icons not showing up.

### DIFF
--- a/util/arch-packages.txt
+++ b/util/arch-packages.txt
@@ -18,6 +18,7 @@ libmpc
 
 qt5-base
 qt5-tools
+qt5-svg
 qwtplot3d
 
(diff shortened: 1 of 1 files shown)


justbuchanan:
@jgkamat, this may be the same issue you were having with SVGs not working in your docker image.  Maybe adding this package to ubuntu-setup would fix it.

jgkamat:
The docker image was missing some packages! I found them, and added them to the ubuntu setup. In a bit i'll submit a pull request which will add that and add some cool arguments to the ubuntu setup, as well as fixing the ppa issues we had with the mbed compiler.

PR #242, opened 2015-03-05
Robot status ui

justbuchanan:
I took the status widget that @charles2000 put together, added and tweaked a few things, and made it display in the soccer GUI with info constantly updated.

A few things todo before this is ready to merge:
- [x] only show robots that we have radio comm with - otherwise we really have no status to show other than that radio doesn't work.
- [x] Add UI for showing dribbler fault
- [x] Add UI for showing kicker fault
- [x] We need to change the way we store robot references once we start showing non-visible robots that are identified by Board ID.  We also need a way to connect this with robots we see through vision.
- [x] remove the status text in FieldView that this replaces.
- [x] get a locked copy of or reference to the RadioRx packet so it doesn't get changed by another thread while we're reading it in the GUI thread.  This causes crashes in the current setup.
- [x] calculate battery charge percent from voltage
- [x] add battery profile for 2015 robot (see #245)
- [x] re-examine battery profile for 2008 battery.  According to Robot.cpp:161, 14.3v is considered low.

Attached is a screenshot.

screen shot 2015-03-04 at 8 18 48 pm


justbuchanan:
I just removed the old status text that was displayed next to each robot on the field view since this info is now shown in the status widget.  The status widget doesn't currently show all of it though, so see commit 0c43828 for reference when it's time to add it all into the status widget.

PR #253, opened 2015-03-11
Switch reboot command on firmware to breaks, and check sha hashes

jgkamat:
I switched the reboot command being sent to the mbed to a break command being sent by python. 

In addition, I also added a check to see if the SHA hashes of the copied files are the same, as we had problems in the past with that. 

I'm pretty sure I added all the dependencies for this, but let me know if anything was missed.

This is another attempted solution of #231.

Also I added a nofirmware option to ubuntu-setup


justbuchanan:
Nice work!

PR #258, opened 2015-03-25
Redesigned Interface

monamohebbi:
Redesigned user interface to combine plays and behavior tabs, while also creating a new space so referee tab and robot status area can be seen at all times


justbuchanan:
I branched off of your branch @monamohebbi, and created a `ui-redesign` branch.  I fixed most of the layout issues - there were a few places where panels were put inside a QWidget, which I rearranged and put inside of Layout objects.  The layout objects expand their contents to fit the available size.

A few things left to do:
- [ ] Some of the panels start out at awkward sizes - it'd be nice if when soccer launched, everything was sized appropriately and stretched nicely based on the window.  This is supported with the layouts, we mostly just need to set the MinimumWidth and MinimumHeight for certain things.
- [ ] It would be nice to have some placeholder text for things that start out empty such as the BehaviorTree and the RobotStatusList.  Right now at launch both show up as blank white rectangles, which is sort of confusing.  If we could have it display a default bit of text describing it is when it's empty, it would clear things up a lot.
- [ ] A lot of things are in horizontal and vertical splitter layouts, meaning that you can click and drag the  dividing bar between the contained widgets to resize.  The splitter "drag zones" are fairly invisible though, so it's hard to tell where to click.  We can fix this by using stylesheets to set a background image for the draggable zone.

Overall looks great Mona, nice work!  It'll be great having all this info visible at the start rather than having to dig around in tabs for it like we've had before.

PR #260, opened 2015-03-25
Playbook feature

ra4king:
- Created playbook.py that loads/saves playbook files
- Added load_playbook and get_enabled_plays_path to play_registry.py
- Added load/save playbook functions to main.py that are called from same function names in GameplayModule.cpp
- Added command line option to main.cpp for loading playbooks that calls GameplayModule function.


soccer/gameplay/play_registry.py
```diff
@@ -216,12 +244,15 @@ def parent(self):
         @parent.setter
         def parent(self, value):
             self._parent = value
-        
+
+        @property
+        def full_play_path(self):
+            path = []
```
justbuchanan: Yo, this method might need some more content.

justbuchanan:
Looks good!  I added a couple comments and fixed a couple minor things, which I'll push up to the `roi-playbooks` branch in a sec.  Go ahead and merge that branch into yours and implement the method I pointed out above and it'll be good to go!

One other thing is that right now the path passed to the `-pbk` flag is relative to the soccer/gameplay directory.  This does make things a bit easier, but it might be slightly confusing since other paths, such as for config files, are passed relative to the current working directory (the run dir when launching soccer).  We could either change it to be relative to the run directory or clarify that it's just the *name* of a playbook and all playbooks are required to be located in soccer/gameplay/playbooks.  The second one is probably easier.

ra4king:
Wow, made quite a few typos, I never got the chance to review my code as I usually does. That function was not meant to be actually, I deleted it. I also changed the behavior to your second option: the playbook specified is the file name inside soccer/gameplay/playbooks. I updated the code and the command line usage string.

justbuchanan:
Ok, cool, thanks for fixing those.  A couple more things:
* It looks like your `get_enabled_plays_paths()` checks node.enabled on Category nodes, not leaf nodes which actually have that property.
* Since we're now just giving the name of a playbook file in the playbooks directory, we might as well leave off the '.pbk' extension when specifying it from the command line (since it's assumed), and have the playbook python module automatically append it.

ra4king:
* That function was never tested, as I still need to put in UI buttons to call it. However, no it checks node.enabled on Node instances: __iter__ only yields Nodes.

* Done

justbuchanan:
Awesome!

It will probably be a couple weeks before the UI changes get merged, so the UI buttons will have to wait til then.  If we want to merge this now, we can either test the save_playbook() stuff and make sure it works or you can make add a commit to delete it, which we can revert later to add it back in when the buttons are added.

justbuchanan:
Also, the travis build is failing, which is weird because everything builds fine on my computer, even after a `make clean`...

jgkamat:
The travis build is failing on ```make robot2015``` not on building soccer. I'm going to look into it.

justbuchanan:
Yeah I see several commits from Jon in the list that may be from another branch or something.  Not sure what's up there.  It seems like something might have gotten rebased and messed up the git history somehow, so I created a fresh branch `roi-playbooks2` off of master and cherry-picked the relevant commits in order.  It now has all of the stuff we want and none of the stuff we don't.  `make robot2015` now works fine as well (at least on my computer).

Roi, to get up-to-date with this, do a `git fetch` from the robojackets repo, then do `git reset robojackets/roi-playbooks2` while you're on your playbooks branch.  I'd recommend doing a similar thing from your master branch to make sure your master is an exact mirror of the robojackets master branch.

jgkamat:
@ra4king If you want to update this pull request with the new history you will need to do a ```git push --force``` as well.

justbuchanan:
@ra4king, any update on this?

ra4king:
All fixed now, forced the branch to point to your commits in roi-playbooks2, fixed the blank lines issue, and removed the save feature for now.

PR #263, opened 2015-03-30
Albert/pid fixes

ashaw596:
Fix PID memory Leaks

### DIFF
--- a/common/Pid.cpp
+++ b/common/Pid.cpp
@@ -27,10 +27,7 @@ void Pid::setWindup(unsigned int w)
 {
 	if (w != _windup) 
 	{
-
 		_windup = w;
-
-		
 
 		if (w > 0)
 		{	
--- a/common/Pid.cpp
+++ b/common/Pid.cpp
@@ -8,11 +8,10 @@
 using namespace std;
 
 
-Pid::Pid(float p, float i, float d, unsigned int windup)
+Pid::Pid(float p, float i, float d, unsigned int windup) : _oldErr()
 {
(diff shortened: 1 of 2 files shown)


common/Pid.cpp
```diff
@@ -35,16 +35,19 @@ Pid::~Pid()
...
 {
-	if (w > 0)
+	if (w != _windup) 
 	{
-		if (w != _windup) {
-			_windup = w;
-			_oldErr = new float[_windup];
-			memset(_oldErr, 0, sizeof(float)*_windup);
+		if (_oldErr) 
+			delete[] _oldErr;
```
justbuchanan: Bro, don't leave me [hanging](http://en.wikipedia.org/wiki/Dangling_pointer)

ashaw596: Wait so what did I leave hanging?

justbuchanan: You have to make sure to set `_oldErr` to nullptr after you deallocate it, otherwise it points to invalid memory.  There are places, like the destructor, that try to access or delete it if it's not nullptr.

ashaw596: Ic Now. Screw Pointers.

justbuchanan:
Looks good!  For future reference, something like [boost's circular buffer class](http://www.boost.org/doc/libs/1_57_0/doc/html/circular_buffer.html) would be good for storing the windup.  And it would be good to use that when it comes time to store past robot commands in the MotionControl class to account for latency.  The way you have this now with std::vector works fine though, so I'll merge it.

PR #272, opened 2015-04-15
Editorconfig

jgkamat:
Added editor-config that enforces 4 space indention across editors. To implement this install an [editorconfig](http://editorconfig.org/) plugin for your editor!


.editorconfig
```diff
@@ -0,0 +1,27 @@
+# Robocup base editorconfig
```
justbuchanan: Robocup -> RoboCup

PR #280, opened 2015-04-17
Albert/composite path

ashaw596:
Finish CompositePath and refactor InterpolatedPath


justbuchanan:
Travis is complaining :construction_worker:

soccer/Robot.cpp
```diff
@@ -640,11 +641,50 @@ void OurRobot::replanIfNeeded(const Geometry2d::CompositeShape& global_obstacles
...
+		
+		Planning::Path *first = _planner->run(pos, angle, vel, _motionConstraints, &full_obstacles);
+		//unique_ptr<Planning::Path> subPath = first->subPath(0,first->getTime()/2.0);
+		//unique_ptr<Planning::Path> subPath2 = first->subPath(first->getTime()/2.0,first->getTime());
+		
+		//Planning::Path *first = _planner->run(pos, angle, vel, test, &full_obstacles);
+		//Planning::Path *second = _planner->run(*test.targetPos, angle, *test.targetWorldVel, _motionConstraints, &full_obstacles);
+		//Planning::CompositePath *composite=new Planning::CompositePath();
+		
+		/*
```
justbuchanan: Go ahead and remove all the commented-out old code - it's fine for development, but shouldn't really get merged into master.

soccer/planning/CompositePath.cpp
```diff
@@ -0,0 +1,147 @@
...
+			if (t<=0 || timeLength == -1) {
+				t += timeLength;
+				path->evaluate(t, targetPosOut, targetVelOut);
+				return true;
+			}
+		}
+		return false;
+	}
+
+	/**
```
justbuchanan: It's probably best not to copy the doxygen comments from the header file into the .cpp file b/c they'll inevitably get out of sync as changes are made.

soccer/planning/RRTPlanner.cpp
```diff
@@ -230,10 +238,15 @@ Geometry2d::Point pow(Geometry2d::Point &p1, float i)
 }
 
 using namespace Eigen;
+
+float getTime(Planning::InterpolatedPath &path, int index, const MotionConstraints &motionConstraints, float startSpeed, float endSpeed) {
```
justbuchanan: Do you think `getDuration()` might be a better name for this?  My first thought when I saw getTime was that it was checking for what point in time something happened like reaching a certain point.

soccer/planning/Path.hpp
```diff
@@ -33,7 +33,7 @@ namespace Planning
 			 * @param[in] 	start The point on the path to start checking from
```
justbuchanan: *startTime

ashaw596:
OK. I think I fixed Everything.

soccer/planning/CompositePath.hpp
```diff
@@ -0,0 +1,89 @@
...
+#include <planning/Path.hpp>
+#include <Geometry2d/Point.hpp>
+#include <Geometry2d/Segment.hpp>
+#include <Geometry2d/CompositeShape.hpp>
+#include <Configuration.hpp>
+
+namespace Planning
+{
+	/**
+	 * @brief Represents a motion path as a series of {pos, vel} pairs.
```
justbuchanan: Looks like you copy and pasted the description for InterpolatedPath here...

ashaw596: Oops. Fine. I'll check my javadocs

soccer/planning/InterpolatedPath.cpp
```diff
@@ -366,11 +366,91 @@ bool Planning::InterpolatedPath::evaluate(float t, Geometry2d::Point &targetPosO
...
+	if (times.size()>0) {
+		return times.back();
+	} else {
+		return 0;
+	}
+}
+
+unique_ptr<Path> Planning::InterpolatedPath::subPath(float startTime, float endTime) const
+{
+	if (endTime > 0 && startTime > endTime) {
```
justbuchanan: You probably shouldn't include && endTime > 0 here

ashaw596: If endtime<0. That means until the end of the path.

soccer/planning/InterpolatedPath.cpp
```diff
@@ -366,11 +366,91 @@ bool Planning::InterpolatedPath::evaluate(float t, Geometry2d::Point &targetPosO
...
+		throw invalid_argument("startTime can't be after endTime");
+	}
+
+	if (startTime<0) {
+		throw invalid_argument("startTime can't be less than zero");
+	}
+
+	int start = 0;
+	while(times[start]<startTime) {
```
justbuchanan: Won't this get an index out of bounds in the case that startTime is greater than the duration of the path?

justbuchanan:
I went through this and have a few notes:

* There are several new calls to `addText()` in `replanIfNeeded()` - we probably don't want these popping up on-screen all the time when we're running soccer since replanning stuff will happen often.  Maybe you could add a `bool pathDebug` that we can set to true when debugging, but leave it off otherwise
* CompositePath.append() should probably throw an exception when you pass a null pointer rather than silently ignoring it
* Right now it looks like a return value of -1 from `getDuration()` indicates an empty path - is there a benefit to doing this as opposed to returning 0?  Unless there's a benefit, it seems like this complicates things a bit
* Let's implement CompositePath.subPath() before merging this
* InterpolatedPath.subPath() could use a few comments explaining it, it's sorta hard to read as it is now
* I added a couple test cases for InterpolatedPath that don't pass right now - double check that I didn't screw up anything in the unit test, but I think there are some issues there

Overall, looks pretty good!  I think having the CompositePath class and the ability to get subpaths will add a lot of flexibility in our path replanning.

ashaw596:
Ok. I'll look at this on tuesday

justbuchanan:
Sounds good.

One more question.  Right now if you pass `subPath()` a negative endTime, it indicates that you want to take the subpath from the given start time up until the end.  What does it do if you pass an end time that's greater than the path duration?  It may make more sense to have an endTime >= the duration just return up until the end, then you can set the default value for the parameter to FLOAT_MAX and not have any negative number stuff going on.

ashaw596:
Hmmm. Yeah that's a good idea.  I'll change it.

ashaw596:
Ok. I fixed it. It was dropping the second point in the path sometimes.

soccer/planning/CompositePath.cpp
```diff
@@ -0,0 +1,115 @@
...
+
+		for (;start<paths.size(); start++) {
+			if (paths[start]->hit(shape)) {
+				return true;
+			}
+		}
+		return false;
+	}
+
+	void CompositePath::draw(SystemState * const state, const QColor &color, const QString &layer) const
```
justbuchanan: What does the const here do?  I think we can leave it off.

justbuchanan: http://stackoverflow.com/questions/10091825/constant-pointer-vs-pointer-on-a-constant-value

ashaw596: I mean I guess its not really necessary for the state variable, but is there a reason not to? It just enforces a little extra thing to make sure we don't somehow lose that pointer in our draw function.

ashaw596:
Did I never get around to changing that subPath function?

ashaw596:
btw. @justbuchanan  Do you know if there is a good way to signify overriding functions like @override in java. I'm just copying doxygen stuff right now.

ashaw596:
By the way, are you seeing a million of these messages when compiling? 'using integer absolute value function 'abs' when argument is of floating point type' Or is that my comp? I just started running this on my mac.

barulicm:
C++11 introduced the override specifier, which can be placed at the end of a function signature to indicate to compilers and other tools that the function is intended to override an inherited interface.

Specifically for Doxygen, the INHERIT_DOCS flag will tell Doxygen to inherit documentation for undocumented members of a class.

ashaw596:
Coolz. I'll use the override keyword.

ashaw596:
Ok. I finished implementing subPath.

soccer/planning/CompositePath.cpp
justbuchanan: Yeah it doesn't hurt anything about how the code runs, just at first sight it looked to me like you were trying to make the thing pointed to un-modifiable.  Not a big deal, it just seems like something that would be confusing to the untrained eye.

common/Utils.hpp
```diff
@@ -13,7 +13,18 @@
 #include "Constants.hpp"
 #include "time.hpp"
 #include <math.h>
+const static bool THROW_DEBUG_EXCEPTIONS = true;
+static inline void debugThrow(std::exception exception) {
```
justbuchanan: If you wanted, you could have this function take a template argument so it would work for every type of exception.
~~~{.cpp}
template<typename T>
static inline void debugThrow(const T &exception)...
~~~

soccer/Robot.cpp
```diff
@@ -630,6 +629,7 @@ void OurRobot::replanIfNeeded(const Geometry2d::CompositeShape& global_obstacles
 		//  TODO: This is Stupid. This should be fixed in the RRT planner or the Bezier Algorithm.
 		if (_path->destination() && (*_path->destination() - dest).mag() > 0.025) {
 			_pathInvalidated = true;
+			addText("Destination Moved");
```
justbuchanan: Let's remove this text, since this shows up pretty often.

soccer/planning/CompositePath.hpp
```diff
@@ -0,0 +1,99 @@
...
+			 *
+			 * @param[in] 	t Time (in seconds) since the robot started the path
+			 * @param[out] 	targetPosOut The position the robot would ideally be at at the given time
+			 * @param[out] 	targetVelOut The target velocity of the robot at the given time
+			 * @return 		true if the path is valid at time @t, false if you've gone past the end
+			 */
+			virtual bool evaluate(float t, Geometry2d::Point &targetPosOut, Geometry2d::Point &targetVelOut) const override;
+
+			/**
+			 * Returns true if the path never touches an obstacle or additionally, when exitObstacles is true, if the path
```
justbuchanan: Isn't this the other way around?  I think it returns true if it *does* hit an obstacle.

justbuchanan: Also, I don't see the `exitObstacles` parameter.

soccer/planning/InterpolatedPath.cpp
```diff
@@ -366,11 +365,111 @@ bool Planning::InterpolatedPath::evaluate(float t, Geometry2d::Point &targetPosO
...
+		float constant = (times[start+1]-startTime)/deltaT;
+		Point startPos = points[start+1]*(1-constant) + points[start]*(constant);
+		Point vi = vels[start+1]*(1-constant) + vels[start]*(constant);
+		path->points.push_back(startPos);
+		path->vels.push_back(vi);
+	}
+	Point vf;
+	Point endPos;
+	size_t end;
+	if (endTime>= getDuration()) {
```
justbuchanan: This `if` is kinda redundant since you do `endTime = min(endTime, getDuration());` above :)

soccer/planning/InterpolatedPath.hpp
```diff
@@ -59,10 +63,10 @@ namespace Planning
 			float length(const Geometry2d::Point &pt) const;
 
 			/** Returns number of waypoints */
-			size_t size() const { return points.size(); }
+			size_t size() const;
```
justbuchanan: Just FYI - for really small methods like this, if you keep it in the header, the compiler will automatically inline all calls to it.  Otherwise, the compiler can only inline calls to it that exist in the same .cpp file where the method is defined.

justbuchanan: It's fine to leave it as is, just an FYI.

soccer/planning/InterpolatedPath.hpp
```diff
@@ -80,7 +84,7 @@ namespace Planning
 			void startFrom(const Geometry2d::Point& pt, Planning::InterpolatedPath& result) const;
 
 			//Returns the destination of this path (the last point in the points array)
-			virtual boost::optional<Geometry2d::Point> destination() const;
+			virtual boost::optional<Geometry2d::Point> destination() const override;
 
 			/**
 			 * Returns true if the path never touches an obstacle or additionally, when exitObstacles is true, if the path
```
justbuchanan: Same comment about `exitObstacles` and the return value as the other `hit()` method.

soccer/planning/Path.hpp
```diff
@@ -30,10 +31,10 @@ namespace Planning
 			 * starts out in an obstacle but leaves and never re-enters any obstacle.
```
justbuchanan: Same here.

soccer/planning/RRTPlanner.cpp
```diff
@@ -137,15 +139,24 @@ Planning::InterpolatedPath* RRTPlanner::run(
 	{
 		// FIXME: without these two lines, an empty path is returned which causes errors down the line.
 		path->points.push_back(start);
+		path->times.push_back(0);
+		path->vels.push_back(Geometry2d::Point(0,0));
 	}
 	return path;
 }
-
+Planning::InterpolatedPath update(
```
justbuchanan: What does this method do?  Doesn't look like it's called anywhere.

common/Utils.hpp
ashaw596: Changed it.

soccer/Robot.cpp
ashaw596: It shouldn't show up that often though does it? Only when the destination is constantly moving. We also have a text there for reusing path anyways.

soccer/planning/CompositePath.hpp
ashaw596: Oh. idk. I copied it from interpolated paths. I'll change it.

soccer/planning/InterpolatedPath.cpp
ashaw596: I guess its the same as ==. idk if its worth changing though.

soccer/planning/InterpolatedPath.hpp
ashaw596: fixed

soccer/planning/Path.hpp
ashaw596: fixed

soccer/planning/RRTPlanner.cpp
ashaw596: Its a skeleton method for when we can use these classes to replan from a certain point in a path.

ashaw596:
Ok. I think I fixed all the CR comments.

soccer/planning/RRTPlanner.cpp
justbuchanan: Oh ok.  Wouldn't we just decide where we want to replan from, then pass that <pos, vel> pair to the planner?  What would `update()` do?

soccer/Robot.cpp
justbuchanan: It doesn't show up often for a play like `LineUp`, but if you run something with the `Mark` behavior, it does something like

~~~{.py}
target = (opponent.pos - ourgoal) * factor + ourgoal
robot.move_to(opponent.pos - ourgoal
~~~

In this case, it'll show up pretty often.

ashaw596: Fine I get rid of it since we already have a replanning text. I still think we should be more specific in our debug text.

soccer/planning/InterpolatedPath.hpp
```diff
@@ -80,22 +84,26 @@ namespace Planning
...
+			 *
+			 * @param[in]	startTime The startTime for from which the subPath should be taken.
+			 * @param[in] 	endTime The endTime from which the subPath should be taken. If it is greater than the duration fo the path,
+								 it should go to the end of the path.
+			 * @return 	A unique_ptr to the new subPath
+			 */
+			virtual std::unique_ptr<Path> subPath(float startTime = 0, float endTime = std::numeric_limits<float>::infinity()) const override;
 
 			/**
 			 * Returns true if the path never touches an obstacle or additionally, when exitObstacles is true, if the path
```
justbuchanan: How about this comment?

soccer/planning/RRTPlanner.cpp
ashaw596: Well I wrote this like 2 months ago 8), but I think I thought the planner could use the information of the previous path we were on to select a good point to start planning the new path from.

Or we could have the logic for that done in Robot or something. I guess it just depends on what we want. I think at that point too we didn't have the CompositePath so it was just going to add splice it on using the interpolatedPath

soccer/Robot.cpp
justbuchanan: Sounds good, it's just kinda crufty when we have a bunch of text on our nice pretty soccer view :).  I do agree though that we should have a better way to see debug text like this when we want to.  Right now, when you call addText("my info"), it puts it in the "RobotText" layer.  Maybe we could put these planner-related things in a "RobotPlanningDebugText" layer and make it invisible by default, so if you want to see it, you could check the box in the layers tab.  Thoughts?  That would probably be best as a separate pull request though, this one is already pretty massive haha.

soccer/planning/InterpolatedPath.hpp
ashaw596: Godddddd. Is there someway to link these better like @override but easy to navigate :/. Headers are so annoying....

soccer/planning/RRTPlanner.cpp
justbuchanan: Ok gotcha, well do you want to remove update() for now, then later when we figure out how replanning is going to work?

soccer/planning/InterpolatedPath.hpp
ashaw596: Waiiitt. I changed that.

justbuchanan: You've got it marked as `override`, so I think it would be fine to leave it undocumented here and the reader can just look at the superclass method where it's declared for the first time and see the description.  Any place where you copy comments, they're bound to get out of sync.

ashaw596: Oh. nvm. Thought that was a different function. How did that ever get there?

soccer/planning/RRTPlanner.cpp
ashaw596: Removed. Its useless anyways until we have an RRT Planner that actually takes start and end speed into account anyways.

soccer/planning/InterpolatedPath.hpp
ashaw596: I guess thats a good point. Maybe I'll remove the subclass documentation.

ashaw596:
I think thats about mostly everything

justbuchanan:
Looks good to me!  Go ahead and merge it whenever travis is done checking it out.

justbuchanan:
For some reason the CompositePath subpath test fails sometimes, but not all the time...  I pulled this branch and was able to get it to fail on occasion.

ashaw596:
I can't get it to fail on my comp :/

ashaw596:
Did that fix it?

ashaw596:
Hey @justbuchanan , could you test it on you're computer I can't get it to fail on mine and I don't think I should keep spamming random commits to try to get travis to fail. :)

justbuchanan:
I ran this snippet for a while and it didn't break, so I think it's good!  Not sure what was up before.

~~~{.sh}
while echo 'testing' && make test-cpp > testing.log 2>&1; do; done; echo 'tests broke :(
~~~

ashaw596:
I added a default value of 0 for the float instance variable. Weird cause I
thought they were default initialized.

ashaw596:
Well coolllzzz! I'll merge it.

justbuchanan:
Ahh ok, that must have been it.  In C++, scalars (floats, ints, etc) default to whatever value was in memory before.  Luckily, this is the type of thing that a static analyzer would catch :)  I'll see if I can get that up and running soon.

Issue #187, opened 2015-01-16
Interface with mechanical base ID chip


justbuchanan:
This is how we will uniquely identify mechanical bases in order to do per-robot calibration.

jjones646:
@justbuchanan We have some of these now.

ra4king:
http://www.digikey.com/product-detail/en/DS2411R%2BT%26R/DS2411R%2BCT-ND/2045729

justbuchanan:
This might be helpful: https://github.com/contiki-os/contiki/blob/master/dev/ds2411/ds2411.c.

It's not written for the MBED, but shouldn't be hard to port.

ra4king:
That's the most cryptic code I've ever seen. I had to hunt down through multitudes of header files to dig for half of those variables/functions!

ra4king:
Got it working.
https://github.com/ra4king/robocup-software/commit/63a8d9287d61e1212b25b6226b291d1c16fd8659

justbuchanan:
Nice, that's awesome! :boom:.  The next step from there will be packaging it up into its own Ds2411 class.

ra4king:
#281

PR #281, opened 2015-04-18
DS2411 ID Chip Driver

ra4king:
#187

### DIFF
--- a/firmware/robot2015/cpu/commands.cpp
+++ b/firmware/robot2015/cpu/commands.cpp
@@ -4,6 +4,7 @@
 #include "reset.hpp"
 #include "LocalFileSystem.h"
 #include "git_version.hpp"
+#include "ds2411.hpp"
 
 #include <algorithm>
 
@@ -375,8 +376,12 @@ void cmd_info(const vector<string> &args)
 
     printf("Commit Author:\t%s\r\n", git_head_author);
 
-
     printf("\r\n");
+
+    DS2411_ID id;
+	ds2411_read_id(ds2411_pin, &id, true);
+
+	printf("\r\n");
 
     // Prints out a serial number, taken from the mbed forms
     // https://developer.mbed.org/forum/helloworld/topic/2048/
@@ -507,4 +512,3 @@ void cancelIterativeCommand(void)
 {
 	executingIterativeCommand = false;
 }
-
--- a/firmware/robot2015/cpu/commands.cpp
+++ b/firmware/robot2015/cpu/commands.cpp
@@ -379,7 +379,7 @@ void cmd_info(const vector<string> &args)
     printf("\r\n");
 
     DS2411_ID id;
-    ds2411_read_id(ds2411_pin, &id, true);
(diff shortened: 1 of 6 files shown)


ra4king:
#187

justbuchanan:
Nice work!  A few suggestions for improvements before we merge this:
* put a little more description in the header file for what this thing actually is.  You and I know, but someone that stumbles upon ds2411.hpp has no idea what it's actually for.
* have ds2411_read_id take in a PinName instead of a DigitalInOut - it's sort of an implementation detail that DigitalInOut is used, the user of the function should only have to specify what pin it's connected to.
* Move the debug printing stuff out of main.cpp and into the ds2411_read_id() function and have a bool parameter that defaults to false for whether or not it should be printed when the function is called.
* remove the blinking led stuff from main.cpp

Other than those things, it looks really good!

ra4king:
Got the first three done.

For the blinking LED, I kept that in as initConsoleRoutine clears the screen. Can I have the LED blink a certain amount of times then continue to the console?

justbuchanan:
Awesome!  The main() function is going to eventually be the actual program for running everything for the robot, so It's probably best to leave the LED thing out since that won't be on the final control board.

I think a good way to be able to test the ID stuff without cluttering main() would be to have `cmd_info()` (in commands.cpp) print the id stuff out after the git version info.  You can add a new pin declaration for the id chip to `robot.hpp` so `cmd_info()` knows which pin it's on.

ra4king:
Ok I'll push what I have now and make those changes later, AFK for a bit.

ra4king:
Hmm I don't know where to put the declaration of the DS2411_ID variable. It shouldn't be in `robot.hpp`, but it doesn't fit in any other C file. What do you suggest?

justbuchanan:
I wouldn't make a global variable for the ID just yet.  We might eventually, but leave it out for now.  Just put a pin declaration in robot.hpp for where the chip is connected, then have the `cmd_info()` function re-read from id chip every time it's called.

ra4king:
Oh so you did want me to re-read from the chip every time. Not the most efficient but it works for now.

ra4king:
Done. The debug info prints out when 'info' is typed as a command in the console, right after the Git info.

justbuchanan:
Awesome, I like it!  Only one thing left and I'll merge it.  Right now all the other pin mappings are #defined in robot.hpp under the comment "PIN DECLARATIONS", so it'd be best to put the ds2411 pin declaration up there too.  I do agree that doing `extern const PinName` for those is a little better than macros, but we should be consistent and do the pin declarations all the same.  If you want to do all the pins that way and put the actual definitions in a `robot.cpp` file, that'd be fine too.

justbuchanan:
Actually I think the #define might be slightly better (even if less c++ "correct") b/c you can just look at the header and know exactly what the pins are actually mapped to instead of having to go back to the .cpp file.

ra4king:
Oh! I never noticed those, the file was so long I just scrolled all the way to the bottom haha. I made it a #define.

justbuchanan:
:+1:

PR #286, opened 2015-04-22
Adds new dependency required by the mbed build tools

guyfleeman:


jgkamat:
Uh, jinja2 should have not been added to the ubuntu-packages file, that package is not in the repositories. 1a34868 should probably be reverted, but 15256f6 should be fine.

jgkamat:
[docker logs](http://pastebin.com/raw.php?i=zrRGCc0G)

PR #287, opened 2015-04-22
Fix build failure created by pull #286

jgkamat:
Sorry for all these one line pull requests, but the build broke. :disappointed:


justbuchanan:
Oops, thanks for catching that!

PR #302, opened 2015-04-27
unkick() now works for the simulator too.  fixes #295

### LINKED ISSUE #295: Unkick not called in simulator after halt
Not sure if this is a simulator only issue or not. See one_touch_pass on the touchpass branch for an instance of this

justbuchanan:
@jgkamat I think this fixes the issue you were having.  Try it out when you get a chance and let me know if it works.


jgkamat:
Yup, this seems to fix the issue for me! I also found a small error in my own play, which I fixed.
This looks good!

PR #306, opened 2015-06-17
Bumped up the compiler version to C++14

justbuchanan:


ashaw596:
Shouldn't you remove the -std=c++11in soccer, simulator, and test too?

justbuchanan:
For some reason they were never there, which was a little surprising haha.  I did a global search for "-std=" and fixed up what was there.

Looks like travis doesn't like my changes... time to investigate.

ashaw596:
It's definitely in soccer/CMakeLists. At least in my repo.

justbuchanan:
I totally missed that, good call.

jgkamat:
This dosent seem to work for me unless my gcc version is > 4.9, not 4.8...
Unfortunately, 4.9 isn't in ubuntu 14.04 repos :cry:

ashaw596:
We should really make our own apt-get repo next year.

justbuchanan:
The fact that travis-ci is still on Ubuntu 12.04 is getting more and more irritating...  I wonder if we can get a version of clang installed on travis-ci that supports c++14.

jgkamat:
[CircleCi](https://circleci.com/integrations/docker) seems like a really good alternative, I've tried it out and i'm loving it! It has all the integration features we currently use with travis (and more like ssh), and it supports docker containers (so we could theoretically get any environment we wanted). I've set up [my docker repo](https://github.com/jgkamat/docker/pull/1) as a proof of concept, but we probably shouldn't mess with anything until after comp.

justbuchanan:
Hmm, that actually sounds pretty sweet!  Another problem that might help fix is speed.  If we cache a docker image with `ubuntu-setup` already run, our test suite would go way faster than it currently does.  Once comp is done this year, let's definitely check it out.

jgkamat:
Hmm, I don't think c++14 is supported by gcc 4.9. I'm not sure if ubuntu 14.04 has anything newer...

justbuchanan:
It might be good if we set the compiler to clang for circle.  I'll try it on this branch and see if that helps.

jgkamat:
If you actually wanted to inject env variables every build, the best place to do that would be in buildcircle.sh
You could add `-e CC=clang` to the arguments

jgkamat:
[you should take a look at this(https://circleci.com/gh/RoboJackets/robocup-software/196), I got it working (but had to add another ppa)

justbuchanan:
Nice!  I guess if we have to we can use a ppa, but maybe we can get clang to do it - I'm pretty sure stock clang on Ubuntu supports c++14.

jgkamat:
That would be great (I really hate adding more ppas...)

justbuchanan:
I've been messing with this a little more and on Ubuntu 14.04, the regular clang package doesn't support c++14.  You have to install `clang-3.5` for 14 support, but that means the binary name is `clang-3.5`, not `clang`, which makes it a little tougher to get everything to work on Ubuntu and everything else :/

jgkamat:
Dammit why do they make everything so difficult :angry: 

We could make a quick bash script to check which one is available, and call that instead of the clang library (or make an environment variable called $CLANG and populate that with the real clang). 

Or we could make ugly symlinks. Maybe theres a way to make apt add those symlinks or something gracefully.

```
if ! command -v clang-3.5; then
  command -v clang
fi

and then in another file, run that like

`./getclang` --std=c++14....
```

justbuchanan:
Yeah I'm becoming less and less of an ubuntu fan...

I sorta like the symlink option the best, but it's still a little ugly.  We could put that in ubuntu-setup after installing packages and it should work fine.

jgkamat:
Uh, could we refrain from symlinking this in all cases? How about

```
if [ "$(clang --version | grep -Eo '[0-9]+\.[0-9]+' | head -n1)" \< "3.5" ]; then
  ln -sf $(which clang-3.5) /usr/bin/clang
fi
```
instead, or perhaps simpler, adding the simlink only if the repo's are added  (I don't want to mess up my system with these symlinks) :confused: 

Debian Testing is :tada:

ashaw596:
Can't we just force ubuntu users to use clang 3.5? (or 3.6)]

```sudo apt-get install clang-3.6```

**Comment by jgkamat** [comment]
The issue is that when you install clang-3.5/6 on ubuntu, the executable is `clang-3.5/6`, not `clang`, as in other systems.

**Comment by justbuchanan** [comment]
@jgkamat looks good to me, you can go ahead and add that to `ubuntu-setup` on this branch.

I agree, the fact that we're using a frozen distro makes this pretty tough.  I do sorta like your proposal of using Docker so we can have exactly the same dev environment on everyone's computers, it would just have to be setup so that people to be dead simple to use.  I think for now we'll continue workarounds to make Ubuntu work, but eventually we should definitely look into actually solving the problem.

**Comment by jgkamat** [comment]
I think that the versioning problem is big enough that it deserves its own issue, I'll make one in a bit (probably just for a long term goal)

**Comment by ashaw596** [comment]
What was the thing that was causing us to use the gcc std library while compiling with clang?

**Comment by ashaw596** [comment]
btw, do we really want to symlink clang? why don't we just specify which compiler cmake should use?

**Comment by jgkamat** [comment]
we could say 'use clang-3.5' but 'clang-3.5' isn't available on most systems, 'clang' is.
that clang version is > 3.5, but because ubuntu froze clang before that, the 'clang' version on ubuntu is not as high as we need.

Alternatively, we could create a small shell script to put in /usr/bin and call that, and that would run the correct version of clang.

**Comment by ashaw596** [comment]
I mean we're going to have to tell cmake to use clang anyways. We can check if clang 3.5 is in the system and tell it to use that or just clang otherwise.

**Comment by ashaw596** [comment]
Ok, I've been messing around with this branch for a week and I just cannot get clang to compile on ubuntu 14. Have you guys managed to get it to compile? I keep getting linking errors and incompatible libs.

**Comment by jgkamat** [comment]
I've gotten the GCC version that supports C++14 working (by adding yet another terry guo). I've never tried to get the clang compiler to work.

**Comment by ashaw596** [comment]
Yeah I saw that on the docker compile. meh. Are we just gonna use that probs?

**Comment by jgkamat** [comment]
please no more 3rd party repos though
:cry: 
It would be great if we could get clang to work as is...

**Comment by ashaw596** [comment]
lol. You added it. idk I keep getting into an issue with some c11 deprecated stuff if I don't switch libs. Then I get into some linking issues if I switch to stdc++ libs with protobuf and I think I got some compile issues with boost. I can't figure it out.

**Comment by ashaw596** [comment]
I think the boost headers installed on ubuntu are from boost 1.0. at least on my computer. Any idea how to update them?

**Comment by jgkamat** [comment]
Huh, isn't the version added by ubuntu setup newer than that? Try `apt-cache search boost` and see if something relevant comes up?

**Comment by ashaw596** [comment]
It is but the headers in usr/include/boost are pretty  old it looks like.
It definitely doesn't have make_unique.

**Comment by jgkamat** [comment]
The version of boost that ubuntu setup installs should be 1.54, maybe it's files are in a different place? You could install 1.55 if you wanted though.
`cat  /usr/include/boost/version.hpp`

**Comment by ashaw596** [comment]
Oh nevermind. I got confused I was trying to use make_unique, but that wasn't introduced until 1.56.

**Comment by justbuchanan** [comment]
It works!  Can someone try this out on their ubuntu install (after running ubuntu-setup of course).  And Jay can you check debian?

**Comment by jgkamat** [comment]
Looking good from here! I'll test this out in an ubuntu VM as well, but all seems well so far.

**Comment by ashaw596** [comment]
Looks good on ubuntu. So did we end up giving up on the clang compiler? Did this install another repo for gcc?

**Comment by justbuchanan** [comment]
Sweet, I'll go ahead and merge it then.  This is unfortunately still using the GCC compiler on some platforms (uses clang on mine).  GCC's not terrible, but clang is better.  We can work on better usage of clang in the future, but having C++14 is good either way.

**Comment by jgkamat** [comment]
This made osx travis unhappy? I think the clang version on that is too old?

**Comment by justbuchanan** [comment]
Good call, didn't notice that.  Everything builds fine on my mac and the travis script does a `brew update` and `brew upgrade` before building, so in theory it should have the same up-to-date packages as my computer...  I wish travis had the same SSH feature circleci does.

**Comment by justbuchanan** [comment]
Looks like the compiler is actually coming from Xcode, not homebrew, so this makes sense unfortunately :/  I have AppleClang 7 and travis has AppleClang 6.

**Comment by jgkamat** [comment]
One more thing, on arch, cmake seems to be complaining about how google/googletest dosen't have a cmake dir in it's root, but it seems fine on ubuntu for some reason. Have you seen this?

**Comment by justbuchanan** [comment]
Hmm, that's wierd...  Can you post the error text?

**Comment by jgkamat** [comment]
```
mkdir -p build
cd build && cmake -GNinja -Wno-dev --target  test-soccer  .. && ninja  test-soccer
-- Found Qt5: /lib64/cmake/Qt5Widgets
-- Found libusb-1.0:
--  - Includes: /usr/include/libusb-1.0
--  - Libraries: /lib64/libusb-1.0.so
-- Boost version: 1.58.0
-- Found the following Boost libraries:
--   system
-- Boost version: 1.58.0
-- Found the following Boost libraries:
--   python3
-- MBED target set to LPC1768
-- Configuring done
-- Generating done
-- Build files have been written to: /home/jay/Code/robocup-software/build
ninja: warning: multiple rules generate common/git_version.cpp. builds involving this target will not be correct; continuing anyway [-w dupbuild=warn]
[14/85] Performing update step for 'googletest'
Current branch master is up to date.
[15/85] Performing configure step for 'googletest'
FAILED: cd /home/jay/Code/robocup-software/build/src/googletest-build && /usr/bin/cmake -DCMAKE_ARCHIVE_OUTPUT_DIRECTORY_DEBUG:PATH=DebugLibs -DCMAKE_ARCHIVE_OUTPUT_DIRECTORY_RELEASE:PATH=ReleaseLibs -GNinja /home/jay/Code/robocup-software/build/src/googletest && /usr/bin/cmake -E touch /home/jay/Code/robocup-software/build/src/googletest-stamp/googletest-configure
CMake Error: The source directory "/home/jay/Code/robocup-software/build/src/googletest" does not appear to contain CMakeLists.txt.
Specify --help for usage, or press the help button on the CMake GUI.
[15/85] Linking CXX static library common/libcommon.a
ninja: build stopped: subcommand failed.
makefile:36: recipe for target 'test-soccer' failed
make: *** [test-soccer] Error 1
```

justbuchanan:
Not sure what's up.  I just did a `make clean`, `arch-setup`, and `make test-soccer` on arch and it worked ok for me.  Probably not related, but I just noticed that the arch-setup script doesn't sync submodules, so I added that in and pushed.

justbuchanan:
Also, I looked in the directory that the error mentioned and there's definitely a CMakeLists.txt in there for me.  It seems like a potential with submodules on your end.  Did you double check that everything is synced and updated?

jgkamat:
Hmm, Im still getting the error. I'm seeing the CMakeLists in external/googletest/googletest, not external/googletest. I'll show you what I'm seeing at the meeting...

PR #311, opened 2015-06-18
Added `static-analysis` makefile target to run the clang static analyzer

### LINKED ISSUE #308: Setup clang static analyzer
The clang project has a tool called the "static analyzer" that magically finds potential bugs in your codebase.  I'd bet that we would find quite a few things if we ran it on robocup-software.  I'll be setting up a `make` target so it's easy to run, then taking the advice it gives when relevant.

justbuchanan:
Closes #308.

Not sure if it's a good thing or a bad thing, but static analysis only found a few minor things in our codebase.  I made those changes as part of this pull request.

Can someone confirm that this works on Ubuntu?  I've only got an Arch machine with me right now.  On Arch, the static analyzer is a separate package, but from what I can tell, it's installed as part of the `clang` package on Ubuntu.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@
 .scons_variables
 /build
 /build-scons
+/static-analysis
 .pydevproject
 /gmon.out
 *.swp
--- a/.gitignore
+++ b/.gitignore
@@ -4,7 +4,6 @@
 .scons_variables
 /build
 /build-scons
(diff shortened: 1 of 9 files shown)


jgkamat:
It looks like its working fine for me on ubuntu1404

justbuchanan:
Cool, thanks!

PR #319, opened 2015-06-27
Kill background simulators when running make run-sim

jgkamat:
I got really sick of the headless simulator being reused when running `make run-sim`, so I added a line to the makefile to kill the headless simulator when run-sim starts. Should I add it to run after the execution instead?

The line should be ignored if it fails

Also, I'm not sure if macs have pkill installed (its not posix). If thats the case, then I can convert this to use standard ps and kill.

### DIFF
--- a/makefile
+++ b/makefile
@@ -10,6 +10,7 @@ static-analysis:
 run: all
 	cd run; ./soccer
 run-sim: all
+	-pkill -f './simulator --headless'
 	cd run; ./simulator --headless &
 	cd run; ./soccer -sim
 
--- a/makefile
+++ b/makefile
@@ -10,7 +10,8 @@ static-analysis:
 run: all
 	cd run; ./soccer
 run-sim: all
-	-pkill -f './simulator --headless'
+	# Kill simulator if running
+	-if [ -n "$$(ps -A -o args | grep './[s]imulator --headless')" ]; then kill "$$(ps -A -o pid,args | grep './[s]imulator --headless' | awk '{print $$1}')"; fi
 	cd run; ./simulator --headless &
 	cd run; ./soccer -sim
 


justbuchanan:
Cool, I like it - simulators fighting each other is a pain in the butt haha.  I looked it up online and I think OSX does come with pkill.  I'd check, but my Mac is having issues right now.  @ashaw596 , can you check if this works on OSX?

I think at some point it would be nice to add a flag to the simulator so doing something like `simulator --timeout=5` would tell it to exit if it goes for five seconds without receiving any new "radio" packets.  That way we don't have to explicitly kill it, it would just gracefully exit after you close soccer.  We can do this sometime after comp though.

jgkamat:
I switched it over to use ps and kill, because we might as well try to support as many systems as possible :smile: 

I had to escape my `$`s like this `$$` though, I'm assuming thats standard....

justbuchanan:
Ok cool.  Well if the first one works on OSX (I think it will), I'd rather go with it since it's much simpler.  If we have to go with option two for compatibility then we can.

I'll go ahead and merge version one and if it breaks OSX, we'll pull version two.

jgkamat:
:+1: sounds good to me!

ashaw596:
I've used pkill on my comp before so it'll probs work. Might have installed
it myself tho. Don't think so so it's probs good.

justbuchanan:
I looked at this again, and I think if we actually did `killall simulator` it might be slightly better because we don't have to include the flags that we passed to it originally.  And I've used killall on OSX plenty of times.

jgkamat:
I originally was going for only stopping headless simulators, but if we're fine with stopping all simulators, then that probably works better!

justbuchanan:
Yeah that's a fair point, let's leave it as is then.  I'll add an issue for adding a timeout option to the simulator, but we can do that sometime after comp.

PR #322, opened 2015-06-30
implemented lqr control for robot.  closes #98

### LINKED ISSUE #98: MOTION CONTROL
We need to suck it up and do some actual math and control theory, we've had crap motion control for too long.

justbuchanan:
Ladies and gentlemen, I present to you the new robot motion controller.  It's an [LQR Controller](https://en.wikipedia.org/wiki/Linear-quadratic_regulator), which takes into account the robot dynamics (mass, moment of inertia, etc) in order to calculate the "optimal" voltages to apply to the motors to achieve our target velocity.  The [readme](https://github.com/RoboJackets/robocup-software/tree/lqr-control/firmware/robot2015/cpu/motion_control) has more info on how it all works.

This is based on the MATLAB Simulink setup in the /modeling folder.  At compile time, it calculates the gain matrices for the robot and stores them in a lookup table to be used at runtime.  Calculating them on the robot would be nice, but it's a very expensive process.

It still needs some tuning as well as updated model parameters once the robot is complete.  The model paramters go [here](https://github.com/RoboJackets/robocup-software/blob/lqr-control/firmware/robot2015/cpu/motion_control/TunableRobotParameters.cpp).

~~The lqr-generation code is currently in /lqr_lookup_table - it will eventually go under /firmware, but that will require some reorganization that will break other open firmware branches, so that will happen later.~~

TODO:
- [ ] Set LQR weights to sane defaults, then tune
- [ ] Test PseudoInverse()
- [ ] Test LQR outputs
- [ ] Test LqrLookupTable
- [ ] run on MBED

cc @barulicm @jjones646 @ChrisDoho


justbuchanan:
Can someone verify that this builds on ubuntu?  You'll need to run ubuntu-setup first.

jgkamat:
`make robot2015

justbuchanan:
Yeah - actually hold off for now - I'm in the process of fixing some matrix ops.  I'll let you know when it's fixed and you can try building robot2015.

jgkamat:
`make robot2015` seems to be failing right now, but 19e93a6cb28bea44554adb4801f0c25ba6fff3f9 seems to build (both `make` and `make robot2015`) fine on debian 8 and ubuntu. It pulled in java though :cry:

justbuchanan:
Unfortunately I have to use the "octave" program to compute LQR matrices, which is java-based.

justbuchanan:
Ok, it should be good to go now.

jgkamat:
:+1: the build looks good!

justbuchanan:
Awesome, that's a good start!  Thanks for checking.  Now to see if the outputs are any good...

justbuchanan:
This is going to be too much work to get up and running for this competition, so we're going with PID control instead ( #703 ).

PR #327, opened 2015-07-02
automatically generate git version info for verilog

justbuchanan:
@jjones646, I added some CMake magic for generating a "git_version.vh" file with the first eight characters of the hash and a dirty bit.  Let me know if you need anything changed.


justbuchanan:
This builds fine on my computer, but travis is failing it for some reason...

PR #328, opened 2015-07-02
Albert/hide layers

### LINKED ISSUE #307: Add the ability to have text layers in soccer default to hidden
When drawing shapes or text on the soccer field view, we pass a string indicating the "layer" that these show up on.  The "layers" tab has a check box for each layer to show or hide the content.  Right now every layer defaults to being enabled, which can lead to a lot of unneeded debug content being drawn that clutters up our pretty soccer field view.  It would be nice if we could add a way to hide some by default, then we can enable them when we're debugging specific things.

ashaw596:
Maybe what you wanted? It hides the motion planning stuff at least.
Closes #307

### DIFF
--- a/soccer/FieldView.hpp
+++ b/soccer/FieldView.hpp
@@ -24,6 +24,11 @@ class FieldView : public QWidget
 			{
 				_layerVisible[i] = value;
 			}
+			else if (i >= _layerVisible.size())
+			{
+				_layerVisible.resize(i+1);
+				_layerVisible[i] = value;
+			}
 		}
 		
 		bool layerVisible(int i) const
--- a/soccer/MainWindow.cpp
+++ b/soccer/MainWindow.cpp
@@ -31,7 +31,7 @@ using namespace Eigen;
 QString LiveStyle("border:2px solid transparent");
 QString NonLiveStyle("border:2px solid red");
 
-std::array<QString, 2> hiddenLayers[]= {"Debug", "Rules"};
+std::vector<QString> hiddenLayers {"MotionControl", "Planning0", "Planning1", "Planning2", "Planning3", "Planning4", "Planning5", "time", "velocity"};
 
 void calcMinimumWidth(QWidget *widget, QString text)
 {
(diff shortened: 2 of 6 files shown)


soccer/FieldView.hpp
```diff
@@ -24,6 +24,11 @@ class FieldView : public QWidget
 			{
 				_layerVisible[i] = value;
 			}
+			else if (i >= _layerVisible.size())
```
justbuchanan: What happens if `i` is greater than _layerVisible.size() by more than one?

soccer/MainWindow.cpp
```diff
@@ -31,6 +31,8 @@ using namespace Eigen;
 QString LiveStyle("border:2px solid transparent");
 QString NonLiveStyle("border:2px solid red");
 
+std::vector<QString> hiddenLayers {"MotionControl", "Planning0", "Planning1", "Planning2", "Planning3", "Planning4", "Planning5", "time", "velocity"};
```
justbuchanan: This doesn't reflect the current visibility of layers in the UI, only the default/initial visibility.  Maybe rename it initiallyHiddenLayers?  Also, should probably be an ivar, not a global.

soccer/MainWindow.hpp
```diff
@@ -71,6 +71,8 @@ class MainWindow : public QMainWindow
 
         QTimer updateTimer;
 
+		void addLayer(int i, QString name, bool checked);
```
justbuchanan: This should probably be private.

soccer/FieldView.hpp
ashaw596: Then it gets resized to that size and the intermediate values gets set whenever those i values get called. It should really never happen out of order though.

soccer/MainWindow.cpp
ashaw596: What do you mean by an ivar?

ashaw596:
So whatcha thing @justbuchanan

justbuchanan:
Ok, looks good to me!

PR #330, opened 2015-07-06
Add a onetouchpass play

### LINKED ISSUE #244: Implement one-touch-pass play
One play that a lot of teams use is to have a robot position itself a little ways away from the goal, then have another robot pass to it.  The stationed robot is positioned so that as soon as it receives the ball, it kicks it at the goal.  This year, our passing accuracy should be significantly improved, so this is something we should aim to have.  If we can get it to work in the simulator now, I think we can get the motion control to catch up.

Here's a video of us doing it a long time ago:

https://www.youtube.com/watch?v=U5mJghTbqKk

jgkamat:
This still needs a lot of cleanup, but I think its about time for this to be reviewed. Currently the only thing left to do is aim for areas that arent the center of the goal when using touchpasses (which is waiting on window evaluator improvements (that I could do), and possibly optimize my function for determining receive points (Right now it creates a list of points, and then iterates through them, we don't have to store this list).

This works best when the ball is on the enemy side of the field. Sometimes the ball won't replan if a robot gets in the way, this is because the field is layed out so all points are blocked (usually). We can increase the amount of checked points to overcome this. I also check for a replan once every 30 frames, as it's relatively slow.

Let me know if anything needs more work/documentation/cleanup, and i'll get to it! :smile: 
I've rebased this onto the latest master so you may need to `reset --hard`. (@rmkeezer).

This closes #244


justbuchanan:
Sounds great!  I'll look through and review it later tonight or tomorrow.

@barulicm is rewriting the WindowEvaluator in C++ (with a python wrapper) and I think we'll add in shot success likelihood into it.  I think we'll wait for that to go through before merging this.

One thing that would probably be good is if you could refactor this into a tactic, so we could easily drop one_touch_pass into an existing play.  This would just require moving most of the code out of the test play you currently have an into a new class that subclasses Tactic.

jgkamat:
@ashaw596 This is how I'm currently offsetting the mouth of the robot so the shot (should) go on target. I first calculate the mag of the vector described by the arrow in the picture below, then I calculate the angle deflection that would be needed to get the X component of the kicker's vector when shooting at 8m/s. Then the X component of the kicker and the arrow should cancel each other out, and the resulting ball velocity vector should go straight into the goal (the y axis in the rotated frame). We can turn this off pretty easily if we wanted. One problem with this is that this is calculated after the ball has been passed, so the robot can only begin deflecting then.

test

jgkamat:
I just moved onetouchpass to a tactic, and made a little play to run it. 

Another issue I just noticed is that if the 1st robot blocks it's future pass point, the point will be recalculated. Is there a way to get that specific robot to pass into the exclude_robots list of the evaluator?

justbuchanan:
Yeah, take a look at CoordinatedPass [here](https://github.com/RoboJackets/robocup-software/blob/867b8ce39903e819e4934c78d65400b83e979052/soccer/gameplay/tactics/coordinated_pass.py#L125).  You can't get a reference to the receiving robot until it's been assigned, but afterwards you can do `receiver.robot` to get a reference to it.

soccer/gameplay/constants.py
```diff
@@ -73,8 +74,10 @@ class Field:
 
     TheirGoalSegment = robocup.Segment(robocup.Point(GoalWidth / 2.0, Length),
                                         robocup.Point(-GoalWidth / 2.0, Length))
+    TheirGoalCenter = robocup.Point(0 , Length)
```
justbuchanan: Let's remove this and the OurGoalCenter value below.  The Segment class has a [center() method](https://github.com/robojackets/robocup-software/blob/master/soccer/gameplay/robocup-py.cpp#L302), so it doesn't add much.  Also, if we wanted to have these, we'd have to modify the `setFieldConstantsFromField_Dimensions()` method below to update their values if we switch field sizes.

soccer/gameplay/evaluation/touchpass_positioning.py
```diff
@@ -0,0 +1,100 @@
+import robocup
+import constants
+import main
+import evaluation.passing
+
+## The Touchpass Positioner finds the best location within a rectangle
```
justbuchanan: Maybe make this just a bit more descriptive by mentioning that it's used for finding positions for the receiver in the touchpass tactic.  It also wouldn't hurt to give a quick description of what you mean by "best" location here.

soccer/gameplay/evaluation/touchpass_positioning.py
```diff
@@ -0,0 +1,100 @@
...
+    def debug(self):
+        return self._debug
+    @debug.setter
+    def debug(self, value):
+        self._debug = value
+
+    ## Returns a robocup.Rect object that is the default location to be evaluated
+    #
+    # Takes a current ball position/initial kick position (robocup.Point)
+    def generate_default_rectangle(self, kick_point):
```
justbuchanan: This could probably be a class method since it doesn't depend on any instance state.

soccer/gameplay/evaluation/touchpass_positioning.py
```diff
@@ -0,0 +1,100 @@
...
+                candiate = robocup.Point(currentx, currenty)
+                if not constants.Field.TheirGoalShape.contains_point(candiate):
+                    outlist.extend([candiate])
+                currentx = currentx + threshold
+            currenty = currenty - threshold
+            currentx = rect.min_x()
+        return outlist
+
+    ## Evaluates a single point, and returns the probability of it making it.
+    def eval_single_point(self, kick_point, receive_point):
```
justbuchanan: It'd be good to clarify that this is the probability of the pass AND the shot succeeding.  Also clarify that the probability is between zero and one.

Having values between zero and one for probabilities or power values is fairly standard throughout our codebase, but I think there may still be a few places where it's 0 to 100.  Those should be changed though :)

soccer/gameplay/evaluation/touchpass_positioning.py
```diff
@@ -0,0 +1,100 @@
...
+import constants
+import main
+import evaluation.passing
+
+## The Touchpass Positioner finds the best location within a rectangle
+#
+# Example usage:
+# tpos = TouchpassPositioner()
+# best_pt = tpos.evaluate(ball.pos, robot2.pos)
+class TouchpassPositioner:
```
justbuchanan: What are your thoughts on changing this from a class to just a set of functions?  It seems like there isn't really any state to keep around, just some evaluation calls.

soccer/gameplay/evaluation/touchpass_positioning.py
```diff
@@ -0,0 +1,100 @@
...
+        # TODO dont only aim for center of goal. Waiting on window_evaluator returning a probability.
+        targetPoint = constants.Field.TheirGoalCenter
+        currentChance = currentChance * evaluation.passing.eval_pass(receive_point, targetPoint)
+        return currentChance
+
+
+    ## Finds the best receive point for a bounce-pass.
+    #
+    # Takes in an initial kick point and an optional evaluation zone.
+    def eval_best_receive_point(self, kick_point, evaluation_zone=None):
```
justbuchanan: I think you can change it from `evaluation_zone=None` to `evaluation_zone=generate_default_rectangle(kick_point)` and avoid having to check for `None` below.

soccer/gameplay/evaluation/touchpass_positioning.py
```diff
@@ -0,0 +1,100 @@
...
+        # Autogenerate kick point
+        if evaluation_zone == None:
+            evaluation_zone = self.generate_default_rectangle(kick_point)
+
+        # print(evaluation_zone.min_x())
```
justbuchanan: Once this is finalized, let's remove the commented-out print statements.

soccer/gameplay/plays/testing/test_one_touch_pass.py
```diff
@@ -0,0 +1,34 @@
...
+import tactics.behavior_sequence
+import robocup
+import constants
+import main
+
+
+## Continually runs a one_touch_pass pass tactic
+class TestOneTouchPass(play.Play):
+
+    ReceiveXCoord = 1
```
justbuchanan: Instead of defining a receive x and y, you can just define a receive point here.

soccer/gameplay/skills/angle_receive.py
```diff
@@ -0,0 +1,259 @@
...
+                    behavior.Behavior.State.failed,
+                    lambda: time.time() - self._ball_kick_time > AngleReceive.ReceiveTimeout,
+                    'ball missed :(')
+
+
+
+    ## set this to True to let the receiver know that the pass has started and the ball's in motion
+    # Default: False
+    @property
+    def ball_kicked(self):
```
justbuchanan: Take a look at `OurRobot.just_kicked` - I think it does the same thing as you're doing with this property, but is baked into the `OurRobot` class.

soccer/gameplay/skills/angle_receive.py
```diff
@@ -0,0 +1,259 @@
...
+        ## being in this state signals that we're ready for the kicker to kick
+        aligned = 2
+
+        ## the ball's been kicked and we're adjusting based on where the ball's moving
+        receiving = 3
+
+    def __init__(self):
+        super().__init__()
+
+        self.ball_kicked = False
```
justbuchanan: Rather than storing ball_kicked as a variable, you can check if the kicker behavior has completed.  See [this transition](https://github.com/robojackets/robocup-software/blob/master/soccer/gameplay/tactics/coordinated_pass.py#L57) in coordinated_pass.

justbuchanan: Scratch that, I misread this, this is fine :)

soccer/gameplay/skills/angle_receive.py
justbuchanan: Scratch this one too

soccer/gameplay/tactics/coordinated_pass.py
```diff
@@ -32,10 +32,16 @@ class State(enum.Enum):
         receiving = 3   # the kicker has kicked and the receiver is trying to get the ball
 
 
-    def __init__(self, receive_point=None):
+    def __init__(self, receive_point=None, skillreceiver=None):
```
justbuchanan: Let's make a note here or at the top about what the skillreceiver is used for / why it's a parameter.

soccer/gameplay/tactics/coordinated_pass.py
```diff
@@ -86,7 +92,7 @@ def receive_point(self, value):
 
 
     def on_enter_running(self):
-        receiver = skills.pass_receive.PassReceive()
+        receiver = self.skillreceiver
```
justbuchanan: I think we need to `restart()` the skillreceiver here - if the CoordinatedPass is restarted as of now, it'll use the same instance of skillreceiver, which will be in some sort of completed state.

soccer/gameplay/tactics/one_touch_pass.py
```diff
@@ -0,0 +1,72 @@
...
+from enum import Enum
+
+## A tactic that causes a robot to pass to another one,
+# who scores on the goal as fast as possible.
+#
+# This class is supplemented by touchpass_positioning and angle_receive
+class OneTouchPass(composite_behavior.CompositeBehavior):
+
+    tpass = evaluation.touchpass_positioning.TouchpassPositioner()
+    THRESHOLD = 0.15 # 15%
```
justbuchanan: Let's make a note on what THRESHOLD does here and/or rename it to something a little more descriptive like ReceivePointChangeThreshold.

soccer/gameplay/tactics/one_touch_pass.py
```diff
@@ -0,0 +1,72 @@
...
+    tpass_execution = 0
```
justbuchanan: This should probably be an instance variable.

justbuchanan:
I made a few notes throughout, so let me know if you have any questions.  Overall, it looks really good, nice work!  This will be a great tactic to have available - a lot of other teams use it and most goals are scored using something like this.

soccer/gameplay/evaluation/touchpass_positioning.py
jgkamat: The default rectangle is dependent on the ball's current position, and using that would cause the rectangle to only be generated once. I did this to make sure that the rectangle would be regenerated every time. Don't want to run into that problem we were having with coordinated_pass again :smile:

jgkamat:
I think I got all the notes! I switched tpass_positioner to be a static file, and moved stuff around to accomodate. Thanks! :smiley:

soccer/gameplay/evaluation/touchpass_positioning.py
```diff
@@ -0,0 +1,93 @@
...
+
+## The Touchpass positioning file finds the best location within a rectangle to ricochet
+# a ball into the goal.
+#
+# By default, this will select a rectangle that is across the field from the current ball position.
+# The best location is found by multiplying the chance the pass from the ball's position to the point will make it
+# with the chance a shot into a goal will make it. The greatest probability is selected, and returned.
+#
+# Example usage:
+# tpos = TouchpassPositioner()
```
justbuchanan: Let's update the usage here :)

soccer/gameplay/plays/testing/test_coordinated_pass.py
```diff
@@ -36,9 +36,11 @@ def execute_running(self):
         pass_bhvr = self.subbehavior_with_name('pass')
 
         if pass_bhvr.is_done_running():
-            pass_bhvr.restart()
+            self.remove_subbehavior('pass')
```
justbuchanan: Any reason for removing, then re-adding the pass behavior instead of just restarting the old one?  The only reason I like `restart()` better here is that it's a good test of whether or not the object state is handled appropriately throughout transitions.

soccer/gameplay/skills/angle_receive.py
```diff
@@ -0,0 +1,257 @@
...
+        # Kick the ball!
+        self.robot.kick(self.kick_power)
+
+        # don't use the move_to() command here, we need more precision, less obstacle avoidance
+        pos_error = self._target_pos - self.robot.pos
+        vel = pos_error * 3.5
+        self.robot.set_world_vel(vel)
+
+        # If the shot took place, end the behavior!
+        if self._shot_time != self.robot.lastKickTime():
```
justbuchanan: Would `robot.just_kicked()`?  It would save from having to store shot time in a variable.

soccer/gameplay/plays/testing/test_one_touch_pass.py
```diff
@@ -0,0 +1,31 @@
...
+            'immediately')
+
+        pass_bhvr = tactics.one_touch_pass.OneTouchPass()
+        self.add_subbehavior(pass_bhvr, 'pass')
+
+    def execute_running(self):
+        pass_bhvr = self.subbehavior_with_name('pass')
+
+        if pass_bhvr.is_done_running():
+            self.remove_subbehavior('pass')
```
justbuchanan: Same `restart()` comment here.

justbuchanan:
Looking good!  I made a couple more notes, but I think it's about good to go.  The only major thing left is switching to WindowEvaluator for targeting, which will probably have to wait until @barulicm ports that to C++ and includes shot probabilities.  We can either merge it now and open an issue for switching to WindowEvaluator or just keep this open until then - your call.

soccer/gameplay/plays/testing/test_coordinated_pass.py
jgkamat: Whoops, I think this was from some old experimentation... I'll put it back the way it was.

soccer/gameplay/skills/angle_receive.py
jgkamat: Yup, this works for me! I didn't know that existed, i'll go ahead and replace this logic.

jgkamat:
I'd rather just keep this open until the windowevaluator is ported. This code is pretty slow because of the windowevaluator, and it would be nice to speed it up before it's merged!

justbuchanan:
Sounds good to me.  The next step would be running our plays and seeing which ones could benefit from this.  I'd probably leave the current plays as is, but make new ones that use a onetouch pass (and potentially refactor common things out if it makes sense).

justbuchanan:
Closing since all of this is a part of the comp2015 PR now.

jgkamat:
I never merged this into comp2015, (but I did merge comp2015 into here), should I go ahead and do that now?

justbuchanan:
My bad, I though we put all of this on comp2015.  I guess you were just running this branch at comp.  Let's just reopen this, comp2015 is big enough :)

jgkamat:
Ok, once comp2015 is merged, I'll fix all the conflicts!

PR #344, opened 2015-07-18
Fix inversion of left/right strafing while accounting for wheel angles

### LINKED ISSUE #152: Fix wheel-angle assumption in firmware
The firmware assumes that the robot wheels are angled at 45 degrees to each other, which is a little off.

See the code here for how we currently translate cmd_x, cmd_y, cmd_w to wheel velocities: https://github.com/RoboJackets/robocup-software/blob/master/firmware/robot/cpu/control.c#L27

jgkamat:
Fixes #152, is a replacement for 3320b55, with Y values fixed.

Tested on a 2008 robot. [See this for a better laid out diff](https://github.com/RoboJackets/robocup-software/compare/5843ebd...09b23ca)

### DIFF
--- a/doc/wheel_layout.txt
+++ b/doc/wheel_layout.txt
@@ -1,9 +1,9 @@
 <!-- Wheels
-	/3   0\
+	/0   1\
 		x > >
-	\2   1/
+	\3   2/
 
-	> > > is positive X with robot center being 0,0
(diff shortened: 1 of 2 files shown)


jgkamat:
Not a critical fix, but it is still nice to have this. This could (possibly) wait until after comp.

PR #350, opened 2015-07-19
Add better failure checking for coordinated_pass

jgkamat:
Added better failure checking for the python coordinated_pass

It draws a triangle from the kick point to the receiver, and fails the pass if outside the area. I've set the angle for this triangle to 0.25 rads, which seems to err on keeping the play running when passes are bad. In the future, this angle should be set to what our robots are capable of reaching.

Either way, we should probably test this before merging.

TODO: These changes need to be applied to angle_receive in some way. It should be very similar to what is being done in pass_receive though.

### DIFF
--- a/soccer/gameplay/robocup-py.cpp
+++ b/soccer/gameplay/robocup-py.cpp
@@ -40,7 +40,7 @@ void translateException(NullArgumentException const& e)
 
 /**
  * NOTES FOR WRAPPER FUNCTIONS/METHODS
- * 
+ *
  * Keep in mind that pointer parameters will be be nullptr/NULL if the value
  * from python was None.  Check for this case so that we don't segfault.
  */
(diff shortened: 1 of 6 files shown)


soccer/gameplay/robocup-py.cpp
```diff
@@ -246,6 +246,10 @@ void State_draw_polygon(SystemState *self, boost::python::list points, boost::py
 	self->drawPolygon(ptVec, Color_from_tuple(rgb), QString::fromStdString(layer));
 }
 
+void State_draw_raw_polygon(SystemState *self, Geometry2d::Polygon points, boost::python::tuple rgb, const std::string &layer) {
```
justbuchanan: If we make `points` a pointer, it'll reduce the overhead of having to make a copy.

PR #352, opened 2015-07-19
Fix role requirement assignment for capture

jgkamat:
Fixed capture not selecting closest robot to ball. Helps out two_side_attack


jgkamat:
I also upped the receive_timeout to 3 for coordinated pass

PR #367, opened 2015-07-21
Let coordinatedpass select the closest robot to the true receive point

jgkamat:
This is pretty tiny, but I want to save it until after comp, this is more of a reminder for myself. 

This tells coordinated pass to select the best robot to intercept the ball during a pass, rather than the closest robot to the agreed target position.


jgkamat:
This no longer works because pass_receive uses capture for receiving. This needs to be moved to capture eventually.

justbuchanan:
Good point.  I've added that as todo list item #375.  It would be best if we did it along with some additions to the role requirements system, otherwise I would have just done it now.

PR #380, opened 2015-08-10
Robot2015 Firmware

jjones646:
everything compiles and runs fine for me, so here's it is.


justbuchanan:
Cool, looks like you've got a lot of firmware stuff done!  I'll go through it tonight and tomorrow and see if I have any suggestions or comments.

In the future, let's try to break pull requests up into smaller chunks - it's pretty tough to sort through all 24,697 additions and 2,446 deletions in this one :P

firmware/common2015/drivers/ring-buffer/ring_buffer.h
```diff
@@ -0,0 +1,201 @@
+/*
```
justbuchanan: Any reason for using this over [boost::circular_buffer](http://www.boost.org/doc/libs/1_58_0/doc/html/circular_buffer.html)?  The boost one is nice because it's C++ and we don't have to copy the source files into our repo.

firmware/robot2015/src-fpga/CMakeLists.txt
```diff
@@ -0,0 +1,161 @@
...
+# Remove anything that isn't a directory
+foreach(TMPVAR ${XILINX_VERSION_DIRS})
+    if(NOT((IS_DIRECTORY ${TMPVAR}) AND (EXISTS ${TMPVAR})))
+        list(REMOVE_ITEM XILINX_VERSION_DIRS ${TMPVAR})
+    endif()
+endforeach()
+
+# Find the highester version 
+set(XILINX_ROOT "0.0")
+foreach(TMPVAR ${XILINX_VERSION_DIRS})
```
justbuchanan: For some reason this is choosing the lowest version available, not the highest...  Haven't looked into why yet.  I'm pretty sure this working before, did you change anything related to this part?

firmware/robot2015/src-fpga/CMakeLists.txt
```diff
@@ -0,0 +1,161 @@
...
+)
+
+# Remove anything that isn't a directory
+foreach(TMPVAR ${XILINX_VERSION_DIRS})
+    if(NOT((IS_DIRECTORY ${TMPVAR}) AND (EXISTS ${TMPVAR})))
+        list(REMOVE_ITEM XILINX_VERSION_DIRS ${TMPVAR})
+    endif()
+endforeach()
+
+# Find the highester version
```
justbuchanan: That's a new vocab word haha

justbuchanan:
One thing that might be nice is to setup our fpga stuff to also compile with `iverilog`.  This way we can have our CI setup verify that our FPGA code compiles.  This obviously wouldn't run the code, but checking that it compiles would be a good start.  Not sure how tough it would be to work this in, it may be way more work that it's worth.

firmware/robot2015/src-ctrl/modules/controller/controller.hpp
```diff
@@ -0,0 +1,5 @@
+
+#include "robot-devices.hpp"
+
+
+void Task_Controller(void const*);
```
justbuchanan: What does this function do?  It looks like it's defined in ControllerTaskThread.cpp, so maybe we can drop the empty controller.cpp.

firmware/robot2015/src-fpga/src/qencoder/README.md
```diff
@@ -0,0 +1,3 @@
+# MyHDL
+
```
justbuchanan: What all does the qencoder stuff do?

firmware/robot2015/src-ctrl/scripts/linker/sections_FLASH.ld
```diff
@@ -0,0 +1,235 @@
+/* Linker script for mbed LPC1768 */
```
justbuchanan: What does this linker script do beyond the default one for mbed?  Is this for the parameter feature?

firmware/robot2015/src-fpga/CMakeLists.txt
```diff
@@ -0,0 +1,161 @@
...
+# show what versions were found and what version we're using
+if(AVAILABLE_XILINX_VERSIONS_COUNT EQUAL 0)
+    message(WARNING
+        "Unable to find Xilinx tools.  Install the ISE WebPACK (14.7) to \
+        '${XILINX_SYSTEM_ROOT}' to synthesize Verilog for the FPGA. \
+        http://www.xilinx.com/support/download/index.html/content/xilinx/en/downloadNav/design-tools.html.")
+else()
+    # print out some version info
+    if(${AVAILABLE_XILINX_VERSIONS_COUNT} EQUAL 1)
+        message(STATUS "${AVAILABLE_XILINX_VERSIONS_COUNT} version of Xilinx tools found:")
```
justbuchanan: Should the variables come after the colons?

firmware/robot2015/src-fpga/robocup2015.ucf
```diff
@@ -0,0 +1,27 @@
+# This is the "User Constraints File" for configuring the fpga
```
justbuchanan: What's the difference between robocup2015.ucf and robocup.ucf?

firmware/robot2015/src-kckr/kicker.h
```diff
@@ -0,0 +1,4 @@
+#pragma once
```
justbuchanan: It may be slightly better to just put the includes in the files that need them and get rid of kicker.h.

firmware/common2015/drivers/ring-buffer/ring_buffer.h
jjones646: no, it can go

firmware/robot2015/src-fpga/CMakeLists.txt
jjones646: Hmm. It wasn't picking the highest version for me initially. If there was 14.6 and 14.7, it would pick the 14.6 version. It was sorting it alphabetically based on the full path to the folder. So I changed it to extract the number from each directory and sort the numbers instead. That seemed to work, and I tested the changes quite a bit too. I just checked and I'm using CMake version 3.0.2 if that info will help at all.

firmware/robot2015/src-ctrl/modules/controller/controller.hpp
jjones646: the controller.cpp was where I was going to put functions that will filter the imu readings and/or integrate the LQR stuff into everything somehow. I briefly looked at the LQR stuff, but never really go a grasp at what was going on or how it would be used on-board, so I'm sure you have better judgement than I would on what to do here.

firmware/robot2015/src-fpga/src/qencoder/README.md
jjones646: I was seeing how different it was writing HDL things in python using the MyHDL library. MyHDL is pretty cool and has pretty decent unit testing support, but I have already gotten use to Verilog. So I just left it in everything for reference.

firmware/robot2015/src-ctrl/scripts/linker/sections_FLASH.ld
jjones646: Yeah it was the parameter linker script that never worked. I removed the target from the cmake builds, but I guess I never removed it from here...oops.

firmware/common2015/utils/logger/logger.hpp
```diff
@@ -0,0 +1,71 @@
...
+#define GEN_LOG_STR(x,y)    x ":" TO_STRING(y)
+#define LOG_LINE            GEN_LOG_STR(__BASE_FILE_NAME__, __LINE__)
+#define LOG(lvl, ...)       log(lvl, LOG_LINE, __func__, __VA_ARGS__)
+
+#else   // RJ_LOGGING_EN
+#define LOG(...)            // Nothing
+#endif
+
+
+#define FOREACH_LEVEL(LEVEL)    LEVEL(LOG_LEVEL_START)      \
```
justbuchanan: This is probably a lot more log levels than we'll need.  Unless we have a well-defined purpose for each one, want to just boil it down to INFO, WARN, and ERROR?

firmware/robot2015/src-fpga/CMakeLists.txt
jjones646: No, it prints out a list of all the versions that were found after it.

firmware/common2015/utils/logger/logger.cpp
```diff
@@ -0,0 +1,66 @@
...
+
+/**
+ * Current log level.
+ */
+volatile uint8_t rjLogLevel;
+
+
+/**
+ * [log The system-wide logging interface function. All log messages go through this.]
+ * @param logLevel [The "importance level" of the called log message.]
```
justbuchanan: Let's avoid copying the comments for things from headers into cpp files - they'll inevitably get out of sync if anything changes :)

firmware/common2015/utils/rj-macros.hpp
```diff
@@ -0,0 +1,14 @@
+#pragma once
+
+
+/* Macros for string conversion pass through 2 sets to help out
+ * with variable expansion. Not really sure if this is needed, but
+ * doing it anyways since there's pratically magic happening between
+ * the pre-processor and the log function in 'logger.hpp'.
+ */
+#define STRINGIFY(x)        #x
```
justbuchanan: If these functions do the same thing, let's remove one of them.

firmware/robot2015/src-fpga/robocup2015.ucf
jjones646: robocup2015.ucf is the one Doho used for testing. robocup.ucf is the one from 2011. Then again, I made the new ucf file for the 2015 stuff today, so they'll both be replaced by that one pretty soon. I left the original one there so it would still generate a bitfile correctly during the build.

firmware/common2015/utils/robot-types.hpp
```diff
@@ -0,0 +1,20 @@
+#pragma once
+
+#include "RTP.hpp"
+
```
justbuchanan: We should get rid of the defines for NULL, TRUE, and FALSE here - nullptr, true, and false are all part of the c++ standard.

firmware/robot2015/CMakeLists.txt
```diff
@@ -1 +1,7 @@
-add_subdirectory(cpu)
+# we'll go ahead and set the AVR toolchain cmake file too for consisteny in defining all toolchain setup files
+set(ATTINY_TOOLCHAIN_FILE ${CMAKE_CURRENT_LIST_DIR}/attiny_toolchain.cmake)
+set(ATTINY_TOOLCHAIN_FILE ${ATTINY_TOOLCHAIN_FILE} PARENT_SCOPE)
```
justbuchanan: I think we can delete this line - we shouldn't need to reference the attiny toolchain anywhere above this directory.

jjones646:
@justbuchanan Yeah, I never want to have to go through a merge like this one again...I didn't even know the common2015 folder was there until I merged the branches...that's when things got real. A lot was definitely learned about the do's/don'ts of merging and submitting pull requests for this one haha.

justbuchanan:
Haha for sure.  It looks like you did a fine job of getting the merge figured out, but I bet it was quite a bitch to sort through it all.  The beauty of version control is that you never lose anything for good, but you may have to do some digging :)

firmware/robot2015/src-fpga/CMakeLists.txt
justbuchanan: Oops, missed that, it's all good.

firmware/common2015/utils/logger/logger.hpp
jjones646: The first and last ones are there for bounds checking because you can change the log level through the console now. most of them can be removed after the first complete firmware build, but I think everyone is used somewhere in the code as of now. The INFx ones are all used in low level stuff and won't be needed for a stable set of code, but there's definitely useful for the time being.

firmware/common2015/utils/assert/assert.cpp
```diff
@@ -0,0 +1,13 @@
+#include "assert.hpp"
+
+/**
+ * [assertFail This is called when an assertion fails. NOTE: THIS HAULTS ALL EXECUTIONS!]
```
justbuchanan: Let's move this comment block to the header.

firmware/common2015/utils/rj-macros.hpp
jjones646: One add commas, one doesn't. The STRINGIFY/TO_STRING thing has to be there for the logging stuff because there's a lot of prepreocessor magic happening for that.

firmware/common2015/utils/robot-types.hpp
jjones646: ok, noted. good to know.

firmware/common2015/utils/logger/logger.hpp
justbuchanan: Ok, makes sense.  One thing I've seen done before (especially in python) that's kinda neat is passing a separate `Logger` instance to different modules.  So in main(), we could pass a `Logger` to the radio module, one to the fpga module, etc.  In debug mode, we'd have all of them set the log level to INFO, but in production, we'd probably bump it up to WARN.  The nice thing with this is that logging level isn't global.  That way once radio is a solved problem, you can set it's logger to WARN, but turn the fpga logger to INFO while you work on it.  Not something for this pull request, but maybe something to consider for later.

Also, @barulicm added some things to the logging utils on the master branch including stream-based logging (like cout rather than printf) that I think got lost in the merge.  Take a look at the diff to see the master version and see if you can retrieve the new bits from that.

firmware/robot2015/CMakeLists.txt
jjones646: ok, and I think we might have to do something weird to compile for the attiny441. I haven't looked into at all really, but gcc doesn't know about it when you pass it as a compiler flag.

justbuchanan:
What was the decision on whether to go with CMSIS vs MBED RTOS?  Doesn't matter to me, I'm equally unfamiliar with both haha.  My vote is just for the simpler one unless the other provides features we want/need.

firmware/common2015/utils/rtos-mgmt/helper-funcs.hpp
```diff
@@ -0,0 +1,8 @@
+#pragma once
+
+#include <cstdint>
+#include <string>
+
+void setISRPriorities(void);
```
justbuchanan: These functions could probably use a couple of comments :)

Also, will the ISR priorities be different for the base station and the robot?  If so, we should probably move this out of common2015 or make it more general to handle both.

firmware/common2015/drivers/radio/radio.hpp
```diff
@@ -0,0 +1,14 @@
+#if 0
```
justbuchanan: Are radio.hpp/cpp no longer used?

jjones646:
It was the mbed stuff, but I realized a few more things about the two now. With the mbed library, you call methods against a thread object. For the CMSIS library, you use regular functions and pass it the thread's ID that you want it to be called on. It's the same way for the all of the data queue structures. So if you're trying to make a call within the thread function itself, it's really hard to do with the mbed library unless you pass it a pointer to its own thread object when you start it. With the CMISIS library, you can call functions that will return the thread ID of the currently running thread. So, inside a thread function it's way easier to use the CMISIS stuff since you can call thread functions using your own ID. And, outside a thread function it's easier to use the mbed stuff because creating an object doesn't involve weird macro things going on. I've been using both depending on what's the simplest solution for whatever's needed in each situation.

firmware/common2015/utils/rtos-mgmt/helper-funcs.hpp
jjones646: I haven't really looked into interrupt priorities. I don't expect the base station will need any interrupt priority changes tho.

firmware/common2015/drivers/radio/radio.hpp
jjones646: I think things are to the point where it's safe to say it won't be used again.

firmware/robot2015/src-ctrl/CMakeLists.txt
```diff
@@ -0,0 +1,84 @@
...
+file(GLOB_RECURSE assm_SRC "*.s")
+set_property(SOURCE ${assm_SRC} PROPERTY LANGUAGE C)
+# comment out below line to add assembly files to the build
+unset(assm_SRC)
+
+# firmware source files
+file(GLOB_RECURSE robot2015_SRC "*.cpp")
+
+# Exclude FreeRTOS stuff if it happens to be a directory here...
+string(REGEX REPLACE ";?[^;]+/FreeRTOS[V]?[0-9\\.]+/[^;]+;?" "${0}" robot2015_SRC "${robot2015_SRC}")
```
justbuchanan: This can probably go if we're not using the FreeRTOS stuff anymore.

firmware/robot2015/src-ctrl/modules/motors/motors.hpp
```diff
@@ -0,0 +1,74 @@
...
+	FETLC_OC
+	*/
+	bool encOK;
+	bool hallOK;
+	std::array<uint16_t, 2> drvStatus;
+};
+
+
+struct motor_t {
+	motorVel_t 	targetVel;
```
justbuchanan: The LQR motion control stuff I have setup calculates a voltage to send to each motor rather than a velocity.  I'll run you through the controls once we're all back at Tech (and we can change it around if needed), but the basic idea is that if the controller takes into account the whole robot's acceleration, it can do a better job if it calculates voltages than velocities.

justbuchanan: Also, what's the adjVel field?

firmware/robot2015/src-ctrl/modules/ControllerTaskThread.cpp
```diff
@@ -0,0 +1,73 @@
...
+
+	if (threadID != NULL)
+		threadPriority  = osThreadGetPriority(threadID);
+	else
+		threadPriority = (osPriority)NULL;
+
+	float gyroVals[3] = { 0 };
+	float accelVals[3] = { 0 };
+
+	MPU6050 imu(RJ_I2C_BUS);
```
justbuchanan: It may be better if we initialize the MPU in main and pass it in here.  That way we can run the imu and get readings out of it without having to run the control loop.

justbuchanan:
Ok that makes sense.  I guess you can't really get a pointer to a Thread from inside those functions unless you pass it through from the top.

justbuchanan:
A few other notes:

* A lot of the stuff in src-ctrl/drivers seems like it's copied partially from somewhere, but most of them #include files that aren't present.  It may be best to leave these out for now, then just copy in stuff as needed.  Also, I haven't looked into it that much, but I think some of those are already pulled down as part of the mbed library through the CMake ExternalProject stuff.  Browse through third_party/mbed_lib_build_tools if you want to see what all's there.
* There are a lot of headers that have reference the GPL, but our project is actually licensed under Apache v2.  We have a LICENSE file at the root of the project, so we can remove any license references from the headers.
* Not something that needs to be a part of this PR, but a thought for later.  If we're doing competition vs development firmware builds, we could add targets for each in the main makefile so you could do `make robot2015` or `make robot2015-debug` to easily switch between the two.  Then throughout the project, we can consolidate all the debug stuff inside of `#ifdef DEBUG`, similar to how it is now.

firmware/robot2015/src-ctrl/CMakeLists.txt
jjones646: agreed

firmware/robot2015/src-ctrl/modules/motors/motors.hpp
jjones646: I have the mpu-6050 getting consistent readings. Each read is triggered in the Controller_Task thread, so getting accel values for x & y overall just involves filtering those readings a little bit. If a few things in the MPU6050 class are changed for the I2C stuff, it's very possible to do 5 or 6 updates to the FPGA between each radio packet using imu readings. The targetVel/adjVel being a bad naming scheme for distinguishing between velocities/voltages computed when there is no imu. More thought needs to go into the motor struct fields, but it just hasn't happened yet. Mostly because of the earlier comment about the LQR stuff, and I'm still working on the SPI interface for the FPGA.

firmware/robot2015/src-ctrl/modules/ControllerTaskThread.cpp
jjones646: ahhh. yes, yes.

jjones646:
- The copied stuff comes from [LPCOpen](https://www.lpcware.com/content/nxpfile/lpcopen-platform) and is mostly there for reference on how to do the low level things when needed.

- The original thought was to have different builds because I thought the console stuff would be too much overhead to keep the real-time response that we need. But ever since I added changing the console thread's priority level based on if there is a command to run or not, I'm thinking that 2 different builds may not be needed. I've done some stability testing since then, and you can constantly print stuff out to the console without anything else being lagged down since that change. So standby, by It's looking like having a single build will work out ok.

firmware/robot2015/src-ctrl/drivers/mcp23017/mcp23017.cpp
```diff
@@ -0,0 +1,281 @@
+/*  MCP23017 library for Arduino
```
justbuchanan: @jjones646, take a look at PR #236 - it's been sitting there a while unmerged b/c it's missing a couple small things, but I've seen it work on a board.  If you've checked that this one works well, we can go ahead and delete the PR for the other one.

One thing to note though is that this file is GPL.  Technically we're not supposed to use GPL'd stuff in an Apache'd project, although I don't think it's really an issue for us here.  http://www.apache.org/licenses/GPL-compatibility.html.

justbuchanan:
* Ok gotcha.  I'll have to take a look at LPCOpen than and get a better idea for how it all works.  Do you think it might be good to more all those to a specific "reference" folder or something?  Any unsuspecting new guys that wander into our repo will probably see it all under src-ctrl and assume it's all in use.  It might be good to separate it out from the drivers we're actually using for the firmware.
* That would be awesome if we can get away with just one build!  There's a decent chance that when everything's up and running, we'll need a lot more of the processing power than is being used now and it might be worth it to disable any non-essential logging when running for real.  But we can figure that out later.

justbuchanan:
One more minor thing - the firmware build targets need to be added to run/.gitignore.

jjones646:
All of the targets can be found with "rj-*.bin" or "rj-*.nib". I honestly can't really remember if that's part of this pull request, but I did go through and make sure all names would work with the local filesystem's name restrictions.

jjones646:
A reference folder would be good. I don't know much about the LPCOpen stuff myself. I just know its published by the same company that makes the microcontroller, so its a good outline for how to initialize and access all of the hardware features.

firmware/robot2015/src-ctrl/drivers/mcp23017/mcp23017.cpp
jjones646: I haven't seen this yet. I'll have time to look at it tonight. The control board does have an interrupt line from the ioexpander connected to it. It should be fine as long as that will work.

justbuchanan:
@jjones646, do you want to keep this PR open or close it in favor of the new firmware branch you're working on?

PR #382, opened 2015-08-12
Setup code reformatting/checkstyling.  Closes #161

### LINKED ISSUE #161: Automatic code formatting
[Uncrustify](http://uncrustify.sourceforge.net/) is a program for automatically formatting code according to a config file.  We write the config with the desired style parameters, then run uncrustify on each C/C++ file in the codebase.  The first time we run it, it will make a ridiculous number of changes, but after that it should only catch occasional things.  I honestly don't care *that* much about code style/formatting, but it makes it easier to read if it's consistent.  Once we agree on a format and get the config file written, code style becomes an objective check against the decided-upon style guide and the fact that it's all automated makes it really easy.

A few todos:
- [ ] write the uncrustify config file so that everyone's happy
- [ ] do initial codebase reformatting - this will happen after comp this year after all branches have re-joined master otherwise the merge conflicts will be ridiculous
- [ ] setup uncrustify as a git pre-commit hook?
- [ ] add an uncrustify checkstyle to travis-ci

I've started this on the `uncrustify` branch although, as mentioned above, it won't be merged into master until after comp this summer.  You can check out the branch and run `make pretty` to see how it works though.

justbuchanan:
This PR adds two new commands/targets to our makefile: `make pretty` reformats code and `make checkstyle` checks to see if code is in accordance with our configs and gives a nonzero exit code if something's off.  This applies to all C/C++ and Python code and uses a separate tool for each.  The main purpose of this is to get consistency across source files.  Having indentation and everything else different from file to file is really obnoxious.  It also prevents most cases where a new feature is added or a bug is fixed and the commit contains several style changes in totally irrelevant parts of the codebase.  The super nice thing with this new setup is that everything is automatic - you don't necessarily have to know specifics on how things should be formatted and you don't have to take the time to manually do it yourself.  Just run a command before you send a PR and you're good to go!

C++ code is formatted with a tool called [clang-format](http://clang.llvm.org/docs/ClangFormat.html), which is a part of the clang compiler project.  Style is specified in the `.clang-format` file at the root of the repo.  Python code is formatted with [yapf](https://github.com/google/yapf), which is a tool from Google.  We can add a config file for this too, but it defaults to [pep8](https://www.python.org/dev/peps/pep-0008/), which is sort of the Python community standard.

The `.clang-format` config right now is basically just the default Google one, which is pretty solid (I'm not biased...), but we can change any of the parameters if you guys have suggestions.  Let the spaces vs tabs war begin!

The makefile is setup to only run the formatter on files that differ from robojackets/master, which is **much** faster than running on every single file in the repo.  If you want to try it out and run it on all files, just copy the command out of the makefile and remove the "--diffbase=robojackets/master" part.

We'll wait to merge this until after the `comp2015` PR since it will cause a shitload of merge conflicts.


jgkamat:
Ubuntu dosent have `clang-format` in the repositories, but it does have `clang-format-3.5`. I don't know why they want to freeze the version for this :confused: 

Even worse, the command on ubuntu is `clang-format-3.5` not `clang-format` and that causes a bunch of errors when running stylize. You can work around this by creating a sketchy symlink (`sudo ln -s `which clang-format-3.5` /usr/bin/clang-format`) but I would rather find a way to work this out properly within stylize (I'm not sure how to do this though). :cry: 

It seems to work as intended when these two hacks are done though.

justbuchanan:
Goddammit Ubuntu...  Not really sure what the best workaround is here.  Stylize just looks for a program named `clang-format` and runs it.  We could modify it, but I don't know what we'd want to make it do differently - there's not really a good way to know the name of the program though.  The symlink thing is a little sketchy, but I sorta like it the best.  Thoughts?

jgkamat:
Oh, I just realized you wrote stylize :smile: 

Is there a way to make clang format a part of the stylize package when installing it? That way, when you `pip3 install stylize` not only does it install stylize itself, it also copies over the [clang-format](https://packagecontrol.io/packages/Clang%20Format) executable, and uses that copied over one (rather than the one on the path). If that dosent work, maybe we could have stylize download the binary on it's first run? I don't know too much about python packages.

We could also check individually for versions, but that's kind of hacky. It would probably be best to have stylize work on ubuntu machines without the symlink though...

justbuchanan:
Haha glad to see I tricked you into thinking it was a legit package :grinning:.  It's surprisingly (and awesomely) easy to register things for use with pip.

This is possible, but not great unfortunately.  To do this, the setup.py script for stylize would have to clone down the clang-format repo and probably do a `make install`.  It's a fairly big codebase though, so it takes a helluva long time (I've tried it).  It would also cause conflicts with the package version of clang-format on everything except Ubuntu unless it also put checks in place for it.  Overall I think it would be kind of a mess.

The nice thing about the symlink is that if we install a specific package version through apt-get, the symlink really can't go wrong.  The only downside would be if you uninstalled clang-format and the symlink was left dangling or if ubuntu gets its shit together and names the binary right in the future.  I think the symlink option is better than the alternative though.

jgkamat:
Hmm, I was thinking about downloading [this binary file](http://107.170.192.246:8001/llvm3.6/ubuntu14.04/clang-format), it seems to work on fedora/ubuntu (after a chmod +x). I don't think we would need any other dependencies but I'm not sure...

justbuchanan:
Ahh ok, that makes more sense.  That has the downside though that it's not cross-platform.  Right now you can install stylize on osx and Linux just the same.  It wouldn't be a big deal, but it would have to check for os first.  The other thing is that this still has potential to conflict with the package manager.  On Arch, clang-format is a package and I'd  rather have it come through `pacman` than be bundled with the pip package.

If you want to mess with the stylize setup, clone the repo and run `sudo python3 setup.py install`.

jgkamat:
Hmm, looks like an error when installing stylize
```
Downloading/unpacking stylize>=0.0.9 (from -r /home/developer/robocup-software/util/requirements3.txt (line 9))
  Downloading stylize-0.0.9.tar.gz
  Running setup.py (path:/tmp/pip_build_root/stylize/setup.py) egg_info for package stylize
      File "<string>", line 17, in <module>
      File "/tmp/pip_build_root/stylize/setup.py", line 10, in <module>
        long_description=open('readme.rst', 'r').read(),
      File "/usr/lib/python3.4/encodings/ascii.py", line 26, in decode
        return codecs.ascii_decode(input, self.errors)[0]
    UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 125: ordinal not in range(128)
    Complete output from command python setup.py egg_info:

  File "<string>", line 17, in <module>

  File "/tmp/pip_build_root/stylize/setup.py", line 10, in <module>

    long_description=open('readme.rst', 'r').read(),

  File "/usr/lib/python3.4/encodings/ascii.py", line 26, in decode

    return codecs.ascii_decode(input, self.errors)[0]

UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 125: ordinal not in range(128)
```

Ubuntu-setup errors should fail the build before anything happens, so thats weird...

jgkamat:
Circle will continue to try to use the broken baseimage though because of caching, you can work around this by adding a newline somewhere.
I think I know why ubuntu-setup passed, but I'll fix that later.

justbuchanan:
That's odd...  running `pip3 install stylize` works fine for me on my computer.  I'll see if I can figure out what's up.

jgkamat:
:cry: 
```
Downloading/unpacking stylize>=0.0.11 (from -r /home/developer/robocup-software/util/requirements3.txt (line 9))
  Downloading stylize-0.0.11.tar.gz
  Running setup.py (path:/tmp/pip_build_root/stylize/setup.py) egg_info for package stylize
      File "<string>", line 17, in <module>
      File "/tmp/pip_build_root/stylize/setup.py", line 10, in <module>
        long_description=open('readme.rst', 'r').read(),
      File "/usr/lib/python3.4/encodings/ascii.py", line 26, in decode
        return codecs.ascii_decode(input, self.errors)[0]
    UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 143: ordinal not in range(128)
    Complete output from command python setup.py egg_info:

  File "<string>", line 17, in <module>

  File "/tmp/pip_build_root/stylize/setup.py", line 10, in <module>

    long_description=open('readme.rst', 'r').read(),

  File "/usr/lib/python3.4/encodings/ascii.py", line 26, in decode

    return codecs.ascii_decode(input, self.errors)[0]

UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 143: ordinal not in range(128)
```

jgkamat:
Hmm, I just tried this on a clean 14.04 VM, and that worked fine, but on a clean dockerimage it does not. I'm not sure whats going on, but I'll look into the problem.

You can run `docker run jgkamat/stylizetest pip3 install stylize` to test the issue on docker.

jgkamat:
Woah, those are really nice crosses in the logs! :+1:

justbuchanan:
Indeed.  And someday soon they'll be checkmarks!

ashaw596:
I'm getting an error on mac when I try make pretty
```
stylize --diffbase=master --exclude_dirs build third_party
make: stylize: No such file or directory
make: *** [pretty] Err
```

justbuchanan:
`stylize` is the tool that does the formatting - it's installed through `ubuntu-setup` or you can do `pip3 install stylize`.

ashaw596:
I'm on mac. Shouldn't we add it to osx-setup?

jgkamat:
It looks like it should be added in osx-setup, have you run it on this branch?

ashaw596:
ah, its failing because it says cmake is already installed.

ashaw596:
Thats weird. ok nvm. It wouldn't automatically update cmake through homebrew.
Should we add brew upgrade to update the brew packages?

ashaw596:
Also, does anyone else like 4 spaces instead of 2 spaces? This is more compact, but its also weirding me out. =)

jgkamat:
+1 on 4 spaces
If we used tabs then we could work in whatever width we wanted :smile:

joshhting:
Also +1 on 4 spaces

justbuchanan:
I'll probably be merging comp2015 later today, then merging this one afterwards.  If anyone has any more suggestions for style configs, it would be best to do those now so we can avoid major reformattings later.

jgkamat:
Please merge onetouchpass (and as many other things) before this
I don't want merge conflicts :cry:

justbuchanan:
Haha sounds good.  I expect a lot less merge conflicts with python code since we do a fairly good job there of actually following style guidelines.

What's the status of onetouchpass?  Should we merge it now and just open a ticket for improving the positioning/aiming later?

jgkamat:
I think it should be good right now, the main problem is testing it to aim properly (and improving coordinated pass). I'll fix the merge conflicts when this is finalized...

justbuchanan:
Sounds good.  Go ahead and open a ticket for that part and I'll merge the PR.

justbuchanan:
This website has a nifty tool for showing exactly what each of the clang-format configs do:

http://clangformat.com/

justbuchanan:
I set it to exclude checkstyle/reformat for firmware/{common2015, robot2015} for now since there's a large firmware branch that would get wrecked by this.  Other than that, I think it's good to go!

justbuchanan:
I think I found a small bug in `clang-format` that's causing inconsistent formatting between different versions.  This doesn't affect people on ubuntu since that's what we're using for checkstyling.  I've noticed the bug in version 3.7 and 3.8.  I filed a ticket here, so hopefully they'll take a look at it: https://llvm.org/bugs/show_bug.cgi?id=24594.

Other than that, I think this is good to go!

PR #386, opened 2015-08-18
switched from make to ninja. closes #204

### LINKED ISSUE #204: Setup Ninja build system
CMake is sort of an intermediate build system in that it doesn't build code directly, but outputs makefiles or other build files that can then be run to build the code.  CMake uses the `make` "Generator" by default, but you can choose other ones by passing it a value using the "-G" flag.  You can see more about CMake generators on their documentation page: http://www.cmake.org/cmake/help/v3.0/manual/cmake-generators.7.html.

Ninja is a build system sort of similar to `make`, but does a better job of parallelizing the build and will compile the project much faster.  I tried Ninja on my computer and it ran almost twice as fast.  This isn't exactly a fair comparison because I ran it against a serial make run, which can be parallelized by passing the "-j" (jobs) flag to make.  Ninja still benchmarks faster overall though because it's a newer system and designed from the ground up for speed and parallelization.

Switching to Ninja on my system (Arch linux) was pretty easy and just involved replacing "make" in our makefile with "ninja" and adding "-GNinja" to all of our CMake calls.  I ran into some trouble when trying this on Ubuntu though, so I can't put this onto master yet.  If someone can figure this out and get it to work reliably on Ubuntu, it'll give us faster builds overall.

justbuchanan:


jgkamat:
Needs `ninja-build` added to the ubuntu packages, otherwise looking good!

`make robot2015` seems to be failing though (on ubuntu) :cry:

justbuchanan:
For some reason `ninja-build` was in there already, so it's good.  There's some sort of dependency issue with googletest that I'm trying to figure out - I think that's what's failing the build.

jgkamat:
Whops my bad, I just looked at the diff :smile:

justbuchanan:
I pulled up the log and this is currently running out of memory and failing :(.  We'll try it after the move to circleci and see if it works - circleci gives us 4G ram vs the 3G with travis.

justbuchanan:
This currently works for everything except firmware builds.  For firmware, we override the usual compiler and settings through a custom "toolchain".  For some reason this all works fine with `make`, but with `ninja` it tries to use the main system compiler to build all of our MBED stuff, which definitely doesn't work...

jgkamat:
Google test is :cry: right now (the link wasn't updated here)

ashaw596:
=(

justbuchanan:
Just added `git submodule sync` before the update in circle.yml, that should fix it.

jgkamat:
It dosent seem to be working, thats weird...
I tried the same thing on my branch :/

jgkamat:
Whats weird is that if we use the SSH feature of circleci, everything works after a sync. Huh...

justbuchanan:
Yeah I was just noticing the same thing...  SSH works perfectly...

jgkamat:
(Don't forget to cancel you'r ssh when your done with them (they hold up a build for 30 min otherwise)

justbuchanan:
Will do :+1: 

Btw, not that it affects this particular problem, bit it might be a good idea to add submodule sync here: https://github.com/jgkamat/DoCIF/blob/dcb216060a3e84429013ed9ab9608930ce68902f/util/maketest.sh#L91

jgkamat:
I'm away from my computer, but could you try submodule update, then a sync, then an update?

justbuchanan:
That worked!  I have no idea why though...

jgkamat:
Probably update scans the gitmodules file, then scan copies that into the git dir, then update copies it in? I don't know, I probably should read up on submodules internals some more

justbuchanan:
Yeah not sure.  It feels like a bug to me, you should never have to run submodule update or sync twice.

Also, I'm getting a permissions error now about the ccache dir.  Not sure if it's related to ninja or the new DoCIF stuff.  Thoughts?

~~~{.sh}
FAILED: ccache /usr/bin/c++    -std=c++11 -g -Werror=return-type -Werror=delete-non-virtual-dtor -Werror=uninitialized -fPIC -Isrc/googletest/include -isystem /usr/include/eigen3 -I../common -Icommon -MMD -MT common/protobuf/CMakeFiles/proto_messages.dir/LogFrame.pb.cc.o -MF common/protobuf/CMakeFiles/proto_messages.dir/LogFrame.pb.cc.o.d -o common/protobuf/CMakeFiles/proto_messages.dir/LogFrame.pb.cc.o -c common/protobuf/LogFrame.pb.cc
ccache: failed to create /home/developer/.ccache/tmp (Permission denied)
~~~

justbuchanan:
Nevermind, I think it's all working now!

jgkamat:
Huh, what was the issue before? (Sorry I wan't able to see this)

justbuchanan:
See this build, it looks like the issue happened on the DoCIF branch too.

https://circleci.com/gh/RoboJackets/robocup-software/599

jgkamat:
Huh, I'm going to hold of on merging the DoCIF PR until I figure out why this happened. Looks like the edge case where the .ccache directory is not present at the start of the build.

justbuchanan:
:+1:.  It seemed like both cases where I saw this, it resolved itself when I just ran the same build again.

justbuchanan:
I think the ninja stuff is good to go unless anyone sees any issues.

jgkamat:
`make robot2015` seems to fail after a clean, other than that this looks good!
Ninja seems to be using ~3-4 gb of ram, is that normal?

justbuchanan:
Hmm, that's odd.  What kind of error do you get?  `make robot2015` works fine for me and for circleci after a clean.  The mbed libraries shouldn't be built during a regular `make`, are you sure they were?

As for memory, I'm not really sure.  It should definitely use more than make since it's doing a lot more at once, but that does seem kinda high.  Not sure what we can do about it though.

jgkamat:
Whoops, I forgot to sync my submodules again. robot2015 looks good. 
I guess I can live with the memory overhead, but we probably need to tell people who run in VM's about that.

justbuchanan:
Ok cool.  I looked into the memory usage some and it looks like the only thing we can change to make it lighter on memory is to pass a smaller number to the `-j` flag.  By default it runs as many builds in parallel as you have cores, but you can set it manually to a smaller number to decrease memory usage.  I tried it on my VM that I only gave 2GB to and it seemed to work fine (without passing the `-j` flag), it was just a little slower.

ashaw596:
Is the memory overhead really an issue? I thought it was suppose to intelligently selected the j value based on system load. Or was it just based on available cores.

justbuchanan:
By default Ninja runs as many jobs in parallel as there are cores.  I don't think it'll be an issue, but as Jay said it's possible to run out of memory on a VM depending on how it's configured.

jgkamat:
On my host machine, running make w/o cache and google chrome running freezes my computer, but If i free memory by killing google chrome, it works fine. I had ~3GB free space when I did this, I think...

justbuchanan:
How much ram do you have overall?

jgkamat:
6GB, I think. Most people usually give 4gb or less to their VM's, so this might be a problem? Firefox is way less memory intensive though. I'll show you what I'm seeing on tuesday.

ashaw596:
Its fine on my 4 gb vm and ist still fine when I switched to 2 gb. I do have a large swap space though I believe.

PR #389, opened 2015-08-20
made soccer's min size a little smaller.  fixes #388

### LINKED ISSUE #388: Make the soccer app smaller in osx
Right now its bigger than my screen on a retina macbook 13

justbuchanan:
@ashaw596, can you run this on your MacBook and see how it looks?


justbuchanan:
@ashaw596 can you try this out and let me know if it works on your MacBoook?

ashaw596:
I don't think its smaller

ashaw596:
Actually, its just a bit too wide now.

justbuchanan:
I set the `maximumSize` property for each of the internal ref buttons and added stretchy spacers between them, which allows soccer to much smaller (in width).  I think this is good to go now.  @ashaw596, can you run it on your MacBook and verify?

PR #390, opened 2015-08-20
Switch from travis to circleci.

### LINKED ISSUE #371: Add CircleCI as an alternative to Travis (Use docker for CI)
We've been having issues with Travis due to:
* The speed of the builds. Travis firstly takes 10 minutes to start up and then we spend a majority of our time getting the system to install the packages we need.
* Travis uses Ubuntu 12.04. We could use docker to emulate 14.04 (it will use the older kernel though), but Travis dosen't support using root inside of docker containers, and we can't sacrifice sudo access at this point.

If we switch to CircleCi, with docker, it will solve these two problems of our CI, by
* Giving us access to 14.04 base machines (and any future systems, should we switch)
* Letting us 'cache' dockerimages. The mechanism to do this seems kind of strange, and this needs to be investigated more, but worst case, we can cache these images on the public docker hub and download them from there during build. Letting us 'cache' dockerimages. The mechanism to do this seems kind of strange, and this needs to be investigated more, but worst case, we can cache these images on the public docker hub and download them from there during build.

The only downside of moving to CircleCi is that CircleCi has a limit of 4 builds running at a time.

Caching builds is going to be a little tricky. My proposed solution for this is to make a sha hash of the util folder/ubuntu-setup and associated files and tag all 'baseimages' using that sha. That way, if we update ubuntu-setup, we will push a new updated 'baseimage'. In addition, when we actually do CI, we can tag images with the git sha of the commit, and push the compiled versions with those tags to the dockerhub, so we can have a full history of all our compiled containers.

tl,dr;
- [x] Run CI in docker on CircleCi (keep our travis builds running as is for now).
- [x] Cache ubuntu-setup images on the docker hub, and tag them with a sha.
- ~~Cache the last run ubuntu-setup image on circleCI servers, to reduce download time (optional)~~ Not worth it.
- [x] Transfer ccache data to docker containers
- ~~Tag built images with their commit sha and push them to the dockerhub.~~ Will be done in automated build
- [x] Move the updated documentation script to circle ci using their secrets setup.
- [x] Add proper artifact links to build outputs.

In the future, we can look into using docker as an optional build platform.

jgkamat:
I wan't to make sure that all the nitpicky stuff runs when it should, but things are looking good now. Everything thats supposed to run when its on master is not tested yet though.

This breaks the documentation updating in travis (as the doxygen versions were conflicting). It now uses the one installed in ubuntu-setup. I haven't tested this though (as it only runs on master), but it looks good to me.

Quick summary of what this does
1. Hash all ubuntu-setup files. Check to see if an image with this sha is found, if so, download it, if not, build one, and push it to robojackets/robocup-baseimage. (Additionally tag with master if on master)
2. Use the image generated/pulled in step 1 to run tests. Ccache and the robojackets folder are mounted to this dependency container, and environment variables are passed through. Tests are run using the maketest script which updates status tokens and saves logs.
3. If (2) completes successfully, then then everything else will pass. The ccache directory will be stored on the circleci servers to be retrieved on next build.

If something in (1) fails, then all the custom checks will still be 'pending' but the circleci check will be a failure. If circleci is red, then something is wrong!

Also the style check is always green as of now.

Closes #371


justbuchanan:
Looks great!  I'll look through it all after lunch and see if I have any suggestions for you.  Can't wait to ditch Travis and have Ubuntu 14 for builds :smile:

util/docker/maketest.sh
```diff
@@ -0,0 +1,73 @@
+#!/bin/bash
+# Runs tests and reports on them to github
+
+DIR=$(cd $(dirname $0) ; pwd -P)
+source ${DIR}/docker_common.sh
+
+# Get sha sum
+USER="jgkamat"
```
jgkamat: Not sure if this username should be changed to georgeburdell, but it seems to be working now?

justbuchanan: Yeah we might as well use georgeburdell.  I actually entered an environment variable GH_USER (set to georgeburdell) that you can use for this.

util/docker/maketest.sh
```diff
@@ -0,0 +1,73 @@
...
+cd ${ROBOCUP_ROOT}
+
+# Clean
+make clean
+git submodule update --init
+sudo chown -R `whoami`:`whoami` ${HOME}/.ccache
+
+start_pending
+
+make | tee "${ARTIFACT_DIR}/makeoutput.txt"
```
justbuchanan: It might be nice to put this type of stuff in a "ci task" function, then just call it at the bottom for each of the things we test.  It could take in a shortname (for logfile name and status context), a description, and a command to run.

util/docker/maketest.sh
```diff
@@ -0,0 +1,73 @@
...
+
+# Get sha sum
+USER="jgkamat"
+SUCCESS=true
+LINK_PREFIX="https://circle-artifacts.com/gh/RoboJackets/robocup-software/$CIRCLE_BUILD_NUM/artifacts/0$CIRCLE_ARTIFACTS/"
+ARTIFACT_DIR="/tmp/build_artifacts"
+
+start_pending() {
+    curl -u $USER:$TOKEN -X POST https://api.github.com/repos/robojackets/robocup-software/statuses/${SHA_SUM} -H "Content-Type: application/json" -d '{"state":"pending", "description": "A check for compiling", "context": "circle/compile", "target_url": '"\"${LINK_PREFIX}makeoutput.txt\"}"
+    curl -u $USER:$TOKEN -X POST https://api.github.com/repos/robojackets/robocup-software/statuses/${SHA_SUM} -H "Content-Type: application/json" -d '{"state":"pending", "description": "A check for cpp tests", "context": "circle/test-cpp", "target_url": '"\"${LINK_PREFIX}testcppoutput.txt\"}"
```
justbuchanan: `test-cpp` runs both `test-soccer` and `test-firmware`.  The way you have it covers both, but if you change it to run them as two separate items, we can get a status for each.

util/docker/maketest.sh
justbuchanan: On a similar note, it might be a little cleaner to have a function for posting the status and then you can just have one curl statement in here.

util/docker/baseimage/Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+    echo "developer:x:${uid}:${gid}:Developer,,,:/home/developer:/bin/bash" >> /etc/passwd && \
+    echo "developer:x:${uid}:" >> /etc/group && \
+    echo "developer ALL=(ALL) NOPASSWD: ALL" > /etc/sudoers.d/developer && \
+    chmod 0440 /etc/sudoers.d/developer && \
+    chown ${uid}:${gid} -R /home/developer
+
+USER developer
+ENV HOME /home/developer
+
+# clone and run ubuntu-setup to shave some time off. This will be run again in the robocup-software to compensate for changes to the setup script.
```
justbuchanan: ?

autoupdate-docs-travis.sh
```diff
@@ -8,16 +8,6 @@ set -e
```
justbuchanan: We can rename this to just autoupdate-docs.sh and change the comment at the top since it's not on travis anymore.

autoupdate-docs-travis.sh
```diff
@@ -8,16 +8,6 @@ set -e
 
 
 if [ "$(git rev-parse master)" = "$TRAVIS_COMMIT" ]; then
```
justbuchanan: Since it's specified in the circle config that this only runs on master, you can drop this if statement.

circle.yml
```diff
@@ -0,0 +1,26 @@
...
+
+dependencies:
+  pre:
+      # set build status to pending on start
+    - ./util/docker/maketest.sh --pending ${GH_TOKEN}
+      # Login to docker
+    - docker login -e $DOCKER_EMAIL -u $DOCKER_USER -p $DOCKER_PASS
+    - ./util/docker/baseimage/buildbaseimage.sh
+      # Actually the test step, but we want to cache it so it can go here.
+    - ./util/docker/circle/buildcircle.sh
```
justbuchanan: Does caching not work if we run this under `test`?  If possible, it'd be nice to have the main circleci status only go green if everything passes.

circle.yml
```diff
@@ -0,0 +1,26 @@
...
+  cache_directories:
+    - ~/.ccache
+
+test:
+  override:
+    - sh -c true
+
+deployment:
+  github:
+    branch: master
```
justbuchanan: If you want to test out the documentation updating, you can change this to your branch for now to have it run and see that it works.  It's fine if our docs are non-master for a few days :)

makefile
```diff
@@ -25,7 +25,7 @@ pylint:
 	cd soccer && pylint -E gameplay
 
 clean:
-	cd build && make $(MAKE_FLAGS) clean
+	cd build && make $(MAKE_FLAGS) clean || true
```
justbuchanan: :+1: I should have done this a while ago.

justbuchanan:
A couple more thoughts:

* Not something to do now, but a thought for later: it may be nice to make this easily portable to other repos.  I've been trying to push IARRC and IGVC to setup something for CI since they currently do nothing.  Eventually we may be able to have a shared CI setup that all the teams can use (and that I may steal for personal projects :smile:).
* When you get around to it, let's update the Ci info on the docs page to describe this setup.  You can probably just copy the summary above.
* Also, try doing a global search for the word "travis" in our repo and replace it with circle.

.dockerignore
```diff
@@ -0,0 +1 @@
+.gitignore
```
justbuchanan: Is this file still needed if we're not including the code repo in the built image?

util/docker/maketest.sh
jgkamat: :+1: this is a really good idea

.dockerignore
jgkamat: The baseimage is still built when the ubuntu setup script changes, so this is needed.

jgkamat:
Hmm, I'll think about that, perhaps I could create a small repo to easily setup dockerized CI. I'll have to think about the structure of that though...
I'll keep working on this throughout the day (And responding to these comments, thanks! :smile:)

justbuchanan:
Sounds good.  It'd be a good bit of work to make it really portable, but I could definitely see it being useful.  It could be a neat project for later.

circle.yml
jgkamat: Caching is only done after the deploy step, so this needs to go here, unfortunately.

The main circeci step will only be green if this passes, if this fails, the execution will stop (sh -c test will never be run, because of the failure)

autoupdate-docs-travis.sh
jgkamat: :+1: its looking good as of now, see f5cfe6b34c722e830

.travis.yml
```diff
@@ -1,24 +0,0 @@
-language: cpp
```
jgkamat: Should I keep this file as is for now? I saw you were trying to use it for osx testing...

justbuchanan: I think it's fine if you delete it, we can re-add it if we merge the OS X testing stuff.  I just found out that caching isn't available on travis for public OS X builds, so each build takes about 45 minutes...  Until they fix that, it's pretty worthless.

jgkamat: That really sucks :cry: 
Maybe we could still run it just for verification? 45 minutes is better than nothing (so we can catch osx mistakes). Your call though...

jgkamat:
@justbuchanan should I make the style check run `true` for now, or `make checkstyle`. We could switch it from true -> make checkstyle in #382, if it gets merged after.

Other than that, I gave this another look and everything looks good! I'm ready to have this merged whenever you are ready! :smiley:

util/docker/maketest.sh
```diff
@@ -0,0 +1,86 @@
...
+
+# A function to run a command. Takes in a command name, shortname and description.
+ci_task() {
+    CMD="$1"
+    SHORTNAME="$2"
+    DESCRIPTION="$3"
+
+    SHORTNAMES+=("${SHORTNAME}")
+
+    if [ "$PENDING" = "true" ]; then
```
justbuchanan: Instead of this, it might be better to just check for PENDING below and only make calls to `ci_task` if it's not true.  That way this function doesn't have to know about the pending stuff.

util/docker/maketest.sh
```diff
@@ -0,0 +1,86 @@
...
+# Clean
+make clean
+git submodule update --init
+sudo chown -R `whoami`:`whoami` ${HOME}/.ccache
+
+ci_task 'make' 'compile' 'A check to see if the code compiles'
+ci_task 'make test-soccer' 'test-soccer' 'A check to see if soccer tests pass'
+ci_task 'make robot2015' 'firmware' 'A check to see if firmware works'
+ci_task 'make test-firmware' 'test-firmware' 'A check to see if firmware tests pass'
+ci_task 'true' 'style' 'A check to see if style passes'
```
justbuchanan: Want to just comment out the one for style now and we'll fix it up once we get that side of things figured out?

justbuchanan:
Cool, looks good!  I made a couple minor notes, but other than that the only thing would be to update [this docs page](https://github.com/RoboJackets/robocup-software/blob/master/doc/UnitTestsAndCI.md) with a short description of how this works.  Once you do that, feel free to go ahead and merge it. :+1:

justbuchanan:
And sounds good with the OS X stuff.  It sucks that travis takes 45 minutes, but I agree that it's better than nothing.  I'll go ahead and merge that one after this - it'll be nice to at least have something to check OS X builds since we want to have that as a fully-supported platform.  I'm hoping the travis guys will get around to enabling caching on OS X before too long and build will go much quicker :)

jgkamat:
Sure thing, I'll update that and merge this. Always worried when CI stuff is changed though :sweat: 

I'll keep checking on it for the next couple days and we'll see how it goes.

justbuchanan:
No worries, if CI broke for a bit it wouldn't be the end of the world :smile:.

jgkamat:
Theres a small issue with the build (it looks like somethings not being happy on the master branch, I'll fix it and push it straight to master)

justbuchanan:
Cool, sounds good :+1:

jgkamat:
Everything looks good now, api autoupdated, and its green
At least we got to test the main status indicator...

Whats interesting is that circleci caches per branch, which is interesting (but what we want)...

justbuchanan:
Awesome!

Yeah I think the caching goes by branch, but all new branches just copy the master branch's current cache to use.  This should work out well though.

jgkamat:
Hmm, it looks like that same bit of code that was supposed to push a 'master' tag to the dockerhub when on master is being super slow. It's instant on my computer, so checking to see if things are pushed before pushing them must have been a later docker version than circleci has...

Do you mind me pushing again to master to just remove that bit entirely, it seems to be adding ~4 minutes to the build :cry:

justbuchanan:
No worries, go for it!

jgkamat:
Looking much better now, I'll keep checking on the builds though! Let me know if you find any problems...

justbuchanan:
I noticed some large blobs of json showing up in the main circle log from the github status api - want to redirect those to /dev/null or something to keep the main log a little cleaner?

jgkamat:
hmm, I'll just silence those.

I also noticed this bug thats causing all the firmware tests to fail (when stylize fails, it causes python2 dependencies to stop working). I'll make ubuntu-setup properly fail like it should, and do this at the same time.

jgkamat:
I pushed that update to master. Now when ubuntu-setup fails, the main circleci job will go red (but the rest of the jobs will remain pending unfortunately). It's much better than pushing broken images though.

It also silences the github statuses and forces a rebuild of ubuntu-setup.

I'm checking out for the night, but i'll look into the rest of these errors tomorrow. I don't think any of them are related to the CI setup though... (other pip installs work, I'm not really sure about ninja, I don't think cmake/gcc versions on 14.04 can support c++14).

justbuchanan:
Sounds good.  I'll mess around and see what I can fix on the other branches that don't build yet.  I'm quite confident the stylize issue is my fault haha.

PR #391, opened 2015-08-21
Use travis ci to test builds on OS X

justbuchanan:
This PR replaces the current travis config with one that runs on OS X and only builds soccer (no unit tests, firmware, checkstyle, etc).  It would be nice to use Circle for OS X as well, but I don't think we can have it do a Linux build and an OS X build for the same repo.

Right now it starts out by removing all homebrew packages, then installs just the ones we need.  This is a workaround for the fact that homebrew gives a nonzero exit code when you tell it to install a package that has a newer version available, but an older one is installed.  We'll need a smarter way to resolve this b/c the build currently takes ~45 minutes...  We may also be able to cache installed packages - haven't checked yet.


justbuchanan:
I'll go ahead and merge this now.  Since it takes 45 minutes to run, I think it's fine to ignore the travis check when looking at pull requests, etc - just make sure the circleci stuff passes.

jgkamat:
Not sure about this, but how would you feel about having the osx tests run only on master? It would make pull requests go green soon (but have the disadvantage that we catch osx breakage until its on master). Otherwise, we can just tell everyone during training to ignore travis checks unless they fail.

justbuchanan:
That's a good idea.  I think the only time OS X builds will break is when we change dependencies, so we can just be a little more careful with those.  Having PR statuses go green in 3 minutes instead of 45 would be a lot better. I'll go ahead and change this.

PR #393, opened 2015-08-25
I fixed the stylize install issue.

jgkamat:
It was because docker dosent use UTF by default. I made it use UTF and things are happy now.
#382

@justbuchanan


justbuchanan:
:+1:

PR #401, opened 2015-08-26
Test for PR's from seperate forks

jgkamat:
DON'T MERGE THIS. IT WILL BE CLOSED LATER.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -1,13 +1,17 @@
 machine:
   services:
     - docker
+  environment:
+    GITHUB_HAS_GREAT_SECURITY: ba71cd719dd509a0ea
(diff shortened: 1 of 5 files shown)


jgkamat:
@justbuchanan these tests are coming off after you're travis only on master commit but it's running travis. No big deal, but I thought I would let you know.

justbuchanan:
Not sure what the deal with travis running on all branches is - I whitelisted master in the travis config, so it should only do that.  I opened a ticket with their support team, so we'll see what they say.  I just disabled travis for pull requests in their web console, but I'm still seeing builds for non-PR non-master branches...

PR #402, opened 2015-08-26
Attempt to allow PR's from seperate forks

jgkamat:
Github apparently is really smart and invalidates api tokens pushed to GH, so I 'worked around that'.

Other than that, this looks good. See #401 for it in action.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -26,4 +26,5 @@ deployment:
     branch: master
     commands:
       - ./util/docker/circle/apidocs.sh
+      - ./util/docker/baseimage/pushbaseimage.sh
 
--- a/circle.yml
+++ b/circle.yml
@@ -2,6 +2,11 @@ machine:
   services:
     - docker
   environment:
+    # In order to allow circleci to push ci statuses to GitHub for pull requests
+    # from external forks, we have to store an access token.  GitHub invalidates
+    # API tokens pushed to repositories for security reasons, so we split it in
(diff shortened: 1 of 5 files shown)


justbuchanan:
Looks good to me.  How GITHUB_HAS_GREAT_SECURITY token work?

jgkamat:
Github actually checks to see if your token is pushed to github, and invalidates it if it sees it. So I just split it in two (and then github can't invalidate the token)

justbuchanan:
Ahh ok, makes sense.  Want to put a comment in the circle file describing that?

justbuchanan:
How do they invalidate it if it's just entered into the circle environment variables, but never committed/pushed to GitHub?

jgkamat:
circle.yml is on github (thats the way to add environmental variables that arent blocked on seperate forks). Normal env. vars are blocked, but things in circle.yml are public and added in every scenario.

justbuchanan:
Makes sense, I forgot circle blocks environment variables on separate forks.

jgkamat:
I added back the pushing to the master tag as part of this PR, but this time, in the deployment step, so it doesn't slow down the results of the builds.

justbuchanan:
Nice, I like it!  Feel free to merge this if it's ready.

jgkamat:
This made master red because [docker hub is down right now](https://status.docker.com/). If this continues we might have to force this to always return true. But this is the first time I've seen this happen...
I'll rebuild this once it goes back up

PR #411, opened 2015-08-30
replaced by-ref usage of Point with by-value.  closes #381

### LINKED ISSUE #381: Replace by-reference usage of Geometry2d::Point with by-value
It turns out that by-value is actually faster, which makes sense since points are very small.  See this gist for proof: https://gist.github.com/justbuchanan/a1594aead5fda4e6e526.

justbuchanan:


justbuchanan:
@ashaw596, let me know when you get a chance to look over this.

ashaw596:
Looks good.

PR #417, opened 2015-08-31
Refactored InterpolatedPath to use MotionInstant and made Path.evaluate() return an optional instant

justbuchanan:
* renamed PathPlanner -> SingleRobotPathPlanner
* InterpolatedPath now has a vector of MotionInstant objects rather than separate vectors of position and velocity
* Path.evaluate() returns a boost::optional<MotionInstant> rather than having an output variable
* Removed unused methods: InterpolatedPath.{getPoint(), startFrom()}

### DIFF
--- a/soccer/Robot.cpp
+++ b/soccer/Robot.cpp
@@ -18,6 +18,7 @@
 
 using namespace std;
 using namespace Geometry2d;
+using Planning::MotionInstant;
 
 /** thresholds for avoidance of opponents - either a normal (large) or an
  * approach (small)*/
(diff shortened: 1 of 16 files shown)


justbuchanan:
@ashaw596, I noticed that `InterpolatedPath`'s `getPoint()` and `startFrom()` methods aren't being used anywhere.  Can we get rid of them or do you think they'll have a use later?

ashaw596:
Maybe? They've been in the class forever. That entire class is a bit messy since it is a point path class modified to work with velocities. Most of it is their just so it will work with RRT planner.

ashaw596:
Yeah, I don't think those 2 are going to be used ever.

ashaw596:
PathTest wasn't refactored btw.

ashaw596:
Nice refactor btw.

justbuchanan:
Cool, thanks.  I removed those two methods and fixed up the tests, so let me know if you see any issues.

ashaw596:
looks good. I think there's still something not compiling in the tests though.

PR #424, opened 2015-09-01
Refactored replanning and created IndependentMultiRobotPathPlanner

### LINKED ISSUE #415: Implement basic version of IndependentMultiRobotPlanner

justbuchanan:
Should be merged after #421.

* Planning is now run by the Processor
* Processor has a single MultiRobotPathPlanner that handles planning for all robots
* Robots no longer store a reference to their Planner
* IndependentMultiRobotPathPlanner is a subclass of MultiRobotPathPlanner that runs a SingleRobotPathPlanner for each robot.
* Field obstacles are drawn to the "Obstacles" layer, which is hidden by default
* Refactored InterpolatedPath to store a single vector of waypoints, rather than separate vectors for positions, velocities, and times
* SingleRobotPathPlanner objects now handle their own replanning.  `run()` is called every iteration and it's up to the individual planner to decide whether to return the same path as before or make a new one.

closes #415

### DIFF
--- a/soccer/Robot.hpp
+++ b/soccer/Robot.hpp
@@ -412,13 +412,6 @@ public:
     double distanceToChipLanding(int chipPower);
     uint8_t chipPowerForDistance(double distance);
 
-    Planning::SingleRobotPathPlanner* pathPlanner() { return _planner.get(); }
-
-    void setPathPlanner(
-        std::unique_ptr<Planning::SingleRobotPathPlanner> planner) {
-        _planner = std::move(planner);
-    }
-
     void setPath(std::unique_ptr<Planning::Path> path);
 
 protected:
@@ -435,9 +428,6 @@ protected:
 
     Planning::MotionCommand _motionCommand;
     MotionConstraints _motionConstraints;
-
-    std::unique_ptr<Planning::SingleRobotPathPlanner>
-        _planner;  /// single-robot RRT planner
 
     std::unique_ptr<Planning::Path> _path;  /// latest path
 
--- a/soccer/Robot.hpp
+++ b/soccer/Robot.hpp
@@ -438,10 +438,6 @@ protected:
 
     std::unique_ptr<Planning::Path> _path;  /// latest path
 
-    /// whenever the constraints for the robot path are changed, this is set
-    /// to true to trigger a replan
-    bool _pathInvalidated;
-
     /**
      * Creates a set of obstacles from a given robot team mask,
      * where mask values < 0 create no obstacle, and larger values
--- a/soccer/Robot.hpp
+++ b/soccer/Robot.hpp
@@ -171,7 +171,7 @@ public:
      * Saving the pointer may lead to seg faults as it may be deleted by the
      * Robot who owns it.
      */
-    const Planning::Path* path() const { return _path.get(); }
+    std::unique_ptr<Planning::Path>& path() { return _path; }
 
     /// clears old radioTx stuff, resets robot debug text, and clears local
     /// obstacles
@@ -313,7 +313,8 @@ public:
     }
     void clearLocalObstacles() { _local_obstacles.clear(); }
 
-    Geometry2d::ShapeSet collectAllObstacles(const Geometry2d::ShapeSet& globalObstacles);
+    Geometry2d::ShapeSet collectAllObstacles(
+        const Geometry2d::ShapeSet& globalObstacles);
 
     void approachAllOpponents(bool enable = true);
     void avoidAllOpponents(bool enable = true);
@@ -358,12 +359,6 @@ public:
      * carrier.
      */
     void shieldFromTeammates(float radius);
-
-    /**
-     * Replans the path if needed.
-     * Sets some parameters on the path.
-     */
-    void replanIfNeeded(const Geometry2d::ShapeSet& globalObstacles);
 
     /**
      * status evaluations for choosing robots in behaviors - combines multiple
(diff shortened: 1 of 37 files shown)


ashaw596:
I think some weird stuff is going on with the planning/replanning for capture right now.

soccer/Robot.hpp
```diff
@@ -467,9 +442,9 @@ class OurRobot : public Robot {
      * or opp from _state
      */
     template <class ROBOT>
```
ashaw596: Do we know why this was a template?

soccer/planning/Path.hpp
```diff
@@ -15,8 +15,11 @@ namespace Planning {
  */
 class Path {
 public:
+    Path() : _startTime(0) {}
     virtual ~Path() {}
 
+    virtual bool valid() const = 0;
```
ashaw596: Why do we need this?

soccer/planning/MultiRobotPathPlanner.hpp
```diff
@@ -0,0 +1,43 @@
...
+#include <planning/Path.hpp>
+
+#include <map>
+#include <memory>
+
+namespace Planning {
+
+/// The PlanRequest encapsulates all information that the planner needs to know
+/// about an individual robot in order to generate a path for it.
+struct PlanRequest {
```
ashaw596: Do we want this in this file?

soccer/planning/RRTPlanner.cpp
```diff
@@ -1,52 +1,136 @@
...
+            path->waypoints.emplace_back(
+                MotionInstant(start.pos, Geometry2d::Point()), 0);
+        }
+        return unique_ptr<Path>(path);
+    } else {
+        return prevPath;
+    }
+}
 
+Geometry2d::Point RRTPlanner::findNonBlockedGoal(
```
ashaw596: Can't we just use the closed nonblocked point nearest to the start point?

soccer/planning/RRTPlanner.cpp
```diff
@@ -1,52 +1,136 @@
...
+        ((float)(timestamp() - prevPath->startTime())) * TimestampToSecs +
+        1.0f / 60.0f;
+
+    MotionInstant target;
+    boost::optional<MotionInstant> optTarget = prevPath->evaluate(timeIntoPath);
+    if (optTarget) {
+        target = *optTarget;
+    } else {
+        // We went off the end of the path, so use the end for calculations.
+        target = *prevPath->destination();
```
ashaw596: destination could return boost::none

soccer/planning/SingleRobotPathPlanner.cpp
```diff
@@ -0,0 +1,36 @@
...
+ConfigDouble* SingleRobotPathPlanner::_replanTimeout;
+
+void SingleRobotPathPlanner::createConfiguration(Configuration* cfg) {
+    _replanTimeout = new ConfigDouble(cfg, "PathPlanner/replanTimeout", 5);
+    _goalChangeThreshold =
+        new ConfigDouble(cfg, "PathPlanner/goalChangeThreshold", 0.025);
+}
+
+std::unique_ptr<Planning::SingleRobotPathPlanner> PlannerForCommandType(
+    Planning::MotionCommand::CommandType type) {
```
ashaw596: SOmething more efficient would be nice here.

soccer/Robot.hpp
justbuchanan: This function is called with a vector of OurRobot* and again with a vector of OpponentRobot*.  Unfortunately you can't just cast a vector of a subclass to a vector of the superclass :cry:

soccer/planning/Path.hpp
justbuchanan: Good point, I think checking that `destination` is non-null is just as good.  Just removed it.

soccer/planning/MultiRobotPathPlanner.hpp
justbuchanan: If `MultiRobotPathPlanner` is the only place it's used, I think it's probably fine to keep it here.  If we want to make `SingleRobotPathPlanner` take this in rather than having separate arguments, then it would probably be best in its own file.

soccer/planning/RRTPlanner.cpp
justbuchanan: Not sure what you mean. This is used when the requested move target is inside an obstacle.  In that case, we try to find the closest non-blocked point to the one requested.

Handling a starting position that's inside an obstacle is a separate case.  The RRT handles this already at each step.  See `FixedStepTree.extend()`.

soccer/planning/RRTPlanner.cpp
justbuchanan: The check at the top for path validity ensures that it has a destination.  I've changed that to check that destination is non-null now that valid() is gone.

soccer/planning/SingleRobotPathPlanner.cpp
justbuchanan: What did you have in mind?

ashaw596: I mean not remaking it each time. Not important right now though

ashaw596:
Can we use the same branch?

soccer/planning/SingleRobotPathPlanner.cpp
justbuchanan: Ahh gotcha.  Well this method is intended to make a new one each time - I think it should be the user of this function's responsibility to re-use the same planner over and over again as needed.  Right now IndependentMultiRobotPathPlanner only calls this when the command type changes.

justbuchanan:
Just fixed the replanning issue you were seeing via abbbafc, I think it all works now.

justbuchanan:
We could use the same branch, but I'd like to try to avoid having a long-running "pathplanner" branch that we keep adding to.  It'd be nice if possible to implement a new feature / do a refactor, then merge that back to master and just work from there.

Let me know if you see any other issues/questions with this PR.  If not, I think it's about ready to merge.

ashaw596:
OK fine. You're right its fine. I still think we should explore the other type of workflow (I guess mostly just because I don't think it will work with more people). But you made this change so freaking fast. Most part looks good.

PR #437, opened 2015-09-09
Refactored path endpoints

justbuchanan:
* renamed Path.destination() -> Path.end()
* added Path.start()
* added Path.valid()
* no longer using boost::optional for endpoints

### DIFF
--- a/soccer/planning/CompositePath.cpp
+++ b/soccer/planning/CompositePath.cpp
@@ -93,11 +93,18 @@ void CompositePath::draw(SystemState* const state, const QColor& color,
 
 float CompositePath::getDuration() const { return duration; }
 
-boost::optional<MotionInstant> CompositePath::destination() const {
+MotionInstant CompositePath::start() const {
     if (paths.empty()) {
-        return boost::none;
+        throw InvalidPathException();
     }
-    return paths.back()->destination();
+    return paths.front()->start();
+}
+
+MotionInstant CompositePath::end() const {
+    if (paths.empty()) {
+        throw InvalidPathException();
+    }
+    return paths.back()->end();
 }
 
 unique_ptr<Path> CompositePath::subPath(float startTime, float endTime) const {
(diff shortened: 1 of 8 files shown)


soccer/planning/CompositePath.cpp
```diff
@@ -93,11 +93,18 @@ void CompositePath::draw(SystemState* const state, const QColor& color,
 
 float CompositePath::getDuration() const { return duration; }
 
-boost::optional<MotionInstant> CompositePath::destination() const {
+MotionInstant CompositePath::start() const {
     if (paths.empty()) {
-        return boost::none;
```
ashaw596: So I'm guess infinite paths aren't a thing anymore? Not that we really every had a real use case for it.

soccer/planning/Path.hpp
```diff
@@ -18,6 +25,9 @@ class Path {
     Path() : _startTime(0) {}
     virtual ~Path() {}
 
+    /// Returns true if the path is non-empty
+    virtual bool valid() const { return true; }
```
ashaw596: I don't really understand this. When are we expecting to have invalid paths?

soccer/planning/CompositePath.cpp
justbuchanan: Infinite paths make things kinda weird... unless you can see a good use case for them, I'd say we restrict them to be finite.

soccer/planning/Path.hpp
justbuchanan: Right now it's possible to construct a CompositePath or an InterpolatedPath that's empty.  The `valid()` check just ensures that the path is an actual path with content.

ashaw596: I mean its possible. But we should really never have a pathPlanner return an empty path.
We can just add checks in the pathplanner and return nullptr if something really irrecoverable happens.
 On another note, we could make Path stuff mostly immutable since we have the subpath and compositePath stuff setup.

justbuchanan: That's fair.  Every path planner should return a valid path, so we shouldn't have to check elsewhere if it's valid.  I'll remove the `valid()`.

justbuchanan: Done.

ashaw596:
Maybe need more docs on start and end. Otherwise looks good.

Issue #436, opened 2015-09-09
Create obstacle escape planner


justbuchanan:
This will be run for robots that don't have an actual motion command.  @ashaw596, when do you think you can have the motion command refactor (#414) done?  Plugging the escape planner into the system will require a "None" type for MotionCommand.

PR #439, opened 2015-09-09
Added TargetVelPathPlanner

### LINKED ISSUE #427: Implement TargetVelPathPlanner
Rather than having MotionControl handle velocity commands separately, we want to have a planner that returns a path corresponding to the target velocity.

justbuchanan:
* TargetVelPathPlanner now creates a Path for target world velocity commands, as used in `Capture` and other behaviors.
* TargetVelPathPlanner respects obstacles, so it should no longer be possible to accidentally crash or enter goal zones when using target world velocity.
    * removed backoff state from Capture that was put in place to avoid issues with obstacles and target world velocity
* Implemented Path.draw() so that subclasses don't need to unless they have good reason.
* Refactored TrapezoidalPath a bit (underscores for ivars)
* Moved common replanning checks from RRTPlanner to SingleRobotPathPlanner
* Unit tests for TargetVelPathPlanner

TODO:
- [x] Use binary search instead of linear search for endpoint
- [x] Config value for vel change threshold
- [x] Handle case where start point is inside an obstacle - done in PR #443 via 2d96736749ab69d06dbedf26e967e97112753c43
- [x] A couple tests?

closes #427


common/Geometry2d/ShapeSet.hpp
```diff
@@ -25,7 +25,11 @@ class ShapeSet {
     std::vector<std::shared_ptr<Shape>> shapes() { return _shapes; }
     const std::vector<std::shared_ptr<Shape>> shapes() const { return _shapes; }
 
-    void add(std::shared_ptr<Shape> shape) { _shapes.push_back(shape); }
+    void add(std::shared_ptr<Shape> shape) {
+        if (!shape)
+            throw std::runtime_error("Attempt to add null shape to ShapeSet");
```
ashaw596: awee. I like debug throw. Not important though.

ashaw596: could also be an assert.

soccer/motion/MotionControl.cpp
```diff
@@ -76,12 +75,16 @@ void MotionControl::run() {
...
+        // deg/sec^2
+        // 	0.1 ,								// time into
+        // path
         // 	_robot->angleVel,					// start speed
-        // 	0,									// final speed
+        // 	0,									// final
+        // speed
         // 	targetAngle,
-        // 	targetW 							// ignored
+        // 	targetW 							//
```
ashaw596: This formatting is so weird.

ashaw596: Maybe just go ahead and delete it.

common/Geometry2d/ShapeSet.hpp
justbuchanan: Done.

soccer/motion/MotionControl.cpp
justbuchanan: Done.

justbuchanan:
I think this is about done.  The only un-handled case that I can see is if the robot starts out in an obstacle.  Right now it will just plan a path through it, which isn't what we want.  I think it would be best there to just run the EscapeObstaclesPathPlanner, which isn't implemented yet (see #436).  I think we can leave it as is, then update it once the EscapeObstaclesPathPlanner is done.

justbuchanan:
@ashaw596 do you have any suggestions for this PR or is it good to merge?

soccer/Robot.cpp
```diff
@@ -458,7 +458,8 @@ Geometry2d::ShapeSet OurRobot::collectAllObstacles(
     if (_state->ball.valid) {
         // _state->drawShape(ball_obs, Qt::gray,
         //                   QString("ball_obstacles_%1").arg(shell()));
-        fullObstacles.add(createBallObstacle());
+        auto ballObs = createBallObstacle();
+        if (ballObs) fullObstacles.add(ballObs);
```
ashaw596: We could just have an empty shapeset instead of handling these nullptrs everywhere.

soccer/planning/Path.cpp
```diff
@@ -0,0 +1,36 @@
...
+    Packet::DebugRobotPath* dbg = state->logFrame->add_debug_robot_paths();
+    dbg->set_layer(state->findDebugLayer(layer));
+
+    // Get the closest step size to a desired value that is divisible into the
+    // duration
+    const float duration = getDuration();
+    const float desiredStep =
+        0.25;  // draw the path by interpolating every x seconds
+    const float segmentCount = ceilf(duration / desiredStep);
+    const float step = duration / segmentCount;
```
ashaw596: Isn't it simpler if this logic is put in one for loop?

soccer/Robot.cpp
justbuchanan: True, although in this case the createBallObstacle() method will only ever return one obstacle, so I don't know if a ShapeSet makes as much sense.

soccer/planning/Path.cpp
justbuchanan: Are you referring to how it handles the endpoint separately from the rest?  That's to avoid a potential issue where we may go slightly off the end of the path due to float rounding error.

ashaw596: No i Mean couldn't this be replaced by a
for(float i=0; i<duration; i+=desriedStep) {
addPoint(time=i.....)
}
addPoint(time=paht->end())

ashaw596:
What controls replanning for getting off the path now? Since the worldvel path should replan pretty actively for that.

soccer/planning/Path.cpp
justbuchanan: I think if I do it that way, there's potential to double-add the endpoint due to float rounding imperfection.  We could get around that by doing `i < duration-0.00001` or similar, but it's a little less clean.

I did refactor the common bits a bit.

justbuchanan:
See [this line](https://github.com/RoboJackets/robocup-software/pull/439/files#diff-df1e174998a752fb786ef108f491e16fR64).  Replanning due to tracking error is now in SingleRobotPathPlanner since it's shared between the subclasses.

soccer/planning/Path.cpp
ashaw596: I mean no matter which way we do that, its the same problem especially with the 0.25*numsegments and the ceilf. 
if duration = 0.250000000001 and the ceilf gets ceilf(1.00000000001), segment count = 2
Then it does t=0.25 and t=0.25000000000001 which would probs be the same in evaluate 
Either way floats are meh. Its just as likely either way.

justbuchanan: The way I have it, 0.25 is the initial *desired* step.  The step that's chosen is dependent on the result of `ceilf()`.  In your example, segmentCount would be 2, and `step` would get set to 0.125.  The loop would then run for i=0 and i=1, corresponding to t=0 and t=0.125.  Then the end would be added afterwards and there's no floating point ambiguity.

justbuchanan: I think that `roundf()` would actually be slightly better than `ceilf()` so we stick as close as possible to the desired step.  I'll change it in a sec.

ashaw596:
Ok looks good.

ashaw596:
Should _replan_threshold be lowered for velplanner though so it acts more like would be expected for direct vel commands?

justbuchanan:
We could, although the places that use the world vel commands calculate the target velocity based on ball placement or something similar.  Letting the robot "drift" by having a smaller replan threshold would probably be less desirable since it would have to change the target vel direction to be towards the ball again.

PR #443, opened 2015-09-10
Added EscapeObstaclesPathPlanner

### LINKED ISSUE #436: Create obstacle escape planner

justbuchanan:
* EscapeObstaclesPathPlanner returns the start position if it's obstacle-free, otherwise uses an RRT to find an unblocked point nearby.
* Fixed shot obstacle creation in _kick.py
* Added drawing of local obstacles (defaults to off - must be checked in the Layers tab)
* Added "None" command type - will need to be changed once the new command stuff is in place (#444)
* Added `operator<<()` for `ShapeSet`

Closes #436.

TODO:
- [x] Unit tests
- [x] Before and after video of shot-blocking
- [x] make goal change threshold a config value
- [x] make rrt step size a config value


justbuchanan:
Here's a before and after comparison showing robots moving out of the way with the EscapeObstaclesPathPlanner vs blocking shots.

Before: http://sendvid.com/yor2l9s6
After: http://sendvid.com/m1zm2qjl

justbuchanan:
@ashaw596, I think this one's about good to go as well.

Btw, check out the before and after video clips.  cc @ashaw596 @joshhting

PR #444, opened 2015-09-10
Refactor motion commands/constraints

ashaw596:
Possibly ready to merge

finished some initial implementation. 
Created MotionCommands, RotationCommands.
Have seperate RotationConstraints and MotionConstraints Kept that for now so that editing the constraints survives changing the motionCommand.

MotionCommand and RotationCommand are implemented polymorphicly right now.
I put the enums back in and doing an if check on enum and then a static cast.

Still not sure about the design for a couple things

unique_ptr vs shared_ptr for the commands
I think that maybe the commands should be cast before they are sent to the pathPlanners since they really shouldn't take any commands they can't accept.
Then a centralized pathPlanner plan function doesn't work but there has to be some immediary anyways which knows which planner takes what commands anyways.

#TODO: 
 - [x] Tons of merge conflicts with your pull requests
 - [x] Pick an implementation for the getCommandType enum. Its implemented with a virtual method in Rotation and a variable in the base class in MotionCommand.
 - [x] Figure out exactly how pivot should be handled. Its a Motion Command right now. It also has rotational component though so another refactor of Pathplanner stuff and Motioncontrol.
 - [x] Clean up stuff.
 - [x] Move the RotationConstraint Configs to the right class file.

Check up on the boost/make_unique thing. Seems to not compile in circle.


soccer/CMakeLists.txt
```diff
@@ -23,7 +23,7 @@ set(ROBOCUP_LIB_SRC
     "motion/TrapezoidalMotion.cpp"
     "NewRefereeModule.cpp"
     "planning/CompositePath.cpp"
-    "planning/DirectTargetPathPlanner.cpp"
+    #"planning/DirectTargetPathPlanner.cpp"
```
justbuchanan: Don't we want to keep DirectTargetPathPlanner for now until we find something better?  It'd be best to avoid degrading goalie performance in the meantime.

soccer/Robot.cpp
```diff
@@ -162,8 +161,10 @@ void OurRobot::resetForNextIteration() {
 }
 
 void OurRobot::resetMotionConstraints() {
+    _rotationConstraints = Planning::RotationConstraints();
     _motionConstraints = MotionConstraints();
-    _motionCommand = Planning::MotionCommand();
+    _motionCommand = make_unique<Planning::WorldVelTargetCommand>(Geometry2d::Point(0,0));
```
justbuchanan: It would be nice if MotionCommand had a "None" type that we could use here.

soccer/CMakeLists.txt
ashaw596: Then we have to implement a DirectTargetCommand, Planner and stuff. We can do it but everything else needs to be merged together first. I'd really prefer not to keep pivot working right now because its really screwy right now in motionControl since we don't have a pivotKick pathplanner or really support for angle control in our pathplanning layer yet.

soccer/Robot.cpp
ashaw596: I thought about that but would it really be different from a worldvel command of nothing. Theres enough madness with the types of commands already. I used nullptr before but then thats more cases that we need to do nullptr checks.

soccer/CMakeLists.txt
justbuchanan: Well we can't merge something into master if it breaks pivot, that's a pretty essential motion :/

Can't we refactor the commands in a way that allows them to still be interpreted as before (MotionControl handles pivot, not a special planner)?  It's ok if this PR doesn't bring the structure of our commands entirely to how we want to have them in the end.  We may have to modify the types, etc a little bit later once the pieces are in-place.

ashaw596: I mean pivot works. I fixed that. Its just MotionControl is still the mess its always been with our rotation stuff.

soccer/Robot.cpp
justbuchanan: You're right that "None" is sort of functionally equivalent to a world vel of zero, but I think having them separate makes it a little cleaner.  If we wanted to print out the inputs to the path-planning layer, it would be a lot nicer to see "None" or `nullptr` for a command and know that the gameplay system never set it to anything rather than seeing a worldvel of zero.  Would it be tough to use `nullptr` and not have a "None" type?

soccer/CMakeLists.txt
justbuchanan: Ok cool.  Yeah I think we should just leave MotionControl stuff alone at the moment, then we can get to it later once we have the more path-oriented stuff in place.

ashaw596: Its just this stuff isn't going to merge with the other changes without alot of cajoling. I would really like use to just work on one branch for a while for these pathplanning changes.  We're still basically rewriting everything with every merge.

soccer/planning/MotionConstraints.hpp
```diff
@@ -57,7 +31,8 @@ struct MotionConstraints {
     static void createConfiguration(Configuration* cfg);
     static ConfigDouble* _max_acceleration;
     static ConfigDouble* _max_speed;
-    static ConfigDouble* _max_angle_speed;
+    static ConfigDouble* _max_rotation_speed;
```
justbuchanan: Let's move these to the RotationConstraints class.

ashaw596: It's in the todo

soccer/CMakeLists.txt
justbuchanan: Yeah it unfortunately requires some merging since we're messing with the same code.  If we work on the same branch, we have the same conflicts, we just end up doing them in smaller pieces.  There aren't really less conflicts though.  If you want, we can wait to merge the TargetVelPlanner and EscapeObstaclesPlanner until after this PR and I'll just handle the merge conflicts that show up with those planners.

ashaw596:
meh yeah merge conflicts. Its k, i can merge mine after. Its just annoying because we have to do so many compatability tweaks to keep the main stuff working in master before we merge and then tweak them again. I still really want to try out multiple branching workflow with multiple pull request layers sometime. It could be interesting for groups working at the same time.

ashaw596:
@justbuchanan Ok. I merged it with the VelPlanner stuff. It thinks this is about ready to merge. Something about googletests submodule is breaking the build I think? I don't get how submodules work.

ashaw596:
Ok, I think that mostly everything.

justbuchanan:
Make sure to merge master into this first.

common/Utils.hpp
```diff
@@ -13,6 +13,13 @@
 #include <typeinfo>
 #include <vector>
 #include <QtWidgets>
+#include <memory>
+namespace std {
```
justbuchanan: Hey, can we avoid putting things in the `std` namespace?  We can have our own `make_unique` until we get c++14 working, but it's kinda bad practice to put things into `std`.

ashaw596: But this way it's easy replaceable when we finally upgrade to c++14. None of the code needs to change except that we delete this function.

ashaw596:
NOOOOO more merge conflicts =(((

soccer/Robot.cpp
```diff
@@ -163,63 +165,48 @@ void OurRobot::stop() {
...
-    // reset other conflicting motion commands
-    _motionConstraints.faceTarget = boost::none;
-    _motionConstraints.pivotTarget = boost::none;
-
-    *_cmdText << "angleVelocity(" << targetAngleVel << ")" << endl;
-}
-
 void OurRobot::pivot(Geometry2d::Point pivotTarget) {
-    _motionConstraints.pivotTarget = pivotTarget;
+    _rotationCommand = nullptr;
```
justbuchanan: Should this set the rotationCommand to face the ball?

soccer/planning/RotationConstraints.cpp
```diff
@@ -0,0 +1,15 @@
...
+#include "planning/RotationConstraints.hpp"
+
+REGISTER_CONFIGURABLE(RotationConstraints);
+
+ConfigDouble* RotationConstraints::_max_rotation_speed;
+ConfigDouble* RotationConstraints::_max_rotation_acceleration;
+
+void RotationConstraints::createConfiguration(Configuration* cfg) {
+    _max_rotation_speed = new ConfigDouble(
+        cfg, "MotionConstraints/Max Rotation Speed", 2.0);  //  radians / second
```
justbuchanan: Should these configs be in the RotationConstraints "namespace"?

justbuchanan:
The way you have this now with constraints and commands being separate, does that mean we'll have to pass those four items + obstacles, etc to path planning?  It seems like kind of a lot of stuff to pass around.  What was the issue with merging commands and constraints together so that MotionCommand has a maxSpeed and maxAccel property and we could get rid of MotionConstraints and AngleConstraints?

ashaw596:
But whats the point of combining the constraints and commands together? The same amount of data that gets passed around no matter what. Also I have them seperately so that the constraints aren't overwritten time a new command is sent. Also, passing around those 20 pointers isn't really our performance constraint.

ashaw596:
@justbuchanan Ok merged with master. Also, should the goalie really be respecting the shooting obstacle when its trying to block shots?

justbuchanan:
Yeah I guess it's not that bad.  I wasn't worried about performance at all with passing around extra pointers, I just think it makes it more readable to not have so many parameters to a function.  Merging the command/constraint pairs together seems like a good way to do that since they're really not used separately.  Not a big deal though.

I wouldn't worry about the shot obstacle thing for now.  It doesn't affect actual gameplay since we won't be shooting on our goal, but it does make our demo play pretty strange.  We should fix it so the demos are good, but I'd leave it for now.

Btw, the style check isn't passing on this PR.

soccer/Processor.cpp
```diff
@@ -435,11 +435,11 @@ void Processor::run() {
                 Geometry2d::ShapeSet fullObstacles =
                     r->collectAllObstacles(globalObstaclesForBot);
 
-                Planning::PlanRequest request(
-                    Planning::MotionInstant(r->pos, r->vel), r->motionCommand(),
-                    r->motionConstraints(), std::move(r->path()),
+                requests[r->shell()] = Planning::PlanRequest(
+                    Planning::MotionInstant(r->pos, r->vel),
+                    r->motionCommand()->clone(), r->motionConstraints(),
```
justbuchanan: Why the `clone()` here?

soccer/Robot.cpp
```diff
@@ -153,8 +152,10 @@ void OurRobot::resetForNextIteration() {
 }
 
 void OurRobot::resetMotionConstraints() {
+    _rotationConstraints = RotationConstraints();
     _motionConstraints = MotionConstraints();
-    _motionCommand = Planning::MotionCommand();
+    _motionCommand = make_unique<Planning::EmptyCommand>();
```
justbuchanan: Should we handle the empty/none case the same for both angle and motion commands?  Maybe they should both be nullptr or both be EmptyCommand.

soccer/Robot.cpp
```diff
@@ -163,63 +164,48 @@ void OurRobot::stop() {
...
-             << goal.y << ")" << std::endl;
-
-    _motionCommand.setPathTarget(MotionInstant(goal, endVelocity));
+        cout << " in OurRobot::moveDirect(goal): adding a goal (" << goal.x
+             << ", " << goal.y << ")" << endl;
 
-    // reset conflicting motion commands
-    _motionConstraints.pivotTarget = boost::none;
+    _motionCommand = make_unique<Planning::DirectPathTargetCommand>(
+        MotionInstant(goal, (pos - goal).normalized() * endSpeed));
```
justbuchanan: Should this be `goal - pos`?

soccer/Processor.cpp
ashaw596: because planrequest has to own the pointer.

soccer/Robot.cpp
ashaw596: fixed

soccer/motion/MotionControl.cpp
```diff
@@ -59,25 +60,44 @@ void MotionControl::run() {
...
+    boost::optional<Geometry2d::Point> targetPt;
+    const auto& motionCommand = _robot->motionCommand();
+    if (motionCommand->getCommandType() == MotionCommand::Pivot) {
+        PivotCommand command = *static_cast<PivotCommand*>(motionCommand.get());
+        targetPt = command.pivotTarget;
+    }
+    if (rotationCommand != nullptr) {
+        try {
+            Planning::FacePointCommand* facePointCommand =
+                dynamic_cast<Planning::FacePointCommand*>(
```
justbuchanan: `dynamic_cast` will return a `nullptr`, not throw a `bad_cast` upon failure in this case.  See http://en.cppreference.com/w/cpp/language/dynamic_cast

soccer/planning/MotionCommand.hpp
```diff
@@ -3,64 +3,66 @@
...
 #include "planning/MotionInstant.hpp"
+#include "Utils.hpp"
 
 namespace Planning {
 
-/**
- * This class contains the motion constraints that the gameplay logic sets for
- * a robot.
- */
 class MotionCommand {
```
justbuchanan: This class could use a comment.

soccer/planning/RRTPlanner.cpp
```diff
@@ -44,13 +44,16 @@ bool RRTPlanner::shouldReplan(MotionInstant start, MotionInstant goal,
...
 std::unique_ptr<Path> RRTPlanner::run(
-    MotionInstant start, MotionCommand cmd,
+    MotionInstant start, const MotionCommand* cmd,
     const MotionConstraints& motionConstraints,
     const Geometry2d::ShapeSet* obstacles, std::unique_ptr<Path> prevPath) {
+    assert(cmd->getCommandType() == Planning::MotionCommand::PathTarget);
+    Planning::PathTargetCommand target =
+        *static_cast<const Planning::PathTargetCommand*>(cmd);
+
     // This planner only works with commands of type 'PathTarget'
```
justbuchanan: Can you move this comment above the assert?

soccer/planning/TargetVelPathPlanner.cpp
```diff
@@ -93,26 +104,28 @@ bool TargetVelPathPlanner::shouldReplan(
 // TODO(justbuchanan): Paths aren't dynamically feasible sometimes because it
 // doesn't account for initial velocity
 std::unique_ptr<Path> TargetVelPathPlanner::run(
-    MotionInstant startInstant, MotionCommand cmd,
+    MotionInstant startInstant, const MotionCommand* cmd,
     const MotionConstraints& motionConstraints,
     const Geometry2d::ShapeSet* obstacles, std::unique_ptr<Path> prevPath) {
-    assert(cmd.getCommandType() == MotionCommand::WorldVel);
-
-    // If the start point is in an obstacle, escape from it
```
justbuchanan: Why did this chunk get deleted?

ashaw596: Because it also takes the Pivot Command right now. It doest the comandType check in the lambda

justbuchanan: Yeah but that's a separate thing.  There was a chunk of code here that handled the case where the start point is inside an obstacle.  Looks like you hit a merge conflict and chose your changes when you should have chosen both.

ashaw596: Oh you meant that chunk. oops.

justbuchanan:
Looks good to me, go ahead and merge when it's ready.

PR #453, opened 2015-09-16
Add an --overwrite-sources-list flag to ubuntu-setup

jgkamat:
Adds gatech mirrors if running `./ubuntu-setup -osf`. Need someone on stock ubuntu14.04 to test this before merge!

### DIFF
--- a/util/ubuntu-setup
+++ b/util/ubuntu-setup
@@ -1,18 +1,20 @@
 #!/bin/bash -e
 
-HELP_STR="Usage: ./ubuntu-setup [-y | --yes] [-n | --noclobber] [-c | --clobber] [-f | --firmware] [-nf | --nofirmware] [-h | --help]
+HELP_STR="Usage: ./ubuntu-setup [-y | --yes] [-n | --noclobber] [-c | --clobber] [-f | --firmware] [-nf | --nofirmware] [-osl | --overwrite-sources-list ] [-h | --help]
 
-\tyes:\t\tassume no user input
-\tnoclobber:\tdont install 3rd party repositories
-\tclobber:\tInstall 3rd party repositories
-\tfirmware:\tInstall firmware repository
-\tnofirmware:\tPrevents installation of firmware deps firmware repository
-\thelp:\t\tprint this message!
+\tyes:\t\t\tassume no user input
+\tnoclobber:\t\tdont install 3rd party repositories
+\tclobber:\t\tInstall 3rd party repositories
+\tfirmware:\t\tInstall firmware repository
+\tnofirmware:\t\tPrevents installation of firmware deps firmware repository
+\toverwrite-sources-list:\tOverwrites the sources.list with gatech mirrors for faster ubuntu-setup at gatech. Works only on Ubuntu 14.04
+\thelp:\t\t\tprint this message!
 
 This script will need root privileges"
 
 # defaults!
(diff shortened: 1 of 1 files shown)


jgkamat:
This will copy `/etc/apt/sources.list` to `/etc/apt/sources.list.old` before adding the gatech one. This preserves the old copy. On future runs, it will not overwrite the old file.

util/ubuntu-setup
```diff
@@ -65,6 +71,31 @@ if [ $UID -ne 0 ]; then
...
+if $OVERWRITE_SOURCES; then
+    read -p "Do you really want to overwrite your sources.list? THIS SHOULD ONLY BE DONE ON UBUNTU 14.04 [yN]: " yn
+    case $yn in
+        [Yy]* ) ;;
+        [Nn]* ) exit;;
+        * ) exit ;;
+    esac
+
+    # Copy sources.list to sources.list.old if sources.list.old is empty
+    if [ -f "/etc/apt/sources.list" ]; then
```
justbuchanan: Should this be the .old one?

jgkamat: This is correct, I think. 
```
If sources.list is found
rename sources.list -> sources.list.old (with no overwrite)
```

If sources.list does not exist, then we should not move it (that would fail). Maybe we should have it error out, as that usually means you are on a non-debian system.

justbuchanan: Ahh ok, I didn't see the "-n".  :+1:

justbuchanan:
Nice, I like it!  I've got a 14.04 vm that I can try this on to double-check.  I'll do that in a bit and let you know.

jgkamat:
Cool, thanks! Let me know how fast everything downloads...

justbuchanan:
I ran it on the VM and it's definitely faster :).  Go ahead and merge when you're ready.

jgkamat:
Cool! Could you give me /etc/apt/sources.list[.old], just so we have a copy of the original file (in case something goes wrong)?

I'm going to see if I can make some other improvements and I'll merge it!

justbuchanan:
Sure, here ya go: https://transfer.sh/13pReb/sources.list.old.

justbuchanan:
Actually that link only lasts 14 days, here it is as a gist: https://gist.github.com/justbuchanan/d045498c95fa1cd2dc70.

jgkamat:
Nice, thanks! I'll merge this sometime later today!

justbuchanan:
Looks like something's up with the CI setup, can you check on it?

jgkamat:
huh thats weird, give me a sec

jgkamat:
UTF locale generation seems to be failing which is...weird, since nothing has changed...

Thats causing stylize to get angry again...

jgkamat:
we're seeing the same issue as [these guys](https://github.com/abevoelker/docker-ubuntu-locale)

I'm reverting the typo commit because that triggered the rebuild. Future changes to ubuntu-setup will fail until I fix this though. Probably an infastructure change in circleci, i'll submit a ticket with them anyway, since this was working a while ago

jgkamat:
@justbuchanan heads up, osx is failing on master, not sure whats causing it though.

justbuchanan:
Looking into it now.  Looks like an issue with an updated python package...

PR #463, opened 2015-10-06
New gameplay docs

justbuchanan:
This is a work-in-progress and isn't ready to merge yet.  At tonight's meeting we'll be discussing plays and having everyone write a simple one.  I've been meaning to writeup some documentation for the python-based gameplay system for a while and this seems like a good time to do so.  Let me know if you have any thoughts on other things to include, etc.


justbuchanan:
I merged this now so it will be live on the docs site for the meeting tonight, but it needs a little improvement and probably has some typos.  Let me know if you have any suggestions.

PR #470, opened 2015-10-18
Added update-alternatives command to ubuntu-setup to configure g++ with 4.9

ra4king:
As recommended here: http://askubuntu.com/questions/372248/downloaded-g-4-8-from-the-ppa-but-cant-set-it-as-default

### DIFF
--- a/util/ubuntu-setup
+++ b/util/ubuntu-setup
@@ -161,8 +161,8 @@ fi
 apt-get install $ARGS $(sed 's/#.*//;/^$/d' $BASE/util/ubuntu-packages.txt)
 
 if [ "$SYSTEM" = "ubuntu-14.04" ]; then
-    sudo update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-4.9 40
-    sudo update-alternatives --config g++
+    sudo update-alternatives --install /usr/bin/g++ g++ $(command -v g++-4.9) 40
+    sudo update-alternatives --set g++ $(command -v g++-4.9)
 fi
 
 if [ "$(clang --version | grep -Eo '[0-9]+\.[0-9]+' | head -n1)" \< "3.5" ]; then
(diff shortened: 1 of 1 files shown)


util/ubuntu-setup
```diff
@@ -156,6 +160,11 @@ fi
 # install all of the packages listed in required_packages.txt
 apt-get install $ARGS $(sed 's/#.*//;/^$/d' $BASE/util/ubuntu-packages.txt)
 
+if [ "$SYSTEM" = "ubuntu-14.04" ]; then
```
justbuchanan: Should this be == instead of =?

Also, can you put a comment above the `if` with a short description of what this is for?

jgkamat: It dosent actually matter, both work. `=` is legacy, so its more 'cross platform', but I prefer to use `==` because I don't want to mess up in other languages. Yah...

jgkamat: Also a message telling people what's going on during this step would be nice (especially for debugging).
This is actually pretty important, this script is going to be rerun many times. We should make another conditional with something like this?
```
if command -v g++; then 
  echo "[WARN] You already have g++ symlinked. Not overwriting."
  exit 2
fi

File: util/ubuntu-setup [hunk 2]
```diff
@@ -156,6 +160,11 @@ fi
...
+    sudo update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-4.9 40
+    sudo update-alternatives --config g++
+fi
+
 if [ "$(clang --version | grep -Eo '[0-9]+\.[0-9]+' | head -n1)" \< "3.5" ]; then
     echo "-- Symlinking clang-3.5 to clang"
```
jgkamat: Something like this is also a possibility, although this is probably overkill for this situation.

jgkamat:
LGTM :+1:

util/ubuntu-setup
justbuchanan: Still could use a small comment above it explaining what this is there for.

PR #480, opened 2015-10-27
Add Coveralls and Make CI Better

jgkamat:
I made a bunch of improvements, the largest being Coveralls.

I have no idea what our real coverage is, but its saying 19% now... If someone could review the blacklist I made that would be :ok_hand:
[Go here for coveralls coverage](https://coveralls.io/github/RoboJackets/robocup-software)

This branch is failing because 'master' on the CI server has been cached to an old commit or something, and it is failing the style checks as a result. (circleci caches repos too). I'll open a separate issue for that. (running make checkstyle w/ an updated master is good).

No rush to review this though, just putting it out there.


justbuchanan:
Cool, it looks good to me!  19% isn't too shabby :).  I didn't do a super thorough check of the coverage exclude list, but it looks like you've got all the main things that matter.

Not entirely sure how to fix the style thing :/.  Maybe we could have it update master before running the checkstyle.  Also, I reset master on my computer to a really old commit to force a checkstyle on the files that fail on the CI and it didn't find any issues.  I think this is because the version of clang-format on the server is different than the one on my computer and there are some edge cases that get formatted differently, which is sort of hard to fix.  A work around that I've used before is to run `make pretty` on an ubuntu install that has the same clang-format as the server.  You could do this with the circleci SSH setup, then use the https login to push from there back to this branch.

jgkamat:
I ssh'd in, and it looked like the cached master had diverged from our master, which was weird. Not sure what could have caused that...

jgkamat:
Also, another thing to keep in mind, this causes everything to be built twice in CI, we might not want that. We could possibly limit coverage checks to just the master branch. With caching though, this seems ok.

jgkamat:
Coverage is actually (closer?) to 15%, I was including the test files themselves in the reporting...

justbuchanan:
Cool, 15% is still not too bad :)

It would be cool to show coverage differences for PRs so we can easily see what effect a particular PR has.  I agree though that if it ends up making CI take a lot longer it's probably not worth it.  With ccache though, I wouldn't expect it to be too bad.  What have you seen so far?

jgkamat:
With ccache, build times wen't from 2m -> 3m, but without they seem to be about 20m (while caching the baseimage). The next run becomes 3 minutes again though.

justbuchanan:
3 minutes sounds fine to me.  With those build times, I think it's fine to build coverage on everything.  Obviously changes to header files that are included by many files will make it take a good bit longer, but I don't think it'll be too much of an issue.

This PR looks good to me, feel free to merge whenever it's ready.

justbuchanan:
One minor thing: circleci has a few different styles of build button available - you can change the main ci button on the readme one so it matches the style of the coveralls one so it looks consistent.

jgkamat:
Just updated the badge style, I'll merge this on sunday's meeting.

PR #489, opened 2015-11-04
Replace binutils gatch mirrors with official gnu mirror

### LINKED ISSUE #487: Can't build robot firmware due to gt mirrors being down
I just tried this again on the latest master (90fb3a0f) and it's failing.  @jgkamat, what did you figure out for this?

jgkamat:
Closes #487

Please test this out before you merge though, :smile:  

It LGTM (but so do the old mirrors so idk).

### DIFF
--- a/firmware/arm.scons
+++ b/firmware/arm.scons
@@ -10,7 +10,7 @@ host_env.AppendENVPath('PATH', Dir('install/bin').abspath)
 # to build it).
 host_env['ENV']['MAKEFLAGS'] = '-j%d' % GetOption('num_jobs')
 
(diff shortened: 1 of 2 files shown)


jgkamat:
This switches our firmware mirror to ftp.gnu.org, which has a higher (lower?) chance of being down? I'm not sure, but the download was faster.

jgkamat:
Coverage is failing because our status token is not present on outside repos. We should find a way to fix that soon...

justbuchanan:
Cool, works on my computer!  And I think your fix for coverage should be fine.

jgkamat:
It's not happy with || true, i'll look into that later, I'll make it true from the makefile for now.

justbuchanan:
LGTM :+1:

PR #492, opened 2015-11-04
fixed udev rules for mbed

justbuchanan:
* if multiple mbeds are plugged in, they each get their own name at /dev/mbed[0-9]+
* {arch, ubuntu}-setup scripts now reload udev rules after installing

### DIFF
--- a/util/arch-setup
+++ b/util/arch-setup
@@ -4,7 +4,7 @@ BASE=$(readlink -f $(dirname $0)/..)
 
 echo "-- Installing udev rules"
 sudo cp -f $BASE/util/robocup.rules /etc/udev/rules.d/
-sudo udevadm control -R # reload rules
+sudo udevadm control --reload || true # reload rules
 
 echo "-- Installing required packages"
 sudo pacman -S --needed --noconfirm $(sed 's/#.*//;/^$/d' $BASE/util/arch-packages.txt)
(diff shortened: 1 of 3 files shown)


justbuchanan:
If two mbeds are plugged in, they'll now show up as `/dev/mbed0` and `/dev/mbed1`.  Previously only one of them would show up as `/dev/mbed`.

jgkamat:
the udevadm control -R is failing with exit code 2, which is why the build is failing. I don't know why though...

justbuchanan:
Yeah I'm not sure why either :/.  It works fine on my ubuntu vm, so I wonder if it's something docker-related.

jgkamat:
Yah, it looks fine, I'll look into this..

jgkamat:
Its complaining that it cannot find 
`/run/udev/control` 
in the docker containers, which is true.

Not sure about what to do about it though...
We could try to symlink the hosts udev control to inside the container if we want.

justbuchanan:
Hmm, that's a possibility.  Not sure how socket communication like that works through container boundaries though.

I could just remove the command to reload udev rules, but it would be sorta nice to have :/

jgkamat:
Worst case scenario, we could || true it, but I don't want to do that too much...

justbuchanan:
Yeah I think `|| true` will work here.  In general we should avoid it, but it's ok for rule reloading to fail here.

PR #500, opened 2015-11-11
Handle digital IO in one wrapper class for io-extender

mosdragon:
Use the same one class to handle both read and writes. Fixed constructor

### DIFF
--- a/firmware/robot2015/src-ctrl/drivers/mcp23017/io-expander.hpp
+++ b/firmware/robot2015/src-ctrl/drivers/mcp23017/io-expander.hpp
@@ -31,9 +31,10 @@ enum IOExpanderPin {
 
 
 /*
- * A digitalOut class meant to replicate basic functionality of the mBed digitalOut
+ * A digitalInOut class meant to replicate basic functionality of the
+ * mBed digitalOut and digitalIn
  */
-class IOExpanderDigitalOut
+class IOExpanderDigitalInOut
 {
  private:
   IOExpanderPin m_pin;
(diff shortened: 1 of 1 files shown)


hifzasakhi:
wiki overview of cache coherence
https://en.wikipedia.org/wiki/Cache_coherence

ashaw596:
Why is cache coherence here?

justbuchanan:
:+1:

mosdragon:
@justbuchanan 2 quick questions.

1. I noticed this is a .hpp file, but it seems to have implementation language in here as well like setting the m_pin field. I know C++ doesn't care about file names, but is that bad practice? Should the hpp and cpp files always be separated out?

2. Also, noticed this line at the end of the file:
```c++
bool IOExpanderDigitalInOut::isInit = false;
```
it was under the ```public``` part of the class. Is this similar to a static variable in Java instead and the isInit value is just being initialized to false? And is there any preference in C++ for making the field publicly accessible rather than making a setter/getter for the field and making it private instead?

justbuchanan:
Good questions!

1. In general it's best to put the actual code/implementation in the corresponding .cpp file.  This is partly so that when you change the code, only that cpp file has to be recompiled; when you change the hpp file, everything that #include's it has to be recompiled.  It's also nice to separate the interface and implementation a bit.  In this case, the IO expander class is just a small wrapper over the MCP23017 stuff, so it's fine to have it in the hpp.
2. Yeah, it's pretty equivalent to a static variable in java.  This should probably be declared private, you can move it into that section if you want.  For things that are accessed outside the class, it's good to add getters and setters, but since this is only used inside the class, you can leave them out.

PR #501, opened 2015-11-11
Implement coordinates display when hovering over field. closes #257

### LINKED ISSUE #257: Show (x, y) location of cursor when hovering over the field

jedyang97:
Implemented by keeping track of MouseMoveEvent and updating QLabel posLabel.

### DIFF
--- a/soccer/SimFieldView.cpp
+++ b/soccer/SimFieldView.cpp
@@ -17,11 +17,6 @@ SimFieldView::SimFieldView(QWidget* parent) : FieldView(parent) {
     _dragRobot = -1;
     _dragRobotBlue = false;
     setMouseTracking(true);
-
-    _posLabel = new QLabel(this);
-    QRect rect =
-        QFontMetrics(_posLabel->font()).boundingRect("X: -9.99, Y: -9.99");
-    _posLabel->setMinimumWidth(rect.width());
 }
 
 void SimFieldView::mousePressEvent(QMouseEvent* me) {
--- a/soccer/SimFieldView.cpp
+++ b/soccer/SimFieldView.cpp
@@ -5,6 +5,7 @@
 
 #include <QPainter>
 #include <QMouseEvent>
+#include <iostream>
 
 using namespace boost;
 using namespace Packet;
@@ -16,6 +17,11 @@ SimFieldView::SimFieldView(QWidget* parent) : FieldView(parent) {
     _dragMode = DRAG_NONE;
     _dragRobot = -1;
     _dragRobotBlue = false;
+    setMouseTracking(true);
+
+    _posLabel = new QLabel(this);
+    QRect rect = QFontMetrics(_posLabel->font()).boundingRect("   X: 999, Y: 999");
+    _posLabel->setMinimumWidth(rect.width());
 }
 
 void SimFieldView::mousePressEvent(QMouseEvent* me) {
@@ -68,6 +74,14 @@ void SimFieldView::mousePressEvent(QMouseEvent* me) {
 }
 
 void SimFieldView::mouseMoveEvent(QMouseEvent* me) {
+
+    _posLabel->move(me->pos());
+    QString s = "   X: ";
+    s += QString::number(me->pos().x());
+    s += " Y: ";
+    s += QString::number(me->pos().y());
+    _posLabel->setText(s);
+
     switch (_dragMode) {
         case DRAG_SHOOT:
             _dragTo = _worldToTeam * _screenToWorld * me->pos();
--- a/soccer/SimFieldView.cpp
+++ b/soccer/SimFieldView.cpp
@@ -19,7 +19,8 @@ SimFieldView::SimFieldView(QWidget* parent) : FieldView(parent) {
     setMouseTracking(true);
 
     _posLabel = new QLabel(this);
-    QRect rect = QFontMetrics(_posLabel->font()).boundingRect("X: 999, Y: 999");
+    QRect rect =
+        QFontMetrics(_posLabel->font()).boundingRect("X: -9.99, Y: -9.99");
     _posLabel->setMinimumWidth(rect.width());
 }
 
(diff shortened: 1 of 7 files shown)


jedyang97:
@justbuchanan

jgkamat:
Theres a small style issue with this PR, could you fix it by running `make pretty` and pushing those changes?

jedyang97:
Thanks for replying. I noticed the style issue too. But there are some errors when I make pretty on my computer. I will try to figure it out.

jedyang97:
OK. I will try to do that! Thx :-p

jedyang97:
Hi Jay,

I am new to this. There seems to be some library issues. Do you know how can I fix this?


YangJianings-MacBook-Pro:robocup-software yangjianing$ make pretty
  File "/usr/local/bin/stylize", line 9, in <module>
    load_entry_point('stylize==0.1.2', 'console_scripts', 'stylize')()
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/Extras/lib/python/pkg_resources.py", line 357, in load_entry_point
    return get_distribution(dist).load_entry_point(group, name)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/Extras/lib/python/pkg_resources.py", line 2394, in load_entry_point
    return ep.load()
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/Extras/lib/python/pkg_resources.py", line 2108, in load
    entry = __import__(self.module_name, globals(),globals(), ['__name__'])
  File "/Library/Python/2.7/site-packages/stylize/__main__.py", line 115
    nonlocal file_scan_count
                           ^
SyntaxError: invalid syntax
make: *** [pretty] Error 1

jgkamat:
It looks like your system is configured to use python2 instead of python3. Could you try setting python3 as the default? (I'm not sure how you would do this) @justbuchanan.

The style error dosen't seem to be your fault, you're code looks fine to me! I'll get back to you when I figure this out...

jedyang97:
👍

jedyang97:
Thanks @ashaw596 . That solves the style issue. BTW, do you know how can I run make pretty on Mac El Capitan, it doesn't seem to work on my computer.

ashaw596:
Not quite sure what the problem is. It works on my computer. We can take a look at it next meeting.

justbuchanan:
Nice, this looks good!  A couple things though:
* The coordinates that we use for soccer are in meters with the origin at the center of our goal, which differs from Qt's coordinates for drawing.  Take a look at the coordinate transforms [here](https://github.com/RoboJackets/robocup-software/blob/master/soccer/FieldView.hpp#L70) - I can show you how to use them if you need.
* Right now if your mouse leaves the field view, the coordinates stay visible at their last location.  It would be best if they auto-hide when the mouse moves away.
* If you find something on our todo list that you're interested in working on and someone's already assigned, be sure to talk to them before assigning yourself.  Not sure if you asked her about it at the meeting, but I think Audrey was looking into doing this.  We just want to avoid any duplicated efforts :smile: 

I'm not quite sure what's up with stylize on your computer.  It does seem like a python 2 vs 3 issue as Jay mentioned, but our install script explicitly installs it for python 3...  You can try uninstalling it for python2 and re-installing for python 3 though and see if that fixes it:
~~~{.py}
pip uninstall stylize
pip3 install stylize
~~~

jedyang97:
OK, I will work on these two problems. And yeah, I should talk to Audrey beforehand. Sorry for that. I will send her an email about this.

For the stylize, I managed to solve the problem by 'brew uninstall python' and reinstall stylize by 'pip3 install stylize' and then 'brew install python'. Not sure what's wrong with my python

jedyang97:
@justbuchanan Pls take a look at the new commit. Let me know if there is any problem. :-p

soccer/SimFieldView.cpp
```diff
@@ -68,9 +69,17 @@ void SimFieldView::mousePressEvent(QMouseEvent* me) {
 }
 
 void SimFieldView::mouseMoveEvent(QMouseEvent* me) {
+    _posLabel->move(QPoint(me->pos().x() - 45, me->pos().y() + 17));
```
justbuchanan: It would probably be best to move this code into `FieldView` now that you've moved the other label-related things there.

justbuchanan:
Looking good overall, I think it's almost done!  I mad a couple minor notes and have one more suggestion.

Right now you're using `QCursor` to check if the mouse is inside the widget or not.  Take a look at [QWidget.leaveEvent](http://doc.qt.io/qt-5/qwidget.html#leaveEvent).  It lets you get the same info, but is slightly better since you don't have to store a `QCursor` object and you don't have to check it at every paint event - you can just override `leaveEvent()` and `enterEvent()` and be notified when those things change.

soccer/SimFieldView.cpp
jedyang97: Hi, thanks for the comment! I have read your suggestion and decide to use leaveEvent. So then I will not be using a QCursor. And I will not be able to move this line to FieldView, because I no longer have a reference to the mouse position without QCursor.

justbuchanan: Yeah sounds good.  I think you can still move this to `FieldView`, you'll just have to reimplement the `mouseMoveEvent()` in that class the same way it's implemented here.  The only caveat is that in order for both `SimFieldView` and it's parent class (`FieldView`) to have the mouseMovedEvent() method called, you'll need to call it here in this class.

I'd recommend moving this code block into `FieldView::mouseMoveEvent()`, then putting `FieldView::mouseMoveEvent(me)` here.

jedyang97: Wow, that sounds like a great solution. Thanks, doing it right now. :+1:

jedyang97:
Hi @jgkamat , please check out this new commit. Thanks so much for the suggestion. It's a more elegant implementation and it performs better!

jedyang97:
Not sure what is failing the tests here. I did pass `make pretty`

justbuchanan:
Yeah I'm not sure why it's failing - it passes the checkstyle on my computer too.  I think it has something to do with the version of the formatter on the server being older than the one on our computers.  We can worry about that later though.

Great work!  I'm merging this in now.

jedyang97:
Great! Thank you!👍

justbuchanan:
Sure thing, Jed!  You did the actual work here, so thank you :smile:

justbuchanan:
@jedyang97, I just realized a small error that I didn't see before.  The current implementation has a small memory leak because `_posLabel` is allocated in the constructor, but is never `delete`d.  It's not a huge deal, but it would be best to fix it.  There are two options:
* you can add a `delete _posLabel` to `FieldView`'s destructor
* OR you can make `_posLabel` be a direct instance variable (not use a pointer)

I think the second option is slightly cleaner since you avoid any potential memory issues and don't have to worry about deleting it later.

jedyang97:
Hi @justbuchanan , thanks for pointing this out. I am new to C++ and QT, so I am not sure, but I found this: http://www.qtcentre.org/threads/17060-Qt-memory-handling, which says QT will automatically destroy all of the children widgets of a QWidget when it is destroyed. So do I still need to destroy the label myself?

justbuchanan:
My mistake, I think you're right here.  Your initialization looks like `QLabel(this)`, so `this` (the FieldView object) becomes the "parent" of the label and will take care of deleting it later.  I think everything's fine as is :+1:

jedyang97:
:+1:

PR #504, opened 2015-11-12
Fix style ci

jgkamat:
Hopefully this will fix the problems we were seeing

I fixed SSH by adding a link into the container when we build the baseimage.


jgkamat:
When your branch is behind master, this causes style to fail sometimes (see above)

justbuchanan:
:+1:

PR #510, opened 2015-11-25
Simplified cc1201 config

justbuchanan:
@jjones646, I started work on simplifying the cc1201 config and rearranged things a bit.  I'm using the C array export format from SmartRF so that setting the config just requires a loop over each of the entries.  This replaces the previous CC1201Config class.  I also consolidated the register read/write methods.

This doesn't quite work yet, but is almost there.  Let me know if you have any suggestions.

TODO

- [x] Fix AES address space issues.  The exported #defines from smartrf studio put aes values outside the range of a uint16.  The user guide seems to show different values than the exported defines.


jjones646:
@justbuchanan. The ti config file sounds excellent. I bet that's quite a bit faster at startup now.
I haven't gotten a chance to look at your changes in too much detail yet, but I will say the extended register access can be tricky. So if you have weird things happening that doesnt make any sense, look over what's happening on an extended read/write.

justbuchanan:
I fixed the main issue with this, which was just a dumb mistake on my part.  I was setting the registers in the constructor right before the call to reset()...

Anyways, I think this is good to go now.  Take a look at it when you get a chance and we can merge it into the firmware branch if there are no issues.

cc @jjones646 @guyfleeman

Issue #186, opened 2015-01-16
Use MBED to interface with ball sense hardware


justbuchanan:
We need to pulse the emitter and read the detector to determine whether or not we have captured the ball.

justbuchanan:
@gtrosawa what are the differences between the ball sensor setup this year and our previous one?

justbuchanan:
@jjones646, @gtrosawa, We're looking at the ball sensor bar from last year and I'm not quite sure which side is which.  One "bulb" is red and one's clear - do you know which side is the detector?

jjones646:
- Purple/Black = Detector
- Clear = Emitter

justbuchanan:
@jjones646, @ryanstrat what's the status on ball sense boards?

jjones646:
@justbuchanan These are ~1 month away I believe. The pcb design is finished - just a matter of getting the order in and then waiting at that point.

ryanstrat:
​We ran into some issues panelizing the board that I think we've worked
through. They'll be ordered within the week with a 2.5 week turnaround.


Ryan Strat
ryan@str.at
(678) 517-3393
RoboJackets | RoboCup Project Manager
Georgia Institute of Technology

JNeiger:
Is there a schematic for the connection between the Detector/Emitter and the mbed?

Right now I have the emitter connected to p20 and ground. The detector is connected to p19 and the same ground.

With this setup, whenever the emitter has power going through it, the detector values are shifted a significant amount (~500 units), almost as if there is some sort of cross talk between them.

justbuchanan:
@jjones646, any ideas what might be causing the shift in values he's seeing here?  I looked at the breadboard setup for this yesterday and it all looked right to me.

@JNeiger, can you post a link to the branch that your code is on so I can check it out?

JNeiger:
@justbuchanan https://github.com/JNeiger/robocup-software/tree/ball_sensor

justbuchanan:
I browsed through your code and it all looks good to me.  Can you also post a pic of your breakout board?

justbuchanan:
I think the solution may be to put a resistor in series with the photoresistor, as shown here: ![photoresistor circuit](http://teach.alimomeni.net/2012spring2/wp-content/uploads/2012/02/analogin_variableresistor.jpeg).

I forgot about having that resistor in there before, but that should fix the issue.  @jjones646, do you know what resistor the final board will have?

PR #516, opened 2016-01-20
Ball sensor

JNeiger:
Adds the basic functionality for #186.


jgkamat:
It looks like a bunch of commits unrelated to this got carried in. Do you know how to use cherry pick to only get the relevant commits?

justbuchanan:
I think it's because he merged in a newer version of `master` than had been pulled into `robot2015-firmware`.  @JNeiger, I updated `robot2015-firmware`, so try pulling in the latest from there and push it back up to your branch again.

JNeiger:
Whoops. I forgot robot2015-firmware was behind master. I'll get this fixed either tonight or tomorrow

JNeiger:
Should be fixed now

justbuchanan:
it's failing the checkstyle at the moment.  Try running `make pretty`, then commit the changes to fix it.

jgkamat:
@justbuchanan there seem to be a ton of files coming up, [see this](https://circle-artifacts.com/gh/RoboJackets/robocup-software/1038/artifacts/0/tmp/circle-artifacts.EUA9GCJ/style.txt)

Not sure what caused this, this PR only changes two files? I tried running make pretty and it made a ton of changes.

justbuchanan:
@jgkamat I think it's because `make pretty` checks all files that differ from master even though this PR intends to merge a few commits onto `robot2015-firmware`.

@JNeiger, I'll run `make pretty` on the firmware branch and update it, then you can merge firmware into yours and update this branch.

JNeiger:
I think I got it

justbuchanan:
Hmm, looks like your submodules are still not quite right.  Try doing:
~~~{.sh}
git fetch robojackets
git checkout robojackets/robot2015-firmware -- external
git add external
git commit -m 'fixed submodules'
~~~

This should reset all submodules to what they are on the firmware branch.

justbuchanan:
Here's the datasheet for the IR emitter: http://www.vishay.com/docs/84228/vsmb294008rg.pdf

firmware/robot2015/cpu/ball_sense.hpp
```diff
@@ -0,0 +1,38 @@
+// Determines if the emitter to reciever beam is broken
+//  while accounting for ambiant light
+#include "mbed.h"
+
+class ball_sense {
```
justbuchanan: Let's rename this to `BallSense` to keep with project conventions.

firmware/robot2015/cpu/ball_sense.hpp
```diff
@@ -0,0 +1,38 @@
...
+    ball_sense();
+
+    // Call once per cycle
+    void update_ball_sensor();
+
+    // Returns true if beam is broken (Ball is contained)
+    bool have_ball();
+
+private:
+    DigitalOut* emitter_pin;
```
justbuchanan: Let's make the DigitalOut and AnalogIn regular variables, rather than pointers.  It's slightly safer and faster

firmware/robot2015/cpu/ball_sense.hpp
```diff
@@ -0,0 +1,38 @@
...
+    AnalogIn* detector_pin;
+
+    // Holds the light sensed when
+    //  emitter is lit (sense_light) and
+    //  dark (sense_dark)
+    int sense_light = 0, sense_dark = 0;
```
justbuchanan: If this value isn't used outside the update function and isn't read except in the iteration it's written, it would probably be better to make it local to the function rather than an instance variable.

firmware/robot2015/cpu/ball_sense.hpp
```diff
@@ -0,0 +1,38 @@
...
+    //  values before the sensor is considered
+    //  broken
+    const int sense_threshold = 8000;
+
+    // Number of consecative "broken" senses
+    //  before we are confident the beam is broken
+    const unsigned int consec_num = 2;
+
+    // Consecative "broken" senses counter
+    unsigned int consec_ctr;
```
justbuchanan: This probably should be initialized to zero, otherwise there's a chance it may falsely detect a ball when this is first initialized.

firmware/robot2015/cpu/ball_sense.hpp
JNeiger: As it is right now, update_ball_sensor "alternates" functionality based on which time it is called so they need to be outside the scope of the local function. The first time it checks the light emission when the emitter is on, the second it checks it when it is off.

I was mostly worried about a false reading if I tried to do them within the same time step. I can test it later with the mock setup that we had running to see if there is any noticeable problem.

JNeiger:
I noticed the file structure has also changed. Do you want me to move these into their own folder under /firmware/robot2015/src-ctrl/drivers?

justbuchanan:
Yeah that sounds good, you can put it in its own folder under `drivers`.

firmware/robot2015/cpu/ball_sense.hpp
justbuchanan: Oops, I meant to just comment on the `sense_light` variable, didn't notice they were on the same line.  I think it's actually fine as it is now.

justbuchanan:
I went through the checkstyle log again and realized I was reading it wrong before... oops

The list it shows is all of the files it checked, the ones with an X next to them are the ones that it made changes to, which was only 3 files.  Go ahead and run the `git apply` command it shows and it should all be good.

justbuchanan:
:+1:

Issue #498, opened 2015-11-11
Soccer UI vibrates because of different font width of timer numbers


ashaw596:
Occurs on my laptop which is OSX el capitan

jgkamat:
Font is different on my system, and I don't see this issue, so it's a font problem.

justbuchanan:
Yeah sounds like a font issue.  I don't see this problem on my Mac, but I'm still on Yosemite.

jgkamat:
![2017-01-13-011123_535x78_scrot](https://cloud.githubusercontent.com/assets/4349709/21920148/386ad868-d92d-11e6-9e02-32f930f967fe.png)

This is still an issue for me, at these numbers above. The font is monospace, so I've got no clue why it's doing this. Sounds like a good beginner issue for someone interested in UI.

PR #517, opened 2016-01-20
Changed UI to allow more space for FieldView. closes #418

### LINKED ISSUE #418: Move the robot status panel into the tab widget
Once our new robots are in good shape, we shouldn't have hardware issues nearly as often as we do now.  It would give us a lot more room to expand the FieldView if the robot status panel was placed into the tab widget and not always visible.  We should also make the tab turn red whenever a robot has a "showstopper" so we can still easily see if there's a problem.

jedyang97:


justbuchanan:
Awesome!  Can you post a screenshot of what the soccer ui looks like now?

jedyang97:
Hi, so I am trying to change the UI to allow more space for FieldView. After I change the layout of MainWindow.ui, the components don't seem to change when the window is resized. Also, how do I set the initial orientation of FV into horizontal? Thanks in advance!

screen shot 2016-01-20 at 14 37 56

jedyang97:
Another issue is that the TabWiget on the right is taking up all the available space. I tried to set the size property of it into minimum expending or minimum, neither option works. Any idea?

justbuchanan:
I'd recommend checking out the [Qt layouts tutorial](http://doc.qt.io/qt-5/layout.html).  There's a lot of info there, but it does a good job of explaining how layouts work and how to effectively set the size policy, min size, max size, etc.  I think if you change the size policy on the field view (or its parent layout) and the tab widget, you can get it to expand the field view when the whole window is expanded.

The initial field view orientation is set in code [here](https://github.com/RoboJackets/robocup-software/blob/robot2015-firmware/soccer/FieldView.cpp#L47).  Set the `_rotate` value to adjust the orientation.

justbuchanan:
Let me know if you get stuck with the size policy settings, it can be kinda tricky.

jedyang97:
👍I‘ll look at it

jedyang97:
@justbuchanan Check this out. I fixed all the problems mentioned above. Also closes #498, where the UI vibrates under OSX El Capitan.
screen shot 2016-01-20 at 19 19 55

justbuchanan:
Nice, looks a lot better!  A couple notes:
* We can get rid of the draggable handle between the FieldView and the referee/team info.  The drag handle is nice to the left of the tab widget so that we can resize it, but since the referee/team info is a fixed vertical size, we won't be adjusting it.
* It would be nice if we can re-arrange the referee info so that it takes up less vertical space.  The main goal is to have a larger FieldView and the limiting factor right now is the ref stuff.

justbuchanan:
Btw, how did you fix the issue with text "vibrating" on OS X?

jedyang97:
I just changed the font of the vibrating word to Arial. I will take care of the draggable handle issue. For the referee panel, can we have two labels in each column to free up more space? Though that could look a little ugly.

jedyang97:
How can I refer to a line in a specific commit?(not in the master branch?)

ashaw596:
Looks good. The font fix worked! Btw, you could probably save some vertical space so that the field can be bigger if you move timeouts left and goalie id to another column instead of keeping everything in one column. If you mess with the margins, everything should still fit horizontally. Also, if you lower the line spacing, the vertical space should also be lowered without cutting readability much.

ashaw596:
You could also probably lower the min-width of the right panel a bit since I don't think any of the tabs require quite that wide of a space.

ashaw596:
Also, the draggable handle for the referee panel is nice, since you can use it to actually minimize the entire panel. It looks great when you do this. 
image

ashaw596:
Just a comment since, the vibration issue is mostly fixed. If there was some way to set a fixed width for the timer box, it would fix the fact that the whole top bar still "vibrates" occasionally when the number of numbers changes. 1 -> 10 -> 100

jedyang97:
👍I will do that.

justbuchanan:
It's looking good!  Just a quick note: since stylesheets cascade in Qt, I think you can set the stylesheet on the MainWindow itself and all labels within it will use that font.  That way we only have to set it in one place.

jedyang97:
screen shot 2016-01-21 at 13 40 05

ashaw596:
Looks great! As a comment, if you want, you could center the field horizontally and vertically. Not really important though. Nice man!

Also, it defaulting to 0 degrees instead of 270 might make the sides slightly more consistent with the referee box if you want.

justbuchanan:
:+1:  for 270 degrees

jedyang97:
Yeah, it would be nice if I could center align it. Any idea of how I can do this? There doesn't seem to be a property of alignment in QT Designer.

I will change the default orientation.

ashaw596:
I belive if you encapsulate it in an layout, you can set an alignment. I think. Maybe

jedyang97:
Hi i found that each time I launch soccer, I got a different number in the "Time left" label in the referee panel. Is this normal? If not, what code should I look at to debug this?

jedyang97:
This also happens to the master branch. Another issue is that the logTime sometimes just jump ahead. Did this happens to the master branch?

jedyang97:
I still cannot figure out how to set FieldView to align to the center. Other than that, I think it's ready to merge.

justbuchanan:
I'm not sure what's up with the "time left" label - I tried launching soccer several times and it shows 0 for me every time.

I don't know offhand how to do the vertical centering.  I think the FieldView widget takes whatever size the parent splitter layout allocates for it, so we may need to do that with code inside the `FieldView` class.  We can probably leave it as is for now though.

It looks like it's failing the style check right now, so try running `make pretty`.

jedyang97:
Notice the Time Left label on the far left side. I have tried make pretty but nothing changes.
screen shot 2016-01-22 at 15 25 08

justbuchanan:
Hmm, that time is definitely not right haha.  Can you try running the `master` version on your computer and see if you get the same result?  I can't reproduce on my computer.  If it happens on master too, we should open a separate issue and include that screenshot.

I think the checkstyle issue is related to #519 - I'll have to look at it and get back to you.

jedyang97:
Hi, somehow my soccer cannot launch any more, I got this:
`TypeError: No to_python (by-value) converter found for C++ type: SystemState*
libc++abi.dylib: terminating with uncaught exception of type std::runtime_error*
./soccer: line 2: 51350 Abort trap: 6           $(dirname $0)/soccer.app/Contents/MacOS/soccer $@
make: *** [run-sim] Error 134
`
I tried to reinstall all the packages and ran os-setup several time and still couldn't solve it. Any ideas?

justbuchanan:
I noticed this on both my Arch and OS X computers earlier today too...  I think it's a problem with boost 1.6, which isn't really something we can fix.  I see some people having a similar issue [here](https://github.com/BVLC/caffe/issues/3494).  Their only solution was to downgrade boost to version 1.59.  Hopefully boost fixes this soon :/

jedyang97:
Thanks! This is pretty frustrating :-( BTW, how do you downgrade to 1.59 with homebrew? I `brew install boost155`, but then when I `make` I got
`CMake Error at /usr/local/Cellar/cmake/3.4.2/share/cmake/Modules/FindBoost.cmake:721 (file):
  file STRINGS file "/usr/local/include/boost/version.hpp" cannot be read.
Call Stack (most recent call first):
  soccer/CMakeLists.txt:77 (find_package)


-- Boost version: 0.0.0
-- Found the following Boost libraries:
--   system
CMake Error at /usr/local/Cellar/cmake/3.4.2/share/cmake/Modules/FindBoost.cmake:721 (file):
  file STRINGS file "/usr/local/include/boost/version.hpp" cannot be read.
`

justbuchanan:
Yeah it's a pain, but luckily it's not too hard to downgrade a package with homebrew.  Run `brew info boost` to see what versions of boost you have in homebrew.  Typically you'll have the latest version installed and linked and a few previous ones in the filesystem, but not linked into place.  Here's a shortened version of the output I get when I run the command:

~~~{.sh}
$ brew info boost
boost: stable 1.60.0 (bottled), HEAD
Collection of portable C++ source libraries
http://www.boost.org
/usr/local/Cellar/boost/1.58.0 (10,718 files, 462.2M) *
  Poured from bottle
/usr/local/Cellar/boost/1.59.0 (10,928 files, 429.2M)
  Poured from bottle
/usr/local/Cellar/boost/1.60.0_1 (11,139 files, 436M)
  Poured from bottle
From: https://github.com/Homebrew/homebrew/blob/master/Library/Formula/boost.rb
~~~

The asterisk shows the one currently in use.  Choose a pre-1.60 version and tell homebrew to use it with something like `brew switch boost 1.59.0`.  Then go back to the robocup folder and do `make clean && make run`.  That worked for me, let me know if you have any issues.

jedyang97:
@justbuchanan Unfortunately I usually `brew cleanup` each time after I `brew update & brew upgrade`. So I do not have the older versions in my homebrew directory. However, I figured out another way to downgrade to a specific version of any software with homebrew when you do not locally have the older version. I'll just jot this down here for future reference:
1. Go to the homebrew github page and search for the file with the name of the software you want to downgrade and click to go to that file (for example, search for boost and then go to boost.rb).
2. Click on history to view all the commits available for it.
3. Find the commit with the version you want and copy the SHA of that commit.
4. `cd` to your homebrew directory (/usr/local/)
5. `git checkout` + the SHA you just copied
6. `brew install boost` (`brew uninstall boost` first if you still have the latest version)
7. `brew update` to get you homebrew back to the latest commit.

jedyang97:
And unfortunately, the "Time Left" problem also happens to master.
screen shot 2016-01-23 at 15 01 48

justbuchanan:
I pulled your code and ran the code formatter on Ubuntu, which has the same version of clang-format on the ci server, so the style check now passes.

The referee labels overlap on ubuntu though - not sure if it's a difference in fonts or something else.
2016-01-23-152403_1367x802_scrot

jedyang97:
Hmm, looks like the spaces between widgets are smaller on Ubuntu. I'll adjust it.

justbuchanan:
@jedyang97, have you had a chance to add some space between the labels yet?

jedyang97:
Hi, I am so sorry. I was too busy these two weeks to do this. I will try to add spaces to it ASAP. It's also OK for me if someone else wants to do it.

justbuchanan:
No worries, I completely understand.  I may take a look at it soon if you don't have time, but it's not urgent to get this implemented.

jedyang97:
Thank you!

PR #521, opened 2016-01-23
initialize ref time left variable to zero.  fixes #520

### LINKED ISSUE #520: "Time Left" label occasionally generates random incorrect numbers.
Notice on the upper left side of the referee pane, the "Time Left" label displays an incorrect number. This doesn't happen every time I launch soccer. It happens about 1 in 5 times. I am on Mac OS X El Capitan.

screen shot 2016-01-23 at 15 01 48

justbuchanan:
@jedyang97, give this a try.  I think the issue is just that the variable is never initialized so it uses whatever was in memory before.  Sometimes this is zero, sometimes it's something ridiculous.

### DIFF
--- a/soccer/NewRefereeModule.hpp
+++ b/soccer/NewRefereeModule.hpp
@@ -163,7 +163,7 @@ public:
     //
     // If the stage runs over its specified time, this value
     // becomes negative.
-    int stage_time_left;
+    int stage_time_left = 0;
 
     // The number of commands issued since startup (mod 2^32).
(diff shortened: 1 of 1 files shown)


jedyang97:
OK👍I'll test it out.

jedyang97:
Yeah I think this fixes the problem. I tested it about 20 times and the problem never happens again.

justbuchanan:
Awesome!  Always nice when it's an easy fix.

PR #522, opened 2016-01-24
Converted Timestamp constant to function

### LINKED ISSUE #378: Make SecsToTimestamp and TimestampToSecs function macros.

### LINKED ISSUE #523: Fix TrapezoidalPath hit method
Its using the wrong types to store time.

### LINKED ISSUE #528: Rect.hit() isn't quite right
cc @ashaw596

chachmu:
Converted Timestamp constant to function. Closes #378
Also Fixes #523 
Fixes #528


common/time.hpp
```diff
@@ -15,4 +15,12 @@ static inline Time timestamp() {
     return (Time)time.tv_sec * 1000000 + (Time)time.tv_usec;
 }
 
+static inline RJ::Time SecsToTimestamp(double secs){
```
ashaw596: Is there a reason this takes in a double. You should also add a comment that this returns a u_int64 and gets rid of decimals.

ashaw596: It might also be nice to create an overload one that takes in an u_int which doesn't lose precision to double multiplication.

soccer/motion/MotionControl.cpp
```diff
@@ -140,9 +140,8 @@ void MotionControl::run() {
 
         // convert from microseconds to seconds
         float timeIntoPath =
-            ((float)(RJ::timestamp() - _robot->path()->startTime())) *
-                TimestampToSecs +
-            1.0 / 60.0;
+            RJ::TimestampToSecs((float)(RJ::timestamp() - _robot->path()->startTime()))
```
ashaw596: The float cast isn't really necessary.

soccer/planning/SingleRobotPathPlanner.cpp
```diff
@@ -52,14 +52,14 @@ bool SingleRobotPathPlanner::shouldReplan(
...
-    const RJ::Time kPathExpirationInterval = replanTimeout() * SecsToTimestamp;
+    const RJ::Time kPathExpirationInterval = RJ::SecsToTimestamp(replanTimeout());
     if ((RJ::timestamp() - prevPath->startTime()) > kPathExpirationInterval) {
         return true;
     }
 
     // Evaluate where the path says the robot should be right now
     float timeIntoPath =
-        ((float)(RJ::timestamp() - prevPath->startTime())) * TimestampToSecs +
+        RJ::TimestampToSecs((float)(RJ::timestamp() - prevPath->startTime())) +
```
ashaw596: This float cast also isn't necessary (Or good really)

justbuchanan:
The build is currently failing, which may be due to #523.  @ashaw596 let us know when you get a chance to look at that.

ashaw596:
@justbuchanan Sooo, I fixed TrapezoidalPath's hit method. However, the pathplanner isn't working correctly so now the tests are failing. Its because the hit(Point) method accounts for the robot size while the hit(Seg,Circle, etc) don's so the FindNoNBlocking goal isn't working correctly I believe.

justbuchanan:
Hmm, the `Shape.hit()` methods take into robot radius, see https://github.com/RoboJackets/robocup-software/blob/master/common/Geometry2d/Circle.cpp#L153

ashaw596:
Well Rect doesn't (which is failing our tests). We really need to finish auditing Geometry2d sometime. So is the expected to take into account the Robot Radius?

ashaw596:
I tried changing hit to use nearSegment in Rect, but TargetVelPlanner and stuff stopped working completely.

common/time.hpp
```diff
@@ -15,4 +15,12 @@ static inline Time timestamp() {
     return (Time)time.tv_sec * 1000000 + (Time)time.tv_usec;
 }
 
+static inline RJ::Time SecsToTimestamp(double secs) {
```
ashaw596: Could you just add a quick comment to doc what these functions do? Other than that. I think its good.

justbuchanan: Yeah that'd be good.  Just mention that it converts from a decimal number of seconds to an integer number of microseconds (a timestamp).  And vice versa for the other one.

ashaw596:
This should close #523 also

justbuchanan:
LGTM :+1:

PR #525, opened 2016-01-27
added config system for normal and damped joystick speeds

### LINKED ISSUE #343: Use config system for damped joystick speeds rather than hard-coded values and increase them

Offboss:
Added the speed of damped and normal joystick translation and rotation to the config system. Closes #343


ashaw596:
This branch doesn't appear to be compiling in test-soccer right now.

justbuchanan:
@Offboss, since I erased the above "Made pretty" commit on your computer and added a couple new commits, you'll have to do a force push to get the new commits onto this branch on github.  Try `git push -f origin configJoystick`.  Generally it's not a good idea to do force pushes (you can ask @jjones646 ), but in this case it's ok.

jjones646:
@justbuchanan @Offboss Even if you are *99.99% certain* that the files are on your other computer, there's a 100% change that they're actually not. :pensive:

PR #526, opened 2016-01-27
Fixed some kicker-attiny merge issues and make pretty.

jpfeltracco:


justbuchanan:
Looks good!  Can you put a readme.md or something in the drivers/attiny driver and just say where the driver files came from?

Also, what version of the attiny chip are we using again?  I've forgotten, we've looked at several haha.  cc @jjones646

jjones646:
@justbuchanan The [ATtiny84A](http://www.atmel.com/images/8183s.pdf)

justbuchanan:
I see that the driver files in here are for AVR910, how does that relate to ATtiny84A?  Is AVR910 the family or something?

jjones646:
The [AVR910](http://www.atmel.com/images/doc0943.pdf) is a programmer for AVR microcontrollers. Not sure why it's included - didn't look that deep into the files merged from the `kicker-attiny` branch. If it is needed, there's a good chance it can be installed along with gcc-avr somehow.

jpfeltracco:
My bad, I brought it back in because the build was failing without it. I'll look into it!

PR #527, opened 2016-01-27
fixed issue with boost 1.6

justbuchanan:
Boost 1.6 seems to require a call to `reigister_ptr_to_python()` that wasn't needed before...  This adds that call, which fixes an issue that caused soccer to crash on launch when built with boost 1.60.

@jedyang97 can you try this out with the latest versions of things from `homebrew` and see if it works for you?

cc @mosdragon


jedyang97:
Hi, I just tested on my computer and this fix works! I'm on OS X El Capitan and I tested this with boost 1.60.0_1 (the latest version on homebrew).

justbuchanan:
:+1:

PR #531, opened 2016-01-31
fixed compatibility issues with boost 1.6

justbuchanan:
@jgkamat, this should fix the errors you were seeing about to_python by value converters when running plays.


jgkamat:
:+1: can I merge this?

PR #533, opened 2016-01-31
CI server now generates a patch that can be applied to fix style

### LINKED ISSUE #519: Checkstyle does different things on mac vs ubuntu
For some reason checkstyle on mac and ubuntu are slightly different. For example, 85721a7 passed checktyle on mac, but on ubuntu running make pretty resulted in, af30f23 which fixed two files.

justbuchanan:
closes #519

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,4 @@ __pycache__/
 
 # IntelliJ
 .idea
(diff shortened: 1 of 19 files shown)


justbuchanan:
closes #519

PR #534, opened 2016-01-31
Created tests for basic Geometry2d classes. Audited and Refactored Point, Segment, Line

ashaw596:
Changes makefile to allow the execution of a subsection of all tests with TEST=* or TEST=Point* notation.
Added comprehensive tests to Point Class.
Added more tests to Line and Segment Class.
Refactored Segment to not be a subclass of Line.
Added nearlyEquals method in Geometry2d::Utils for float comparisons.
Fixed, simplified, and audited most methods of Line and Segment.

Minor cleanup of Rect.


ashaw596:
@justbuchanan If you get a chance, can you take a look at some of the changes.

common/Geometry2d/Line.hpp
```diff
@@ -5,20 +5,19 @@
...
-        pt[1] = other.pt[1];
+    explicit Line(Point p1, Point p2) : pt{p1, p2} {
+        // assert(p1 != p2);
     }
 
-    Line(Point p1, Point p2) {
-        pt[0] = p1;
-        pt[1] = p2;
-    }
+    explicit Line(const Line& other) : Line(other.pt[0], other.pt[1]) {}
```
justbuchanan: Any reason for `explicit` here?  Copy constructors usually omit it

common/Geometry2d/Line.hpp
```diff
@@ -33,10 +32,12 @@ class Line {
     Applies a transformation matrix to the line.
     @param t the transformation matrix to perform on the line
     */
+    /*
```
justbuchanan: Can we keep this method?

common/Geometry2d/Line.hpp
```diff
@@ -56,6 +62,7 @@ class Line {
      * tells you which side of the line you are on
      * @return the sine of the non-zero portion of the cross product
      */
+    // TODO check this
```
justbuchanan: x

common/Geometry2d/Point.hpp
```diff
@@ -26,16 +32,23 @@ class Point {
...
     @param y the y coordinate
     */
-    Point(float x, float y) {
-        this->x = x;
-        this->y = y;
-    }
+    Point(float x, float y) : x(x), y(y) {}
+
+    /**
+     * Implicity constructor for creating a Point from a Packet::Point
```
justbuchanan: Implicity -> Implicit

common/Geometry2d/Point.hpp
```diff
@@ -26,16 +32,23 @@ class Point {
...
+     */
+    Point(const Packet::Point& other) : x(other.x()), y(other.y()) {}
+
+    /**
+     * Implicity constructor for creating a Point from a QPointF
```
justbuchanan: same as above

common/Geometry2d/Point.hpp
```diff
@@ -51,25 +64,32 @@ class Point {
...
-    Point operator/(const Point& other) const {
+    Point operator/(Point other) const {
         return Point(x / other.x, y / other.y);
     }
 
     /**
      * see operator+
+     * does vector multiplication, note the operator
+     */
+    Point operator*(Point other) const {
```
justbuchanan: This one might need a little more explanation since it's not the dot or cross product.  It's sort of an unusual operation to perform.

common/Geometry2d/PointTest.cpp
```diff
@@ -1,19 +1,219 @@
 #include <gtest/gtest.h>
+// #include <../../common/Point.hpp>
```
justbuchanan: x

common/Utils.hpp
```diff
@@ -33,6 +35,11 @@ inline void debugThrow(const T& e) {
 inline void debugThrow(const std::string& string) {
     debugThrow(std::runtime_error(string));
 }
+
+inline void debugAssert(bool b) {
```
justbuchanan: This doesn't look like it's being used anywhere.  Want to remove it?

common/Geometry2d/PointTest.cpp
```diff
@@ -1,19 +1,219 @@
...
+        Point(0 - FLOAT_EPSILON / 2, 5 + FLOAT_EPSILON / 2)));
+    EXPECT_TRUE(Point(0, 5).nearlyEquals(
+        Point(0 + FLOAT_EPSILON / 2, 5 - FLOAT_EPSILON / 2)));
+
+    EXPECT_FALSE(Point(0, 5).nearlyEquals(
+        Point(0 - FLOAT_EPSILON * 1.1, 5 + FLOAT_EPSILON * 1.1)));
+    EXPECT_FALSE(Point(0, 5).nearlyEquals(
+        Point(0 + FLOAT_EPSILON * 1.1, 5 - FLOAT_EPSILON * 1.1)));
 }
+// TODO(ashaw596) Add tests for angleTo and angleBetween once those changes are
```
justbuchanan: status on this?

common/Geometry2d/Rect.hpp
```diff
@@ -63,6 +63,7 @@ class Rect : public Shape {
 
     Point center() const { return (pt[0] + pt[1]) / 2; }
 
+    /* Assumes that pt[0] <= pt[1] for both x and y */
```
justbuchanan: If we make this assumption here, would it be best to force this in the constructor?  Most Rect objects I've seen elsewhere are either defined by an upper-left `origin` along with `width` and `height` OR a top-left and bottom-right point.  It might be good if we change things with ours to swap the points in the constructor so that they're ordered.

common/Geometry2d/Segment.cpp
```diff
@@ -18,6 +18,13 @@ Rect Segment::bbox() const {
 }
 
 float Segment::distTo(const Point& other) const {
+    float dist = nearestPoint(other).distTo(other);
+    if (nearlyEqual(dist, 0)) {
+        return 0;
+    }
+    return dist;
+    /*
```
justbuchanan: x

common/Geometry2d/Segment.cpp
```diff
@@ -18,6 +18,13 @@ Rect Segment::bbox() const {
 }
 
 float Segment::distTo(const Point& other) const {
+    float dist = nearestPoint(other).distTo(other);
+    if (nearlyEqual(dist, 0)) {
```
justbuchanan: Any reason for the special case for zero here?  We might as well just return `dist` always.

common/Geometry2d/Segment.cpp
```diff
@@ -155,6 +154,7 @@ bool Segment::nearPoint(const Point& point, float threshold) const {
     return d >= 0 && d <= 1;
 }
 
+/*
```
justbuchanan: x

common/Geometry2d/Segment.hpp
```diff
@@ -31,11 +36,14 @@ class Segment : public Line {
...
 
+    /* Returns the relative vector */
+    Point delta() const { return pt[1] - pt[0]; }
+
     /* returns the length of the segment */
     float length() const { return (pt[1] - pt[0]).mag(); }
 
     bool nearPoint(const Point& point, float threshold) const;
-    bool nearPointPerp(const Point& point, float threshold) const;
+    // bool nearPointPerp(const Point& point, float threshold) const;
```
justbuchanan: x

makefile
```diff
@@ -1,4 +1,5 @@
 MAKE_FLAGS=--no-print-directory
+TEST = *
```
justbuchanan: Maybe rename this to GTEST_FILTER so its clearer what it's for

soccer/SystemState.hpp
```diff
@@ -70,9 +70,14 @@ class SystemState {
      */
 
     /** @ingroup drawing_functions */
-    void drawLine(const Geometry2d::Line& line, const QColor& color = Qt::black,
+    void drawLine(const Geometry2d::Segment& line,
```
justbuchanan: Should this be `Line` since we have a separate `drawSegment()` below?

soccer/gameplay/robocup-py.cpp
```diff
@@ -255,15 +272,26 @@ void State_draw_arc(SystemState* self, const Geometry2d::Arc* arc,
     self->drawArc(*arc, Color_from_tuple(rgb), QString::fromStdString(layer));
 }
 
+// TODO(ashaw596) Fix this lie of a function
```
justbuchanan: lol what does this mean?

soccer/gameplay/robocup-py.cpp
```diff
@@ -255,15 +272,26 @@ void State_draw_arc(SystemState* self, const Geometry2d::Arc* arc,
...
 void State_draw_line(SystemState* self, const Geometry2d::Line* line,
                      boost::python::tuple rgb, const std::string& layer) {
     if (line == nullptr) throw NullArgumentException("line");
-    self->drawLine(*line, Color_from_tuple(rgb), QString::fromStdString(layer));
+    self->drawLine(Geometry2d::Segment(*line), Color_from_tuple(rgb),
```
justbuchanan: Why convert from `Line`?

common/Geometry2d/Rect.hpp
ashaw596: I'm not sure that we actually do make this assumption since this function is only used in one place where pt[0]==pt[1] at the start. I thought about that but wasn't sure what we wanted since our pts are mutable anyways so constructors can be bypassed. (I also realized this about the non=existent line asserts) idk if its worth since its such a specific case. (really I would prefer that this not be a function of point since its such an oddly specific use case where its used in the Polygon class).

common/Geometry2d/Segment.cpp
ashaw596: Its assumed by some other functions that distTo()==0 means the point is on the line. On the otherhand I could make an actual onLine function.

makefile
ashaw596: But thats so longgg.

soccer/SystemState.hpp
ashaw596: This is mostly backwards compatibility problems. and meh. I think that drawLine should draw an actually Line as instead of just a line segment, but thats not the behavior expected right now so idk.

soccer/gameplay/robocup-py.cpp
ashaw596: drawLine vs drawSegment. Infinite lines vs segments. idk what to do about these semantics when they're everywhere. Slight semantic clarification and hard to track down python bugs or just leaving it slightly confusing with naming.

common/Geometry2d/Line.hpp
ashaw596: no. Not really a reason to even have an implementation of the implicit copy constructor really.

common/Geometry2d/Line.hpp
ashaw596: Ok. Its not used though. =/

common/Geometry2d/PointTest.cpp
ashaw596: Well, those changes are in my rotation planner stuff. So I can't do this until I merge that branch.

ashaw596:
I think I fixed most of the things. Do you think this is ready @justbuchanan?

README.md
```diff
@@ -87,6 +87,9 @@ Note: The doxygen documentation site above is updated automacally using circle-c
 ## Testing
 We use [gtest](https://code.google.com/p/googletest/) for unit-testing our software, which can be run by running `make tests`.  To add a test to be run with the rest of the bunch, add a new file in soccer/tests.
 
+The soccer tests can be run using `make test-soccer` or firmware tests with `make test-firmware`.
+The TESTS name filter to run only certain tests. For example `make test-soccer TESTS=Point*` runs only the tests for the Point class.
```
justbuchanan: I tried running the command like you have here with the TESTS filter set to Point* and it still ran all of the tests.  I think it's because you set the variable within the makefile.  Maybe you can check in the makefile if it's set and only change it if it's empty?

common/Geometry2d/PointTest.cpp
```diff
@@ -1,19 +1,218 @@
 #include <gtest/gtest.h>
-#include <Geometry2d/Point.hpp>
-
+#include "Geometry2d/Point.hpp"
+#include "Geometry2d/Util.hpp"
 using namespace std;
 using namespace Geometry2d;
 
+bool floatEqual(float a, float b) {
```
justbuchanan: Let's delete this function and use the one you put in Utils.hpp

common/Geometry2d/Segment.hpp
```diff
@@ -9,17 +9,22 @@
...
 public:
+    /** the Segement consists of two points */
+    Point pt[2];
+
     Segment() {}
 
-    Segment(const Segment& other) : Line(other) {}
-    Segment(Point p1, Point p2) : Line(p1, p2) {}
+    Segment(Point p1, Point p2) : pt{p1, p2} {
+        // assert(p1 != p2);
```
justbuchanan: x

README.md
justbuchanan: Nevermind, I was on the wrong branch... This works fine :+1:

common/Geometry2d/PointTest.cpp
```diff
@@ -1,19 +1,218 @@
...
+    float mag = p0.mag();
+    p0.clamp(1);
+    EXPECT_FLOAT_EQ(1, p0.mag());
+    EXPECT_PRED2(pointEqual, p0, Point(-20.5 / mag, 1.5 / mag));
 }
 
-//  if you clamp a vector to a value that's bigger than it's magnitued, it
-//  shouldn't change
+//  if you clamp a vector to a value that's bigger
+//  than it's magnitued, it shouldn't change
```
justbuchanan: *magnitude

soccer/SystemState.hpp
justbuchanan: Hmm, if we have both a drawLine and a drawSegment method, we should probably make them do what the name says.  You're right though that we almost always want to draw a segment.  Can we change the calls to drawLine to be drawSegment?

justbuchanan: Meh, I guess it's ok as is.  I don't particularly like it, but it's not really causing any issues.  Maybe just put a comment above drawLine and drawSegment to clarify what each does.

justbuchanan:
I made a couple minor notes, but looks good!  Go ahead and merge when you're ready.

common/Geometry2d/PointTest.cpp
ashaw596: removed

common/Geometry2d/Segment.hpp
ashaw596: xd

common/Geometry2d/PointTest.cpp
ashaw596: ^

justbuchanan: ?

ashaw596: Nothing. it means what you said. (its pointing at your comment)

PR #540, opened 2016-02-10
Adds configurable values for robot speed and rotation, closes #343

### LINKED ISSUE #343: Use config system for damped joystick speeds rather than hard-coded values and increase them

jcarn:
Also adds configurable values for the damped joystick rotation and translation. The starting values of the configurable values may be adjusted in Joystick.cpp

EDIT: Just realized that when I rolled back some changes I also deleted all my comments, going through and putting those in right now


justbuchanan:
Looks good!  Let me know when you get the comments back in and I'll look over it again.

Issue #435, opened 2015-09-09
Better logging for soccer


justbuchanan:
In our firmware, we have functions/macros for logging that allow specifying the log level (DEBUG, INFO, WARNING, ERROR) - we should use a similar setup for soccer.  Bonus points if we can use the exact same code, but compile it for both platforms.

The main benefit of this is that we can set the log level lower when we want a lot of debug info, but set it higher for normal operation so we only see important messages.

Issue #166, opened 2014-12-03
Create routine for estimating camera latency


justbuchanan:
It is possible with some fancy motion and math to calculate the delay between the camera taking a picture and our system receiving the coordinates of each robot.  If we know this delay more accurately, we can do a better job of using the camera data in our motion control system and predicting the state of the current field.  We can use the camera data + our prediction for what has happened during the delay.

Stanford described their method for doing this in their TDP, which can be found here:

http://robocupssl.cpe.ku.ac.th/_media/robocup2014:tdp:stanfordrc-2014.pdf

justbuchanan:
@ashaw596 what's the latest on this?  Are we just waiting to test on real robots?  Also, what branch is this on?

ashaw596:
Yeah. Basically just waiting for testing on real robots

ashaw596:
It's on the albert/latency

ashaw596:
@jpfeltracco @CollinAvidano 
Bump before I forget about this again. I had a mostly implemented play for helping calculate latency as well as something  implemented to account for it a bit.

JNeiger:
@liquidmetal9015 This is implemented in your benchmark thing right?

PR #544, opened 2016-02-17
Add resizing of soccer based off of the geometry vision packets (if they are filled out)

### LINKED ISSUE #334: Pull field dimensions from vision packet
SSL-Vision now includes field geometry in vision packets. We should at least add the option of pulling field constants from this data. This will help smooth over the pains of the inevitably erroneous fields at competition.

jgkamat:
I think I've gotten most of the kinks worked out, but I haven't tested this out in any way whatsoever. I'm going to try it out on the field and maybe try to get grsim working with it. This is just what I have so far.

Closes #334


justbuchanan:
Looks good so far!

One way you could test it, in addition to actually trying it out, is to break the geometry-packet parsing stuff into its own function that takes in a packet and returns a Field_Dimensions object.  Then you can add a unit test for it and make sure it's getting the arcs and lines right.  Protobuf has a text format that more or less looks like json, so you could hard-code a sample packet into the unit test file.

justbuchanan:
@jgkamat did you still want me to look at something with the protobuf messages or did you get it figured out?

jgkamat:
Yah I'm still working on it unfortunately...

[this is where I traced it to](https://github.com/RoboJackets/robocup-software/blob/master/soccer/VisionReceiver.cpp#L84) (and here I'm still seeing the problem of not seeing the repeated feilds be filled...)

justbuchanan:
@jgkamat  I messed around with this some and I think I got it working.  I noticed we were listening on the old vision port, which doesn't transmit the new version of the geometry packets.  Switching ports seemed to fix it and I verified that lines/arcs are coming through.  Here's a commit: 6a78d5a8b35e5b9b5f34b169146cc03fa20785e1

jgkamat:
Ah I should have looked at that way sooner... I'll get to work on finishing up the finishing touches on this pr soon!

jgkamat:
I think I've finally got it! :tada: 
One thing to note though, the `Border` feild is not updated automatically as no information about it is included in the vision packet (although it's just the distnace from the lines to the edge of the field, so it shouldnt be a problem).

Ready for review (minus style I guess).

common/Field_Dimensions.cpp
```diff
@@ -9,4 +9,7 @@ const Field_Dimensions Field_Dimensions::Double_Field_Dimensions(
     1.000f, 0.500f, 1.000f, 0.500f, 10.400f, 7.400f);
 
 Field_Dimensions Field_Dimensions::Current_Dimensions =
-    Field_Dimensions::Double_Field_Dimensions;
\ No newline at end of file
+    Field_Dimensions::Double_Field_Dimensions;
+
+const Field_Dimensions Field_Dimensions::Default_Dimensions =
```
justbuchanan: This doesn't look like it's used anywhere.  We might as well leave it out and just use Current_Dimensions.

soccer/Processor.cpp
```diff
@@ -40,6 +41,15 @@ RobotConfig* Processor::robotConfig2015;
 std::vector<RobotStatus*>
     Processor::robotStatuses;  ///< FIXME: verify that this is correct
 
+Field_Dimensions* currentDimensions = &Field_Dimensions::Current_Dimensions;
+
+// Joystick speed limits (for damped and non-damped mode)
```
justbuchanan: Are these joystick values a possible merge conflict artifact?  I think these values used to live here, but have since become config values in Joystick.cpp.

soccer/Processor.cpp
```diff
@@ -295,10 +305,26 @@ void Processor::run() {
             log->CopyFrom(packet->wrapper);
 
             curStatus.lastVisionTime = packet->receivedTime;
+
+            if (packet->wrapper.has_geometry()) {
+                // DEMO: Test out field sizes
+                const SSL_GeometryFieldSize fieldSize = packet->wrapper.geometry().field();
+                // FIXME - Account for network latency
+
+                if (fieldSize.field_length() != 0 &&
```
justbuchanan: It would be best if we do a more thorough comparison here.  One potential area I could see this going wrong is if the length of the field at comp matches our default setup, but the field construction people screwed up the width (they always screw up something).  As is, we wouldn't detect and would just stick with our incorrect default width.

A good solution would be to implement `operator==()` on `Field_Dimensions` and use it here.

soccer/Processor.cpp
```diff
@@ -582,6 +608,59 @@ void Processor::run() {
...
+    if (fieldSize->field_lines_size() == 0) {
+        return;
+    }
+
+    const SSL_FieldCicularArc* penalty = nullptr;
+    const SSL_FieldCicularArc* center = nullptr;
+    float displacement = 0.500f;  // default displacment
+
+    for (int i = 0; i < fieldSize->field_arcs().size(); i++) {
+        if (boost::iequals(fieldSize->field_arcs().Get(i).name(), "CenterCircle")) {
```
justbuchanan: The `name()` method returns a `std::string`, so you can just use `==` here instead of boost.

soccer/Processor.cpp
```diff
@@ -582,6 +608,59 @@ void Processor::run() {
...
+            // Assume center circle
+            center = &fieldSize->field_arcs().Get(i);
+        } else if (boost::iequals(fieldSize->field_arcs().Get(i).name(), "LeftFieldLeftPenaltyArc")) {
+            penalty = &fieldSize->field_arcs().Get(i);
+        }
+    }
+
+    for (int i = 0; i < fieldSize->field_lines().size(); i++) {
+        if (boost::iequals(fieldSize->field_lines().Get(i).name(), "RightPenaltyStretch")) {
```
justbuchanan: Same as above

soccer/Processor.cpp
```diff
@@ -295,10 +305,26 @@ void Processor::run() {
...
+                // FIXME - Account for network latency
+
+                if (fieldSize.field_length() != 0 &&
+                    // Compare floats with a small threshold
+                    (currentDimensions->Length() - (fieldSize.field_length() / 1000.0f)) > 0.001f) {
+                    // Set the changed field dimensions to the current ones
+                    decodeGeometryPacket(&fieldSize);
+                }
+                else  {
+                    decodeGeometryPacket(&fieldSize); // TODO REMOVE FORCE OVERRIDE
```
justbuchanan: x

soccer/Processor.cpp
```diff
@@ -295,10 +305,26 @@ void Processor::run() {
...
+            if (packet->wrapper.has_geometry()) {
+                // DEMO: Test out field sizes
+                const SSL_GeometryFieldSize fieldSize = packet->wrapper.geometry().field();
+                // FIXME - Account for network latency
+
+                if (fieldSize.field_length() != 0 &&
+                    // Compare floats with a small threshold
+                    (currentDimensions->Length() - (fieldSize.field_length() / 1000.0f)) > 0.001f) {
+                    // Set the changed field dimensions to the current ones
+                    decodeGeometryPacket(&fieldSize);
```
justbuchanan: It would be slightly better if `decodeGeometryPacket()` just did the decoding as its named.  Then you can do `setFieldDimensions(decodeGeometryPacket(&fieldSize))` here

soccer/Processor.cpp
```diff
@@ -582,6 +608,59 @@ void Processor::run() {
...
+                               penalty->radius() / 1000.0f + adj,       // PenaltyDist
+                               0.010f,                                  // PenaltyDiam
+                               penalty->radius() / 1000.0f + adj,       // ArcRadius
+                               center->radius() / 1000.0f + adj,        // CenterRadius
+                               (center->radius()) * 2 / 1000.0f + adj,  // CenterDiameter
+                               displacement / 1000.0f,                  // GoalFlat
+                               (fieldSize->field_length() / 1000.0f + (fieldBorder) * 2),
+                               (fieldSize->field_width() / 1000.0f + (fieldBorder) * 2)));
+    } else {
+        // TODO log errors here
```
justbuchanan: This would be good to do.  You can just do something like: `cerr << "Error: failed to decode SSL geometry packet" << endl;`

soccer/Processor.cpp
justbuchanan: Actually since `decodeGeometryPacket()` only sets it if decoding is successful, maybe keep it as-is, but rename it to reflect the fact that it's updating current dimensions in addition to doing the decoding.

soccer/Processor.cpp
```diff
@@ -582,6 +608,59 @@ void Processor::run() {
...
+        // Force a resize
+        // TODO fix hardcoded values here
+        setFieldDimensions(Field_Dimensions (
+                               fieldSize->field_length() / 1000.0f, fieldSize->field_width() / 1000.0f,
+                               fieldBorder,
+                               thickness,
+                               fieldSize->goal_width() / 1000.0f, fieldSize->goal_depth() / 1000.0f,
+                               0.160f,                                  // Goal Height
+                               penalty->radius() / 1000.0f + adj,       // PenaltyDist
+                               0.010f,                                  // PenaltyDiam
```
justbuchanan: For this and goal height, let's add `static const float Default___` values to the Field_Dimensions class so we don't have these constants both here and in `Field_Dimensions.cpp`.

justbuchanan:
It's looking good!  It'll be nice to not have to manually change constants when the field is inevitably built wrong at comp haha.

common/Field_Dimensions.cpp
jgkamat: I think that I could use this in the cases where I have to hardcode the values (I think that would be cleaner than adding a bunch of variables), would that be ok?

justbuchanan: Yeah, I like it!

justbuchanan: It may make more sense then to initialize these with: default = double; current = default;

soccer/Processor.cpp
jgkamat: yup, not sure how that got in here, I'm removing it.

common/Field_Dimensions.hpp
```diff
@@ -96,6 +103,28 @@ struct Field_Dimensions {
...
+                 std::abs(Border() - a.Border()) > FLT_EPSILON ||
+                 std::abs(LineWidth() - a.LineWidth()) > FLT_EPSILON ||
+                 std::abs(GoalWidth() - a.GoalWidth()) > FLT_EPSILON ||
+                 std::abs(GoalDepth() - a.GoalDepth()) > FLT_EPSILON ||
+                 std::abs(GoalHeight() - a.GoalHeight()) > FLT_EPSILON ||
+                 std::abs(PenaltyDist() - a.PenaltyDist()) > FLT_EPSILON ||
+                 std::abs(PenaltyDiam() - a.PenaltyDiam()) > FLT_EPSILON ||
+                 std::abs(ArcRadius() - a.ArcRadius()) > FLT_EPSILON ||
+                 std::abs(CenterRadius() - a.CenterRadius()) > FLT_EPSILON ||
+                 std::abs(CenterRadius() - a.CenterRadius()) > FLT_EPSILON ||
```
justbuchanan: duplicate

soccer/Processor.cpp
```diff
@@ -2,6 +2,7 @@
 #include <QMutexLocker>
 #include <poll.h>
 
+#include <boost/algorithm/string.hpp>
```
justbuchanan: x

soccer/Processor.cpp
```diff
@@ -296,10 +299,19 @@ void Processor::run() {
             log->CopyFrom(packet->wrapper);
 
             curStatus.lastVisionTime = packet->receivedTime;
+
+            if (packet->wrapper.has_geometry()) {
+                // DEMO: Test out field sizes
```
justbuchanan: x

soccer/Processor.cpp
```diff
@@ -296,10 +299,19 @@ void Processor::run() {
...
+                const SSL_GeometryFieldSize fieldSize =
+                    packet->wrapper.geometry().field();
+                // FIXME - Account for network latency
```
justbuchanan: We don't have to do anything with latency here, right?

soccer/Processor.cpp
```diff
@@ -296,10 +299,19 @@ void Processor::run() {
             log->CopyFrom(packet->wrapper);
 
             curStatus.lastVisionTime = packet->receivedTime;
+
+            if (packet->wrapper.has_geometry()) {
+                // DEMO: Test out field sizes
+                const SSL_GeometryFieldSize fieldSize =
```
justbuchanan: If you make this a reference, it'll save us from making a copy.

soccer/Processor.cpp
```diff
@@ -590,6 +602,74 @@ void Processor::run() {
...
+    if (fieldSize->field_lines_size() == 0) {
+        return;
+    }
+
+    const SSL_FieldCicularArc* penalty = nullptr;
+    const SSL_FieldCicularArc* center = nullptr;
+    float displacement =
+        Field_Dimensions::Default_Dimensions.GoalFlat();  // default displacment
+
+    for (int i = 0; i < fieldSize->field_arcs().size(); i++) {
```
justbuchanan: Note that an easier way to do loops over elements in a collection is with this syntax:

~~~{.cpp}
for (const SSL_FieldCircularArc& arc : fieldSize->field_arcs()) {

}
~~~

soccer/Processor.cpp
```diff
@@ -590,6 +602,74 @@ void Processor::run() {
...
+        if (fieldSize->field_arcs().Get(i).name() == "CenterCircle") {
+            // Assume center circle
+            center = &fieldSize->field_arcs().Get(i);
+        } else if (fieldSize->field_arcs().Get(i).name() ==
+                   "LeftFieldLeftPenaltyArc") {
+            penalty = &fieldSize->field_arcs().Get(i);
+        }
+    }
+
+    for (int i = 0; i < fieldSize->field_lines().size(); i++) {
```
justbuchanan: You can make this a little shorter by combining this loop with the one above.

soccer/Processor.cpp
```diff
@@ -590,6 +602,74 @@ void Processor::run() {
...
+            penalty->radius() / 1000.0f + adj,       // ArcRadius
+            center->radius() / 1000.0f + adj,        // CenterRadius
+            (center->radius()) * 2 / 1000.0f + adj,  // CenterDiameter
+            displacement / 1000.0f,                  // GoalFlat
+            (fieldSize->field_length() / 1000.0f + (fieldBorder)*2),
+            (fieldSize->field_width() / 1000.0f + (fieldBorder)*2));
+
+        if (newDim != *currentDimensions) {
+            // Set the changed field dimensions to the current ones
+            cerr << "Updating field geometry based off of vision packet."
```
justbuchanan: This should go to `cout` rather than `cerr`.

soccer/Processor.cpp
justbuchanan: nevermind, I can't read...

soccer/Processor.cpp
jgkamat: This was there before I started, I think i moved it up here by mistake. It was [here](https://github.com/RoboJackets/robocup-software/pull/544/files#diff-fefaade3f2ed6066fb659cd4ff181154L302). I'm not sure if it's relevant there, but I'll put it back if it is.

soccer/Processor.cpp
jgkamat: Whoops, thanks!

justbuchanan: We don't have much of a logging system setup, although it would be nice!  See #435

soccer/Processor.cpp
justbuchanan: Ok gotcha.  Yeah if you want to just delete it, that would be fine.  Latency calculation has been a long-running todo item #166 , #293

justbuchanan:
LGTM :+1: .  Go ahead and merge when ready.

jgkamat:
Nice, thanks! Fingers crossed that this dosent break anything at comp :P

PR #545, opened 2016-02-18
added `make robot2015-gdb`, removed gdb-server.py

justbuchanan:


jjones646:
@justbuchanan I think I figured out how to fix the whole infinite breakpoint thing. I moved a few things around and it's currently setup to use a flag I added when there's multiple boards (my pr [here](https://github.com/mbedmicro/pyOCD/pull/221)).

Give it a try and see if everything still works for you. You can also call `make robot2015-gdb GDB_NO_CONN=1` to start the pyocd-server without actually connecting to it. Then anyone can connect to it over ip in gdb.

makefile
```diff
@@ -84,6 +84,31 @@ robot2015:
...
 
+GDB_PORT ?= 3333
+.INTERMEDIATE: build/robot2015-gdb.pid
+build/robot2015-gdb.pid:
+# this will cache sudo use without a password in the environment
+# so we won't enter the gdb server and skip past the password prompt.
+	@sudo echo "starting pyocd-gdbserver, logging to build/robot2015-gdb.log"
+# now we can refresh the sudo timeout and start up the gdb server
+	sudo -v && { sudo pyocd-gdbserver --allow-remote --port $(GDB_PORT) --reset-break \
+	--target lpc1768 -S -G -B 0 > build/robot2015-gdb.log 2>&1 & sudo echo $$! > $@; }
```
justbuchanan: @jjones646, I think we should be using the `>>` operator rather than `>` here.  The double caret one appends to the file's existing contents, while the single caret overwrites the previous contents.

justbuchanan: Also, I'm getting an error saying that the `-B 0` option isn't recognized.  Should it be lowercase `b` to select the board number?

makefile
```diff
@@ -84,6 +84,31 @@ robot2015:
...
+
+GDB_NO_CONN ?= 0
```
justbuchanan: What's `GDB_NO_CONN` for?

PR #547, opened 2016-02-22
reran formatters on codebase now that version changed

justbuchanan:
The python code formatter has been updated several versions since we originally ran it on the codebase, so it has many changes it wants to make.  I went ahead and re-ran `stylize` on all files (not just those that differ from master) to take care of it once rather than having occasional changes come through when files are touched.

@jgkamat For some reason this is passing the style check, which surprises me.  The version inconsistencies between my computer and the CI should definitely cause issues.  The log from `make checkstyle` shows that zero files were checked...  Any thoughts on why this might happen?  It may be a problem with stylize, but could also be something CI or git-related.  It's using the --diffbase=master flag, so it should be comparing as many files as are changed in this PR.


ashaw596:
By the way, is there a way to make the diffbase master...my_branch_name so we diff from the point we branch off instead of random changes from the diffs in master?

justbuchanan:
Hmm, maybe so.  We could change it so that it finds the most recent commit it has in common with master and uses that.

jgkamat:
It seems like [only the python files were checked](https://circle-artifacts.com/gh/RoboJackets/robocup-software/1259/artifacts/0/tmp/circle-artifacts.Lbhv1VC/style.txt).

I'm not sure what would cause this... It seems to check the cpp files for other PR's.

jgkamat:
I think I found our problem, the master branch update was getting cached to a version newer (?!) than the actual master. This should fix it.

justbuchanan:
Interesting...  When I first pushed this branch to github, my local master did have this commit on it.  I removed it right afterwards since I didn't want it on master yet, but maybe something very strange happened.  This should be a good fix though.

jgkamat:
Got it now, the final result should be up soon.

justbuchanan:
:+1:

PR #548, opened 2016-02-22
Added make robot2015-gdb

justbuchanan:
Re-opening this.  The previous PR got invalidated because the `robot2015-firmware` branch disappeared.


justbuchanan:
@jjones646, I removed the `-B` flag for now since it's not in the released version of pyOCD, but go ahead and put it back in once your PR is merged.

jjones646:
@justbuchanan It might not get merged. You can read the comments in the pr, but they don't want to add it - which is pretty understandable. So I kinda explained what the actual problem is and am hoping they have a better suggestion.

justbuchanan:
Ok, I guess we'll see what they say.

Btw, I've been having an issue where it will occasionally fail to remove the pid file.  When that happens, it sees that the pid file exists next time I try to run the make command and it never starts up the gdb server.  I get around it by just deleting the pid file, but I'm not sure how to actually fix the issue.

justbuchanan:
@jjones646, I've noticed that the `ls` command doesn't work when I'm running code via gdb.  The command runs but doesn't print anything.  Any ideas?

jjones646:
@justbuchanan It's possibly because each target in a makefile is run in its own subshell? I also had a lot of problems with the pid file. Launching it with screen or something similar may work better. I'm not a big fan of having a pid file in the repo since it'll gone if you run make clean during the session. I forgot why I made 2 targets for it, but it could be brought back into a single target block for sure with the trap setup.

justbuchanan:
Yeah I'm not sure.  I verified that the trap part does get called at least sometimes.  And the makefile INTERMEDIATE bit usually works.  It doesn't fail often, so we can just keep it as is unless you can find something better.

PR #549, opened 2016-02-23
fixed issues with diffbase and ci checkstyle

justbuchanan:
@jgkamat, the fix you had worked great for non-master branches, but seems to fail when the branch being CI'd is the master branch.  See the current [checkstyle log](https://circle-artifacts.com/gh/RoboJackets/robocup-software/1270/artifacts/0/tmp/circle-artifacts.dwZ0bAA/style.txt).

I think this should fix it up.  origin/master should always be up-to-date, it's just the ci-local version of master that wasn't always updated when the CI was running for a non-master branch.


jgkamat:
Hmm, whoops, I didn't think about that. I don't think this would keep the behavior of master being updated though (we would need to add a `git fetch origin` before the command), but with that, I think this would work.

jgkamat:
I don't think origin would be up to date by default, because circle caches git repos and then checks out a speicific commit (it never does a fetch origin or a pull).

justbuchanan:
Ahh ok, yeah that sounds right.  I'll go ahead and add a fetch back to the beginning.

PR #550, opened 2016-02-24
Quicksave and Quickload

### LINKED ISSUE #530: Add ability to save and restore robot locations to soccer
Currently, we can drag robots and reset the robot positions. It would be great if there was a way to save the current state of the robots and restore it in to allow easier testing plays and things in simulator.

chachmu:
Solves issue with Stop Robots not working. Adds quicksave and quickload functions under Simulation. Closes  #530

### DIFF
--- a/soccer/FieldView.hpp
+++ b/soccer/FieldView.hpp
@@ -41,7 +41,10 @@ public:
 
     void rotate(int value);
 
-    const Geometry2d::TransformMatrix& getTeamToWorld() { return _teamToWorld; }
+    const Geometry2d::TransformMatrix& getTeamToWorld() const {
+        return _teamToWorld;
+    }
+
     // True if this control is showing live (vs. historical) data.
     // If false, it will draw a red border.
     bool live;
--- a/soccer/FieldView.hpp
+++ b/soccer/FieldView.hpp
@@ -41,6 +41,9 @@ public:
 
     void rotate(int value);
 
+    const Geometry2d::TransformMatrix& getTeamToWorld(){
+        return _teamToWorld;
+    }
     // True if this control is showing live (vs. historical) data.
     // If false, it will draw a red border.
     bool live;
@@ -50,6 +53,7 @@ public:
     bool showCoords;
     bool showDotPatterns;
     bool showTeamNames;
+
 
 protected:
     virtual void paintEvent(QPaintEvent* e) override;
--- a/soccer/FieldView.hpp
+++ b/soccer/FieldView.hpp
@@ -41,9 +41,7 @@ public:
 
     void rotate(int value);
 
-    const Geometry2d::TransformMatrix& getTeamToWorld(){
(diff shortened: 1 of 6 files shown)


soccer/FieldView.hpp
```diff
@@ -41,6 +41,7 @@ class FieldView : public QWidget {
 
     void rotate(int value);
 
+    const Geometry2d::TransformMatrix& getTeamToWorld() { return _teamToWorld; }
```
justbuchanan: Let's put a `const` in between `()` and `{` to indicate that calling the method doesn't modify `FieldView`.

soccer/MainWindow.cpp
```diff
@@ -878,24 +877,81 @@ void MainWindow::on_actionStopRobots_triggered() {
...
-        r->mutable_vel()->set_x(0);
-        r->mutable_vel()->set_y(0);
-        r->set_w(0);
+        if (robot->visible) {
+            SimCommand::Robot* r = cmd.add_robots();
+            r->set_shell(robot->shell());
+            r->set_blue_team(processor()->blueTeam());
+            Geometry2d::Point newPos =
+                _ui.fieldView->getTeamToWorld() * robot->pos;
+            r->mutable_pos()->set_x(newPos.x);
```
justbuchanan: Any reason for setting position here?  If we just set the velocity to zero it should keep them wherever they are already.

soccer/MainWindow.hpp
```diff
@@ -204,4 +208,6 @@ private Q_SLOTS:
     std::vector<QPushButton*> _logPlaybackButtons;
 
     RJ::Time _firstLogTimestamp = -1;
+
+    Packet::SimCommand quickLoadCmd;
```
justbuchanan: Rename this to `_quickLoadCmd` to match the naming conventions with the other instance variables.

soccer/MainWindow.cpp
```diff
@@ -878,24 +877,81 @@ void MainWindow::on_actionStopRobots_triggered() {
...
+            r->set_w(0);
+        }
     }
     _ui.fieldView->sendSimCommand(cmd);
 }
 
+void MainWindow::on_actionQuicksaveRobotLocations_triggered() {
+    _ui.actionQuickloadRobotLocations->setEnabled(true);
+    quickLoadCmd.reset();
+    int i = 0;
```
justbuchanan: I don't think the `i` variable is used anywhere

justbuchanan:
Nice work!  I made a few minor notes, but overall looks great!

soccer/MainWindow.cpp
chachmu: For whatever reason SimCommand doesn't work if the position is not set. It just prints "Trying to override non-existent robot".

justbuchanan: Ok, sounds good.

PR #551, opened 2016-02-27
Refactored some firmware bits

justbuchanan:
In my search for the segfault bug in firmware, I ended up re-arranging and cleaning up a few things throughout the firmware codebase.  Here are the main highlights:

* Converted many class methods to instance methods (especially CommModule and Console)
* Switched some things over from raw CMSIS RTOS to MBED's C++ wrapper
* using map instead of a vector to store CommPort objects.  This removes the need to sort
* removed `blackhole_port`
* In a few places, `cleanup()` was called in a class and its superclass, which would have caused a segfault if the objects had been deallocated
* `CommPorts` no longer inherits from `CommPort`
* removed this line: `return *&(*pIt);`
* changed several for loops to the c++11 syle `for (elem : coll)`
* removed demo packet sending form CommTaskThread
* removed several unused instance variables
* removed calls to `osThreadTerminate()` at end of thread functions
* removed `define_thread()` from rtos helpers
* fixed a bug in MCP23017::pinMode() that caused it to always set pins as input
* changed code style to match conventions we've been using throughout the rest of the project
  * lowercase camel-cased methods/variables
  * leading underscore for ivars

@jjones646, take a look through the changes and let me know if I've done anything really stupid or if something was a particular way before for a good reason and I screwed it up.


jjones646:
Is `return *&(*pIt);` a bit too ambiguous? Haha, but for real - I gave this a glance over last night and most of it looks really good. Having someone else look over all of this stuff was really needed, so thanks for doing that.

There's a few things happening with the removal of the `blackhole` port, but I don't think it's anything major and can look over it a bit more with you tomorrow.

Also, is there a doc somewhere for the coding style?

justbuchanan:
Haha I read it about 10 times trying to decide if it was doing something that I didn't understand...  And no problem, it always helps to get a second set of eyes on things.  Overall though I didn't change anything major, just a bunch of small things here and there.

As for the blackhole port stuff, that was the object returned whenever `CommPorts` is asked for a port that doesn't actually exist.  I was thinking we might as well just have it throw an exception since any reference to an invalid port is a bug.  Thoughts?

Hmm, there used to be a doc/example for coding style, but I can't seem to find it...  I'll look around some more and see if we can get something up on our docs site.  In general though, I've just been following the conventions throughout the soccer C++ codebase just to keep things consistent.  We follow the [google style guide](https://google.github.io/styleguide/cppguide.html) for the most part.

justbuchanan:
@jjones646, I think this is good to go.  Let me know if you see any issues.

PR #555, opened 2016-02-28
Add firmware compiler and mercurial to dependencies. Adds a TAP

ashaw596:
Fixes OSX build errors


justbuchanan:
Travis works again!

PR #560, opened 2016-02-29
Made the field edge configurable

### LINKED ISSUE #392: Make the field edge obstacle a config value rather than a hard-coded constant
The code for this is in `soccer/gameplay/GameplayModule.cpp`.

jcarn:
closes #392

### DIFF
--- a/soccer/gameplay/GameplayModule.cpp
+++ b/soccer/gameplay/GameplayModule.cpp
@@ -32,7 +32,7 @@ void GameplayModule::createConfiguration(Configuration* cfg) {
     _fieldEdgeInset = new ConfigDouble(cfg, "Field Edge Obstacle", .3);
 }
 
-const bool GameplayModule::hasFieldEdgeInsetChanged() {
+bool GameplayModule::hasFieldEdgeInsetChanged() const {
     if (abs(_fieldEdgeInset->value() - _oldFieldEdgeInset) >
         numeric_limits<double>::epsilon()) {
         return true;
--- a/soccer/gameplay/GameplayModule.cpp
+++ b/soccer/gameplay/GameplayModule.cpp
@@ -8,9 +8,14 @@
 
 #include <stdio.h>
 #include <iostream>
+#include <cmath>
 
 // for python stuff
 #include "robocup-py.hpp"
+
+using namespace Gameplay;
+
+REGISTER_CONFIGURABLE(GameplayModule);
 
 using namespace std;
 using namespace boost;
@@ -18,12 +23,27 @@ using namespace boost::python;
 
 using namespace Geometry2d;
 
+ConfigDouble* GameplayModule::inset;
+
+void GameplayModule::createConfiguration(Configuration* cfg) {
+    inset = new ConfigDouble(cfg, "Field Edge Obstacle", .3);
+}
+
+bool GameplayModule::hasInsetChanged() {
+    if (abs(inset->value() - oldInset) > numeric_limits<double>::epsilon()) {
+        oldInset = inset->value();
+        return true;
+    }
+    return false;
+}
+
 Gameplay::GameplayModule::GameplayModule(SystemState* state)
     : _mutex(QMutex::Recursive) {
     _state = state;
 
     calculateFieldObstacles();
 
+    oldInset = .3;
     _goalieID = -1;
 
     //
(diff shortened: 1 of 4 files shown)


justbuchanan:
@jcarn, It's currently failing the build because some submodules got changed to older versions.  If you open up the "Files Changed" tab on here, you can see the two files in the "external" folder that have changed.

To fix this, you should be able to do:
~~~{.sh}
git checkout master -- external
git commit -m 'reverted submodule changes'
git push
~~~

To keep submodules from causing issues in the future, check out my post in our [gitter chatroom](https://gitter.im/RoboJackets/robocup-software?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge) about adding post-checkout and post-merge hooks to git to handle them automatically.

soccer/gameplay/GameplayModule.hpp
```diff
@@ -101,6 +101,10 @@ class GameplayModule {
 
     void calculateFieldObstacles();
 
+    bool hasInsetChanged();
```
justbuchanan: Mark this as `const` since it doesn't modify the `GameplayModule` object that it's called on.

soccer/gameplay/GameplayModule.hpp
```diff
@@ -120,6 +124,9 @@ class GameplayModule {
     /// This is held while plays are running.
     QMutex _mutex;
 
+    static ConfigDouble* inset;
+    double oldInset;
```
justbuchanan: Rename this with a leading underscore for consistency.

justbuchanan: Also, since it isn't necessarily clear to someone reading this header file what `inset` refers to, maybe a better name would be `fieldEdgeInset` to give it a little more context.

soccer/gameplay/GameplayModule.cpp
```diff
@@ -286,6 +305,8 @@ void Gameplay::GameplayModule::run() {
     bool verbose = false;
     if (verbose) cout << "Starting GameplayModule::run()" << endl;
 
+    /// check to see if field constants have been changed, recalculate if
```
justbuchanan: ?

soccer/gameplay/GameplayModule.cpp
```diff
@@ -82,26 +102,25 @@ void Gameplay::GameplayModule::calculateFieldObstacles() {
                  TransformMatrix::rotate(M_PI);
 
     // TODO(barulicm): describe what @inset is
```
justbuchanan: x

soccer/gameplay/GameplayModule.cpp
```diff
@@ -8,22 +8,42 @@
...
+    return false;
+}
+
 Gameplay::GameplayModule::GameplayModule(SystemState* state)
     : _mutex(QMutex::Recursive) {
     _state = state;
 
     calculateFieldObstacles();
 
+    oldInset = .3;
```
justbuchanan: It would probably be better to set it equal to the config value here rather than a constant.

soccer/gameplay/GameplayModule.cpp
```diff
@@ -8,22 +8,47 @@
...
+    // this sets the disance from the field boundries to the edge of the global
+    // obstacles, which the
+    // robots will not move through or into
+    _fieldEdgeInset = new ConfigDouble(cfg, "Field Edge Obstacle", .3);
+}
+
+const bool GameplayModule::hasFieldEdgeInsetChanged() {
+    if (abs(_fieldEdgeInset->value() - _oldFieldEdgeInset) >
+        numeric_limits<double>::epsilon()) {
+        _oldFieldEdgeInset = _fieldEdgeInset->value();
```
justbuchanan: Setting the value of `_oldFieldEdgeInset` should probably be at the end of `calculateFieldObstacles()` instead of here.  As it is now, if the config value changes, a call to `hasFieldEdgeInsetChanged()` will return true, but another call right afterwards will return false.  It would be better if it returned true until the obstacles have been updated.

soccer/gameplay/GameplayModule.hpp
```diff
@@ -101,6 +102,10 @@ class GameplayModule {
 
     void calculateFieldObstacles();
 
+    const bool hasFieldEdgeInsetChanged();
```
justbuchanan: The const should go at the end here: `bool hasFieldEdgeInsetChanged() const;`

PR #561, opened 2016-02-29
O SX n' Firmware

jjones646:
#392 Definitely have to be easy with merging this into master, but it actually works this time around...I think.

Still a slight bit of cleaning up to do, but let's see if this will work for others at the moment.

- [x] merge the mbed-copy` scripts into one and make the cmake change for it
- [x] Add colorama to requirements2.txt
- [ ] Fix build warning: `warning: unknown warning option '-Werror=return-local-addr'; did you mean '-Werror=return-stack-address'? [-Wunknown-warning-option]`
- [x] cite source for `MbedTargets.cmake`


util/requirements2.txt
```diff
@@ -4,4 +4,9 @@ jinja2   # A dependency for the mbed build library
 colorama # A dependency for the mbed build library
 cpp-coveralls # A dependency for coverage testing
 
-pyOCD # for debugging code on MBED
+# firmware - required
+mbed-ls
```
justbuchanan: Until they merge your PR, you can change this to:

~~~{.sh}
git+https://github.com/jjones646/mbed-ls@issue_38_squashed
~~~

util/mbed/mbed_cmake_vargen.py
```diff
@@ -0,0 +1,131 @@
+#!/usr/bin/env python
```
justbuchanan: With the "shebang" set like this, it will run on some systems with python2 and others with python3.  This is because on many newer systems, `python` is just a symlink to `python3`.

This script looks like it only works with python2, so if you cange the first line to `#!/usr/bin/env python2`, it will work on all systems without issue.

util/mbed/mbed-copy-osx.py
```diff
@@ -0,0 +1,57 @@
+#!/usr/bin/env python
```
justbuchanan: Same comment about python versioning, if you can put a 2 or 3 after `python` to indicate which one it works with, it'll prevent version issues on different systems.

util/mbed/mbed-copy-osx.py
```diff
@@ -0,0 +1,57 @@
...
+
+import os
+import time
+import shutil
+import serial
+import argparse
+import mbed_lstools as mbedls
```
justbuchanan: Now that we're using mbed_lstools, would it be easy to use a single script for copying on OSX and Linux?

firmware/robot2015/hw-test/CMakeLists.txt
```diff
@@ -36,15 +36,15 @@ if(HW_TEST_UNIT AND NOT(HW_TEST_UNIT MATCHES "default"))
...
+        ${ARM_OBJCOPY} -O binary robot2015_test_ELF ${TEST_UNIT_BIN_FILE}
+        COMMAND ${CMAKE_COMMAND} -E copy ${CMAKE_CURRENT_BINARY_DIR}/${TEST_UNIT_BIN_FILE} ${PROJECT_SOURCE_DIR}/run/
         DEPENDS robot2015_test_ELF
         COMMENT "Objcopying for making the ${HW_TEST_UNIT} hardware test"
     )
     set_directory_properties(PROPERTIES ADDITIONAL_MAKE_CLEAN_FILES ${PROJECT_SOURCE_DIR}/run/${TEST_UNIT_BIN_FILE})
 
     add_custom_target(robot2015-test-prog
-        arm-none-eabi-objcopy -O binary robot2015_test_ELF ${TEST_UNIT_BIN_FILE}
+        ${ARM_OBJCOPY} -O binary robot2015_test_ELF ${TEST_UNIT_BIN_FILE}
```
justbuchanan: This objcopy should be removed since it's already handled in the `robot2015-test` target above.

justbuchanan:
What is the `firmware/mbed/mbedmicrontroller_141212.if` file that was added?

cmake/MbedTargets.cmake
```diff
@@ -0,0 +1,1299 @@
...OT_F411RE" "SAMD21J18A" "KL46Z" "SAML21J18A" "ARCH_LINK_OTA" "NRF51_DK_BOOT" "NRF51_DONGLE" "EFM32HG_STK3400" "LPC11U68" "LPC11U35_401" "DELTA_DFCM_NNN40_BOOT" "LPCCAPPUCCINO" "XADOW_M0" "LPC11U37H_401" "NRF51_MICROBIT_B_OTA" "XBED_LPC1768" "NUCLEO_L152RE" "DISCO_F429ZI" "KL26Z" "EFM32LG_STK3600" "LPC810" "LPC812" "NRF51_MICROBIT_B_BOOT" "NUCLEO_F070RB" "LPC1114" "NUCLEO_F072RB" "SEEED_TINY_BLE_BOOT" "MICRONFCBOARD" "MTS_DRAGONFLY_F411RE" "WIZWIKI_W7500ECO" "EFM32WG_STK3800" "ARM_MPS2_M0P" "DISCO_F051R8" "NUCLEO_F446RE" "ARM_MPS2_BEID" "NUCLEO_F410RB" "NUCLEO_F042K6" "NUCLEO_F103RB" "EFM32ZG_STK3200" "EFM32GG_STK3700" "NUCLEO_L476RG" "ARM_MPS2" "ARCH_LINK" "NRF51_MICROBIT_B" "NUCLEO_F030R8" "NRF51_DK_OTA" "RBLAB_BLENANO_OTA" "DISCO_F100RB" "NUCLEO_F303K8" "LPC11U35_501_IBDAP" "WIZWIKI_W7500P" "K64F" "LPC11U35_501" "NZ32SC151" "LPC2368" "ARCH_LINK_BOOT" "WALLBOT_BLE_OTA" "ARCH_BLE_OTA" "NUCLEO_L073RZ" "MTS_GAMBIT" "HRM1017" "LPC1768" "LPC4088" "WIZWIKI_W7500" "SAMD21G18A" "NRF51_DONGLE_OTA" "SAMR21G18A" "NRF51_DONGLE_BOOT" "NRF51_MICROBIT_OTA" "LPC11U35_Y5_MBUG" "LPC11U37_501" "DISCO_F469NI" "NRF51_DK" "HRM1017_BOOT" "HRM1017_OTA" "KL25Z" "NUCLEO_F031K6" "SSCI824" "MTS_MDOT_F405RG")
```
justbuchanan: We should probably cite a source for this file in a comment at the top, just for future reference.

justbuchanan:
I'm not sure why this is failing the firmware build on circleci right now, it builds fine on my computer.  The error from the logs is `make[1]: *** No rule to make target `robot2015'.  Stop.`, which is strange because there definitely is a target with that name... @jgkamat, any thoughts?

Overall looks good though!  It will be nice being able to do everything from both OS X and Linux.

I think we can definitely merge the two platform-specific mbed-copy scripts into one without too much trouble, especially now that we're using mbed-ls.  Want me to take a look at doing this, or do you want to @jjones646?

justbuchanan:
Also, you should add `make robot2015` and `make kicker2015` to travis-ci, so we automatically test that those compile on OS X.  If you open up `.travis.yml`, you can add these to the `script: ` line.

justbuchanan:
I looked into the make issues again, and I think we need to either change the CMake setup around a good bit or go back to putting the firmware builds in its own `firmware` subdirectory under build.

Running `make robot2015` works fine, but if you run it after running regular `make` it doesn't.  It seems to be because the firmware and non-firmware targets are essentially separate CMake setups.

cmake/MbedTargets.cmake
jjones646: :thumbsup:

firmware/robot2015/hw-test/CMakeLists.txt
jjones646: ahh, missed this one

util/osx-packages.txt
```diff
@@ -21,5 +22,22 @@ bullet --with-shared
...
 clang-format
 graphviz
-px4/px4/gcc-arm-none-eabi
-hg
\ No newline at end of file
+
+# firmware - required
+hg
+python
+gcc-arm-none-eabi-49    # requires `brew tap PX4/homebrew-px4`
```
jjones646: `avr-libc` needs to be changed to the brew cask name, and `gcc-arm-none-eabi` reverted back to what it was originally

util/mbed/mbed-copy-osx.py
jjones646: yes, it probably will work as it stands now - haven't tested it on anything besides os x tho

makefile
```diff
@@ -13,8 +13,8 @@ endef
 # because CMake can only handle one toolchain at a time, so we build the MBED-
 # targeted code separately.
 define cmake_build_target_fw
-	mkdir -p build/firmware
-	cd build/firmware && cmake -Wno-dev --target $1 $2 ../.. && make $1 $(MAKE_FLAGS)
```
jjones646: @justbuchanan Do you know if this would have anything to do with the tests failing?

justbuchanan: Yeah I think this is the issue.  The comment above the function describes the issue a bit.  There may be a better way to do this, but this is the best solution I know of.

util/mbed/mbed-copy-osx.py
jjones646: `#!/usr/bin/env `:snake::snake:, got it

makefile
jjones646: alright. I moved it since the build files looked kinda awkward being double nested in a firmware directory, but didn't think about the tests. I think we'll be able to live with this getting reverted to how it used to be.

jjones646:
> I think we can definitely merge the two platform-specific mbed-copy scripts into one without too much trouble, especially now that we're using mbed-ls. Want me to take a look at doing this, or do you want to @jjones646?

Hack away at it. There still may be some issues with the mbed resetting to the new binary too.

jjones646:
> What is the firmware/mbed/mbedmicrontroller_141212.if file that was added?

It's the most up to date mbed interface firmware binary from mbed.com. I don't believe it's anywhere besides their website so I threw it in here since it can't be compiled with gcc as it stands now. It would be good if the prog script could detect the mbed interface version and do an auto update of it when needed with it.

jjones646:
This should be mostly ready besides some cleanup things.

justbuchanan:
Sweet, I tried it out on my Mac and it seems to all be working!

I added a couple minor TODOs to the top, but overall it looks good.

I tried to start on a unified mbed-copy script, but quickly realized that `mbed-ls` doesn't list mbeds that aren't mounted to the filesystem, which is kind of a bummer for me since I have auto-mounting turned off on my Linux machine.  I opened an issue on their github page (https://github.com/ARMmbed/mbed-ls/issues/56), but haven't heard anything yet.  If they accept it, I'll send them a PR and then take a look at improving our copy script.

justbuchanan:
I went ahead and sent them a PR for listing unmounted MBEDs.  It seems like it'll probably take a while to (maybe) get merged, so in the meantime, want to make a branch on your fork that contains your Capitan stuff and [my changes](https://github.com/justbuchanan/mbed-ls/tree/list-unmounted-mbeds) and we can just use that version for now?

justbuchanan:
@jjones646 how close is this to being ready to merge?

jjones646:
@justbuchanan really close. I need to merge your pr into this for the unified copy script and push it into here. I'll do that in just a bit then it should be ready i think

jjones646:
@justbuchanan It's merged. I just realized there was a big green button to do that on GitHub...

justbuchanan:
Haha yep, it's pretty handy.  Looks good to me!  I'll go ahead and click the big green button

PR #562, opened 2016-03-03
Radio comm works, added ping/pong commands for testing

justbuchanan:
I think I've finally fixed the last few issues with radio communication!  We can now reliably send and receive data between MBEDs with radios.

One issue I fixed was with FIFO access.  The `writeReg()` method being used to write to the FIFO was handling access differently because the address of the TX FIFO is in the "extended address range".  The datasheet seems to say that the FIFOs are handled differently than the other things in the extended address range.  See commit 616d956d5089b83cba779d5d9b7ad2e84ba2156b.

The other change to get things to work was strobing the radio into IDLE before filling the TX FIFO and strobing into TX.  See commit ade6e7f22446ce49e595f3c88481d33934ee5efa.  I honestly don't know why this is needed, but without it, it doesn't seem to send data.  I've browsed the data sheet some more, but haven't found anything relevant.

Also, the `CC1201` constructor now puts the radio in RX mode at the start.  This combined with the register settings that place it into RX after a successful TX or RX means that the radio is pretty much always in RX except on the occasion that we're transmitting something.

TODO:
- [x] determine why radio needs to go into IDLE before filling FIFO and going to TX.  I gave up on figuring this out, but put a comment next to the code describing the mystery.
- [x] clean up RX/TX FIFO access
- [x] Add better radio demo commands (see PR #563)
- [x] verify that `CommLink.twos_compliment()` is correct (I changed it in 616d956d5089b83cba779d5d9b7ad2e84ba2156b)
- [x] `rtp::packet.recv() currently ignores header`
- [x] remove `Thread::wait()` calls in `sendData()` before/after strobes?
- [x] What to do with cc1101 code?  It doesn't quite work and I don't think we'll need it.  (deleted in 27f67524126809a1431cf0dc189404aa6d327d27)
- [x] better CC1201/CommLink docs
- [x] rename cs to nCs in CommLink
- [x] clean up code
- [x] figure out `rtp::packet` endianness-handling
- [x] sort out `port_fields` vs packed struct in `rtp::header_data`
- [x] review fields of `rtp::packet` (subclass, sfs, ack, etc)
- [x] figure out packet addressing.  Are we using addresses?  - I'll leave as-is for now and later we can actually implement it.  Right now the packet has that field set, but I don't think it's actually used by the radio?
- [x] Restore old serial connection `ping` command
- [x] Document the ping/pong commands a bit
- [x] Something is off with the timing in our commands.
- [x] fix or remove RX/TX flush logging
- [x] fix port printing for `radio` command
- [x] init and de-init for iterative commands?
- [x] fix radio `up`/`down` commands now that port api has changed
- [x] how to use keyboard to break out of command loop?
- [x] clean up CommInitialization
- [x] verify that things work without a radio plugged in

cc @jjones646 @guyfleeman


jjones646:
Holy hot damn. This causes for a celebration. :beers:

The whole IDLE thing before going into TX may be the cause from how an FIFO error is handled in the CC1201 driver code. And FIFO errors could be the result of incorrect threshold settings for the size of each. That's all a pretty wild guess tho because going to IDLE before entering TX wasn't something I ever looking into I don't believe.

[I warned you](https://github.com/RoboJackets/robocup-software/pull/510#issuecomment-159491827) about the extended register access...it's way trickier than the datasheet makes it out to be. The FIFO extended read/write stuff from f10df82d0 definitely works for reference. Something critically important to consider is that it will be a shared SPI bus with the FPGA. And that is using 8 bit transfers at the moment. But it's the same hardware registers at the end of the day, and I'd be highly surprised if the mbed class reconfigured the bit width for every spi->write() call. It's pretty easy to move the FPGA over to 16 bit transfers, but definitely something to keep in mind if it doesn't work at all on the control board as everything stands now.

justbuchanan:
@jjones646, The `packet` class has an `address` field, but it's not currently setup to do anything from what I can tell.  Is this something we plan on using?  Maybe just have all packets from robots be addressed to the base station and all packets from the base station will send to broadcast?

jjones646:
@justbuchanan The address field is useful for filtering the `robot -> base station` packets from other robots. You have it right, `base station -> robots` via broadcast and `robots -> base station` via base station address.

justbuchanan:
@jjones646, @guyfleeman I've been trying to setup the ping/pong commands as non-iterative and have them break out of their infinite loops when ctrl+c is pressed.  Right now that doesn't work because [in ConsoleTaskThread.cpp](https://github.com/RoboJackets/robocup-software/blob/justin/radio-feb29/firmware/robot2015/src-ctrl/modules/ConsoleTaskThread.cpp#L63), interrupts are disabled before calling `execute_line()` for the current command.  Can I remove the IRQ disabling there or is there another way to allow input to be read while a command is executing?

jjones646:
@justbuchanan give it a shot. the reasons it's there is because it's there and there is really no other reason. I have no clue whether or not it's actually needed (or _supposed_ to be needed for that matter).

justbuchanan:
Haha sounds good.  The [commit](3f089522a20ae051ea06dd33fdd405e9c6e6d95b) message doesn't mention it, so it's probably it's too important...

jjones646:
[Here's](https://github.com/RoboJackets/robocup-software/commit/498aafdd14b5c1826c9de24c5f724a99949463bd) the original first commit for that. It's been in an out throughout commits from debugging console things. [This](https://github.com/RoboJackets/robocup-software/commit/7844be01cbf087c66ae810ea57d11edd4167e2e7) one's pretty relevant too.

justbuchanan:
I think I've got this working pretty well.  It now detaches the Console's RXCallback while executing a command, which allows the command to read from stdin.  The ping/pong commands now run indefinitely until a key is pressed, then cleanup and return stdin back to the Console.  See commit e7a1875fa62993ab9088f8f8c9f5fdd4c8686497.  Do you see any potential problems with this setup?

justbuchanan:
I have this all completed and working with one exception.  One major change I made was to shrink the stack size for a few threads (see commit 16a9182) because I was occasionally getting an out-of-memory segfault.  Do you see any problems with changing the stack size?  Is there a smarter way to decide how big they should be?

jjones646:
There is, from [here](https://developer.mbed.org/handbook/RTOS):

```
Get the total stack memory size for this Thread.
uint32_t 	stack_size ()
```

```
Get the currently unused stack memory for this Thread. 
uint32_t 	free_stack ()
```

```
Get the currently used stack memory for this Thread.
uint32_t 	used_stack ()
```

```
Get the maximum stack memory usage to date for this Thread.
uint32_t 	max_stack ()
```

justbuchanan:
Nice!  I'll try adding max stack usage info to the thread info command.  I should be able to run things for a while and see what max usage is, then maybe set stack size to 1.5x the max or something.

jjones646:
You could probably get away with something even lower than 1.5x. I came across this in [`cmsis_os.h`](https://github.com/mbedmicro/mbed/blob/master/libraries/rtos/rtx/TARGET_CORTEX_M/cmsis_os.h), indicating that it would have a hard set upper bound maybe?

```cpp
// The stack space occupied is mainly dependent on the underling C standard library
#if defined(TOOLCHAIN_GCC) || defined(TOOLCHAIN_ARM_STD) || defined(TOOLCHAIN_IAR)
#    define WORDS_STACK_SIZE   512
#elif defined(TOOLCHAIN_ARM_MICRO)
#    define WORDS_STACK_SIZE   128
#endif

#define DEFAULT_STACK_SIZE         (WORDS_STACK_SIZE*4)
```

The multiplication by `4` is interesting here and threw me off for a minute, because the default value for an mbed RTOS Thread object is `DEFAULT_STACK_SIZE`. See [Thread.h:42](https://github.com/mbedmicro/mbed/blob/master/libraries/rtos/rtos/Thread.h), which is an `uint32_t` and would be `512 x 4 x 4` bytes in total. But is then divided again by the size of `uint32_t` when allocated in [`Thread.cpp:39`](https://github.com/mbedmicro/mbed/blob/master/libraries/rtos/rtos/Thread.cpp), which would come out to `2048` bytes being allocated for a default sized thread.

And it looks like the stack is initially filled with `0xE25A2EA5` in the constructor. So that could be very helpful if you see that showing up in gdb or something for no reason.

justbuchanan:
Interesting.  I may not get to this until after break, but I'll see what I can do.  It seems kind of strange that the stack size is based on the toolchain and not the target platform.  I'm also not entirely sure how to add this to the commands since getting the stack stats requires a reference to the thread itself and not just the thread id.  There should be a way to do it though.

justbuchanan:
I just added max stack size to the `ps` command.  I'm only seeing two threads though, which I think is because we're iterating through the `os_rdy` linked list, which only contains active threads.  Any threads waiting on timers, signals, etc are not in the list.  Do you know how we can print those?  I tried referencing `os_dly`, but it said it was undefined.

Also, the `ps` command printed `B_PRIOR` in the header, but wasn't printing it for each thread.  I deleted it for now, do you want to add it back in?

jjones646:
The `os_rdy` linked list is for the active threads, but I don't remember ever finding an easy way for listing all defined threads. I think it was because the pointers are moved between 2 lists in runtime.

I don't even know what `B_PRIOR` represents.

justbuchanan:
Without enumerating waiting threads, I can't get any stats on most of the relevant threads since they're almost always in a waiting state.  I'll go ahead and merge this now and open a ticket for looking into smarter stack sizing in the future.  There should be a way to do what we want by holding pointers to all the `Thread` objects, it's just kind of a pain.  The sizes I have now seem to work well enough.

PR #563, opened 2016-03-06
Added ping command for testing radio

justbuchanan:
TODO:

- [x] Restore old serial connection `ping` command
- [ ] Document the ping/pong commands a bit
- [ ] Something is off with the timing in our commands.
- [x] fix or remove RX/TX flush logging
- [ ] fix port printing for `radio` command
- [x] init and de-init for iterative commands?
- [ ] fix radio `up`/`down` commands now that port api has changed
- [ ] how to use keyboard to break out of command loop?

Note: I also removed the `CommPorts` class since it was basically just a wrapper around `std::map`.  Also, removed the open/closed flag for ports.  To close a port now, just remove the TX/RX handlers.

### DIFF
--- a/firmware/common2015/drivers/cc1201/CC1201.cpp
+++ b/firmware/common2015/drivers/cc1201/CC1201.cpp
@@ -48,7 +48,7 @@ int32_t CC1201::sendData(uint8_t* buf, uint8_t size) {
         return COMM_FUNC_BUF_ERR;
     }
 
-    flush_tx();
+    strobe(CC1201_STROBE_SFTX);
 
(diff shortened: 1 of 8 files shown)


justbuchanan:
@jjones646, right now the ping/pong commands are setup as "iterative commands" because I wanted them to run indefinitely until Ctrl+C is pressed.  This setup is a little messy though.  Is there a way to have a "non-iterative" command that loops until it detects that Ctrl+C is pressed?  I tried doing something like this by having it check the Console to see if the IterCmdBreakReq flag was set, but it didn't seem to be getting updated.

PR #564, opened 2016-03-06
Added Bezier Velocity Profile

ashaw596:


justbuchanan:
@ashaw596 I've gone through this briefly and so far it looks pretty good.  I'll look through it in more depth (hopefully later tonight), it's a little dense digging through beziers, rrts, etc.

At the moment, it's a little hard for me to follow exactly what's going on.  If you could add a description in the header file somewhere that gives a basic high-level step-by-step of how this all works, I think it would be really helpful!  Also, consider trying to improve the method names a bit.  Right now you've got:
* `generateCubicBezier()`
* `optimize()`
* `cubicBezier()`
* `generateCubicBezierPath()`
* `generateVelocityPath()`
* `generateNormalCubicBezierPath()`

It's a little tough to tell what the intent of each is, especially the cubic bezier ones.  Also, could you add some more details on what the parameters to each are?

common/Geometry2d/Point.hpp
```diff
@@ -258,16 +258,17 @@ class Point {
 
     /**
     * Returns a vector with the same direction as this vector but with magnitude
-    * one.
+    * given,
+    * unless this vector is zero.
     * If the vector is (0,0), Point(0,0) is returned
     */
-    Point normalized() const {
+    Point normalized(float magnitude = 1.0) const {
```
justbuchanan: I'd kinda prefer if we kept this the way it was before.  The definition of "normalized" is that it's a unit vector.  You can achieve the same thing with `myVector.normalized() * magnitude`.

soccer/motion/MotionControl.cpp
```diff
@@ -134,16 +141,8 @@ void MotionControl::run() {
 
     // Position control ///////////////////////////////////////////////
     ////////////////////////////////////////////////////////////////////
-    MotionInstant target;
+    MotionInstant target = optTarget->motion;
 
-    // convert from microseconds to seconds
-    if (!optTarget) {
```
justbuchanan: Can `optTarget == boost::none` still happen here?

soccer/planning/MotionCommand.hpp
```diff
@@ -17,6 +17,7 @@ class MotionCommand {
     virtual ~MotionCommand() = default;
     CommandType getCommandType() const { return commandType; }
     virtual std::unique_ptr<Planning::MotionCommand> clone() const = 0;
+    int debug;
```
justbuchanan: Let's drop this, it doesn't seem like something that really belongs here.

soccer/planning/InterpolatedPath.hpp
```diff
@@ -21,7 +21,7 @@ class InterpolatedPath : public Path {
     /// Each entry in InterpolatedPath is a MotionInstant and the time that the
     /// robot should be at that position and velocity.
     struct Entry {
-        Entry(MotionInstant inst, float t) : instant(inst), time(t) {}
+        explicit Entry(MotionInstant inst, float t) : instant(inst), time(t) {}
```
justbuchanan: This is fine, but I think that the `explicit` keyword is only relevant for single-parameter constructors.  I don't think it actually changes anything here.

http://en.cppreference.com/w/cpp/language/explicit

soccer/planning/RRTPlanner.cpp
```diff
@@ -72,9 +73,11 @@ std::unique_ptr<Path> RRTPlanner::run(
     // Replan if needed, otherwise return the previous path unmodified
     if (shouldReplan(start, goal, motionConstraints, obstacles,
                      prevPath.get())) {
+        int smoothingIterations = 2;  // int smoothingIterations = target.debug;
```
justbuchanan: rm trailing comment.  Also maybe add a `const`

soccer/planning/RRTPlanner.cpp
```diff
@@ -183,10 +186,30 @@ InterpolatedPath* RRTPlanner::optimize(
...
-}
 
+    // Generate Cubic Bezier with Velocity Profile
+    vector<Point> points;
+    for (InterpolatedPath::Entry& entry : pts) {
+        points.push_back(entry.pos());
+    }
+    return generateCubicBezier(points, *obstacles, motionConstraints, vi, vf,
+                               smoothingIterations);
+}
```
justbuchanan: I'll see if we can add it to our style config, but it'd be nice to have newlines between function definitions :)

soccer/planning/RRTPlanner.cpp
```diff
@@ -195,37 +218,380 @@ float getTime(InterpolatedPath& path, int index,
...
+
+    const float endSpeed = vf.mag();
+
+    if (times) {
+        assert(times->size() == points.size());
+        for (int i = 0; i < curvesNum; i++) {
+            ks[i] = 1.0 / (times->at(i + 1) - times->at(i));
+            ks2[i] = ks[i] * ks[i];
+            if (std::isnan(ks[i])) {
+                debugThrow("Meh");
```
justbuchanan: ...

soccer/planning/RRTPlanner.cpp
```diff
@@ -195,37 +218,380 @@ float getTime(InterpolatedPath& path, int index,
...
+        }
     } else {
+        for (int i = 0; i < curvesNum; i++) {
+            ks[i] = 1.0 / (getTime(points, i + 1, motionConstraints, startSpeed,
+                                   endSpeed) -
+                           getTime(points, i, motionConstraints, startSpeed,
+                                   endSpeed));
+            ks2[i] = ks[i] * ks[i];
+            if (std::isnan(ks[i])) {
+                debugThrow("Meh");
```
justbuchanan: ...

justbuchanan: What scenarios would lead to getting a NaN k value?  Is there something we can do to handle or prevent this better?

soccer/planning/RRTPlanner.hpp
```diff
@@ -71,25 +89,54 @@ class RRTPlanner : public SingleRobotPathPlanner {
...
-    static Planning::InterpolatedPath* cubicBezier(
-        Planning::InterpolatedPath& path, const Geometry2d::ShapeSet* obstacles,
+    static std::unique_ptr<Planning::InterpolatedPath> cubicBezier(
+        std::unique_ptr<Planning::InterpolatedPath> path,
+        const Geometry2d::ShapeSet* obstacles,
+        const MotionConstraints& motionConstraints, Geometry2d::Point vi,
+        Geometry2d::Point vf);
+
+    /**
+     * Generates a Cubic Bezier Path based on Albert's random Bezier Velocity
```
justbuchanan: What does the "random" refer to here?  Is this just because it's smoothing a path that's originally from an RRT?

soccer/planning/RRTPlanner.cpp
ashaw596: Oh, it doesn't really happen anymore. It used to be a bigger problem when the getTime function had a bug.

soccer/planning/RRTPlanner.hpp
ashaw596: Yes. Its based on the smoothing path I wrote a couple years ago.

common/Geometry2d/Point.hpp
ashaw596: I think it's a good change since I think it looks cleaner than seperately multiplying by magnitude and everything works the same with original functionality.

soccer/motion/MotionControl.cpp
ashaw596: It gets handled at line 65

soccer/planning/InterpolatedPath.hpp
ashaw596: true. Meh

soccer/planning/MotionCommand.hpp
ashaw596: Fine

ashaw596:
Ok, I think got everything mostly good.

soccer/planning/RRTPlanner.cpp
```diff
@@ -195,109 +212,358 @@ float getTime(InterpolatedPath& path, int index,
         motionConstraints.maxAcceleration, startSpeed, endSpeed);
 }
 
-std::unique_ptr<Planning::InterpolatedPath> RRTPlanner::generateVelocityPath(
+std::unique_ptr<InterpolatedPath> RRTPlanner::generatePath(
     std::vector<Geometry2d::Point>& points,
     const Geometry2d::ShapeSet& obstacles,
     const MotionConstraints& motionConstraints, Geometry2d::Point vi,
     Geometry2d::Point vf) {
+    return nullptr;
```
justbuchanan: What's the deal with `generatePath()`?

justbuchanan:
I'd prefer to remove the changes to the LineUp play before merging this.  What was the issue you were having before you changed it?

soccer/planning/RRTPlanner.cpp
```diff
@@ -195,109 +212,349 @@ float getTime(InterpolatedPath& path, int index,
         motionConstraints.maxAcceleration, startSpeed, endSpeed);
 }
 
-std::unique_ptr<Planning::InterpolatedPath> RRTPlanner::generateVelocityPath(
-    std::vector<Geometry2d::Point>& points,
+std::unique_ptr<InterpolatedPath> RRTPlanner::generatePath(
```
justbuchanan: Can we just delete this method?  It seems pretty useless as is.

justbuchanan: Yeah it makes sense that this could be applied to paths generated in other ways.  I'm just asking about this method in particular.  If all it's doing is calling `generateCubicBezier()` by forwarding the parameters exactly as is, why not just cut it out?  I think the name `generateCubicBezier()` is more descriptive anyways.

ashaw596: Because encapsulation. The methodology for the method isn't specified. Also not sure about that implementation yet. Not sure if it should optimize and RRT plan. Its fine.

PR #566, opened 2016-03-09
Replaced python Field class with pointer to Field_Dimensions object

### LINKED ISSUE #337: Remove python Field class in favor of passing a Field_Dimensions object
Keep it DRY

chachmu:
Removed useless copy constructor and equality setter. Also made the python constants.Field object contain a pointer to the current c++ Field_Dimensions object. Closes #337


common/Field_Dimensions.hpp
```diff
@@ -40,7 +63,9 @@ struct Field_Dimensions {
     static Field_Dimensions Current_Dimensions;
 
     Field_Dimensions()
-        : Field_Dimensions(0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0) {}
+        : Field_Dimensions(0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0) {
+        updateGeometry();
```
justbuchanan: Because the no-argument constructor here calls the constructor that takes in a bunch of arguments, the `updateGeometry()` method actually ends up getting called twice.  Once because we call the multi-parameter contructor and once right afterwards.

Since there's no need to call `updateGeometry()` after initialization, let's get rid of the method and just put its contents inside of the multi-parameter constructor below.

soccer/gameplay/GameplayModule.cpp
```diff
@@ -404,18 +410,6 @@ boost::python::object Gameplay::GameplayModule::getRootPlay() {
...
-            getMainModule().attr("set_field_constants")(
-                Field_Dimensions::Current_Dimensions);
-        } catch (error_already_set) {
-            PyErr_Print();
-            throw new runtime_error(
-                "Error trying to pass field dimensions to python");
-        }
-    }
-    PyGILState_Release(state);
+boost::python::object Gameplay::GameplayModule::getConstantsModule() {
```
justbuchanan: Since we only need to access the constants module one time in the GameplayModule constructor, let's remove the `getConstantsModule()` method and just access the python module in the constructor directly.

ashaw596: I mean I don't think it hurts to have it if its ever needed in future. Not likely to be needed though.

soccer/gameplay/constants.py
```diff
@@ -26,145 +26,4 @@ class Ball:
...
-                        robocup.Point(Field.Width / 2.0, Field.Length),
-                        robocup.Point(Field.Width / 2.0, 0)), robocup.Line(
-                            robocup.Point(Field.Width / 2.0, 0), robocup.Point(
-                                -Field.Width / 2.0, 0))
-    ]
-
-    Field.FieldRect = robocup.Rect(
-        robocup.Point(-Field.Width / 2.0, 0), robocup.Point(Field.Width / 2.0,
-                                                            Field.Length))
+Field = None
```
justbuchanan: Can you add a comment above `Field = None` saying that it gets set to the global `Field_Dimensions` object by the `GameplayModule`?

ashaw596:
Python tests are failibg.
`ERROR: test_our_goal_zone (test_constants.TestConstants)
----------------------------------------------------------------------
  File "/home/developer/project/soccer/gameplay/tests/test_constants.py", line 9, in test_our_goal_zone
    in_zone = robocup.Point(0, constants.Field.PenaltyDist / 2.0)
AttributeError: 'NoneType' object has no attribute 'PenaltyDist'`

soccer/gameplay/GameplayModule.cpp
justbuchanan: It doesn't hurt, it's just cleaner and more concise without it.  Given that the chances of needing this method in the future are very small, I'd rather cut it out now and add it back later if needed.

common/Field_Dimensions.hpp
```diff
@@ -33,6 +40,22 @@ struct Field_Dimensions {
     inline float FloorLength() const { return _FloorLength; }
     inline float FloorWidth() const { return _FloorWidth; }
 
+    inline Geometry2d::Point CenterPoint() const { return _CenterPoint; }
+    inline Geometry2d::CompositeShape OurGoalZoneShape() const {
+        return _OurGoalZoneShape;
+    }
+    inline Geometry2d::CompositeShape TheirGoalShape() const {
```
justbuchanan: Let's rename this to `TheirGoalZoneShape` to keep naming consistent.

justbuchanan:
@ashaw596 @chachmu, I just made a change that should fix the unit tests.  I just changed the initial value of `constants.Field` from `None` to `Single_Field_Dimensions`.

ashaw596:
The python tests  are failing with ``` name 'FieldDimensions' is not defined```

soccer/gameplay/GameplayModule.hpp
```diff
@@ -115,6 +115,8 @@ class GameplayModule {
     /// gets the instance of the main.py module that's loaded at GameplayModule
     boost::python::object getMainModule();
 
+    boost::python::object getConstantsModule();
```
justbuchanan: x

justbuchanan:
:+1:

Issue #546, opened 2016-02-22
Calls to printf() in firmware aren't threadsafe


justbuchanan:
Some corruption is caused in the log output due to calls to `printf()`.  Probably the best solution would be to add a threadsafe wrapper around printf and call it from `log()` and any other place we want to print.

Somewhat related to issue #491 

cc @JNeiger, @jjones646

jjones646:
From issue #491, yes, `printf()` is already thread safe. Thread safe in the sense that two calls to `printf()` will not cause a memory race condition and crash the microcontroller. However, there is no guarantee about keeping buffers in order.

The mutex used in the log function ensures that two 'parallel' calls to `log(...)` will be sent to `printf()` at different times. The log function also ensures that a new line character is appended to the end and the entire thing is flushed before removing the mutex lock. This is because there is additional buffering on the mbed interface chip that will only flush the uart buffer when filled over a threshold or at a newline character.

For all firmware code I've written, I've tried very carefully to only make `printf()` calls at times when it's reasonably safe to say that the mbed interface's uart buffer is empty, and that no other threads have recently called `log(...)`. All `printf()` calls in command.cpp are there because the thread's priority is increased above all others when executing a line. This essentially suppresses all `log(...)` calls during that time - and it's safe to assume that everything will be executed in order for any of those functions.

The only real solution to all of this is to make our own `printf()` function that is aware of the rtos threads. As explained [here](https://github.com/RoboJackets/robocup-software/issues/491#issuecomment-185375853), except you don't want to wrap a mutex around `fflush()` because that can still happen if the stdout buffer is filled beyond a certain threshold (don't take that as fact, but I'm pretty sure). You'll have to wrap the mutex around everything - pretty much exactly how the `log()` function is written.

Lastly, a lot of good stuff can be found in [external/mbed_lib_build_tools/libraries/mbed/common/retarget.cpp](https://github.com/mbedmicro/mbed/blob/master/libraries/mbed/common/retarget.cpp), so it may be possible to redirect stdout to our own defined buffers where everything could be handled correctly.

JNeiger:
I ended up making a general [`log_printf` function](https://gist.github.com/JNeiger/555a20c2418197121f68) that I think I will help solve both this and #491. I think I have everything in there setup right, but I just want to make sure.

justbuchanan:
[...]

Issue #491, opened 2015-11-04
stream-based logging for firmware


justbuchanan:
The more or less "standard" way of doing logging in C++ is with streams.  We implemented something this a while ago, but I think some of it got lost due to merge conflicts.  It'd be nice to be able to something like this in firmware:
~~~{.cpp}
LOG(WARN) << "Something happened to: " << myObject << endl;
~~~

The real power here is that you can implement `operator<<(ostream)` on different classes and very easily print them to the log.

justbuchanan:
As discussed yesterday, the version of the firmware logger on the `master` branch has this feature, but it looks like it got lost on the `robot2015-firmware` branch.  We should just be able to copy the version on `master` to `robot2015-firmware` and verify that it works.

This will need to be tested on the MBED since we can't run our firmware code on the computer itself.  @hmistry1, we can get you an MBED to use at the next meeting if you remind me.

justbuchanan:
Are you having trouble getting the branch onto GitHub or making a pull request once the branch is there?

And sure, we can get you an MBED to test on next time.

JNeiger:
So I overloaded the `log(...)` function within the [`logger.hpp`](https://github.com/RoboJackets/robocup-software/blob/robot2015-firmware/firmware/common2015/utils/logger/logger.hpp) class to allow for a `std::ostream&` type overload (mirroring [this](https://github.com/RoboJackets/robocup-software/blob/master/firmware/common2015/utils/Logger/logger.hpp#L40) function in the old code).

Since both of these are going to be called through the [`LOG(...)`](https://github.com/RoboJackets/robocup-software/blob/robot2015-firmware/firmware/common2015/utils/logger/logger.hpp#L24-L25) macro, I think the next step is to "overload" the macro so it can call both depending on there are any arguments besides the log level. Most likely loosely mirroring this style of [design pattern](http://stackoverflow.com/questions/8886812/overload-c-macros). 

@justbuchanan, Would that be the right way to go on this or is there a better way?

justbuchanan:
Yeah that sounds about right.  We could just name the stream-based macro function to something slightly different.

[...]

PR #567, opened 2016-03-09
Stream logger

JNeiger:
Solves issue: #491, #546


firmware/common2015/utils/logger/logger.hpp
```diff
@@ -61,6 +64,24 @@ extern bool isLogging;
 extern uint8_t rjLogLevel;
 
 /**
+ * [Collects the stream log message into a single string to print]
```
justbuchanan: Doxygen, the program that converts our code comments into the docs website, doesn't handle square brackets specially, so they'll show up in the text on the site just as they are here.  It would be best to get rid of them.

I noticed that several other comments have these and curly braces in the comments and I think they're an artifact of a comment template snippet that does something like:
~~~{.cpp}
/**
 * @brief [your description here]
 */
~~~

Anyways, I'll make a note to go through and remove them from the other comments throughout our firmware since they're sort of a placeholder.

firmware/common2015/utils/logger/logger.hpp
```diff
@@ -61,6 +64,24 @@ extern bool isLogging;
...
+ * @param source   [The source of the message.]
+ * @param format   [The string format for displaying the log message.]
+ */
+class LogHelper : public std::stringstream {
+public:
+    LogHelper(uint8_t logLevel, const char* source, int line, const char* func);
+    ~LogHelper();
+
+private:
+    uint8_t m_logLevel;
```
justbuchanan: Let's drop the leading `m` to keep naming consistency with the rest of our code.

firmware/common2015/utils/logger/logger.cpp
```diff
@@ -18,6 +18,24 @@ uint8_t rjLogLevel;
 Mutex log_mutex;
 
 /**
+ * [Collects the stream log message into a single string to print]
```
justbuchanan: I'd remove this comment since you have a copy of it in the header file already.  The two comments will inevitably get out of sync if anything changes.

firmware/common2015/utils/logger/logger.hpp
JNeiger: Just the `m`? So it leaves `_logLevel` etc? Or the entire `m_`?

justbuchanan: Yeah just the `m`, so you'll have `_logLevel`, etc.

firmware/common2015/utils/logger/logger.hpp
```diff
@@ -24,6 +25,8 @@
 #define LOG(lvl, ...) \
     log(lvl, __BASE_FILE_NAME__, __LINE__, __func__, __VA_ARGS__)
 
+#define S_LOG(lvl) LogHelper(lvl, __BASE_FILE_NAME__, __LINE__, __func__)
```
justbuchanan: One last thing.  Can you add an example usage above this macro/function?  Some people may not be very familiar to stream-based logging or how this is setup, so it'd probably be helpful to show quick example.

JNeiger: Something like `// S_LOG(INIT) << "Example" << std::endl;` good enough or do I need to add a bit more?

justbuchanan: Yeah that sounds good to me :+1:.  Maybe add a `// Example usage:` line above

justbuchanan: Actually, in order for it to show up on our docs site, use three slashes `///` for the comment or use a multiline comment starting with `/**`.  Doxygen ignores comments with only two slashes.

justbuchanan: Also, the `std::endl` isn't needed since the `log()` function already appends one, right?

justbuchanan:
LGTM :+1:

PR #569, opened 2016-03-09
New Radio Base Station and Protocol

### LINKED ISSUE #154: Setup MBED OTA install
Our current (2008/2011) firmware can be updated wirelessly with the `make robot-prog-ota` command.  It'd be cool to do something similar with the MBED too if possible.

### LINKED ISSUE #155: Develop new radio protocol for high-performance and reliable transfer
We're looking to change our radio protocol for the new control board.

A few objectives:
* MBED-based (for both the base-station and on the robot)
* wireless serial port so we can open a shell onto the board without being plugged in
* Suppport for more than six robots (right now we only use six, but in the near future, we'll need to support more)

Right now we're considering having two radio channels that run simultaneously using two separate hardware radios per robot and base station.  One channel would be for motion commands coming to the robot and the other would be a session-oriented reliable channel that would be for firmware updates, robot status info, and serial comm for a terminal session.

A few things to investigate:
* 900MHz (cc1101 chip) vs 2.4GHz (NRF chip)
* Can we do everthing we want using a single 2.4GHz radio?
* How do we implement the session-oriented reliable portion of the protocol?  What MBED libraries are available?

Relevant info:
* https://docs.google.com/document/d/1Ma9RRYa4p4PcaC-CYx8j00_UTisCI85ou5XQsW_cU_A/edit?usp=sharing
* http://wiki.robojackets.org/w/RoboCup_Radio_2009

cc @bstrong1218 @ryanstrat @ChrisDoho

### LINKED ISSUE #278: Write firmware for MBED-based radio base station
A few things todo that I can think of right now:
- [x] receive commands from `soccer` and broadcast them
- [x] status LEDs - the on-board ones will be fine
- [x] place received packets in a buffer to be read from `soccer`
- [x] move MBED-related CMake stuff out of `robot2015` and into `firmware` so they can be accessed by both robot firmware and base station firmware
- [x] change the `USBRadio` class in soccer to use the new base station

@jjones646 at some point it'd be good to get another radio board or two populated for developing this.

I think for this time around for radio stuff we should pretty much just copy the old radio protocol, then put our ambitious dual-channel base station etc. on the back-burner until everything else works.  The cool stuff will probably be a project for next year.

cc @guyfleeman

### LINKED ISSUE #304: Add parameter passing feature to radio protocol
For this year we'll be more or less copying the previous radio protocol since we're running low on time to implement some of the cool features we've talked about.  One thing we're looking at adding is a way to be able to change certain values on the robot at runtime.  This would be used for motion control parameters mostly, but would be flexible for other use cases.  @jjones646 and I like the way this works in the [crazyflie firmware](https://github.com/bitcraze/crazyflie-firmware/blob/56250b5ded22a25e8dc46eda95aa85bccd8e53b1/modules/interface/param.h).

guyfleeman:
Brief Todo List:
- [x] write new protocol into protobuf
- [ ] bring old semantics up to date with new protocol
- [ ] implement new protocol features
- [x] investigate USB libraries for new base station
- [x] get base station communicating with computer
- [ ] interface new base station with soccer startup
- [ ] write protocol support into base station
- [ ] tag old radio stuff prior to merge, no backwards compatibility

Notes and Questions (more to come):
* Does the base station need to be hot-pluggable and re-pluggable while soccer is running? 

Assumed Priorities:
1. base station and protocol (resolve #155, resolve #278)
2. parameter tuning (resolve #304)
3. firmware (resolve #154)


external/DoCIF
```diff
@@ -1 +1 @@
-Subproject commit e52def19124bec4c9b743a8f7276ef2b14f64b51
+Subproject commit 1dec20b1995d4735b7ec295a6ec4ae4d584ee187
```
justbuchanan: Did you mean to change this?  I think this change is breaking the build on circleci

firmware/base2015/PCLink.cpp
```diff
@@ -0,0 +1,57 @@
+#include <string>
+#include "PCLink.hpp"
+
+PCLink::PCLink(void)
```
justbuchanan: I think you can delete this constructor.  Because the other one has default values for the parameters, the result should be the same.

firmware/base2015/PCLink.hpp
```diff
@@ -0,0 +1,24 @@
...
+
+    PCLink(void);
+    PCLink(uint16_t vendorID, uint16_t productID, uint16_t release);
+    virtual ~PCLink(void);
+    void setSerialDebugging(Serial *pc);
+    void setLed(DigitalOut *led);
+    void read(void);
+    void reply(void);
+private:
+    Serial *pc;
```
justbuchanan: Let's name instance variables with a leading underscore to keep consistency with the rest of our c++.

soccer/radio/USB2015Radio.cpp
```diff
@@ -0,0 +1,57 @@
+#pragma once
```
justbuchanan: `#pragma once` doesn't do anything in .cpp files.

justbuchanan:
I think it would be best if the radio is hot-pluggable as long as that's not tough to do.  Our current setup is that way for the most part and it's a pretty nice feature to have :).

For the packets in the common/protobuf folder.  Want to put the protobufs in separate `robot2011` and `robot2015` folders for organization?

soccer/radio/USB2015Radio.hpp
```diff
@@ -0,0 +1,28 @@
...
+    virtual bool isOpen() const override;
+    virtual void send(Packet::RadioTx& packet) override;
+    virtual void receive() override;
+    virtual void switchTeam(bool blueTeam) override;
+
+    std::string getProtocolVersion();
+
+
+private:
+    static volatile bool hidAPIInit = false;
```
justbuchanan: I think `volatile` is probably unnecessary here.  It's only needed when an external device is changing variable values.

soccer/radio/USB2015Radio.cpp
guyfleeman: derped on this one

soccer/radio/USB2015Radio.hpp
guyfleeman: i added it so the hid_exit call would be thread safe, but I haven't decided if I want to handle hid_init and hid_exit in this class for sure yet.

external/DoCIF
guyfleeman: I wasn't sure, It caused a conflict and I thought I used the current one off master. Either way, I'll change it back.

soccer/radio/USB2015Radio.hpp
justbuchanan: Ok gotcha.  `volatile` doesn't do anything for thread-safety though, it's only for variables that might be changed by a device other than the processor.

guyfleeman: Been reading up on this, didn't realize volatile worked differently in C++ than in Java. Hindsight 20/20. I'll refactor the design so this isn't a problem.

PR #572, opened 2016-03-10
fixed issue with unassigned optional behaviors holding up progress

justbuchanan:
@ashaw596 Here's an alternative fix to #571 that I think will solve similar issues in more scenarios.  Can you test it out and make sure it works for you?


justbuchanan:
Good catch, it turned out to be a little more complicated than I thought.  The issue was that the state machine was getting run before robots had initially been assigned, so it transitioned inappropriately.

I deleted that and swapped to a new fix that just changes the transition check for the line up tactic.  I think this works now.

ashaw596:
Looks good

PR #575, opened 2016-03-16
Added config descriptions

### LINKED ISSUE #345: Add descriptions to config values that can be seen from the GUI
It's hard to tell what some of our config values do just from the name, it'd be nice if we could hover over it in the GUI and get a description.

chachmu:
Added option for tooltip descriptions to config values. Closes #345

### DIFF
--- a/soccer/Configuration.cpp
+++ b/soccer/Configuration.cpp
@@ -6,6 +6,8 @@
 #include <stdio.h>
 #include <assert.h>
 
+#include <iostream>
+
 std::list<Configurable*>* Configurable::_configurables;
 
 // Role for tree column zero for storing ConfigItem pointers.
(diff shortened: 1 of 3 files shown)


soccer/Configuration.cpp
```diff
@@ -151,6 +158,12 @@ void Configuration::addToTree(ConfigItem* item) {
     item->_treeItem->setFlags(item->_treeItem->flags() | Qt::ItemIsEditable);
     item->_treeItem->setData(0, ConfigItemRole, QVariant::fromValue(item));
     item->_treeItem->setText(0, path.back());
+
+    if (item->_description != "") {
+        std::cout << item->_description << std::endl;
```
justbuchanan: x

PR #576, opened 2016-03-17
Setup error LEDs

justbuchanan:
I got bored of debugging radio problems, so I took a break and looked into the error leds.  I think I've got them mostly setup, although I don't have a control board to test on.  Btw, I noticed there wasn't an error led for the fpga.  Are we just using the rgb led for that?

TODO:

- [ ] add imu error
- [ ] ball sense
- [x] radio connected/unconnected
- [ ] errors from motors (this is sent back in the first byte for all FPGA SPI transfers)
- [x] remove motor `encOK` flag

cc @jjones646

### DIFF
--- a/firmware/robot2015/src-ctrl/config/pins-ctrl-2015.hpp
+++ b/firmware/robot2015/src-ctrl/config/pins-ctrl-2015.hpp
@@ -97,23 +97,23 @@
 #define RJ_IO_EXPANDER_I2C_ADDRESS (0x42)
 
 // Port A bit masks
-#define RJ_IOEXP_A0 (0)
-#define RJ_IOEXP_A1 (1)
-#define RJ_IOEXP_A2 (2)
-#define RJ_IOEXP_A3 (3)
-#define RJ_IOEXP_A4 (4)
-#define RJ_IOEXP_A5 (5)
-#define RJ_IOEXP_A6 (6)
-#define RJ_IOEXP_A7 (7)
+#define RJ_IOEXP_A0 0
+#define RJ_IOEXP_A1 1
+#define RJ_IOEXP_A2 2
+#define RJ_IOEXP_A3 3
+#define RJ_IOEXP_A4 4
+#define RJ_IOEXP_A5 5
+#define RJ_IOEXP_A6 6
+#define RJ_IOEXP_A7 7
 // Port B bit masks
-#define RJ_IOEXP_B0 (8)
-#define RJ_IOEXP_B1 (9)
-#define RJ_IOEXP_B2 (10)
-#define RJ_IOEXP_B3 (11)
-#define RJ_IOEXP_B4 (12)
-#define RJ_IOEXP_B5 (13)
-#define RJ_IOEXP_B6 (14)
-#define RJ_IOEXP_B7 (15)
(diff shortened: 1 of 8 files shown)


jjones646:
Yeah, the RGB led is for the FPGA. I think it does orange at the moment if it's just the FPGA and RED if it thinks the SPI bus is down all together.

justbuchanan:
Sounds good.  I kept the rgb stuff as-is, just moved it into the main loop so that it gets updated when things change.

I have the motor error leds setup to reflect the values of the motor's `status` field (encOK, hallOK).  I think I'll leave the setting of those values from the fpga to you since I haven't really done anything with the fpga SPI interface.  How close are we to being able to read hallOK, encOK from the fpga?

jjones646:
There's been a lot of things done with the verilog since that data struct was originally created. So here's a brief update:

As everything stands now, there is no way to know anything for `encOK`. It could be done, but it's a non trivial thing to do and most likely will never happen considering the amount of logic space it would require.

For `hallOK`, the verilog can detect _hall error_ vs _hall not connected_, but I don't think I setup where you could tell the difference from the mbed side of things.

Check out the _motorscroll_ command to see how to do motor related things. The things in `fpga.cpp` have been complete for some time now, and all of the functions there do work. There's a watchdog timer on the FPGA that'll turn off the motors after ~114ms of not receiving a velocity update if I recall correctly.

----
Very recently, I've been working on adding directional control for the motors in the verilog. I haven't pushed any of it yet, but it's looking like duty cycles are going to be 9 bit values with the top bit being direction. I believe this is how it was before, so nothing too surprising there - bidirectional, 0 to 512 resolution. It's looking like the minimum duty cycle that a motor will spin with is ~50.

Since the radio changes you've worked on now use 16 bit SPI transfers, I've been meaning to update the fpga protocol to reflect that. Just haven't gotten around to doing it yet.

justbuchanan:
Ok, maybe we should just remove `encOK` for now and add it back if we ever get around to actually implementing it.

Btw, the radio SPI is [setup for 8-bit transfers](https://github.com/RoboJackets/robocup-software/blob/justin/radio-feb29/firmware/common2015/modules/CommLink/CommLink.cpp#L18) so you shouldn't have to change that for the fpga if it's already on 8.  We do probably need to add a mutex though so that the SPI bus is only used for one thing at a time though.

jjones646:
Ahh, you're right. 8-bits it is. I think the SPI transfers are blocking already. I'll do a bit of research into that and see if we'll actually need something for it.

justbuchanan:
Individual calls to `spi.write()` may be threadsafe, but most of our spi code looks something like this:

~~~{.cpp}
chipSelect = 1;
spi.write();
spi.write()
...
chipSelect = 0;
~~~

We'll need to have a mutex around code like this so that we don't select one device and begin transfers, then have another thread start writing different data to it in-between.

justbuchanan:
I'll go ahead and merge this now as-is since everything's working at least as well as before and we can keep the other items on the todo list.

jjones646:
### Trello Card
[For hardware testing](https://trello.com/c/SjEFbLN0/160-populate-control-board-for-justin)

PR #579, opened 2016-03-18
Refactored some parts of firmware CMake

justbuchanan:
This is a work in progress, but the goal is to make it easier to add new libraries and executables and to clean up our existing stuff a bit.  I've added a few functions/macros to help with this.


firmware/mbed/mbed_executable.cmake
```diff
@@ -0,0 +1,96 @@
...
+endfunction(mbed_env)
+
+function(_rj_configure_mbed_binary name)
+    # only build this if specifically instructed
+    set_target_properties(${name} PROPERTIES EXCLUDE_FROM_ALL TRUE)
+
+    # link and include official mbed libs
+    target_include_directories(${name} PUBLIC ${MBED_INCLUDE_DIR})
+    add_dependencies(${name} mbed_libraries)
+endfunction()
```
jjones646: Why not place in the repo's `./cmake` directory?

justbuchanan: True, that may be a better place for it.  I just dropped it here for now since it's mostly mbed-related.

justbuchanan:
Closing for now - I may get around to looking at this later

Issue #279, opened 2015-04-17
Add verilog code to circle-ci


justbuchanan:
We can use icarus verilog for this - we don't need to test much, just see if it compiles successfully.

jgkamat:
Is this as simple as just running ```iverilog robocup.v``` in firmware/robot/fpga?

justbuchanan:
I think it might be actually.  @ChrisDoho, can you confirm?  Is there anything you had to do to specify a netlist or anything and did you have to compile the files separately?

Also, I think we'll leave the old fpga code alone and just test our new stuff, which will be in the robot2015 folder.  The new stuff still lives on the `mbed-fpga` branch and hasn't been merged into master yet, so we'll put the travis config changes there too.

ChrisDoho:
For simulating code in icarus, you do not need any .ucf files. You do need to make a "testbed" for the module you want to include. The test bed simulates the inputs and  writes outputs to specified .vcd file. You can look at any of the files ending in "_tb" in /robot2015/VerilogTest/ for an example.

If you just want to test if it compiles you don't even need the test bed, you just tell icarus to compile the .v files you have. I've never used icarus with a ucf file before so I don't know how it will handle it. As a heads up, Icarus is stupid about how it handles preprocessing. You have to run one command to create the preprocessed verilog files and another to compile those...

PR #582, opened 2016-03-27
Added verilog compilation to circleci

justbuchanan:
This adds a target to cmake for compiling our 2015 fpga code with [iverilog](http://iverilog.icarus.com/).  To try this out on your computer, run `make fpga2015-test` or `make fpga2015-test-strict`.  They both compile our verilog code, but the strict version fails if any warnings are generated.  Most of the things I've seen iverilog warn about are things that we should probably fix such as size mismatches.

TODO:
- [ ] fix warnings generated from verilog code
- [ ] run test benches?

related: issue #279


justbuchanan:
@jjones646, [here](https://circle-artifacts.com/gh/RoboJackets/robocup-software/1472/artifacts/0/tmp/circle-artifacts.szeA5IE/synthesize-verilog.txt) are the compilation warnings it's showing right now - it looks like something that may be an issue, but probably a quick fix.

~~~{.sh}
warning: Port 5 (duty_cycle) of BLDC_Motor expects 10 bits, got 9.
robocup.v:612:        : Padding 1 high bits of the port.
~~~

jjones646:
This isn't fixed, but story time:

@petersonev and I concatenated one of the 32 bit values instead of shifting it in the Verilog (making it 33 bits effectively), and the synthesizer completely ignored this. No pins ended up being what they should have been. Inputs were outputs and outputs were inputs. It was wild.

So, yeah, it brings up some serious concern into how the synthesizer is actually handing these implicit bit paddings.

justbuchanan:
Closing in favor of #598

PR #585, opened 2016-03-28
Added a mutex for spi bus access

justbuchanan:
This PR adds a `SharedSPI` class that bundles a `Mutex` and an `SPI` instance.  The FPGA and CC1201 are now initialized with a shared pointer to a `SharedSPI` instance and lock the mutex when they're using the bus.  This should ensure thread-safety when using the spi bus.

TODO:

- [x] Do we need different spi frequencies for our different devices?  If so, this PR will need to change a good bit.  The fpga was previously using 1000000Hz and the cc1201 was on 5000000.  I currently have them both using 5000000.
- [x] Chip select doesn't look quite right in `FPGA::configure()` (prevously named `FPGA::Init()`)
- [x] Do we need `FPGA.mutex` anymore? (removed in 13d82c3cd7b1a77812c715c9fe09f789ac20ecb6)
- [x] why is chip select toggled in `FPGA.git_hash()`?
- [x] switch `KickerBoard` class to use `SharedSPI`
- [x] verify that fpga configuration still works
- [x] why use `SoftwareSPI` when configuring fpga?  We'll need to lock the shared mutex here.
- [x] kicker nCs vs reset?

### DIFF
--- a/firmware/CMakeLists.txt
+++ b/firmware/CMakeLists.txt
@@ -21,6 +21,8 @@ set(MBED_COPY_SCRIPT sudo ${PROJECT_SOURCE_DIR}/util/mbed/mbed-copy.py)
 
 # enable C++ exceptions
 set(MBED_CMAKE_CXX_FLAGS "${MBED_CMAKE_CXX_FLAGS} -fexceptions")
+# the mbed libraries are compiled with -fno-rtti, so we need to as well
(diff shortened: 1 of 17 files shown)


justbuchanan:
@jjones646 I just added a `frequency` parameter to `SharedSPIDevice` so each of our devices can use different frequencies.

justbuchanan:
@jjones646 can you try this out on the control board with the fpga and make sure the spi stuff there still works?  Other than testing that, I think this is ready to merge.

PR #589, opened 2016-03-28
BallSense now uses an RtosTimer to update at regular intervals

### LINKED ISSUE #587: Initialize ball sensor instance and set leds based on status
See PR #576

justbuchanan:
I also added it to the `main()` function and set the control board led to reflect whether or not the sensor is triggered.  closes #587 

TODO:

- [ ] how often should the sensor update?
- [x] rename LED pin define since it's not an error

cc @JNeiger @jjones646


justbuchanan:
@jjones646 any thoughts on how frequently we should update the ball sensor?

PR #595, opened 2016-04-03
Made Geometry2d::Point's x, y methods

justbuchanan:
This allows syntax compatibility with Eigen::Vector2f.  Code (such as the rrt repo) that is templated to allow for different point classes can use `Point` anywhere it can use `Eigen::Vector2f`.

- [x] for some reason, these changes seem to break path planning in some cases.  need to investigate


common/Geometry2d/Point.hpp
```diff
@@ -12,75 +12,65 @@ Simple class to represent a point in 2d space. Uses floating point coordinates
 */
 class Point {
 public:
-    /** the x coordinate */
-    float x;
-
-    /** the y coordinate */
-    float y;
+    const float& x() const { return _x; }
```
ashaw596: Maybe add an inline specifier. Its not really necessary, but it would be sort of annoying if the compiler didn't inline for some reason

common/Geometry2d/Point.hpp
```diff
@@ -214,19 +208,19 @@ class Point {
     * rotates the point around the origin
     */
     Point& rotate(float angle) {
-        float newX = x * cos(angle) - y * sin(angle);
-        float newY = y * cos(angle) + x * sin(angle);
-        x = newX;
-        y = newY;
+        float newX = x() * cos(angle) - y() * sin(angle);
```
ashaw596: Sort of inconsistent here. with between x() or _x. Either should be fine in this file though.

soccer/gameplay/robocup-py.cpp
```diff
@@ -459,6 +459,11 @@ void WinEval_add_excluded_robot(WindowEvaluator* self, Robot* robot) {
 
 BOOST_PYTHON_MEMBER_FUNCTION_OVERLOADS(Point_overloads, normalized, 0, 1)
 
+float Point_get_x(const Geometry2d::Point* self) { return self->x(); }
+float Point_get_y(const Geometry2d::Point* self) { return self->y(); }
```
ashaw596: I'm not sure I love all the functions we have just just forward functions...
Why don't we just use the syntax
const &float (Point::*pointGetX)() const = &Point::x;
void (Point::*pointSetX)(float x) = &Point::x;

common/Geometry2d/Point.hpp
justbuchanan: Using `inline` for this purpose is more or less deprecated - any reasonable compiler will inline this automatically.

common/Geometry2d/Point.hpp
justbuchanan: :+1:

soccer/gameplay/robocup-py.cpp
justbuchanan: I agree that having to wrap functions like this is pretty crappy, but I don't think your alternative syntax works.  It works fine for the getter, but the setter still needs to be a wrapper function.  Are you able to get something like this to work?

justbuchanan:
@ashaw596 I think this is good to go, let me know if you see any other issues.

soccer/gameplay/robocup-py.cpp
ashaw596: Yeah I didn't check. Yeah the setter won't work since you don't have a setter in Point.

ashaw596:
Looks Good I think. Just some checkstyle issues i think.

PR #597, opened 2016-04-04
Setup interrupts on MCP23017

### LINKED ISSUE #400: Enable support for interrupts on the digital IO expander
The io expander (mcp23017) has 8 inputs and 8 outputs.  We need to set up a system to interrupt the mbed when one of the input lines change.

cc @jjones646

justbuchanan:
closes #400 

TODO:
- [ ] test on actual hardware
- [ ] set MIRROR bit so that an interrupt on either port triggers an interrupt on both pins.  this is good because we are only connected to one of the interrupt lines


jgkamat:
Since firmware has been migrated to RoboJackets/robocup-firmware, this PR dosen't make sense. I'll close this, but keep the branch in case anyone needs it.

Issue #583, opened 2016-03-28
Smarter sizes for firmware thread stacks


justbuchanan:
Right now we initialize threads with a defined stack size that's smaller than the default in order to conserve ram.  These sizes are pretty arbitrary though, it would be better if we find out how much stack each thread is actually using and set it to something slightly higher.

See PR #562

justbuchanan:
closed by #707

PR #599, opened 2016-04-04
Adding delay tasks to ps command

jjones646:


jjones646:
@justbuchanan

justbuchanan:
related to #583

JNeiger:
I'm only seeing 2 threads (ID: 1 and 3) when I run it. Shouldn't there be 3? The `Main`, `Controller`, and `Console` thread?

justbuchanan:
The `CommModule` has 2 threads too, I think there's 6 threads total.

PR #602, opened 2016-04-06
Control Protocol 2016 Update

guyfleeman:
Nothing too major but had to refactor lots of thing to prepare for the new features.


external/DoCIF
```diff
@@ -1 +1 @@
-Subproject commit e52def19124bec4c9b743a8f7276ef2b14f64b51
+Subproject commit 1dec20b1995d4735b7ec295a6ec4ae4d584ee187
```
justbuchanan: This was probably changed on accident.  Do `git checkout robojackets/master --external`, then commit to revert it back to what it was.

soccer/CMakeLists.txt
```diff
@@ -41,7 +41,8 @@ set(ROBOCUP_LIB_SRC
     "Processor.cpp"
     "ProtobufTree.cpp"
     "radio/SimRadio.cpp"
-    "radio/USBRadio.cpp"
+#    "radio/USBRadio.cpp"
+    "radio/USB2015Radio.cpp"
```
justbuchanan: I think this is breaking the build as of now.

soccer/CMakeLists.txt
```diff
@@ -112,6 +117,7 @@ target_link_libraries(robocup GL GLU glut)
 target_link_libraries(robocup pthread)
 target_link_libraries(robocup spnav)
 target_link_libraries(robocup ${SDL_LIBRARY})
+target_link_libraries(robocup "/usr/lib/x86_64-linux-gnu/libhidapi-libusb.so")
```
justbuchanan: x

soccer/CMakeLists.txt
```diff
@@ -98,12 +98,17 @@ include_directories(SYSTEM ${LIBUSB_1_INCLUDE_DIRS})
 # SDL
 include_directories(SYSTEM ${SDL_INCLUDE_DIRS})
 
+# HID API
+include_directories(SYSTEM "/usr/include/hidapi")
```
justbuchanan: x

common/protobuf/Tuning.proto
```diff
@@ -0,0 +1,8 @@
+package Packet;
```
justbuchanan: Let's leave the tuning-related things out of this PR and wait until the next PR.

common/protobuf/Control.proto
```diff
@@ -0,0 +1,34 @@
...
+    {
+        KICK = 1;
+        CHIP = 2;
+    }
+
+    enum TriggerMode
+    {
+        IMMEDIATE = 1;
+        ON_BREAK_BEAM = 2;
+        ARM = 3;
```
justbuchanan: I don't think `ARM` is something we'll need for now.

guyfleeman:
Working on this, looks like the tests are using protobuf 2 when we use 3.

justbuchanan:
Yeah I think you're right, everything compiles fine on my system.  How did you get protobuf 3 installed on your computer?

guyfleeman:
Worked out of the box for me. Looking into it.

guyfleeman:
Found it:
Unions (oneof) were added in 2.6.0. Ubuntu 15.04 and later are on protobuf 2.6.0. Earlier is on 2.5.0. Should I rewrite it without unions or try to get the package on 14.04?

justbuchanan:
If it's not too much hassle, protobuf 3 would be nice.  If it looks like it will be tough to get to work though, I'd go with using something other than unions there.  Once the team moves to 15.04, it should be easy to hop back to protobuf 3, but that won't happen until after comp.

common/protobuf/Control.proto
```diff
@@ -0,0 +1,33 @@
...
+    }
+
+    required float xVelocity = 1         [default = 0];
+    required float yVelocity = 2         [default = 0];
+    required float aVelocity = 3         [default = 0];
+    required float dVelocity = 4         [default = 0];
+    required uint32 kcStrength = 5       [default = 255];
+    required ShootMode shootMode = 6     [default = KICK];
+    required TriggerMode triggerMode = 7 [default = STAND_DOWN];
+    required bool firmwareEnable = 8     [default = false];
```
justbuchanan: What's `firmwareEnable` for?

common/protobuf/RadioTx.proto
```diff
@@ -1,37 +1,32 @@
...
-		// Translational velocity in body coordinates, m/s
-		required float body_x = 6;
-		required float body_y = 7;
-		
-		// Angular velocity, rad/s
-		required float body_w = 8;
-		
-		optional float accel = 9;
-		optional float decel = 10;
+		UNICAST   = 1;
```
justbuchanan: Which packets will be sent as unicast?

common/protobuf/RadioTx.proto
```diff
@@ -1,37 +1,32 @@
...
+		MULTICAST = 2;
+	}
+
+	enum TxType
```
justbuchanan: Instead of having a TxType field, we can probably just check to see what TxBody is set.

common/protobuf/Control.proto
```diff
@@ -0,0 +1,33 @@
...
+    enum Song
+    {
+        STOP = 1;
+        FIGHT_SONG = 2;
+    }
+
+    required float xVelocity = 1         [default = 0];
+    required float yVelocity = 2         [default = 0];
+    required float aVelocity = 3         [default = 0];
+    required float dVelocity = 4         [default = 0];
```
justbuchanan: What's dVelocity for?

soccer/Processor.cpp
```diff
@@ -588,33 +588,44 @@ void Processor::run() {
...
+            control->set_firmwareenable(false);
+            control->set_song(Packet::Control::STOP);
         }
     }
 
     // Add RadioTx commands for visible robots and apply joystick input
     for (OurRobot* r : _state.self) {
         if (r->visible || _manualID == r->shell()) {
-            Packet::RadioTx::Robot* txRobot = tx->add_robots();
+            // Packet::RadioTx::Robot* txRobot = tx->add_robots();
```
justbuchanan: x

common/protobuf/Control.proto
guyfleeman: It tells the robot it needs to prepare to receive firmware.

soccer/Processor.cpp
```diff
@@ -588,33 +588,44 @@ void Processor::run() {
...
             // number of motors.
-            txRobot->CopyFrom(r->radioTx);
+            txRobot->CopyFrom(*(r->robotPacket));
 
             if (r->shell() == _manualID) {
-                JoystickControlValues controlVals = getJoystickControlValues();
-                applyJoystickControls(controlVals, txRobot, r);
+                const JoystickControlValues controlVals =
+                    getJoystickControlValues();
+                // myTestFunc();
```
justbuchanan: x

soccer/Processor.cpp
```diff
@@ -640,20 +650,21 @@ void Processor::applyJoystickControls(const JoystickControlValues& controlVals,
...
     // rotation
-    tx->set_body_w(controlVals.rotation);
+    tx->set_avelocity(controlVals.rotation);
 
     // kick/chip
     bool kick = controlVals.kick || controlVals.chip;
-    tx->set_kick_immediate(kick);
-    tx->set_kick(kick ? controlVals.kickPower : 0);
-    tx->set_use_chipper(controlVals.chip);
+    tx->set_triggermode(Packet::Control::IMMEDIATE);
```
justbuchanan: This should be IMMEDIATE only if `kick == true`

common/protobuf/RadioTx.proto
guyfleeman: Okay. I've got to change a few things to remove the unions. I'll fix this in the process.

soccer/Processor.cpp
```diff
@@ -640,20 +650,21 @@ void Processor::applyJoystickControls(const JoystickControlValues& controlVals,
...
+    tx->set_kcstrength(kick ? controlVals.kickPower : 0);
```
justbuchanan: We probably don't need to check if `kick == true` here since it will only kick/chip if the triggermode is set to do so.

common/protobuf/Control.proto
guyfleeman: Dribbler. Is it always at max? Rename?

common/protobuf/RadioTx.proto
guyfleeman: Everything but firmware for now. The header is extensible.

soccer/motion/MotionControl.cpp
```diff
@@ -33,7 +33,7 @@ void MotionControl::createConfiguration(Configuration* cfg) {
 MotionControl::MotionControl(OurRobot* robot) : _angleController(0, 0, 0, 50) {
     _robot = robot;
 
-    _robot->radioTx.set_robot_id(_robot->shell());
+    _robot->robotPacket->set_uid(_robot->shell());
```
justbuchanan: We'll have to figure out how we want to handle the difference between uid and shell id.

soccer/radio/USBRadio.cpp
```diff
@@ -237,11 +237,11 @@ void USBRadio::send(Packet::RadioTx& packet) {
...
-                                   (robot.kick_immediate() << 1) |
-                                   (robot.sing() << 2) | (robot.anthem() << 3);
-        forward_packet[offset++] = robot.accel();
-        forward_packet[offset++] = robot.decel();
+        forward_packet[offset++] = (robot.shootmode() == Packet::Control::CHIP) |
+                                   ((robot.triggermode() == Packet::Control::IMMEDIATE) << 1) |
+                                   (robot.song() << 2) | (0 /*robot.anthem()*/ << 3);
+        forward_packet[offset++] = 10; //robot.accel();
+        forward_packet[offset++] = 10; //robot.decel();
     }
```
justbuchanan: It would be nice if we could find a way to use the same serialization/deserialization code in firmware and soccer.  It may be hard to do though.

common/protobuf/Control.proto
justbuchanan: Gotcha.  Do we want that in the control packet or can the robot infer that when it receives the first firmware packet?

Also, it would be best to keep this PR to just the control packet refactor, then have another PR later that contains the firmware update stuff.  It may be a while until the firmware stuff gets done.

justbuchanan:
One thing to consider on the firmware-update side of things is that we currently do that outside of soccer.  We have a python script that does the ota-update and soccer just handles the control packets.  We could potentially integrate firmware updates into soccer, but it may be more work than doing it with a script like we have now for the old robots.

common/protobuf/Control.proto
guyfleeman: I wasn't sure if the bot would be able clean up everything it needs to between the first packet arriving, handing th data, and the second packet arriving. It'd also allow a status panel for it in soccer

common/protobuf/Control.proto
justbuchanan: :+1: sounds good, I missed that.  Maybe just put a comment for avelocity and dvelocity just to clarify what they are.

guyfleeman: I have lots of comments to add :)

guyfleeman:
I'd eventually like at least a status panel or something in soccer with the option to flash. For development reasons, we'd of course need to keep the script. Perhaps they can share some python.

justbuchanan:
Yeah it would be really cool if soccer could broadcast the git checksum of the latest firmware and automatically update robots running old code.  We could have the robot status panel show this stuff.

guyfleeman:
That's exactly what I had in mind

guyfleeman:
I was going to use a .bin checksum but we generate a header with git checksum and authoring info don't we. Even better

justbuchanan:
Sounds good.  Either checksum would work, although checksumming the file like you suggested would probably be better.  If we use the git checksum, it will try and update firmware whenever any irrelevant files in the repo are changed.

common/protobuf/Control.proto
```diff
@@ -0,0 +1,33 @@
...
+        IMMEDIATE = 1;
+        ON_BREAK_BEAM = 2;
+        STAND_DOWN = 3;
+    }
+
+    enum Song
+    {
+        STOP = 1;
+        FIGHT_SONG = 2;
+    }
```
justbuchanan: Right now in the `GameplayModule`, we call `sing()` once when we score a goal.  We might need a third entry in the enum for "keep playing" so that we don't tell it to play the fight song in one packet, then tell it to stop in the next packet.

common/protobuf/Control.proto
```diff
@@ -0,0 +1,34 @@
...
+
+    required float xVelocity = 1         [default = 0];
+    required float yVelocity = 2         [default = 0];
+    required float aVelocity = 3         [default = 0];
+    required float dVelocity = 4         [default = 0];
+    required uint32 kcStrength = 5       [default = 255];
+    required ShootMode shootMode = 6     [default = KICK];
+    required TriggerMode triggerMode = 7 [default = STAND_DOWN];
+    required bool firmwareEnable = 8     [default = false];
+    required Song song = 9               [default = STOP];
```
justbuchanan: This should probably default to CONTINUE

common/protobuf/Control.proto
justbuchanan: Let's leave `firmwareEnable` out for now

common/protobuf/RadioTx.proto
```diff
@@ -1,37 +1,27 @@
...
-	repeated Robot robots = 2;
-	optional uint32 sequence = 3;
+
+	message RobotCollection {
+		repeated Robot robots = 2;
+	}
+
+	required TxMode txMode = 1;
+	required TxType txType = 2;
+	optional RobotCollection robotCollection = 4;
```
justbuchanan: Since `RobotCollection` just has one field, want to just remove it and replace this line with `repeated Robot robots`?

common/protobuf/RadioTx.proto
```diff
@@ -1,37 +1,27 @@
...
 
-		// measurement from the camera
-		optional Point position = 14;
-		optional float latency = 15;
-		
-		optional bool kick_immediate = 11;
-		
-		optional bool sing = 12;
-		optional bool anthem = 13;
+	enum TxType
```
justbuchanan: Do we need a TxType field?  I think we can get away with just checking which packet body/message type is set.

soccer/Robot.cpp
```diff
@@ -300,7 +311,13 @@ void OurRobot::unkick() {
     *_cmdText << "unkick()" << endl;
 }
 
-void OurRobot::kickImmediately(bool im) { radioTx.set_kick_immediate(im); }
+void OurRobot::flagFlash() { flashingFw = true; }
```
justbuchanan: Let's leave these methods out until a firmware-specific PR.

soccer/Robot.hpp
```diff
@@ -408,9 +415,21 @@ class OurRobot : public Robot {
...
      */
-    void sing() {
+    void sing() { sing(Packet::Control::FIGHT_SONG); }
+
+    /**
+     * @brief start the robot playing a song
+     * @param song
+     */
+    void sing(Packet::Control::Song song) {
+        // std::string songName = "";
```
justbuchanan: x

soccer/radio/USBRadio.cpp
```diff
@@ -202,7 +202,7 @@ void USBRadio::send(Packet::RadioTx& packet) {
 
     // Build a forward packet
     forward_packet[0] = _sequence;
-    packet.set_sequence(_sequence);
+    // packet.set_sequence(_sequence);
```
justbuchanan: x

soccer/radio/USBRadio.cpp
```diff
@@ -237,11 +242,12 @@ void USBRadio::send(Packet::RadioTx& packet) {
...
-                                   (robot.kick_immediate() << 1) |
-                                   (robot.sing() << 2) | (robot.anthem() << 3);
-        forward_packet[offset++] = robot.accel();
-        forward_packet[offset++] = robot.decel();
+        forward_packet[offset++] =
+            (robot.shootmode() == Packet::Control::CHIP) |
+            ((robot.triggermode() == Packet::Control::IMMEDIATE) << 1) |
+            (robot.song() << 2) | (0 /*robot.anthem()*/ << 3);
+        forward_packet[offset++] = 10;  // robot.accel();
+        forward_packet[offset++] = 10;  // robot.decel();
```
justbuchanan: What do we want to do for accel/decel values?

guyfleeman: are these constraints or dynamic values I found them hard coded in one place and dynamic here?

justbuchanan: I searched through the firmware and it looks like they're read out of the packet, but never actually used.  For now maybe just leave a TODO comment here to remove them.  Eventually we can remove them from the packet format and shrink the packet size.

common/protobuf/Control.proto
justbuchanan: x

simulator/physics/Robot.cpp
```diff
@@ -245,10 +245,13 @@ void Robot::getWorldTransform(btTransform& chassisWorldTrans) const {
...
-void Robot::radioTx(const Packet::RadioTx::Robot* data) {
-    velocity(data->body_x(), data->body_y(), data->body_w());
-    _controller->prepareKick(data->kick(), data->use_chipper());
-    _controller->prepareDribbler(data->dribbler());
+void Robot::radioTx(const Packet::Control* data) {
+    velocity(data->xvelocity(), data->yvelocity(), data->avelocity());
+    _controller->prepareKick(data->triggermode() == Packet::Control::IMMEDIATE
+                                 ? data->kcstrength()
+                                 : 0,
+                             data->shootmode());
```
justbuchanan: Should this be `shootmode() == CHIP`?  I'm not sure if the protobuf enum value is guaranteed to be the right thing here.

simulator/physics/Robot.cpp
```diff
@@ -245,10 +245,13 @@ void Robot::getWorldTransform(btTransform& chassisWorldTrans) const {
     _robotChassis->getMotionState()->getWorldTransform(chassisWorldTrans);
 }
 
-void Robot::radioTx(const Packet::RadioTx::Robot* data) {
-    velocity(data->body_x(), data->body_y(), data->body_w());
-    _controller->prepareKick(data->kick(), data->use_chipper());
-    _controller->prepareDribbler(data->dribbler());
+void Robot::radioTx(const Packet::Control* data) {
+    velocity(data->xvelocity(), data->yvelocity(), data->avelocity());
+    _controller->prepareKick(data->triggermode() == Packet::Control::IMMEDIATE
```
justbuchanan: We should call `prepareKick()` for both IMMEDIATE and ON_BREAK_BEAM.

soccer/Robot.cpp
```diff
@@ -300,7 +305,9 @@ void OurRobot::unkick() {
     *_cmdText << "unkick()" << endl;
 }
 
-void OurRobot::kickImmediately(bool im) { radioTx.set_kick_immediate(im); }
+void OurRobot::kickImmediately(bool im) {
```
justbuchanan: Since we're not using the bool parameter anymore, we should probably get rid of it

soccer/radio/USBRadio.cpp
```diff
@@ -212,21 +211,24 @@ void USBRadio::send(Packet::RadioTx& packet) {
...
+        float bodyVelW =
+            robot.avelocity() * Seconds_Per_Cycle / Radians_Per_Tick;
 
         int outX = clamp((int)roundf(bodyVelX), -511, 511);
         int outY = clamp((int)roundf(bodyVelY), -511, 511);
         int outW = clamp((int)roundf(bodyVelW), -511, 511);
 
-        uint8_t kick = robot.kick();
-        uint8_t dribbler = max(0, min(255, robot.dribbler() * 2));
+        uint8_t kick =
```
justbuchanan: Should this be kick strength instead?

simulator/physics/Robot.cpp
```diff
@@ -245,10 +245,13 @@ void Robot::getWorldTransform(btTransform& chassisWorldTrans) const {
...
-void Robot::radioTx(const Packet::RadioTx::Robot* data) {
-    velocity(data->body_x(), data->body_y(), data->body_w());
-    _controller->prepareKick(data->kick(), data->use_chipper());
-    _controller->prepareDribbler(data->dribbler());
+void Robot::radioTx(const Packet::Control* data) {
+    velocity(data->xvelocity(), data->yvelocity(), data->avelocity());
+    _controller->prepareKick(data->triggermode() != Packet::Control::STAND_DOWN
+                                 ? data->kcstrength()
+                                 : 0,
+                             data->shootmode());
```
justbuchanan: I think this needs to be `shootmode() == CHIP`.  Just passing shootmode relies on CHIP's enum value being 1 and KICK's enum value being 0, which I don't think is the case.

PR #604, opened 2016-04-06
Added gui to show if the usb radio base station is connected and fixed Race Condition Bug

### LINKED ISSUE #283: Add widget to soccer that shows whether or not a radio base station is plugged in

chachmu:
Closes #283 and #607


soccer/ui/MainWindow.ui
```diff
@@ -830,35 +830,70 @@
...
+           </size>
+          </property>
+          <property name="frameShape">
+           <enum>QFrame::Box</enum>
+          </property>
+          <property name="frameShadow">
+           <enum>QFrame::Sunken</enum>
+          </property>
+          <property name="text">
+           <string>Simulating</string>
```
justbuchanan: Can you change this to "Simulator Radio" just to clarify that this is a radio status indicator?  Otherwise it's not very clear what it's indicating.

justbuchanan:
Looks good!  I made one minor note, but other than that it's good to go.

soccer/Processor.hpp
```diff
@@ -177,6 +177,8 @@ class Processor : public QThread {
 
     void setFieldDimensions(const Field_Dimensions& dims);
 
+    const bool isRadioOpen();
```
justbuchanan: The `const` is in the wrong place here.  As is, it says that the return value shouldn't be modified, but I think we'd prefer if it said that calling the method doesn't modify the `Processor` object.  To do this, change it to `bool isRadioOpen() const;`

soccer/MainWindow.cpp
```diff
@@ -793,6 +797,17 @@ void MainWindow::status(QString text, MainWindow::StatusType status) {
     }
 }
 
+void MainWindow::radioBaseStatus(bool usbRadio) {
```
ashaw596: Change name to be more informative

soccer/MainWindow.hpp
```diff
@@ -157,6 +157,7 @@ private Q_SLOTS:
     typedef enum { Status_OK, Status_Warning, Status_Fail } StatusType;
 
     void status(QString text, StatusType status);
+    void radioBaseStatus(bool usbRadio);
```
ashaw596: Same thing

soccer/main.cpp
```diff
@@ -181,6 +181,9 @@ int main(int argc, char* argv[]) {
 
     processor->start();
 
+    while (!processor->isInitialized()) {
+    }
```
ashaw596: Same line?
Also add comment

ashaw596:
Looks good

PR #605, opened 2016-04-07
2015 base station firmware

justbuchanan:
This hasn't been thoroughly tested yet, but I think it's most of the way there.  We subclass MBED's `USBDevice` class to provide our own device descriptor that defines the input and output "endpoints" (usb terminology).  We have one input endpoint and one output.  Whenever we receive input, we copy the contents into a packet and send it out over the radio.  Whenever a packet is received, we push its contents to the pc via the output endpoint.  The nice thing about this implementation is that it's fairly simple and almost identical to the previous base station, so we don't have to change the code in soccer much at all.  Right now the only real change I made to soccer was to make it look for different usb vendor/product ids.

TODO:

- [x] add radio control commands to endpoint zero
  - [x] stall the endpoint when an invalid command/request is sent
- [x] test tx
- [ ] test rx
- [x] figure out how packet headers (i.e. port number) will be sent from soccer
  - [x] figure out size byte prepended to packet content
- [ ] more documentation
- [x] switch USBRadio to use cc1201 commands
  - [ ] figure out channel-setting for cc1201 and fix `USBRadio.channel()`
- [x] fix rx/tx leds
- [x] soccer crashes on startup, not sure why
- [x] reset old configuration when a new one is set
- [x] define control request enum in usb-interface.hpp
- [x] base station occasionally stops transmitting and requires a reboot...

see #278


justbuchanan:
@jjones646 do you know the best way to set the radio channel for the cc1201?  The 1101 had a `CHANNR` register that made it really easy, but the 1201 doesn't seem to have it.

justbuchanan:
Update: I've got this mostly all working.  We can now send packets from soccer to new robots!  I still need to figure out channel-setting and test that we can receive packets from robots, but other than that it should be good to go.

PR #613, opened 2016-04-11
Removed task-signals

justbuchanan:
@jjones646 I don't think we need the back and forth signaling between the main and task threads.  The initialization they do should be independent so they can initialize in parallel.  Do you see any issues with this?

### DIFF
--- a/firmware/robot2015/src-ctrl/main.cpp
+++ b/firmware/robot2015/src-ctrl/main.cpp
@@ -11,7 +11,6 @@
 #include <assert.hpp>
 
 #include "robot-devices.hpp"
-#include "task-signals.hpp"
 #include "commands.hpp"
 #include "fpga.hpp"
 #include "io-expander.hpp"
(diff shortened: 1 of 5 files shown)


jjones646:
There used to be pretty major trouble with the ordering of startup configs. But the firmware is a lot more stable now, so it's very possible this might not be needed anymore.

I would thoroughly check every combination of mbed + control board + radio board before merging this. But if all that checks out, then I'd say it's fine.

justbuchanan:
Sounds good, I'll give this some more thorough testing :+1: .  Just from looking through the code for each thread, it doesn't look like they're touching anything shared so I think it's good.  Part of the issue before may have been because there was no Mutex around the spi bus.

justbuchanan:
I noticed that running the `su` command with these changes results in an exception and a reboot.  Pretty much the same thing as in #639.  It also only happens when it's on the control board...

I'll leave this unmerged for now since the `su` issue isn't present on master.  We should figure out what's up with the control board affecting the console, I'm not sure what it could be :confused:

PR #614, opened 2016-04-11
Fixed red status leds on control board.

justbuchanan:
I think these are all working now.  I noticed some of the pin mappings are off though.  @jjones646 do you have a schematic showing which leds are connected to which lines on the io expander?


justbuchanan:
Nevermind, I got the leds pins remapped.

PR #625, opened 2016-04-12
Attempt to fix "ls" command

### LINKED ISSUE #610: ls command in firmware is broken
Here's what I get when I run it:

~~~{.sh}
Justin Buchanan@robot $ ls

ls:
    Description:        List contents of current directory
    Usage:              ls
Justin Buchanan@robot $
~~~

Any ideas @jjones646?

justbuchanan:
@jjones646 I uncommented the `LocalFileSystem` line you pointed out and the `ls` command now prints "Exception" and reboots...  Any ideas?

closes #610

### DIFF
--- a/firmware/robot2015/src-ctrl/modules/commands/commands.cpp
+++ b/firmware/robot2015/src-ctrl/modules/commands/commands.cpp
@@ -50,7 +50,7 @@ int (*iterative_command_handler)(cmd_args_t& args);
 }  // end of anonymous namespace
 
 // Create an object to help find files
-// LocalFileSystem local("local");
+LocalFileSystem local("local");
 
 /**
  * Commands list. Add command handlers to commands.hpp.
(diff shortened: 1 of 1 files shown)


jjones646:
Did that happen when you were using the control board with the mbed?

justbuchanan:
Interesting... I tried it without the control board and it works perfectly!  What's the deal?

jjones646:
I'm not sure. Weird interrupt timing most likely.

jjones646:
This is interesting though...
The only difference really is that the FPGA will successfully configure, so the bitfile is opened, sent, etc. Maybe a quirk in the mbed's implementation for closing a file..? That's a pretty wild guess though.

jjones646:
If you can get a baudrate of 115200 to work, give that a try. If that's it, then there may be some room for interrupt priority improvements. 

Every time I've used a 115200 baudrate with the mbed, the console characters get corrupted pretty frequently. But you might have been luck, and it's worth trying out for this.

justbuchanan:
Yeah I'm not sure.  I was able to track it down a little bit and it's the `opendir()` method that's throwing the exception.

I'll go ahead and merge this since it fixes `ls` for the general case and open a new issue for whatever hardware/software bug this is.

justbuchanan:
Sounds good, I'll try changing the baud rate and see what happens.

PR #627, opened 2016-04-13
Fixed Race Condition Bug

### LINKED ISSUE #607: Fix race condition in soccer
Not sure what the cause is yet, but occasionally (not always) soccer crashes at startup with this error message:

~~~{.sh}
  File "/home/justin/src/rj/robocup-software/soccer/gameplay/play_registry.py", line 298, in data
    return str(node.play_class.score())
  File "/home/justin/src/rj/robocup-software/soccer/gameplay/plays/offense/basic_122.py", line 58, in score
    return 10 if main.game_state().is_playing() else float("inf")
AttributeError: 'NoneType' object has no attribute 'is_playing'
/bin/sh: line 1: 14152 Aborted                 (core dumped) ./soccer
makefile:24: recipe for target 'run' failed
make: *** [run] Error 134
~~~

note: this can be reproduced by putting a small delay in `USBRadio.open()` and having a radio connected.

chachmu:
Made main.cpp wait until processor is finished setting up to continue. Closes #607

### DIFF
--- a/soccer/Processor.hpp
+++ b/soccer/Processor.hpp
@@ -177,6 +177,10 @@ public:
 
     void setFieldDimensions(const Field_Dimensions& dims);
 
+    const bool isRadioOpen();
+
+    const bool isInitialized();
+
     ////////
 
     // Time of the first LogFrame
@@ -274,4 +278,6 @@ private:
     bool _useFieldOrientedManualDrive = false;
 
     VisionReceiver vision;
+
+    bool _initialized;
 };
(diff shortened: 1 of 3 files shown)


soccer/Processor.hpp
```diff
@@ -177,6 +177,10 @@ class Processor : public QThread {
 
     void setFieldDimensions(const Field_Dimensions& dims);
 
+    const bool isRadioOpen();
+
+    const bool isInitialized();
```
justbuchanan: This could use a brief comment to give some context to what is initialized and the fact that start/run is what does it.

PR #629, opened 2016-04-13
Tuning Protocol 2016 Feature

guyfleeman:
Will address discussion in #304.

### DIFF
--- a/common/protobuf/Control.proto
+++ b/common/protobuf/Control.proto
@@ -24,8 +24,11 @@ message Control
 
     required float xVelocity = 1         [default = 0];
     required float yVelocity = 2         [default = 0];
+    //angular velocity
     required float aVelocity = 3         [default = 0];
+    //dribbler velocity
     required float dVelocity = 4         [default = 0];
+    //kick/chip strength
     required uint32 kcStrength = 5       [default = 255];
     required ShootMode shootMode = 6     [default = KICK];
     required TriggerMode triggerMode = 7 [default = STAND_DOWN];
--- a/common/protobuf/Robot.proto
+++ b/common/protobuf/Robot.proto
@@ -1,9 +1,11 @@
 package Packet;
 
 import "Control.proto";
+import "Tuning.proto";
 
 message Robot
 {
     required uint32 uid = 1;
-    required Control control = 3;
+    required Control control = 2;
+    optional Tuning tuning = 3;
 }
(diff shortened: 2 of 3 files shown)


justbuchanan:
@guyfleeman how's progress on this?

justbuchanan:
@guyfleeman if you want to push what you have for the soccer side of this feature, I can take a look at getting the mbed side up and running.  What have you figured out / implemented with ensuring that the values are in sync given that our communication link is unreliable?

The majority of the mbed side of the radio protocol is [here](https://github.com/RoboJackets/robocup-software/blob/b36793ac6c34f64f996e3735e0e067089d1aff7d/firmware/robot2015/src-ctrl/modules/RadioProtocol.hpp) with a little more in the main.cpp file.

guyfleeman:
@justbuchanan Most of my work is outside of soccer trying to figure out the XML stuff. Its not in a great state to push, but I can get it there soon. I'll look at the radio stuff in the meantime.

justbuchanan:
Sounds good.  I don't think you'll need to mess with any xml though.  The config files are saved on disk as xml, but for syncing the values to the robot it would be best to read soccer's global `Configuration` object rather than the saved file.

guyfleeman:
@justbuchanan will that class support the UID based system for 2015? Or are we waiting on that?

justbuchanan:
The config system right now supports storing values by keypath such as "Planning/MaxAccel", etc.  In order do individual tuning values for each robot, we could do something like "Robot1/Pid/P", "Robot2/Pid/P", etc.  I think for now though we'll probably just use one set of values that apply to all robots and leave per-robot unique values as a later task.

I think we'll basically want to setup a mapping between the keypaths in the config system and the names of the parameters on the robot side.  Then when we do a sync, we'll just read the value for each keypath we want and send them out to the robots.

guyfleeman:
@justbuchanan in that case, it shouldn't be too bad, I'll get the soccer side up this weekend. Sorry for the delay on this one...

guyfleeman:
@justbuchanan to start I'm just going to write the state of a RobotConfig instance to a packet and send that. As far as I can tell, everything is using the RobotConfig class. These are the values we need to send right?

justbuchanan:
Yeah that sounds good to me! The RobotConfig class should have it all.

jgkamat:
Closing due to timeout. If anyone starts work on this, please re-open :smile:

PR #630, opened 2016-04-13
Created standard_play.py and made all plays functional and inheriting standard_play.py

### LINKED ISSUE #366: Find a good way to run test plays without defense
Right now we have to comment out lines of code to make this happen

jcarn:
From what I can tell everything works like it should, but I would appreciate if people with more experience could verify that the plays are running normally. I'll continue checking in the meantime. 

I created new class standard_play. It inherits from play.py. Right now, the only thing it does is handle defense for every play based on the status of the "Use Defense" checkbox in the UI. For every play that inherited play.py, In the future, any action that a normal (but perhaps not every play) play should implement can be placed into standard_play instead of play. I modified every play that inherits from play to instead inherit from standard_play. I also modified the plays to let standard_play handle defense.
closes #366


justbuchanan:
Looks good :+1: .  I'll go through this in more depth hopefully later today, but one quick note for now.  A lot of the testing plays just use one or two robots and never used to use defense.  These should be changed back to how they were so they don't get defense added to them.

jcarn:
Also closes #366

soccer/gameplay/plays/tuning/test_jittery_motion.py
```diff
@@ -47,7 +47,7 @@ def execute_running(self):
         self.robot.set_world_vel(vel)
 
 
-class TestJitteryMotion(play.Play):
+class TestJitteryMotion(standard_play.StandardPlay):
     def __init__(self):
```
jgkamat: These tuning plays shouldn't have defense added to them (I'm guessing either continue to extend from Play, or add an option somehow to turn off defense in defense).

soccer/ui/MainWindow.ui
```diff
@@ -1559,6 +1559,16 @@ p, li { white-space: pre-wrap; }
                   </widget>
                  </item>
                  <item>
+                  <widget class="QCheckBox" name="useDefenseCheckBox">
```
jgkamat: It would be nice to have defense automatically turn off/on when you check/uncheck the box, w/o restarting the play, but thats more of a stretch goal.

jgkamat:
Overall, looking very good! A few of the plays shouldn't have defense added to them under any circumstance (reverting them to extending normal Play should be fine), but otherwise looking good. 

Also we should consider renaming standardPlay to something along the lines of standardCompPlay (this is still a pretty bad name too though...) The 'standardPlay' (i'm guessing) is a root class for any top level plays we actually run in comp. Maybe rootPlay or topLevelPlay? I'm not really sure.

soccer/ui/MainWindow.ui
jcarn: Yeah that is a good idea. I'll look into it, I think it would require messing around in rootplay but it shouldn't be too hard. Also it does stop sometimes go on/off automatically when you check/uncheck the box, but that is sporadic and I don't know why it happens.

jcarn:
I'll get right on the reverting.

Rootplay already exists, although it isn't a root class for top level plays at all. I like the name topLevelPlay, we can discuss later today at the meeting.

I should have known that, thanks for the assist

justbuchanan:
You can remove the class method `handles_goalie()` from the plays that now inherit from `StandardPlay`

soccer/gameplay/standard_play.py
```diff
@@ -0,0 +1,46 @@
...
+
+    #Handles activity while the play is active. A play wishing to utilize this
+    #method in additionto having an "execute_running" method of its own must call
+    #it via super
+    def execute_running(self):
+        self.use_standard_defense()
+
+    #If using defense, it defaults to play's method, otherwise returns True
+    @classmethod
+    def handles_goalie(cls):
```
justbuchanan: This should just be `return True`.  The superclass in this case is `Play`, so calling it's `handles_goalie()` method doesn't get you much.

soccer/gameplay/standard_play.py
```diff
@@ -0,0 +1,46 @@
...
+class StandardPlay(play.Play):
+
+    #Performs actions that all "Standard Plays" should do on initialization
+    #Note: This method is called many times during the duration of a play,
+    #Not just on selection
+    def __init__(self, continuous):
+        super().__init__(continuous)
+        self.use_standard_defense()
+
+        #If the "Use Defense" checkbox is checked and the play isn't already running
```
justbuchanan: It looks a little odd that this is the only comment indented away from the method's indentation

soccer/gameplay/ui/main.py
```diff
@@ -16,6 +16,18 @@ def getMainWindow():
 _has_setup_ui = False
 
 
+def defenseEnabled():
+    global _has_setup_ui
```
justbuchanan: Since you don't use `_has_setup_ui`, you don't have to declare it global here

soccer/gameplay/plays/restarts/kick_penalty.py
```diff
@@ -29,7 +29,3 @@ def __init__(self):
     def score(cls):
         gs = main.game_state()
         return 0 if gs.is_our_penalty() else float("inf")
-
-    @classmethod
-    def is_restart(cls):
```
justbuchanan: This should be put back in

soccer/gameplay/standard_play.py
justbuchanan: bump

PR #631, opened 2016-04-13
Replaced ready and delay task lists with active task list

JNeiger:
Relevant to PR #599 and #583


justbuchanan:
Looks great!  I just ran this on an mbed and the printout is exactly what I'd expect to see, nice work!  It's good to see that several threads have stacks way bigger than they need, we've got plenty of space we can free up :)

~~~{.sh}
JNeiger@robot $ ps                                  STACK SIZE (bytes)
ID      PRIOR   STATE   Δ TIME          MAX             ALLOC           CURRENT
1       4       3       0               23128           23128           1248      
2       6       3       221             2044            2048            112       
3       6       3       7               772             1024            200       
4       5       2       0               1212            2048            880       
5       5       8       0               716             1024            176       
6       6       8       0               716             1024            176       
7       4       6       0               732             1024            176       
==============
Total Threads:  7
~~~

firmware/common2015/utils/rtos-mgmt/helper-funcs.hpp
```diff
@@ -1,3 +1,5 @@
+/** @file */
```
justbuchanan: x

PR #643, opened 2016-04-13
dont use sudo for git submodules

### LINKED ISSUE #642: Make gitsubmodules commands run as normal user and chown all the external directories
Not terribly important. But it be nice to prevent problems in the future that make drastic changes to our submodules. The owners on those are all mixed with root right now.

justbuchanan:
closes #642 

cc @jgkamat

### DIFF
--- a/util/ubuntu-setup
+++ b/util/ubuntu-setup
@@ -179,5 +179,6 @@ pip2 install -r $BASE/util/requirements2.txt
 # This script is run as sudo, but we don't want submodules to be owned by the
 # root user, so we use `su` to update submodules as the normal user
 echo "-- Updating submodules"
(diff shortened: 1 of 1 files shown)


ashaw596:
Um, I think this will break on most peoples computers if you don't chown everything first and if theres any submodule changes.

ashaw596:
su $(whoami) will also probably come out as su root

justbuchanan:
I tried it out and `su $(whoami)` returns the non-root user.  I was surprised too :)

You're right that people will have to chown before running this though.

ashaw596:
We could throw in a chown $(whoami) in there for a while then so we don't have too many issues if people skip a meeting.

justbuchanan:
Yeah that's not a bad idea.  Just added it.

PR #648, opened 2016-04-15
Sublime Project Organization

jjones646:
Not sure how everyone feels about changing up the sublime project file a bit. But figured I would open this and see if anyone likes it?


justbuchanan:
Interesting, I didn't realize you could do stuff like that in sublime.  Personally, I kinda like the hierarchy in sublime to match the actual folder layout for the project - it makes it a little confusing having things organized in two different ways and using both layouts often.  Maybe we could look at rearranging the folders some though.  Also, :+1: for the readme changes.

PR #651, opened 2016-04-17
Implement Ball Placement Challenge

### LINKED ISSUE #542: Implement ball placement for 2016 technical challenge
One of this year's technical challenges is to receive ball-placement commands from the referee system and have our robots carry them out.  The intention is that in the future, ref calls will be all automated and the robots on the field will set the ball up in the right position for restart plays.

Details here: http://wiki.robocup.org/wiki/Small_Size_League/RoboCup_2016/Technical_Challenges/Autonomous_Ball_Placement.

chachmu:
Closes #542.


chachmu:
I need to figure out how to send test protobuf messages to see which parts of this don't work, also, the play that I am working on is definitely not done but I can't figure out exactly what needs to be changed without testing it.

soccer/gameplay/tactics/placement.py
```diff
@@ -0,0 +1,71 @@
...
+            lambda: self.subbehavior_with_name('capture').state==behavior.Behavior.State.completed,
+            'done capturing')
+
+        self.add_transition(Placement.State.placing, Placement.State.standby,
+            lambda: self.subbehavior_with_name('placing').state==behavior.Behavior.State.completed,
+            'done placing')
+
+        self.add_transition(
+            Placement.State.standby, behavior.Behavior.State.completed,
+            lambda: self.subbehavior_with_name('standby').state == behavior.Behavior.State.completed,
```
jgkamat: Not sure about this, but I think there needs to be something here about turning off the intake roller. Also we might want to make another move command to move away from the ball w/o touching it, before moving to the standby location. (You could just add the chosen robot to the line_up play for that, possibly).

jgkamat:
I'm not sure how we could test this, but I would expect it to be part of the refbox (but I couldn't find it). @justbuchanan?

justbuchanan:
Looks good!  It took a long time of messing around with the ssl-refbox code, but I figured out how you can test this.

The ball placement commands aren't a part of the ssl-refbox gui, but are done via an external program.  This external program called the "remote control client" tells the refbox program when to send ball placement commands.

Here's a step-by-step:

* Download, compile (`make`), and run ssl-refbox
* Enable remote control by clicking on the "Config" menu item and then "Enable Remote Control"
* Compile and run `rcon-client`.  `cd` into the `rcon-client` directory and run `make`, then `rcon-client localhost`.  Type `help` to see a list of commands and how to use the program.
* Run `soccer`
* Enter the "Stopped" state by clicking "Stop Game" in ssl-refbox
* Send a ball placement command from `rcon-client`

Also, another helpful tool is the example refbox client provided in the ssl-refbox repo (in the "client" directory).  This just listens for refbox commands and prints to the console whenever it receives anything.  You can use this to help debug problems and to verify that the right commands are coming out of the refbox.

soccer/gameplay/plays/restarts/our_placement.py
```diff
@@ -0,0 +1,35 @@
...
+
+        self.add_transition(
+            behavior.Behavior.State.running, behavior.Behavior.State.completed,
+            lambda: self.kicker.is_done_running(), 'when kicker finishes.')
+
+        self.kicker = tactics.penalty.Penalty()
+        self.add_subbehavior(self.kicker, 'kicker', required=True, priority=10)
+
+        line = robocup.Segment(robocup.Point(1.5, 1), robocup.Point(1.5, 2.5))
+        line_up = tactics.line_up.LineUp(line)
```
justbuchanan: Make sure to call `add_subbehavior()`

chachmu:
Ok, so the play technically works for a placement from our team but it kind of overshoots the target. I still need to:

- [ ] Figure out why the robot slightly veers as it travels causing the ball to roll off the side
- [ ] Limit the robots approach velocity somehow, right now it slows down to quickly when it stops so the ball rolls away
- [ ] Make the other robots stay out of the way during our placement
- [ ] Make a play to handle the other team having a placement

#makefile#
```diff
@@ -0,0 +1,206 @@
+MAKE_FLAGS = --no-print-directory
```
justbuchanan: This file looks like it may have been an accidental copy of the makefile?

soccer/gameplay/plays/restarts/placement.py
```diff
@@ -0,0 +1,50 @@
...
+            behavior.Behavior.State.running, behavior.Behavior.State.completed,
+            lambda: self.placer.is_done_running(), 'when placer finishes.')
+
+        self.placer = tactics.our_placement.OurPlacement()
+        self.add_subbehavior(self.placer, 'placer', required=True, priority=90)
+
+        line = robocup.Segment(robocup.Point(1.5, 1), robocup.Point(1.5, 2.5))
+        line_up = tactics.line_up.LineUp(line)
+
+    def execute_running(self):
```
justbuchanan: You should add `super().execute_running()` at the top of this method so that superclass behavior gets run.

soccer/gameplay/plays/restarts/placement.py
```diff
@@ -0,0 +1,50 @@
...
+import tactics.line_up
+import tactics.our_placement
+
+
+# one robot places the ball, the others just line up and wait
+class Placement(standard_play.StandardPlay):
+    def __init__(self):
+        super().__init__(continuous=True)
+
+        self.remove_subbehavior('defense')
```
justbuchanan: Why remove defense here?

soccer/gameplay/plays/restarts/placement.py
```diff
@@ -0,0 +1,50 @@
...
+
+        self.add_transition(
+            behavior.Behavior.State.running, behavior.Behavior.State.completed,
+            lambda: self.placer.is_done_running(), 'when placer finishes.')
+
+        self.placer = tactics.our_placement.OurPlacement()
+        self.add_subbehavior(self.placer, 'placer', required=True, priority=90)
+
+        line = robocup.Segment(robocup.Point(1.5, 1), robocup.Point(1.5, 2.5))
+        line_up = tactics.line_up.LineUp(line)
```
justbuchanan: The line_up behavior doesn't get run because it's not added as a subbehavior.

soccer/gameplay/plays/restarts/placement.py
```diff
@@ -0,0 +1,50 @@
...
+    def score(cls):
+        gs = main.game_state()
+        return 0 if gs.is_our_placement() else float("inf")
+
+    @classmethod
+    def is_restart(cls):
+        return True
+
+    @classmethod
+    def handlesgoalie(cls):
```
justbuchanan: You can leave out this method - the `StandardPlay` class implements this already and handles the goalie/defense behaviors.

soccer/gameplay/evaluation/double_touch.py
```diff
@@ -39,7 +39,7 @@ def __init__(self):
         self.add_transition(
             DoubleTouchTracker.State.restart_play_began,
             DoubleTouchTracker.State.kicking,
-            lambda: any(bot.has_ball() for bot in main.our_robots()),
+            lambda: (any(bot.has_ball() for bot in main.our_robots())) and not main.game_state().is_placement,
```
justbuchanan: The `is_placement` needs parentheses since it's a method rather than a property.  Should be `is_placement()`.

common/Field_Dimensions.hpp
```diff
@@ -10,56 +10,58 @@
...
 #include <Geometry2d/Rect.hpp>
+#include <Geometry2d/Line.hpp>
 
 /// This class contains constants defining the layout of the field.
 /// See the official SSL rules page for a detailed diagram:
 /// http://robocupssl.cpe.ku.ac.th/rules:main
 struct Field_Dimensions {
-    inline float Length() const { return _Length; }
-    inline float Width() const { return _Width; }
+    float Length() const { return _Length; }
```
ashaw596: Not terribly important, but any particular reason all the inlines were removed?

soccer/gameplay/root_play.py
```diff
@@ -39,10 +39,18 @@ def execute_running(self):
         ################################################################################
 
         if main.game_state().is_stopped():
-            if not isinstance(self.play, plays.stopped.Stopped):
-                logging.info("Running 'Stopped' play due to game state change")
-                self.play = plays.stopped.Stopped()
-                self._currently_restarting = True
+            if main.game_state().is_placement():
```
justbuchanan: Is placement a "sub state" of stopped?  If so, be sure the referee module sets the GameState to stopped when there's a placement command.  If not, then this shouldn't go inside the "if stopped" above.

soccer/gameplay/robocup-py.cpp
```diff
@@ -94,6 +95,11 @@ void OurRobot_move_to_direct(OurRobot* self, Geometry2d::Point* to) {
     if (to == nullptr) throw NullArgumentException("to");
     self->moveDirect(*to);
 }
+void OurRobot_move_to_direct_end_vel(OurRobot* self, Geometry2d::Point* to,
```
ashaw596: Is there a particular reason we need this now? The moveDirect function really isn't greatly designed.

soccer/gameplay/robocup-py.cpp
```diff
@@ -457,6 +463,16 @@ void WinEval_add_excluded_robot(WindowEvaluator* self, Robot* robot) {
     self->excluded_robots.push_back(robot);
 }
 
+struct VecToList {
```
ashaw596: What is this for?

common/Field_Dimensions.hpp
justbuchanan: That was actually part of PR #656.  The compiler automatically inlines small methods, so for 99% of cases, you shouldn't have to specify inline.

soccer/gameplay/tactics/our_placement.py
```diff
@@ -0,0 +1,62 @@
...
+
+    ## prefer to get assigned robot closest to ball
+    def role_requirements(self):
+        reqs = super().role_requirements()
+        if isinstance(reqs, role_assignment.RoleRequirements):
+            reqs.destination_shape = main.ball().pos
+
+        return reqs
+
+#change moveplay to something else
```
justbuchanan: x

soccer/gameplay/root_play.py
chachmu: The refbox can only send a placement command if it is in the stopped state already.

soccer/gameplay/plays/restarts/placement.py
chachmu: If the placement is set to be near our goal, the goalie will come out and kick the ball across the field. This is detrimental if its our placement and I think breaks the rules if it is theirs.

soccer/gameplay/robocup-py.cpp
chachmu: If I use the normal move function it tries to path plan around the ball instead of moving with it.

justbuchanan:
Looks pretty good so far!

* To address the issue with the robot turning and losing the ball as it travels, you should look at the robot.face() method.  The `Aim` method you're using right now in `Dribble` is more for aiming when we're about to kick the ball.  It's goal is to precisely aim at a target.  You can remove `Aim` and just call `face()` in your `execute_running()` (or execute_OtherState) and it will keep the robot facing towards whatever point you specify.
* Albert's work on robot-robot (PR #663) avoidance may help here with keeping other robots out of the way.  Another thing that we could do is add an obstacle that the other robots avoid, but the placement robot ignores.
* During the opponent team's placement, we can probably just run the regular `Stopped` play.  Is there anything special we need to do?

Btw, I merged master into this branch so I could see the diff without the stopped play fixes from @jcarn, so you'll need to do a `git pull` before making changes.

soccer/gameplay/plays/restarts/placement.py
justbuchanan: Ahh ok, makes sense.  Maybe we should have this play subclass the normal `Play` class then.  The `StandardPlay` exists mostly to handle defense.

soccer/gameplay/robocup-py.cpp
ashaw596: I mean with the end velocity. The existing OurRobot_move_to_direct ignored the obstacles.

soccer/gameplay/skills/move_direct.py
```diff
@@ -1,8 +1,9 @@
 import single_robot_behavior
 import behavior
 
-
 ## Behavior that moves a robot to a specified location, circumventing the path planner.
+
```
ashaw596: A bit of a random change. Its probably against checkstyle?

soccer/gameplay/robocup-py.cpp
justbuchanan: Yeah the moveDirect is sort of a hack.  The reason it normally avoids the ball is because we add a large circular obstacle around it whenever we're in the STOPPED state.  See this method: https://github.com/RoboJackets/robocup-software/blob/master/soccer/Robot.cpp#L407

justbuchanan: It may be best if we make "ball placement" its own game state alongside stopped and restart, since it's not really either one of those.  Then we can just not put an obstacle around the ball when it's our placement and you can use the normal move() method.

chachmu:
I was using the aim method since it already handles getting the ball between the robot and the target. If I try to move around the ball and then use face sometimes it bumps into the ball and it rolls away. Is there a better skill for maneuvering around the ball than aim?

jgkamat:
When the opponent team does a ball placment, we probably want to run line up instead, to avoid interfering with them (the stopped play would track the ball as it moved).

soccer/gameplay/skills/dribble.py
```diff
@@ -0,0 +1,125 @@
...
+            lambda: self.subbehavior_with_name('aim').state == skills.aim.Aim.State.aimed,
+            'done aiming')
+
+        self.add_transition(
+            Dribble.State.aim, Dribble.State.capture,
+            lambda: self.subbehavior_with_name("aim").state == behavior.Behavior.State.failed,
+            'fumbled')
+
+        self.add_transition(Dribble.State.drive, Dribble.State.capture,
+                            lambda: self.fumbled() or self.robot.vel.mag() > 1,
```
jgkamat: I'm not sure exactly why this transition happens if the robot's velocity is >1? If you want to keep the robot moving slowly, you might want to add a max speed  parameter to dribble instead?

soccer/gameplay/skills/dribble.py
```diff
@@ -0,0 +1,125 @@
...
+        self.add_transition(
+            Dribble.State.drive, behavior.Behavior.State.completed,
+            lambda: (main.ball().pos - self.pos).mag() < self.threshold,
+            'finished driving')
+
+        self.last_ball_time = 0
+
+    def fumbled(self):
+        return not self.robot.has_ball() and time.time(
+        ) - self.last_ball_time > 0.1
```
jgkamat: If the style checker lets you, can you try to keep the method parenthasis on the same line :cry:

jgkamat:
I would guess that the face command would bump it away because it was turning too fast(?) Maybe you could generate a point in between the target point and the actual ball point to smooth out the movement. (but weighted towards the current ball point, so we keep the ball).

justbuchanan:
@jgkamat good point, line up would probably be best when the opponent team is placing the ball.  We just need to make sure the line is far away from the point the ball is being placed.

There's a method to set the max angle speed of the robot.  You can set it to something very low, then tell the robot to face and it should handle it fine without fumbling the ball.

https://github.com/RoboJackets/robocup-software/blob/master/soccer/gameplay/robocup-py.cpp#L616

soccer/gameplay/skills/dribble.py
chachmu: I wasn't sure how to do that so I was just trying a few different things.

chachmu:
Ok, how do I set the maximum velocity and angular velocity?

justbuchanan:
@chachmu I just realized that there wasn't a method for setting max linear velocity, so I added it in PR #666.

chachmu:
Ok great, I have the hard part of the placement working pretty much perfectly now, but the robot decelerates too quickly when it gets to the target so the ball rolls away. Is there a way to add a max acceleration in the motion constraints like with the velocity?

chachmu:
I fixed that previous issue by having the robot actually make sure the ball had stopped moving before letting go, I no longer need a way to control acceleration.

chachmu:
Ok so I have everything pretty much working but it won't let me push. It fails with the error 
`! [rejected]        ballPlacementCommand -> ballPlacementCommand (non-fast-forward)`

justbuchanan:
Awesome!  One thing to keep in mind is that things that work in the simulator may not work in real life just because our physics aren't quite 100% accurate.  Not much we can do about it now though, we'll just have to test this later on when we have robots and maybe make some adjustments.

Hmm, when you did a git pull, did you make sure to specify the remote branch?  Try again with `git pull robojackets ballPlacementCommand`.

chachmu:
I switched it back to the original play.Play type but now for whatever reason I can't seem to get rid of the goalie behavior. I printed out a list of everything that is running and it says that there is no behavior for the goalie so I'm not really sure how it is still running.

chachmu:
Ok, these are pretty much my last two major questions.

1. I checked the refbox documentation and it says that placement commands are given using the ssl-vision coordinate system. How do I convert that to team coordinates? I'm guessing it's _worldToTeam but I'm not sure.

2. When checking if the ball is in the required radius should I make sure that the radius of the ball doesn't go outside of it or should I just stick with the center coordinates of the ball that we get from ssl-vision?

jgkamat:
Don't worry about the goalie play, thats added outside of your play (because the goalie, if assigned, is the only robot that can move into the goal area, and that is strictly enforced).  

Its probably safer to make sure that the radius of the ball is completely inside, but the ball is probably small enough that it dosent matter in real life. Just make sure if the requested radius is the size of the ball/smaller the play is still able to complete.

chachmu:
I need to make sure there isn't a goalie because it comes out and kicks the ball  away if the placement point is too close. 

The radius of the placement is set, it has to be 10 cm from the point specified by the command. I guess I could just make it take into account the radius anyway just to be safe.

justbuchanan:
To get rid of the goalie, implement the `handles_goalie()` class method in your play and make it `return True`.

soccer/gameplay/plays/restarts/placement.py
```diff
@@ -0,0 +1,72 @@
...
+
+    def execute_running(self):
+        #if the ball placement command changes, reset the behaviors accordingly
+        if self.check_update():
+
+            #print(self._our_restart!=main.game_state().is_our_restart())
+            self._pos = main.game_state().get_ball_placement_point()
+            self._our_restart = main.game_state().is_our_restart()
+
+            self.remove_all_subbehaviors()
```
justbuchanan: Instead of removing, then re-adding subbheaviors, you can implement the on_enter and on_exit methods.  For example:

~~~{.py}
def on_enter_running():
    self.add_subbehavior(placer.Placer(), 'placer')

def on_exit_running():
    self.remove_subbehavior('placer')
~~~

soccer/NewRefereeModule.cpp
```diff
@@ -328,6 +334,18 @@ void NewRefereeModule::updateGameState(bool blueTeam) {
             break;
         case Command::GOAL_BLUE:
             break;
+        case Command::BALL_PLACEMENT_YELLOW:
```
justbuchanan: Should we add `_state.gameState.state = GameState::Stop;` for this and the blue placement to make sure we're in the correct state?

soccer/gameplay/skills/dribble.py
```diff
@@ -0,0 +1,126 @@
...
+    @property
+    def vel(self):
+        return self._vel
+
+    @pos.setter
+    def pos(self, value):
+        self._pos = value
+
+    @vel.setter
+    def vel(self, value):
```
justbuchanan: What does the `vel` property do?

soccer/gameplay/skills/dribble.py
```diff
@@ -0,0 +1,126 @@
...
+        capture = skills.capture.Capture()
+        self.add_subbehavior(capture, 'capture', required=True, priority=100)
+
+    def on_exit_capture(self):
+        self.remove_subbehavior('capture')
+
+    def execute_aim(self):
+        self.robot.set_max_angle_speed(3)
+        self.robot.pivot(self._pos)
+        self.robot.set_dribble_speed(100)
```
justbuchanan: Let's make `DribbleSpeed` a class variable at the top, so if/when we need to tune it, we can change it in just one place.

justbuchanan:
I ran the diagram generator just for fun and here's what it shows for the Dribble skill, looks pretty good!

dribble

soccer/gameplay/skills/dribble.py
chachmu: Jay wanted to make this skill to be usable outside of just a placement so I had set it up to allow you to define an end velocity but when  I removed move_direct_end_vel it became useless so I will go ahead and take it out.

soccer/NewRefereeModule.cpp
chachmu: The refbox cant send that command without already being in the stop state but if it were to be set to ignore the rules this could break it so I will add that.

chachmu:
Ok, everything works except if the ball is in the goal area, the robot won't go capture it because of the global obstacle. It's fine for placing it there because I used move_direct but the capture method uses the normal move so I need to find a way to make it ignore that global obstacle.

justbuchanan:
This line in Processor.cpp is where we set the obstacles that each robot respects: https://github.com/RoboJackets/robocup-software/blob/master/soccer/Processor.cpp#L436.  As of now, it only allows the robot into the goal zone if it's the goalie or if it's the penalty kicker, but we can add a new check for the placement robot.  This would require adding a `bool isBallPlacer` to the `OurRobot` class and setting that flag within the relevant play/behavior.  The only difference is that the check in Processor.cpp only changes whether or not it's allowed into *our* goal zone, but for the placement robot we'll want it to be able to go into both our goal zone and the opponent's.

chachmu:
Everything works now although I did notice that the current code doesn't prohibit the goalie from entering the opponents goalzone. It would be nice to have a chance to test this with the actual robots just to have a chance to tweak the maximum speeds and stuff since it's pretty different in the simulator.

justbuchanan:
Awesome, nice work!  It will need a good bit of tuning for use with actual robots, but that'll have to wait until later.  I'll go ahead and merge this now and open an issue for tuning this later.  It could also use a little more tuning for the simulator since it tends to require a couple tries to get it into the right spot.

Btw, I wrote a script to send referee placement commands that you can check out here: https://gist.github.com/justbuchanan/8bf62d9a05c98a2ef6666474acfe7d37.  Using it is much easier/faster than loading the referee and using their custom remote control client setup.

PR #655, opened 2016-04-18
Updated initial value of fieldEdgeInset to reflect 2016 rules

### LINKED ISSUE #638: Double-check field sizing against new SSL rules
The new rules add a wall between where the refs walk and where the robots are.  We should double-check and make sure that the field border size we have set in `Field_Dimensions` matches the new setup.

jcarn:
In past years the inner 250mm has been free space for the robots to move, with the outer 450mm being reserved for referee walking. This year, they are placing a small wall around the field, and have changed the border so that we now have the inner 300mm to work with. I have changed the value of the configurable fieldEdgeInset variable (which controls how far the robots venture from the painted border of the field) to reflect this change. Closes #638

### DIFF
--- a/soccer/gameplay/GameplayModule.cpp
+++ b/soccer/gameplay/GameplayModule.cpp
@@ -31,7 +31,7 @@ void GameplayModule::createConfiguration(Configuration* cfg) {
 +     * The value is given in meters. As of April 20, 2016 the inner 300mm
 +     * is free space for the robots.
 +     */
-+     fieldEdgeInset = new ConfigDouble(cfg, "Pathplanner/Field Edge Obstacle", .33);
++     fieldEdgeInset = new ConfigDouble(cfg, "PathPlanner/Field Edge Obstacle", .33);
 }
 
 bool GameplayModule::hasFieldEdgeInsetChanged() const {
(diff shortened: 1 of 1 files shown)


soccer/gameplay/GameplayModule.cpp
```diff
@@ -26,10 +26,12 @@ using namespace Geometry2d;
...
-    // this sets the disance from the field boundries to the edge of the global
-    // obstacles, which the
-    // robots will not move through or into
-    _fieldEdgeInset = new ConfigDouble(cfg, "Field Edge Obstacle", .3);
+    /*this sets the disance from the field boundries to the edge of the global
+     * obstacles, which the robots will not move through or into.
+     * The value is given in meters. As of April 20, 2016 the inner 300mm
+     * is free space for the robots.
+     */
+     fieldEdgeInset = new ConfigDouble(cfg, "Field Edge Obstacle", .33);
```
jcarn: GIven that we have an extra 50mm, I thought I would increase the initial value by 50mm. However, the consequences of violating this boundry are now much more severe (high speed coillision) so I wanted to add a bit of safety, I only increased it 30mm. However, it seems to me as if this is still 30mm beyond the actual wall. The last value, .3, was 50mm beyond where the rules state we are allowed to go, but it didn't have a wall to run into. Essentially, should I changed the inset to be more than the value in the rules, exactly equal to it, or less than it to provide some space?

jcarn:
Ignore the changes to the files that aren't gameplay.cpp, those got accidentally added because I am still not the best at git, but I will fix that before it is ready to merge.

justbuchanan:
Looks good to me, other than some git issues :).  Let me know if you need any help figuring this out.

Also, not entirely related, but go ahead and put this config value under the "PathPlanner" category.  You can do that by changing "Field Edge Obstacle" to "PathPlanner/Field Edge Obstacle".

PR #656, opened 2016-04-19
Fixes stop play bug

### LINKED ISSUE #652: Stop play no longer works
This is pretty important, we need the stop play to complete. Press the stop button in the top right, and assign a goalie.

Introduced in a10654f07fdf73f2951ae9a0081b865a23102880

```
ERROR:root:Play 'Stopped' encountered an exception: <class 'AttributeError'>.  Dropping and temp. blacklisting current play...
  File "/home/jay/Code/robocup-software/soccer/gameplay/composite_behavior.py", line 81, in spin
    bhvr.spin()
  File "/home/jay/Code/robocup-software/soccer/gameplay/composite_behavior.py", line 63, in spin
    super().spin()
  File "/home/jay/Code/robocup-software/soccer/gameplay/fsm.py", line 81, in spin
    StateMachine.spin(self)
  File "/home/jay/Code/robocup-software/soccer/gameplay/fsm.py", line 76, in spin
    self.transition(next_states[0])
  File "/home/jay/Code/robocup-software/soccer/gameplay/fsm.py", line 121, in transition
    state_method()
  File "/home/jay/Code/robocup-software/soccer/gameplay/plays/stopped.py", line 56, in on_enter_normal
    idle = tactics.stopped.circle_near_ball.CircleNearBall()
  File "/home/jay/Code/robocup-software/soccer/gameplay/tactics/stopped/circle_near_ball.py", line 32, in __init__
    for pt in self.get_circle_points(6):
  File "/home/jay/Code/robocup-software/soccer/gameplay/tactics/stopped/circle_near_ball.py", line 47, in get_circle_points
    for i in constants.Field.FieldBorders:
AttributeError: 'Field_Dimensions' object has no attribute 'FieldBorders'
```

Seems like something introduced in the move from constants.py to the Field_Dimensions object.

I don't think it would be too hard to fix, but I don't have time to look into it right now.

Whoever fixes this should look for instances of `constants.Field` in the python.

I'll look into this next week if no one else has by then.

jcarn:
Added the missing FieldBorders array of lines that form the border to the FieldDimensions class and made the C++ to python converter reflect that change. Closes #652


jcarn:
I've added the array of lines but before it will work I need to do some things in the converter. I have some questions that I'll ask tomorrow about it, but it should be a pretty simply process.

justbuchanan:
Looks good :+1:

common/Field_Dimensions.hpp
```diff
@@ -56,6 +57,9 @@ struct Field_Dimensions {
     inline Geometry2d::Rect OurHalf() const { return _OurHalf; }
     inline Geometry2d::Rect TheirHalf() const { return _TheirHalf; }
 
+    //We play on a 2D rectangular field, hence the length "4" of the array
+    inline std::array<Geometry2d::Line, 4> FieldBorders() const { return _FieldBorders; }
```
justbuchanan: If you change this to return a reference, it will be slightly faster because it won't have to copy the field border array.  This would look like:

~~~{.cpp}
inline std::array<Geometry2d::Line, 4>& FieldBorders() const { return _FieldBorders; }
~~~

justbuchanan: Also, it's not a problem as-is, but the `inline` keyword isn't really needed for any of these - see this [stack overflow answer](http://stackoverflow.com/questions/1759300/when-should-i-write-the-keyword-inline-for-a-function-method).  The compiler will automatically inline all of these methods since they're very small, regardless of whether or not we use the `inline` keyword.

jcarn: I'll go ahead and remove the inline keyword then. Should I add the & to the other fields in Field_Dimensions as well (i.e. _ourHalf) or are they good as is?

justbuchanan: Yeah sounds good, go ahead and remove the inline keyword.  I think this one is the only one that definitely benefits from the ampersand.  The others might, but it's hard to tell without actually benchmarking it.  And in this case, that's way more hassle than it's worth - performance isn't that important to us here :)

justbuchanan: For reference, I did a benchmark of by-reference vs by-value for function parameters a while ago.  It's not exactly the same as this case, but pretty similar.

https://gist.github.com/justbuchanan/a1594aead5fda4e6e526

jcarn:
I got it working, but it isn't very pretty or refined right now. If anyone really needs it right now I can push a rough version, but otherwise I will make it slightly less painful to read, etc.

PR #659, opened 2016-04-20
Making 'ps' command more useful

jjones646:
Added percentages and mapped the states to an acronym vs a number


justbuchanan:
I'm getting an exception when I run this.  Does it work on yours?

jjones646:
Hmm. I don't have anything to test it on at the moment, but I don't believe I ran into any issues. I added the std::map at the last minute, so that may be causing issues somewhere.

jjones646:
Just updated this back to master. It's working with a bare mbed for me. Still need to test it with the boards.

```
jjones646@robot $ ps
                                   |        STACK SIZE (bytes)       |
ID      PRIOR   STATE   Δ TIME     MAX      ALLOC    CURRENT (NOW|MAX)
1         4      WDL    0          21952    21952    1192    (5%|100%)
2         6      WMX    1          2044     2048     120     (6%|100%)
3         6      WDL    130        804      1024     248     (24%|79%)
4         5      RUN    0          1164     2048     896     (44%|57%)
5         5      WMX    0          716      1024     192     (19%|70%)
6         6      WMX    0          716      1024     192     (19%|70%)
7         4      WSM    0          184      1024     184     (18%|18%)
==============
Total Threads:  7
jjones646@robot $
```

PR #660, opened 2016-04-20
Add handling of IO expander interrupt inputs

### LINKED ISSUE #400: Enable support for interrupts on the digital IO expander

jjones646:
resolves #400

### DIFF
--- a/firmware/robot2015/src-ctrl/main.cpp
+++ b/firmware/robot2015/src-ctrl/main.cpp
@@ -21,8 +21,8 @@
 #include "SharedSPI.hpp"
 #include "KickerBoard.hpp"
 #include "RtosTimerHelper.hpp"
-#include "io-expander.hpp"
 #include "RotarySelector.hpp"
+#include "robot-expander-inputs.hpp"
 
 using namespace std;
 
@@ -139,9 +139,11 @@ int main() {
     // sets the first 8 lines to input and the last 8 to output.  The pullup
     // resistors and polarity swap are enabled for the 4 rotary selector lines.
     MCP23017 ioExpander(RJ_I2C_SDA, RJ_I2C_SCL, RJ_IO_EXPANDER_I2C_ADDRESS);
-    ioExpander.config(0x00FF, 0x00f0, 0x00f0);
+    ioExpander.config(0x00FF, 0x00F0, 0x00F0);
     ioExpander.writeMask((uint16_t)~IOExpanderErrorLEDMask,
                          IOExpanderErrorLEDMask);
+
+    RobotExpanderInputs ioInputs(&ioExpander, RJ_IOEXP_INT);
 
     // rotary selector for shell id
     RotarySelector<IOExpanderDigitalInOut> rotarySelector(
--- a/firmware/robot2015/src-ctrl/modules/robot-expander-inputs.hpp
+++ b/firmware/robot2015/src-ctrl/modules/robot-expander-inputs.hpp
@@ -1,0 +1,105 @@
+#pragma once
+
+#include <mbed.h>
+#include <rtos.h>
+
+#include <mcp23017.hpp>
+#include <io-expander.hpp>
+#include <helper-funcs.hpp>
+#include <logger.hpp>
+#include <Console.hpp>
+
+#include "pins-ctrl-2015.hpp"
+#include "task-signals.hpp"
+
+/**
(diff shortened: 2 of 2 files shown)


firmware/robot2015/src-ctrl/modules/robot-expander-inputs.hpp
```diff
@@ -0,0 +1,105 @@
...
+    RobotExpanderInputs(MCP23017* mcp, PinName interruptPin)
+        : interruptThread_(&RobotExpanderInputs::ThreadHelper, this,
+                           osPriorityBelowNormal, 0.33 * DEFAULT_STACK_SIZE),
+          mcp23017_(mcp),
+          interrupt_(interruptPin) {
+        interrupt_.fall(this, &RobotExpanderInputs::ISR);
+        interrupt_.mode(PullUp);
+    }
+
+    uint8_t RotarySelector() {
```
justbuchanan: Keep in mind we also have the RotarySelector class that does the same thing as this method.  The only benefit over the other implementation is that it works with io expander pins as well as regular mbed pins.

firmware/robot2015/src-ctrl/main.cpp
```diff
@@ -139,10 +139,12 @@ int main() {
...
     MCP23017 ioExpander(RJ_I2C_SDA, RJ_I2C_SCL, RJ_IO_EXPANDER_I2C_ADDRESS);
-    ioExpander.config(0x00FF, 0x00f0, 0x00f0);
+    ioExpander.config(0x00FF, 0x00F0, 0x00F0);
     ioExpander.writeMask((uint16_t)~IOExpanderErrorLEDMask,
                          IOExpanderErrorLEDMask);
 
+    RobotExpanderInputs ioInputs(&ioExpander, RJ_IOEXP_INT);
+
     // rotary selector for shell id
     RotarySelector<IOExpanderDigitalInOut> rotarySelector(
```
justbuchanan: If we're going with a new inputs class, we can remove the rotary selector here.

firmware/robot2015/src-ctrl/modules/robot-expander-inputs.hpp
```diff
@@ -0,0 +1,105 @@
...
+        const std::array<MCP23017::ExpPinName, 4> kPins = {
+            RJ_HEX_SWITCH_BIT0, RJ_HEX_SWITCH_BIT1, RJ_HEX_SWITCH_BIT2,
+            RJ_HEX_SWITCH_BIT3};
+
+        for (size_t i = 0; i < kPins.size(); i++)
+            reading |= mcp23017_->readPin(kPins[i]) << i;
+        return reading;
+    }
+
+    uint8_t DipSwitches() {
```
justbuchanan: If you want to, you can factor out most of this method and the RotarySelector method into a "read pin array" or something so that you don't have to write the same code twice.

firmware/robot2015/src-ctrl/modules/robot-expander-inputs.hpp
```diff
@@ -0,0 +1,105 @@
...
+        uint8_t dip_switches = DipSwitches();
+        bool push_btn = PushButton();
+
+        // TODO: get these values to where they need to go
+    }
+
+    void Task_InputHandler() {
+        while (true) {
+            // refresh system variables
+            RefreshEnvironment();
```
justbuchanan: With this setup, our code that responds to a button press has to poll whatever shared variable this info gets written to, right?  If it doesn't poll that value fast enough, it could easily miss it before it turns off again.

We may want to enable code that needs interrupts to register a callback for when something happens.

justbuchanan:
I didn't get around to testing it because the IO expander was broken, but I put this together a while ago too: https://github.com/RoboJackets/robocup-software/pull/597.

firmware/robot2015/src-ctrl/modules/robot-expander-inputs.hpp
jjones646: yeah, the push button would need some kind of callback handler.

jjones646:
The main difference with this and #597 is that this runs on it's own thread. Which is both good and bad. Good in that it hands it off nicely into the rtos world; Bad in that it's another thread.. I imagine the end code will end up being bits and pieces from both.

PR #663, opened 2016-04-25
Albert/dynamic hit

### LINKED ISSUE #117: Add check to see if the paths intercept in the future to the Run Function
Maybe even figure out a way to modulate speed to dodge?

ashaw596:
Check if Paths hit in the future! ~~Not~~ Mostly cleaned up. The planned paths look good I think.
Closes #117


soccer/planning/SingleRobotPathPlanner.hpp
```diff
@@ -20,7 +21,9 @@ class SingleRobotPathPlanner {
     virtual std::unique_ptr<Path> run(
         MotionInstant startInstant, const MotionCommand* cmd,
         const MotionConstraints& motionConstraints,
-        const Geometry2d::ShapeSet* obstacles,
+        Geometry2d::ShapeSet& obstacles,
```
justbuchanan: Should `obstacles` be `const`?

soccer/planning/Path.cpp
```diff
@@ -32,4 +37,42 @@ void Path::draw(SystemState* const state, const QColor& color,
...
+                                                  float deltaT) const {
+    return std::move(
+        std::make_unique<ConstPathIterator>(this, startTime, deltaT));
+}
+
+bool Path::pathsIntersect(const std::vector<const Path*>& paths, float* hitTime,
+                          Geometry2d::Point* hitLocation,
+                          RJ::Time startTime) const {
+    const float deltaT = 0.05;
+    const float hitRadius = Robot_Radius * 2.5f;
```
justbuchanan: Should we use the "robot avoid radius" from the robot mask stored in `OurRobot`?

soccer/planning/SingleRobotPathPlanner.hpp
ashaw596: Well, I changed it to not const cause I add to the static obstacles. We create unique obstacle sets for each robot anyways.

soccer/planning/Path.cpp
ashaw596: I don't know, it seems a bit clunky. What do we need individual robot avoid radius's for?

justbuchanan:
Overall looks pretty good!  A couple questions/suggestions:

* Right now it looks like the robots are planned in order of shell id.  It would be best if we could always plan the ball carrier first so that it has highest priority and others move out of its way.
* Is there a way we can incorporate dynamic obstacles into the actual RRT?  The current method of planning without them, then trying again (potentially several times) if it intersects a path seems like it could be improved.  Our RRT doesn't incorporate time, so it would have to be based on a heuristic - maybe a rough conversion from distance to start time.  This may end up being a lot of work, but maybe there's a simpler way to do this.

ashaw596:
Yeah, I think I can do a better priority system pretty easily.

Yeahhh its not a very efficient system at all right now. But, incorporating the dynamic obstacles would take a good bit of effort at the moment and it doesn't seem to be necessary at the moment since we don't really have performance issues. Also, did you merge your RRT changes yet? The main efficiency improvement want is to implement a good iterator for the paths that don't have to iterate through time. But it doesn't seem to be an important consideration at the moment. I think I could just make issues for those for now until we have more time or we have issues with performance.

soccer/planning/Path.cpp
justbuchanan: Well we use it [here](https://github.com/RoboJackets/robocup-software/blob/master/soccer/gameplay/plays/offense/basic_122.py#L167) and [here](https://github.com/RoboJackets/robocup-software/blob/master/soccer/gameplay/plays/restarts/their_restart.py#L91) currently.

It may be ok to not have per-robot avod radii, but if we decide to stop using the self avoid mask for anything, we should get rid of it and the related methods.

justbuchanan:
Sounds good.

Ok, I think you're right that it would be a lot of work to incorporate it into the rrt, it's just kinda hacky to have it plan without it, then retry several times.  I can't think of a better approach at the moment though.  Go ahead and open an issue for incorporating it into the rrt and we can come back to it later.  I haven't merged my rrt changes just yet... I need to merge #595 first, but I'm having trouble tracking down the issues in that.  I've also been distracted by trying to get radio firmware up and running :/.

soccer/planning/IndependentMultiRobotPathPlanner.cpp
```diff
@@ -33,6 +35,14 @@ std::map<int, std::unique_ptr<Path>> IndependentMultiRobotPathPlanner::run(
         staticRobotObstacles[shell] = std::make_shared<Geometry2d::Circle>(
             request.start.pos, Robot_Radius);
     }
+
+    auto comparator = [&](int &lhs, int &rhs) {
+            return sortByPriority(requests[lhs], requests[rhs]);
```
justbuchanan: It would be a little shorter if you get rid of `sortByPriority()` and just do `return requests[lhs].priority > requests[rhs].priority` here.

soccer/planning/IndependentMultiRobotPathPlanner.cpp
```diff
@@ -33,6 +35,14 @@ std::map<int, std::unique_ptr<Path>> IndependentMultiRobotPathPlanner::run(
...
+        };
+
+    std::sort(std::begin(staticRequests), std::end(staticRequests), comparator);
+    std::sort(std::begin(dynamicRequests), std::end(dynamicRequests), comparator);
+
     std::vector<int> inOrderRequests;
```
justbuchanan: Does it make sense to always plan the static requests first?  Won't this allow a low-priority robot with a static request to be planned before a high-priority robot with a dynamic request?

justbuchanan:
I think it might make sense to add a file that defines constants for planner priorities.  Since planning priorities are relative, the value that is set for a robot's priority only has meaning relative to what value is set for the other robots.  It would make it a lot easier to tell what the values actually mean if we define them all  in one file rather than having to search around different behavior files to see what they do.

It could look something like:

~~~{.py}
GOALIE_PLANNING_PRIORITY = 20
KICKER_PLANNING_PRIORITY = 15
...
~~~

And then we could use those constants within the behaviors.

soccer/planning/IndependentMultiRobotPathPlanner.cpp
ashaw596: Well, the problem is that the things planned first don't consider the other robots as obstacles since we expect the later robots to plan around us. Its a edge case, but if we planned the dynamics first, the dynamics would run into the static robots. The dynamics would ignore the static since its planned later and the statics can't check where robots will be in the future.

soccer/planning/IndependentMultiRobotPathPlanner.cpp
ashaw596: done

ashaw596:
Ok, I think thats about everything.

PR #666, opened 2016-04-28
added set_max_speed() for OurRobot in python

justbuchanan:
I realized we have a similar method for max angular speed, but not for linear speed.


jgkamat:
Just a future idea, maybe we could add a 'resetMaxSpeed/angularSpeed' function so the called play dosen't have to remember what that value was before it gets changed.

justbuchanan:
We've actually got something like that setup :).  This method gets called every iteration and resets most everything that plays touch: https://github.com/RoboJackets/robocup-software/blob/master/soccer/Robot.cpp#L126

ashaw596:
looks good

PR #667, opened 2016-04-28
Added MotionConstraints class to robocup-py

jcarn:
I only added the config values from the class, since that is what I need. Adding other values should be pretty straightforward if anyone needs them


justbuchanan:
@jcarn I realized we were missing a couple more python wrappers in order for you to actually use this from python.  I added them in 3e3c82c52374a48b7506c225976c418c54ed33c3, so just be sure to merge master into your work and things should work.

PR #673, opened 2016-05-19
Fix bug and switch to simplest acceleration algorithm fix #669

### LINKED ISSUE #669: Path-planning occasionally gets "stuck"
I'm not sure exactly what the cause is, but I'm seeing some path-planning issues with the defenders near the goal zone.  Below is a screenshot showing the issue.  Robot 3 is supposed to be moving into position to block a portion of the red ball-projection triangle, but it isn't moving at all and has been there for several seconds.  I'm guessing it has something to do with the goal zone obstacle.  You can see in the "Tree" panel that robot 3's command velocity is essentially zero.

2016-05-17-211108_1332x1034_scrot

This is present on the master branch (I ran this on f4b82bebe5).

cc @ashaw596

ashaw596:


justbuchanan:
Nice, works perfectly!

Issue #408, opened 2015-08-26
Fix and test onetouchpass aiming/scoring


jgkamat:
Right now onetouchpass was not able to work on real robots. This is in part due to coordinated pass not working, but the aiming in onetouchpass needs to be reevaluated if it continues not to work on the new robots.

jgkamat:
This is close enough for now. I'll take another look at it if we have time.

PR #675, opened 2016-05-22
Start to fix touchpass

jgkamat:
Touchpass was broken from some point during last year in china, but I think I've fixed it (at least in the simulator for now). The cause of this was using capture in the receiving play, which I made some modificiations to fix. This is still a work in progress, but it's mostly done.

The main improvement I was thinking about was hijacking the 'fine-approach' state in capture to lunge towards the ball, as well as improve detecting when a touchpass fails earlier if possible. I've already modified course approach to suggest points to intercept on that allow for a touchpass.

Right now, only style is complaning, but I'll fix that once we have everything else wrapped up.

See #408

### DIFF
--- a/soccer/gameplay/skills/angle_receive.py
+++ b/soccer/gameplay/skills/angle_receive.py
@@ -6,7 +6,7 @@ import main
 import math
 import enum
 import time
-import skills.touchball
+import skills.touch_ball
 import skills._kick
 import skills.pass_receive
 
@@ -143,7 +143,7 @@ class AngleReceive(skills.pass_receive.PassReceive):
 
         self.remove_subbehavior('capture')
 
-        capture = skills.touchball.TouchBall()
+        capture = skills.touch_ball.TouchBall()
         capture.dribbler_power = skills.pass_receive.PassReceive.DribbleSpeed
         self.add_subbehavior(capture, 'capture', required=True)
 
--- a/soccer/gameplay/skills/angle_receive.py
+++ b/soccer/gameplay/skills/angle_receive.py
@@ -19,7 +19,7 @@ import skills.pass_receive
 # Kick is a single_robot_behavior, so no need to import both
 class AngleReceive(skills.pass_receive.PassReceive):
     def __init__(self):
-        super().__init__()
+        super().__init__(captureFunction=(lambda: skills.touch_ball.TouchBall()))
         self._target_point = None
         self.kick_power = 1
         self.target_point = constants.Field.TheirGoalSegment.center()
@@ -135,17 +135,6 @@ class AngleReceive(skills.pass_receive.PassReceive):
         if self._kick_line != None:
             main.system_state().draw_line(self._kick_line,
                                           constants.Colors.Red, "Shot")
-
-    # This depends on the superclasse's implementation somewhat
-    # If we stop using capture in pass_receive, this needs to be modified.
-    def on_enter_receiving(self):
-        super().on_enter_receiving()
-
-        self.remove_subbehavior('capture')
-
-        capture = skills.touch_ball.TouchBall()
-        capture.dribbler_power = skills.pass_receive.PassReceive.DribbleSpeed
-        self.add_subbehavior(capture, 'capture', required=True)
 
     def execute_receiving(self):
         super().execute_receiving()
--- a/soccer/gameplay/skills/angle_receive.py
+++ b/soccer/gameplay/skills/angle_receive.py
@@ -6,6 +6,7 @@ import main
 import math
 import enum
 import time
+import skills.touchball
 import skills._kick
 import skills.pass_receive
 
@@ -142,7 +143,7 @@ class AngleReceive(skills.pass_receive.PassReceive):
 
         self.remove_subbehavior('capture')
 
-        capture = skills.capture.Capture(onlyApproach=True)
+        capture = skills.touchball.TouchBall()
         capture.dribbler_power = skills.pass_receive.PassReceive.DribbleSpeed
         self.add_subbehavior(capture, 'capture', required=True)
 
--- a/soccer/gameplay/skills/angle_receive.py
+++ b/soccer/gameplay/skills/angle_receive.py
@@ -135,6 +135,17 @@ class AngleReceive(skills.pass_receive.PassReceive):
             main.system_state().draw_line(self._kick_line,
                                           constants.Colors.Red, "Shot")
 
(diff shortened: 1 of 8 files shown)


soccer/gameplay/skills/touchball.py
```diff
@@ -0,0 +1,49 @@
+import single_robot_behavior
```
justbuchanan: This file should be named touch_ball.py to keep with the camel-case/underscore naming scheme

soccer/gameplay/skills/touchball.py
```diff
@@ -0,0 +1,49 @@
...
+import skills.capture
+import math
+
+
+class TouchBall(skills.capture.Capture):
+    def __init__(self):
+        super().__init__(onlyApproach=True)
+
+    # Move back so we hit the mouth, not the side
+    adjDist = constants.Robot.Radius * 2
```
justbuchanan: Constants typically start with a capital letter

soccer/gameplay/skills/touchball.py
```diff
@@ -0,0 +1,49 @@
...
+
+    ## Override so this so we only transition when the ball is in front
```
justbuchanan: grammar/wording

soccer/gameplay/skills/capture.py
```diff
@@ -23,7 +23,10 @@ class State(Enum):
         course_approach = 1
         fine_approach = 2
 
-    def __init__(self):
+    ## Capture Constructor
+    # onlyApproach - If true, any turning functions are turned off,
+    # useful for using capture to reflect/bounce moving ballls.
+    def __init__(self, onlyApproach=False):
```
justbuchanan: Maybe rename `onlyApproach` to `faceBall` or something similar for clarity.  The name right now doesn't really convey that it affects facing and nothing else.

soccer/gameplay/skills/touchball.py
```diff
@@ -0,0 +1,49 @@
...
+    # distance from the ball if possible, and move forward to hit the ball at
+    # the last minute.
+    # To do this, let's move the intercept point capture found back a bit.
+    #
+    # In addition, lets try to keep this point stable by choosing the closest
+    # point, instead of the point that we can reach in time closest to the ball
+    def find_intercept_point(self):
+        approach_vec = self.approach_vector()
+
+        adjFactor = robocup.Point(
```
justbuchanan: This could also be written as `adjFactor = robocup.Point.direction(self.robot.angle) * -TouchBall.adjDist`

soccer/gameplay/skills/angle_receive.py
```diff
@@ -135,6 +136,17 @@ def execute_running(self):
             main.system_state().draw_line(self._kick_line,
                                           constants.Colors.Red, "Shot")
 
+    # This depends on the superclasse's implementation somewhat
+    # If we stop using capture in pass_receive, this needs to be modified.
+    def on_enter_receiving(self):
```
justbuchanan: It may be a little cleaner if we have PassReceive's constructor take in a capture behavior.  That way we don't have to override this transition and replace the behavior after it's been added.

justbuchanan:
Nice, it looks like this works a good bit better than before!

Looking through this, I wonder if it makes sense to move some of the capture overrides/changes into AngleReceive itself.  This would completely override PassReceive's implementation of the "receiving" state.  It seems like the main objective of the receiver after the ball is kicked is to face the appropriate angle and to stay on line with the ball's predicted trajectory.  This is pretty different from the capturing that PassReceive implements. Thoughts?

jgkamat:
Hmm, I think that would be a good idea, but I still would like to keep the touchball logic in its own play (Maybe skill?) that doesn't extend from capture. There are a lot of underlying similarities between this and capture, but the result is very different...

justbuchanan:
Sounds good, whatever works best.  Which parts of capture do you want to keep for TouchBall?

jgkamat:
Probably very little (I'm planning on having touchball only go to the nearest intersection point and stay there), so I'll probably not need to copy anything from capture. I'll see what happens though.

soccer/gameplay/skills/angle_receive.py
jgkamat: Pass Receive needs a fresh object every time it runs (this would break multiple coordinated passes, as in testCoordinatedPass). I could pass in a lambda that generates a Capture object. That probably would be a lot cleaner.

jgkamat: actually this is a good idea, I'm going to do it.

jgkamat:
I think I'm ready for round 2, I refactored all the touch-capture logic away, and I removed that mod I made to angle_receive that depended on its superclass.

I think some of the logic in touch_ball could be improved, but I think we should wait until we have real robots to test on before we assume anything additional about this behavior.

soccer/gameplay/skills/capture.py
```diff
@@ -101,11 +105,16 @@ def execute_course_approach(self):
         # don't hit the ball on accident
         self.robot.set_avoid_ball_radius(Capture.CourseApproachAvoidBall)
         pos = self.find_intercept_point()
-        self.robot.face(main.ball().pos)
+
+        if (self.faceBall):
```
justbuchanan: The facing logic may be better to put in `execute_running()` since it applies to all substates of `running`.

soccer/gameplay/skills/touch_ball.py
```diff
@@ -0,0 +1,140 @@
...
+                constants.Colors.White, "TouchBall")
+        else:
+            main.system_state().draw_circle(pos, constants.Ball.Radius,
+                                            constants.Colors.White,
+                                            "TouchBall")
+            self.robot.move_to(pos)
+            self.lastApproachTarget = pos
+
+    def on_exit_course_approach(self):
+        self.lastApproachTarget == None
```
justbuchanan: Should this be `=` instead of `==`?

justbuchanan:
Nice, looks good to me :+1:

soccer/gameplay/skills/touch_ball.py
jgkamat: aaah that should be a warning in python

soccer/gameplay/skills/capture.py
jgkamat: yup, good idea

jgkamat:
I think I've fixed the above issues, and I'm good on them. If you're good, you can merge it!

Issue #645, opened 2016-04-14
Radio calibration doesn't work right


justbuchanan:
When packets are sent/received infrequently (I tried ~1/sec), things work great.  If they're sent at 60Hz, they work for several packets (~20), then stop working.  The radio seems to be transitioning through the correct states, etc and no errors are reported, so I think it's a calibration issue.  @jjones646 do you have any suggestions for things to try here?  I've messed around with the `FS_AUTOCAL` register, but haven't had much luck with any of the three settings (page 76 of the [user guide](http://www.ti.com/lit/ug/swru346b/swru346b.pdf)).

justbuchanan:
@jjones646 Here's the branch we were working from earlier tonight: https://github.com/RoboJackets/robocup-software/tree/justin/new-protocol.  The last few commits should be looked at again - probably 1 in every 10 line changes was actually useful.

Issue #644, opened 2016-04-14
Radio commands should fail if the radio failed to initialize


justbuchanan:
right now they silently do nothing, making it tough to track down...

Issue #578, opened 2016-03-18
Implement address filtering in CC1201


justbuchanan:
TODO:

- [ ] Set the cc1201 address register based on the board/base id
- [ ] Ensure that register configs are setup for address transmission/filtering
- [ ] verify address placement in packet binary format

PR #678, opened 2016-05-23
Lots of radio fixes

### LINKED ISSUE #672: CC1201 fails if other radios are transmitting when firmware boots up
Maybe we need to flush buffers and strobe it into RX after everything initializes

cc @jjones646

justbuchanan:
This fixes many issues we've had with our radios and enables robots to show up in soccer.  Fixes #672, #645, #644.

TODO:

- [ ] remove rate-limiting in USBRadio - we're currently only sending every other packet because the system can't seem to handle 60Hz. (#683)
- [x] Fix TODOs in USBRadio
- [x] USBRadio should address the correct robot - it's currently hard-coded to id 2
- [x] remove test scripts: reload.sh, radio-test.sh
- [x] rebase some commits - the git log is disgusting
- [ ] should `CC1201.sendData()` wait for transmission to complete before returning?  This code is currently commented out
- [x] remove unused `dev.xml`, `new.xml` cc1201 configs
- [x] revert hacks in main.cpp
- [ ] don't hardcode robot id 2 in main.cpp
- [x] fix reply delay in RadioProtocol.hpp
- [x] fix reverse size declaration for usb device config
- [x] send packet header to soccer?  right now we just send the packet's payload
- [x] double-check `rtp::Reverse_Size`
- [ ] figure out what happens with 3+ radios
- [x] value for invalid robot uid?  currently using 0, which is an actual valid shell id
- [ ] robot uid vs shell id
- [x] remove USBRadio._sequence

cc @jjones646 @guyfleeman


jjones646:
> should CC1201.sendData() wait for transmission to complete before returning? This code is currently commented out

Does it still make things not work if the waiting is enabled?


The last real hurdle for the radio is figuring out the interrupt pin's configuration. For the 2011 firmware, the interrupt pin from the CC1101 is toggled according to detecting a preamble (`0x06` ), and [that's what it is now](https://github.com/RoboJackets/robocup-software/blob/justin/radio-fixes/firmware/common2015/drivers/cc1201/cfg/cc1101-compatible.xml#L120) for the 2015 stuff.
However, I've never understood why the interrupt line for the 2011 robots doesn't get triggered from the other robot's reverse packets ([here's](https://github.com/RoboJackets/robocup-software/blob/master/firmware/robot2011/trace_captures/trace_capture_04.logicdata) a logic capture I took a while back that confirms this, you need [this](https://www.saleae.com/downloads) to view it).
All of this is the reasoning behind why using `0x07` with address filtering was originally intended - the interrupt line would only trigger on non-corrupt packets that contained useful information a robot.
I don't really have recommendations on what to do other than adding in a 3rd radio and seeing what works.

justbuchanan:
I tried re-enabling the wait in `sendData()` and it seemed to work fine for the most part.  The only issue was that it seemed like it was able to transmit more frequently without the wait.  It currently doesn't send packets at 60Hz, so I have it sending only every other packet.  I think I got it to a higher send rate though without the wait.

I think you're right that `0x07` would be better.  I haven't done anything with address-filtering yet, but that's on the todo list (#578).  Once that's in place, we should be fine to go with the `PKT_CRC_OK` (or the one that's similar, but not quite the same) interrupt trigger.

PR #679, opened 2016-05-23
Added file logging

JNeiger:
TODO:
- [x] Keep file size in check during run
- [x] Remove log file once new code is pushed to mbed
- [ ] Check call times to make sure performance isn't affected

I ended up opening and closing it each `log` call because we are unable to push code onto the mbed while the file is open.

### DIFF
--- a/firmware/common2015/utils/logger/logger.cpp
+++ b/firmware/common2015/utils/logger/logger.cpp
@@ -19,7 +19,6 @@ uint8_t rjLogLevel;
 Mutex log_mutex;
 
 LocalFileSystem local2("local");
-FILE* fp = NULL;
 
 LogHelper::LogHelper(uint8_t logLevel, const char* source, int line,
                      const char* func) {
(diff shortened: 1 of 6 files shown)


firmware/common2015/utils/logger/logger.cpp
```diff
@@ -17,6 +17,10 @@ uint8_t rjLogLevel;
 
 Mutex log_mutex;
 
+LocalFileSystem local2("local");
+FILE* fp = NULL;
```
justbuchanan: Since `fp` is only used inside the `log()` function, it'd be better to declare it in there rather than globally.

justbuchanan:
Looks good.  Another thing that would be nice to add is a #define flag for enabling/disabling logging to files, similar to how we have one for enabling/disabling logging entirely.

JNeiger:
Also, I would almost leave this unmerged until we fix #686. It gets annoying quick with all the constant file io conflicts

justbuchanan:
Cool, looks good to me.  I'll mess around with #686 next week and see if I can fix it so we can get this merged too.

jgkamat:
@JNeiger is this still relevant to robocup-software, or has everything been moved to RoboJackets/robocup-firmware#26?

If it is, flag me on sunday, and I'll help you rebase onto master or deal with these merge conflicts.

JNeiger:
Huh, I thought I had done this before. I guess this is where it went, Haha.

The only thing that the new version does not have is the enable statement. (The other random files were leftover from a `make pretty` IIRC)
I'm going to add that now to https://github.com/RoboJackets/robocup-firmware/pull/26 and this PR will be out of scope.

jgkamat:
Closing this now, go ahead and delete the branch when you're 100% sure youre ok with deleting it.

PR #687, opened 2016-05-28
Add a `make debug` and `make debug-sim` task

### LINKED ISSUE #506: Create a make debug task in the makefile
Not sure how easy this would be, but we need a make debug to compile with debug flags (and maybe run gdb)

jgkamat:
Closes #506

Kind of trivial but yah. Can someone check to see if this works on mac? I'm not 100% sure it does, but it should. 

I got to experience true `/bin/sh` scripting... :cry:

### DIFF
--- a/makefile
+++ b/makefile
@@ -2,6 +2,8 @@ MAKE_FLAGS = --no-print-directory
 TESTS = *
 FIRMWR_TESTS = -i2c -io-expander -fpga -piezo -neopixel -attiny -led
 
+GDB=$(shell if [ "`uname`" = "Linux" ]; then echo gdb; else echo lldb; fi)
+
 # build a specified target with CMake and Ninja
 # usage: $(call cmake_build_target, target, extraCmakeFlags)
 define cmake_build_target
@@ -31,11 +33,18 @@ run-sim2play: all
 	cd run; ./simulator --headless &
 	cd run; ./soccer -sim -y & ./soccer -sim -b
 
+debug: all
+	$(GDB) ./run/soccer
+
+debug-sim: all
+	-pkill -f './simulator --headless'
+	cd run; ./simulator --headless &
+	gdb ./run/soccer
+
 # Run both C++ and python unit tests
 tests: test-cpp test-python
 test-cpp: test-soccer test-firmware
 test-soccer:
-
 	$(call cmake_build_target, test-soccer)
 	run/test-soccer --gtest_filter=$(TESTS)
 test-firmware:
@@ -132,7 +141,7 @@ fpga2015:
 	$(call cmake_build_target_fw, fpga2015)
 fpga2015-prog:
 	$(call cmake_build_target_fw, fpga2015-prog)
-	
+
 # Base station 2015 firmware
 base2015:
 	$(call cmake_build_target_fw, base2015)
--- a/makefile
+++ b/makefile
@@ -2,8 +2,6 @@ MAKE_FLAGS = --no-print-directory
 TESTS = *
 FIRMWR_TESTS = -i2c -io-expander -fpga -piezo -neopixel -attiny -led
 
-GDB=$(shell if [ "`uname`" = "Linux" ]; then echo gdb --args; else echo lldb --; fi)
-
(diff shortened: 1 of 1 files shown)


jgkamat:
I think the style errors that are happening are in root_play for some reason, I think we can ignore those for now.

makefile
```diff
@@ -31,11 +33,18 @@ run-sim2play: all
 	cd run; ./simulator --headless &
 	cd run; ./soccer -sim -y & ./soccer -sim -b
 
+debug: all
+	$(GDB) ./run/soccer
+
+debug-sim: all
+	-pkill -f './simulator --headless'
+	cd run; ./simulator --headless &
+	gdb ./run/soccer
```
justbuchanan: gdb -> $(GDB)

justbuchanan: Also, if you do `gdb --args run/soccer -sim`, you won't have to type `-sim` after gdb starts.

justbuchanan:
I made a couple notes, but other than that it looks good.  I don't have my mac with me to test this at the moment though.  @ashaw596?

makefile
jgkamat: ooh thats pretty neat! I'll go ahead and fix it.

justbuchanan: Sounds good.  Double-check that that works with `lldb` too - I'm not sure offhand if they have the same flags for that.

jgkamat: Turns out lldb is in the repos. Assuming my lldb is the same as mac's:
`lldb -- ./run/soccer -sim` works, --args dosent unfortunately. I'll mess with it and get it to work in every scenario.

jgkamat:
I think I've gotten lldb and gdb both working, someone on mac, let me know what you find out!

justbuchanan:
Hmm, I'm having trouble with this on OS X :/.  Here's what I get when I run `make debug-sim`:

~~~{.sh}
lldb -- ./run/soccer -sim
(lldb) target create "./run/soccer"
error: './run/soccer' doesn't contain any 'host' platform architectures: x86_64, i386
(lldb) Loading config:  "/Users/justbuchanan/src/rj/robocup-software/run/simulator.cfg"
New Ball: 0.000000 0.000000
New 2008 Robot: 0 : 1.000000 0.000000
New 2008 Robot: 1 : 1.000000 1.000000
New 2008 Robot: 2 : 1.000000 -1.000000
New 2008 Robot: 3 : 2.000000 1.000000
New 2008 Robot: 4 : 2.000000 -1.000000
New 2008 Robot: 5 : 2.000000 0.000000
New 2008 Robot: 0 : -1.000000 0.000000
New 2008 Robot: 1 : -1.000000 1.000000
New 2008 Robot: 2 : -1.000000 -1.000000
New 2008 Robot: 3 : -2.000000 1.000000
New 2008 Robot: 4 : -2.000000 -1.000000
New 2008 Robot: 5 : -2.000000 0.000000
libpng warning: iCCP: known incorrect sRGB profile

error: './run/soccer' doesn't contain any 'host' platform architectures: x86_64, i386
(lldb) r
error: invalid target, create a target using the 'target create' command
~~~

I don't think this has anything to do with the changes you've - it may be something to do with our CMake setup for OS X.  I'll mess with this a little tomorrow and see what I can figure out.

justbuchanan:
I checked this out and the problem is that `run/soccer` on OS X is just a bash script that launches `soccer.app`.  I think the best fix may be to call `lldb soccer.app` on OS X instead of just `soccer`.

justbuchanan:
I just pushed a fix for this - let me know if you see any issues.  I also moved the "uname" call to within each make target since the file path has to change too - not sure if there's a cleaner way to do this.

jgkamat:
The conditional was always returning false for me, I think I've gotten it though (or maybe I've locked it to only true -_-). I don't know why, but makefile dosen't like quotes around strings, `$(cmd)` or `$$(cmd)`, but it does like `$(shell cmd)`. Let me know if this works, if it does, :+1:

justbuchanan:
Good catch!  I tried it on my Mac after I changed it, but didn't bother to retry on Linux...  Your fix works though :+1:

PR #688, opened 2016-05-28
Fixed KickerBoard memory checking and programming.

### LINKED ISSUE #584: Fix avr-isp
TODO:

- [ ] Fix `checkMemory()` function - it checks to see if the avr's flash memory contents are equal to a given binary file, but sometimes incorrectly says they're equal.
- [ ] Fix `enableProgramming()` function

See PR #559

jpfeltracco:
The AVR910::program(...) didn't fseek to the beginning of the file before writing. This led to issues when checkMemory was called before programming which advances our seek position to the end of the file.

Also, in KickerBoard::flash(...) we were calling checkMemory with page size first and num pages second. This was consistent with AVR910::program(File* fp, int pageSize, int numPages), but in AVR910::checkMemory the arguments were actually flipped. I changed AVR910::checkMemory to maintain consistency.

This will hopefully resolve #584

### DIFF
--- a/firmware/common2015/drivers/avr-isp/AVR910.cpp
+++ b/firmware/common2015/drivers/avr-isp/AVR910.cpp
@@ -80,6 +80,7 @@ bool AVR910::program(FILE* binary, int pageSize, int numPages) {
     int c = 0;
     int highLow = 0;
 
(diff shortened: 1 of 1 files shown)


justbuchanan:
Nice! :+1: 

closes #584

PR #689, opened 2016-05-29
Know battery levels from a robot in Soccer

jjones646:
The firmware side of everything is in place with this. The conversion constant was computed, then tweaked slightly after testing in hardware. Also estimated a set of 20%, 50%, and 100% levels for the battery profile. It should be accurate enough for what we're using it for - at least for now.

The motors shouldn't be spun if the battery is critically low, but that's somewhat of a separate issue. And it'd be much easier to do that after most of the motor/radio things are merged into master.

cc @justbuchanan

### DIFF
--- a/firmware/robot2015/src-ctrl/main.cpp
+++ b/firmware/robot2015/src-ctrl/main.cpp
@@ -176,6 +176,11 @@ int main() {
     // Make sure all of the motors are enabled
     motors_Init();
 
+    // setup analog in on battery sense pin
+    // the value is updated in the main loop below
+    AnalogIn batt(RJ_BATT_SENSE);
+    uint8_t battVoltage = 0;
+
     // Setup radio protocol handling
     const uint8_t robotID = 2;  // TODO: remove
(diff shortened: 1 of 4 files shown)


firmware/robot2015/src-ctrl/main.cpp
```diff
@@ -245,6 +245,14 @@ int main() {
             errorBitmask |= !status.hallOK << pair.second;
         }
 
+        // get the battery voltage
+        AnalogIn batt(RJ_BATT_SENSE);
+
+        uint8_t batt_vol = (batt.read_u16() >> 8);
+
+        const float batt_vol_in_soccer = batt_vol * 0.09884f;
```
justbuchanan: It'd be good to make this a constant in `rtp.hpp` so the magic number only exists in one place.

jgkamat:
@justbuchanan Something messed up in the merge and the `make fpga2015-test-strict` test is failing on master. On my machine its fine on this branch but on master its broken. Its probably fine for now, but just a heads up

justbuchanan:
Interesting... thanks for the heads up, I'll check it out!

PR #697, opened 2016-06-09
Switch to RoboJackets RRT library

### LINKED ISSUE #428: Replace old RRT implementation with the new one in the rrt repo

justbuchanan:
This isn't quite done yet, but is fairly close.  This ideally shouldn't affect behavior or performance much at all and is mostly a swap-in replacement.  After this is merged, I'd like to look into enabling and making use of some of the more advanced features this rrt library provides, such as ~goal-biased growth~, waypoint-biased growth, and ASC.  closes #428 

TODO:
- [x] testing
- [x] double-check `RRT.getPath()`
- [x] double-check EscapeObstaclesPlanner to ensure it can actually escape obstacles
- [x] ensure behavior is correct when start point is inside of an obstacle - see RoboCupStateSpace.transitionValid()
- [x] RoboCupStateSpace.stateValid()


chachmu:
@justbuchanan What's the status on this PR? Are you still working on this at all or should I go ahead and test this?

justbuchanan:
Yeah I can go ahead and finish this up.  Once comp2016 gets merged, I'll get it working and let you guys know.

justbuchanan:
This needs a little more testing, but other than that should be about done.  Let me know if you see any issues.

Note that this doesn't setup any of the fancy features such as:
* waypoint bias (grow path biased towards previous path)
* goal bias
* adaptive stepsize

It wouldn't be too hard to add those, but I'll leave that as a task for someone else :smile:

jgkamat:
Nice, thanks! I'm going to figure out why the build is failing now (I think it's due to the submodules somehow)...

@joshhting

justbuchanan:
Yeah I'm not sure why the build doesn't work.  It looks related to the fact that I changed `submodule update` calls to `submodule update --recursive` because the rrt library has submodules that we have to pull.  I don't know why that breaks though.

jgkamat:
It seems like `git submodule --recursive` stores links with absolute paths, instead of relative ones (whaat) somehow. On my system, these links are relative (try opening up `external/rrt/third_party/DoCIF/.git`, and see if it's a relative or absolute path). 

Right now, we could either stop syncing submodules recursively on the host before ubuntu-setup, or we could try to figure out if a newer version of git fixes this (and upgrade on the circleci host).

jgkamat:
this oneliner also does it for us (but it kills me)
`find -type f -name .git -exec bash -c 'f="{}"; cd $(dirname $f); echo "gitdir: $(realpath --relative-to=. $(cut -d" " -f2 .git))" > .git' \;`
https://stackoverflow.com/questions/10953953/ensuring-relative-git-paths

justbuchanan:
Hmm, that's strange that git hardcodes absolute paths for submodules since it prevents moving a repo :/.  I have the same version of git as you and the submodule paths are all relative on my local repo.

Not sure how hard it would be to update git on circleci, but that should fix this.  Maybe this would do it?

~~~
sudo add-apt-repository ppa:git-core/ppa
sudo apt-get update
sudo apt-get install git
~~~

jgkamat:
This unfortunately won't work because the submodules are initially cloned on the true target machine (circleci) with an old git version. I think the best course of action is simply to not submodule init recursively at the start, I'll see if that works now.

jgkamat:
This fix won't work when builds are cached though, since we don't run ubuntu-setup when we're cached. If we keep going with this we'll need to run the submodule update --recursive inside the container every time.

justbuchanan:
Ahh darn :/.  So updating git in the circle.yml doesn't get us a new enough version to avoid the issue?

I can look into making the rrt library able to compile without the submodules - it only depends on googletest for testing and docif for its own ci.  This will break if the rrt library requires submodules in the future though.

jgkamat:
I think I got it, I just upgraded the git version in the circle container (above). That should solve the problem (as the newest git makes relative links, which can be read inside the container). It's kind of ugly, but this works. If it ever breaks, we can just use the one liner that was in that stackoverflow thread.

justbuchanan:
Nice, I think it's all working now!  You can probably rebase the last several commits and wipe my 'git hacks' commit that didn't work if you want to :).

cc @ashaw596 You know the RRT stuff pretty well so if you could look over this too, that would be great!

jgkamat:
I just cleared out the last few commits, make sure to reset your repo :P

soccer/planning/RoboCupStateSpace.hpp
```diff
@@ -0,0 +1,68 @@
...
+        for (const auto& shape : _obstacles.shapes()) {
+            if (shape->hit(Geometry2d::Segment(from, to)) && !shape->hit(from))
+                return false;
+        }
+        return true;
+    }
+
+private:
+    const Geometry2d::ShapeSet& _obstacles;
+    const Field_Dimensions _fieldDimensions;
```
ashaw596: Do we really have to copy this? Its not that important, but it is around 15 floats and shapes you're copying each frame.

soccer/planning/EscapeObstaclesPathPlanner.cpp
```diff
@@ -53,20 +54,24 @@ Point EscapeObstaclesPathPlanner::findNonBlockedGoal(
     Point goal, boost::optional<Point> prevGoal, const ShapeSet& obstacles,
     int maxItr) {
     if (obstacles.hit(goal)) {
-        FixedStepTree goalTree;
-        goalTree.init(goal, &obstacles);
-        goalTree.step = stepSize();
+        auto stateSpace = make_shared<RoboCupStateSpace>(
```
ashaw596: Alternatively, you can save the stateSpace instead of making a new shared_pointer each iteration.

soccer/planning/RRTPlanner.cpp
```diff
@@ -182,57 +182,28 @@ vector<Point> RRTPlanner::runRRT(MotionInstant start, MotionInstant goal,
...
-            if (tb->connect(newPoint->pos)) {
-                // trees connected
-                // done with global path finding
-                // the path is from start to goal
-                // runRRT will handle the rest
-                break;
-            }
-        }
+    // Initialize bi-directional RRT
+    auto stateSpace = make_shared<RoboCupStateSpace>(
```
ashaw596: Alternatively, you can save the stateSpace instead of making a new shared_pointer each iteration.

soccer/planning/RoboCupStateSpace.hpp
jgkamat: We could have this function hook into the geometry packet changing mechanism if we wanted to avoid copying it around unless absolutely needed.

ashaw596:
Hmm, its the RRT is is sort of cause some weird giant paths to be planned compared to the old planner. Like this https://www.dropbox.com/s/nhtkrdz7j1z1wha/Screenshot%202016-11-05%2003.31.01.png?dl=0

jgkamat:
Hmm, maybe we could use the RRT submodule from the same version that we're using it with now, and then tweak the rrt until it performs well?

ashaw596:
Yeah. its probs fine.

soccer/planning/EscapeObstaclesPathPlanner.cpp
justbuchanan: Yeah I mostly did it this way to avoid the complexity of handling field size changes.

justbuchanan:
Yeah the giant paths thing is pretty strange.  I looked into it some and am not sure what the cause is, but I'll look into it some more.  This RRT should generate paths pretty much the same as before.

justbuchanan:
I finally got back around to working on this and I think it's working well!  Give it a try and see if you can get it to generate any paths worse than usual.

The reason it was getting such large paths before was because the previous implementation always ran for the maximum number of iterations and took the best path it found.  The new rrt just returns the first path it finds.  I improved the performance significantly by setting the goal bias to 0.3 (it was 0 before), so it should do a lot better job now of finding short paths.  I opened an issue (https://github.com/RoboJackets/rrt/issues/52) to add a feature to the rrt library for continuing after finding the first path in order to try for a better one.

I also added some debug visualizations so that you can see the actual rrt trees in soccer.  It's fairly hacky though and needs a little work.  Note that in the screenshot, most robots aren't using the rrt planner because a straight-line path is available

TODO:
- [ ] draw rrt visualizations on a separate debug layer for each robot rather than all on one.  This will probably require adding a shell id property to `SinglePlanRequest`
- [ ] disable drawing by default by changing the constant at the top of `RRTPlanner.cpp`
- [ ] uncheck the rrt drawing layers by default
- [ ] is there a better way to pass `SystemState` to the planner?

2016-11-30-221316_1338x987_scrot

ashaw596:
ooh, nice figuring that out. That looks awesome man! It's really useful to see the trees in soccer!!!! Also I was messing around with partial replanning and I think the waypoint bias with the new RRT will help alot. Even with partial replanning, paths in alternating directions still basically make the robot stand still since they take time to accelerate in every direction. I'll try it out ..... after I finish failing all these hell week tests....

justbuchanan:
Thanks!  Yeah I think being able to see the rrts in the ui will be pretty helpful.  I was stuck on figuring out why the paths were so bad before, but once I added the tree drawing, it was pretty easy to see what was going on.  The waypoint bias should definitely be good for re-planning.  I haven't touched that for this setup, but it shouldn't be hard for you to get it up and running.

jgkamat:
This version seems a lot better than the previous ones (the paths planned seem much more similar to the old version). I'm going to test it out after this week though. I'm not sure about this, but it seems to be slightly slower than the previous version? Is this expected?

justbuchanan:
Drawing the rrts in soccer is super slow.  There's a `const bool` at the top of `RRTPlanner.cpp` that you can flip to turn the drawing off and it'll be fast again.  I'll turn it off before merging.

jgkamat:
Hmm, is it slow even when we disable the RRT Path planning layer in the visible layers? That might be an optimization we can make if so.

justbuchanan:
Yeah even if it doesn't draw it on-screen, it still takes a while for it to record it all to the LogFrame.

justbuchanan:
Other than some code cleanup, I think this is about done.  I crossed off most of my TODOs:

* RRT goal bias, step size, and logging enable/disable are now all in the config gui
* The RRTs used by the EscapeObstaclesPlanner are now logged
* Each robot's RRTs are drawn on a separate debug layer

soccer/planning/RoboCupStateSpace.hpp
justbuchanan: I think unless changes to the global field dimensions object are protected by the processor's mutex, we need to copy it.  Otherwise the field dimensions might change mid-plan and screw things up.

justbuchanan:
Any idea why ci is failing on this?  All I see is `c++: internal compiler error: Killed (program cc1plus)`, which doesn't look good... But probably not a problem with our code - everything builds fine on my machine.  We could try updating to a newer compiler version.

jgkamat:
Yah, I've been getting that a few times, I tried tweaking the caching a bit, and I think it may be better on master, but most likeley this is something that was caused with circleci changing their hosts at some point. It could be that the hosts are running out of ram, but I'm not sure. 

Rerunning the test might fix the issue? I'll take a closer look at this...

jgkamat:
I'm also going to merge master in (there were a few issues with caching which I think I resolved a few commits back), hopefully that will fix the problem on this branch.

jgkamat:
Should be fixed now (I hope) going forward minus style. Let me know if you see it again, which is a possibility.

justbuchanan:
Nice!  Looks like everything's working now.

justbuchanan:
I think this is good to merge whenever you guys are ready.

jgkamat:
Cool, I'll give it a final look over and I'll merge it in! (probably later today).

PR #698, opened 2016-06-09
Fix ball sense issue

### LINKED ISSUE #658: Determine range of lighting conditions that work for ball sensor
The ball sensor code for the old robots checks to ensure that the detector isn't receiving too much light.  If it is, we know that it's too bright for the sensor to actually tell if the emitter is on or off.

justbuchanan:
The logic for detecting a break in the beam was somewhat circular because it updated the "dark" reading every other update and compared it to the "light" reading.  For the first break in the beam, this would trigger a break, but after that, the dark and light readings would always be the same because of the ball being in the way.

This changes it to consider the beam broken if the light and dark readings are the same.

fixes #658 

TODO:

- [x] tune `BallSense.sense_threshold` - I set it to `20` for now, but that's pretty arbitrary

cc @JNeiger

### DIFF
--- a/firmware/common2015/drivers/ball_sense/BallSense.cpp
+++ b/firmware/common2015/drivers/ball_sense/BallSense.cpp
@@ -17,11 +17,13 @@ void BallSense::update_ball_sensor() {
         emitter_pin.write(0);
 
         // Possible break in beam
-        if (sense_light - sense_dark > sense_threshold) {
+        if (std::abs(sense_light - sense_dark) < sense_threshold) {
             consec_ctr++;
         } else {
             consec_ctr = 0;
         }
+        // printf("{'light': %d, 'dark': %d, 'ball': %s},\r\n", sense_light,
+        // sense_dark, have_ball() ? "True" : "False");
     } else  // Emitter off
     {
         // Update value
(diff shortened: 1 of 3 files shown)


justbuchanan:
Here's a list of ball sense readings (sense_light - sense_dark) when there is no ball:

~~~{.sh}
ball diff: 16
ball diff: 48
ball diff: 16
ball diff: 32
ball diff: 48
ball diff: 16
ball diff: 16
ball diff: 16
ball diff: 0
ball diff: 0
ball diff: 0
ball diff: 16
ball diff: 16
ball diff: 48
ball diff: 16
ball diff: 16
ball diff: 16
ball diff: 32
ball diff: 32
ball diff: 16
ball diff: 32
~~~

I'm not sure yet why we're gettings zero readings occasionally - those should only happen when there's a ball.

justbuchanan:
For reference, here are the digikey pages for the ball sense parts:
* [Emitter](http://www.digikey.com/product-detail/en/vishay-semiconductor-opto-division/VSMB294008G/VSMB294008GCT-ND/5012875)
* [Detector](http://www.digikey.com/product-detail/en/fairchild-semiconductor/QSB363ZR/QSB363ZRCT-ND/3041757)

justbuchanan:
I bumped up the update frequency and the readings actually make sense now!  Not sure what the issue was before.

PR #700, opened 2016-06-09
Flip SPI lines back for the FPGA

### LINKED ISSUE #590: Remove FPGA's SoftwareSPI for configuration with new Control Boards
On the first PCB board design, the pins are wrong - so the SoftwareSPI has to be used for initially configuring the FPGA. This is fixed for the boards that will come in next, so the firmware can be switched to use regular SPI for those.

jjones646:
In [jon/fpga-spi-fix](https://github.com/RoboJackets/robocup-software/compare/jon/fpga-spi-fix)

fixes #590 

TODO:
- [ ] Confirm this will work on the final PCB design


jjones646:
this apparently didn't get changed...closing.

PR #703, opened 2016-06-13
PID control for robots

### LINKED ISSUE #619: Setup PID control in robot firmware
Eventually we're planning to use an LQR controller (see PR #322), but it would be nice to have a PID controller setup in the meantime because it's much simpler to implement.

justbuchanan:
This adds PID control for the robot, running one controller per wheel.  The `PidMotionController` class is initialized in `ControllerTaskThread.cpp`, where the `run()` method is called at regular intervals.  The controller takes in the encoder ticks since the last run and the elapsed time, which it uses to calculate the current wheel velocities.  It then calculates the desired wheel velocities from the desired bot velocity (x, y, w) and runs pid.

TODO:

- [x] test on a robot
- [x] more documentation
- [x] remove or explain matlab code for generating transformation matrices
- [ ] tune pid weights
- [x] connect to radio protocol
- [x] fix eigen includes in CMake to reduce compile warnings in soccer
- [x] fix eigen compile error on circleci - it can't find std::round for some reason...
- [x] re-enable command velocity timeout
- [x] explain pid analysis python script
- [x] fix soccer command velocities being backwards/rotated?
- [x] remove `_robot->config->angleVelMultiplier`?

closes #619


justbuchanan:
@jjones646 I've been working on this and I noticed that my values for estimated wheel velocity are significantly higher reality.  I've poked around and I'm wondering the calculation for `dt` is giving a number that's too small.  Here's an excerpt from `ControllerTaskThread.cpp` that you wrote:

~~~{.cpp}
/*
 * The time since the last update is derived with the value of
 * WATCHDOG_TIMER_CLK_WIDTH in robocup.v
 *
 * The last encoder reading (5th one) from the FPGA is the watchdog
 * timer's tick since the last SPI transfer.
 *
 * Multiply the received tick count by:
 *     (1/18.432) * 2 * (2^WATCHDOG_TIMER_CLK_WIDTH)
 *
 * This will give you the duration since the last SPI transfer in
 * microseconds (us).
 *
 * For example, if WATCHDOG_TIMER_CLK_WIDTH = 6, here's how you would
 * convert into time assuming the fpga returned a reading of 1265 ticks:
 *     time_in_us = [ 1265 * (1/18.432) * 2 * (2^6) ] = 8784.7us
 *
 * The precision would be in increments of the multiplier. For
 * this example, that is:
 *     time_precision = 6.94us
 *
 */
const float dt = enc_deltas.back() * (1 / 18.432e6) * 2 * 64;
~~~

Any chance there's a mistake in here somewhere?  Where did the (1/18.432) come from?

jjones646:
Hmmm. So there's a pretty major bug in the verilog...The 1/18.432e6 is the FPGA's clock speed.

See [this change](https://github.com/RoboJackets/robocup-software/commit/658963a56e24800054b7f9befc5b1572e818a837) that I just pushed. Don't synthesize with [jon/new-ucf](https://github.com/RoboJackets/robocup-software/tree/jon/new-ucf) directly however - the ucf file is slightly different on it.

Once that's updated, remove the 2x multiplier in the cpp. It will be dead on 5ms 99% of the time - just tested it all.

As a side note, the time value is latched at the same time that the encoder deltas are also latched - after the 1st byte in the SPI transfer. The duty cycles are updated at the end of the SPI transfer, all at once.

justbuchanan:
Ok cool.  Being one bit off makes quite a difference haha.  I think for now I'll go ahead and just use `CONTROL_LOOP_WAIT_MS` as `dt` since it's 5ms, and just make a note to change back to this once that branch is merged and figured out.  Thanks!

justbuchanan:
We had our first robot tornado tonight!

I had been working on this branch last night and got PID pretty much ready to go and everything was running great.  Tonight I went to do some more work on it, so I re-flashed the robot to make sure it had the latest code, put it on the field, then started soccer.  The moment I touched the joystick to drive it around, it started spinning in place at full speed, completely out of control.  Unplugging the base station didn't stop it, so I wrestled it down and managed to flip off the power switch.

It took a little debugging, but I found the issue...  The mbed that was on it today was a different one than yesterday and it didn't have the updated fpga code with [this commit](https://github.com/RoboJackets/robocup-software/commit/333bee3be5e7962297d35c53ad3649e0287f8031) that handles negative encoder values appropriately.  Two of the wheels were supposed to spin backwards to make the bot move forwards, but since the fpga passed on encoder absolute values (no negatives), the backward wheel motion registered as forward.  PID compensated for this by increasing the duty cycle, which made the wheels report faster forward speed, which led to further compensation until the motors maxed out at about 7000rpm...  This graph sums it up - see wheels 1 and 3 (legend is bottom right):

2016-06-17-005539_948x1034_scrot

Lessons learned:
* a single wrong character is all it takes
* our motors are f***ing powerful
* make sure that the fpga code is up-to-date


cc @jjones646 @barulicm

jjones646:
I can fully understand why high school teachers marks the whole problem wrong for an incorrect sign now...

We should add a play for the robots that does that. Sounds exciting!

firmware/robot2015/src-ctrl/modules/ControllerTaskThread.cpp
```diff
@@ -78,41 +79,34 @@ void Task_Controller(void const* args) {
...
 
     while (true) {
-        imu.getGyro(gyroVals);
-        imu.getAccelero(accelVals);
+        // imu.getGyro(gyroVals);
+        // imu.getAccelero(accelVals);
+
+        // note: the 4th value is not an encoder value.  See the large comment
+        // below for an explanation.
+        array<int16_t, 5> enc_deltas;
```
ashaw596: Zero initialize. array<int16_t, 5> enc_deltas{}

justbuchanan: They get set to legit values by the `set_duty_get_enc()` call a few lines down.

firmware/robot2015/src-ctrl/modules/ControllerTaskThread.cpp
```diff
@@ -78,41 +79,34 @@ void Task_Controller(void const* args) {
...
     osSignalSet(mainID, MAIN_TASK_CONTINUE);
     Thread::signal_wait(SUB_TASK_CONTINUE, osWaitForever);
 
-    const uint16_t ENC_TICKS_PER_TURN = 2048;
-
-    Pid motor2pid(0.0, 0.0, 0.0);
-
-    std::vector<int16_t> duty_cycles;
+    array<int16_t, 5> duty_cycles;
+    duty_cycles.fill(0);
```
ashaw596: Maybe just use the default initializer.

firmware/robot2015/src-ctrl/modules/control/PidMotionController.hpp
```diff
@@ -0,0 +1,83 @@
...
+
+    void setPidValues(float p, float i, float d) {
+        for (Pid& ctl : _controllers) {
+            ctl.kp = p;
+            ctl.ki = i;
+            ctl.kd = d;
+        }
+    }
+
+    void setTargetVel(Eigen::Vector3f target) { _targetVel = target; }
```
ashaw596: std::move

firmware/robot2015/src-ctrl/modules/control/PidMotionController.hpp
```diff
@@ -0,0 +1,83 @@
...
+#endif
+
+        return dutyCycles;
+    }
+
+    static const uint16_t ENC_TICKS_PER_TURN = 2048;
+
+private:
+    /// controllers for each wheel
+    Pid _controllers[4];
```
ashaw596: std::array?

firmware/robot2015/src-ctrl/modules/control/RobotModel.hpp
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    /// (wheel rad/s desired) * DutyCycleMultiplier = duty cycle
+    /// Note that this is an approximation, as the relationship isn't exactly
+    /// linear
+    float DutyCycleMultiplier;
+};
+
+/// Model parameters for 2015 robot
+static const RobotModel RobotModel2015 = []() {
```
ashaw596: Change to constructor?

firmware/robot2015/src-ctrl/modules/ControllerTaskThread.cpp
ashaw596: Well, there is a small edge case where if duty_cycle errors for some reason and is too high, set_duty_get_enc would return early. Not very likely, but it feels in general just zero initializing things is safer as a default since it really doesn't change effeciency that much.

justbuchanan: Good point, didn't notice that :+1:

firmware/robot2015/src-ctrl/modules/control/PidMotionController.hpp
justbuchanan: I think in the case a move is equivalent to a copy since the values in the vector are allocated directly in the object, rather than on the heap.

firmware/robot2015/src-ctrl/modules/control/RobotModel.hpp
justbuchanan: Yeah you're right, `static` isn't what we want here.  http://stackoverflow.com/questions/92546/variable-declarations-in-header-files-static-or-not

I did it this way as opposed to using a constructor for all of the parameters for readability and less code.  This way I don't have to write a constructor that does `x = x` for the 5 params.  It also is easier to read (I think) the initialization parameters this way as opposed to a constructor that takes in 4 floats and an array - it's hard to tell what's what.

Issue #686, opened 2016-05-27
mbed crashes & reboots periodically


justbuchanan:
It might be that it's running out of memory, but we're not quite sure.

cc @jjones646

jjones646:
Almost certainly a memory issue. I've been running everything through soccer for 15+ min now using ca04d6f, and it's been doing great.

jjones646:
This line in Console.cpp definitely isn't helping the issue...

```
// reserve space for 5 lines in rx buffer
_rxBuffer.reserve(400);
```

This seems to be heap allocation.

jjones646:
Another thing to note is that `_rxBuffer` is a `std::string` object. And it looks like under allocation is fine since all additions use `_rxBuffer.push_back`. So a reservation of 10 or 15 is more reasonable.

justbuchanan:
This file in the mbed repo has some info on stack sizes for the main thread and timer thread: https://github.com/mbedmicro/mbed/blob/master/libraries/doc/rtos.txt.

We would have to fork the mbed repo, but it may be worth looking into reducing both of those threads' sizes - main is definitely a lot bigger than it needs to be since most of our stuff happens in other threads.

justbuchanan:
Mostly fixed in #707

PR #707, opened 2016-06-14
Reduce stack sizes for main and timer threads

justbuchanan:
This doesn't fix the crash/reboot problem entirely, but it seems to make it happen less often.  The mbed build script allows you to pass macro definitions on the command line, so I added entries for OS_MAINSTKSIZE and TIMERSTKSZ in arm_mbed.cmake.  I'm not sure how small we can get away with for these because the `ps` command shows that those threads use 100% of their stack.  I'm sure this isn't correct, but I don't know how to get better numbers.

I also added a "heapfill" command that lets you get a rough idea of how much space is available on the heap.

Related to #686 

cc @JNeiger


justbuchanan:
Also, it seems like changing these values in the cmake file doesn't trigger a rebuild of the mbed library, so you'll have to do a `make clean` before rebuilding for the changes to take effect.

Issue #90, opened 2014-07-15
RoleAssignment 'required' and 'priority' do nothing for composite behaviors


justbuchanan:

kfu02:
old

PR #714, opened 2016-06-16
Added a subbehavior to submissive defender to clear the ball.

### LINKED ISSUE #228: Improve Defender and SubmissiveDefender tactics to clear ball
Right now if a ball gets into the defense zone, the defenders just hover around it.  It would be much better if they had a "clear" state so they get the ball and kick it out of the way or pass it to another robot.
screen shot 2015-02-21 at 8 51 31 pm

chachmu:
Submissive Defenders will now clear the ball if it comes within a certain radius of the goal and they can reach the ball before an opponent. Closes #228

### DIFF
--- a/soccer/gameplay/role_assignment.py
+++ b/soccer/gameplay/role_assignment.py
@@ -19,7 +19,7 @@ class RoleRequirements:
         self.require_kicking = False
 
         # multiply this by the distance between two points to get the cost
-        self.PositionCostMultiplier = 1.0
+        self.position_cost_multiplier = 1.0
 
     def __str__(self):
         props = []
@@ -52,12 +52,12 @@ class RoleRequirements:
         self._destination_shape = value
 
     @property
-    def PositionCostMultiplier(self):
-        return self._PositionCostMultiplier
+    def position_cost_multiplier(self):
+        return self._position_cost_multiplier
 
-    @PositionCostMultiplier.setter
-    def PositionCostMultiplier(self, value):
-        self._PositionCostMultiplier = value
+    @position_cost_multiplier.setter
+    def position_cost_multiplier(self, value):
+        self._position_cost_multiplier = value
 
     @property
     def has_ball(self):
@@ -243,7 +243,7 @@ def assign_roles(robots, role_reqs):
                 cost = MaxWeight
             else:
                 if req.destination_shape != None:
-                    cost += req.PositionCostMultiplier * req.destination_shape.dist_to(
+                    cost += req.position_cost_multiplier * req.destination_shape.dist_to(
                         robot.pos)
                 if req.previous_shell_id != None and req.previous_shell_id != robot.shell_id(
                 ):
--- a/soccer/gameplay/role_assignment.py
+++ b/soccer/gameplay/role_assignment.py
@@ -17,6 +17,9 @@ class RoleRequirements:
         self.required = False
         self.priority = 0
         self.require_kicking = False
+        
+        # multiply this by the distance between two points to get the cost
+        self.PositionCostMultiplier = 1.0
 
     def __str__(self):
         props = []
@@ -47,6 +50,14 @@ class RoleRequirements:
             raise TypeError("Unexpected type for destination_shape: " + str(
                 value))
         self._destination_shape = value
+
+    @property
+    def PositionCostMultiplier(self):
+        return self._PositionCostMultiplier
+    
+    @PositionCostMultiplier.setter
+    def PositionCostMultiplier(self, value):
+        self._PositionCostMultiplier=value
 
     @property
     def has_ball(self):
@@ -140,9 +151,6 @@ class ImpossibleAssignmentError(RuntimeError):
 # the munkres library doesn't like infinity, so we use this instead
 MaxWeight = 10000000
 
-# multiply this by the distance between two points to get the cost
-PositionCostMultiplier = 1.0
-
 # how much penalty is there for switching robots mid-play
 RobotChangeCost = 1.0
 
@@ -235,8 +243,10 @@ def assign_roles(robots, role_reqs):
                 cost = MaxWeight
             else:
                 if req.destination_shape != None:
-                    cost += PositionCostMultiplier * req.destination_shape.dist_to(
+                    cost += req.PositionCostMultiplier * req.destination_shape.dist_to(
                         robot.pos)
+                    if req.PositionCostMultiplier>1:
+                        print(req.PositionCostMultiplier)
                 if req.previous_shell_id != None and req.previous_shell_id != robot.shell_id(
                 ):
                     cost += RobotChangeCost
--- a/soccer/gameplay/role_assignment.py
+++ b/soccer/gameplay/role_assignment.py
@@ -17,7 +17,7 @@ class RoleRequirements:
         self.required = False
         self.priority = 0
         self.require_kicking = False
-        
+
         # multiply this by the distance between two points to get the cost
         self.PositionCostMultiplier = 1.0
 
@@ -54,10 +54,10 @@ class RoleRequirements:
     @property
     def PositionCostMultiplier(self):
         return self._PositionCostMultiplier
-    
+
     @PositionCostMultiplier.setter
     def PositionCostMultiplier(self, value):
-        self._PositionCostMultiplier=value
+        self._PositionCostMultiplier = value
 
     @property
     def has_ball(self):
@@ -245,8 +245,6 @@ def assign_roles(robots, role_reqs):
                 if req.destination_shape != None:
                     cost += req.PositionCostMultiplier * req.destination_shape.dist_to(
                         robot.pos)
-                    if req.PositionCostMultiplier>1:
-                        print(req.PositionCostMultiplier)
                 if req.previous_shell_id != None and req.previous_shell_id != robot.shell_id(
                 ):
                     cost += RobotChangeCost
--- a/soccer/gameplay/skills/pivot_kick.py
+++ b/soccer/gameplay/skills/pivot_kick.py
@@ -172,5 +172,6 @@ class PivotKick(single_robot_composite_behavior.SingleRobotCompositeBehavior,
             if self.use_chipper:
                 r.chipper_preference_weight = role_assignment.PreferChipper
             r.require_kicking = True
+            r.PositionCostMultiplier = 1.5
 
         return reqs
--- a/soccer/gameplay/skills/pivot_kick.py
+++ b/soccer/gameplay/skills/pivot_kick.py
@@ -172,6 +172,8 @@ class PivotKick(single_robot_composite_behavior.SingleRobotCompositeBehavior,
             if self.use_chipper:
                 r.chipper_preference_weight = role_assignment.PreferChipper
             r.require_kicking = True
-            r.PositionCostMultiplier = 1.5
+
+            #Increase the importance of Position to make robots closer to the ball clear it during defense
+            r.position_cost_multiplier = 1.5
 
         return reqs
--- a/soccer/gameplay/tactics/defense.py
+++ b/soccer/gameplay/tactics/defense.py
@@ -29,8 +29,6 @@ class Defense(composite_behavior.CompositeBehavior):
     def __init__(self, defender_priorities=[20, 19]):
         super().__init__(continuous=True)
 
-        self.go_clear = False
-
         # we could make the Defense tactic have more or less defenders, but right now we only support two
         if len(defender_priorities) != 2:
             raise RuntimeError(
@@ -44,10 +42,11 @@ class Defense(composite_behavior.CompositeBehavior):
                             Defense.State.defending, lambda: True,
                             "immediately")
         self.add_transition(Defense.State.defending, Defense.State.clearing,
-                            lambda: self.go_clear,
+                            lambda: self.should_clear_ball(),
                             "when it is safe to clear the ball")
         self.add_transition(Defense.State.clearing, Defense.State.defending,
-                            lambda: not self.go_clear, "done clearing")
+                            lambda: not self.should_clear_ball(),
+                            "done clearing")
 
         goalie = tactics.positions.submissive_goalie.SubmissiveGoalie()
         goalie.shell_id = main.root_play().goalie_id
@@ -76,20 +75,35 @@ class Defense(composite_behavior.CompositeBehavior):
     def debug(self, value):
         self._debug = value
 
-    def execute_defending(self):
+    def should_clear_ball(self):
+        #Returns true if our robot can reach the ball sooner than the closest opponent
+        safe_to_clear = False
+        if main.ball().pos.mag() < constants.Field.ArcRadius * 2:
 
-        defender1 = self.subbehavior_with_name('defender1')
-        defender2 = self.subbehavior_with_name('defender2')
-        close_defender = defender1
-        if (main.ball().pos.mag() < constants.Field.ArcRadius * 2):
+            defender1 = self.subbehavior_with_name('defender1')
+            defender2 = self.subbehavior_with_name('defender2')
             if (defender1.robot != None and defender2.robot != None):
-                if (not defender1.go_clear and not defender2.go_clear):
-                    if close_defender.should_clear_ball(
-                            evaluation.ball.time_to_ball(
-                                close_defender.robot)):
-                        close_defender.go_clear = True
-                        self.go_clear = True
+                #TODO: change this to use the config system when we figure out how to do that
+                max_vel = 3.5
+                max_accel = 1.8
 
+                for robot in main.system_state().their_robots:
+                    their_dist_to_ball = robot.pos.dist_to(main.ball().pos)
+                    #if their robot is moving faster than ours, assume it is at its maximum speed, otherwise assume its max speed is the same as ours
+                    their_max_vel = max(max_vel, robot.vel.mag())
+
+                #calculate time for the closest opponent to reach ball based on current /vel/pos data * .9 for safety
+                their_time_to_ball = (
+                    their_dist_to_ball /
+                    their_max_vel) * defender1.safety_multiplier
+                if their_time_to_ball > evaluation.ball.time_to_ball(
+                        defender1.robot) or their_time_to_ball > evaluation.ball.time_to_ball(
+                            defender2.robot):
+                    safe_to_clear = True
+
+        return safe_to_clear
+
+    def execute_running(self):
         self.recalculate()
 
         goalie = self.subbehavior_with_name("goalie")
@@ -105,11 +119,14 @@ class Defense(composite_behavior.CompositeBehavior):
     def execute_clearing(self):
         defender1 = self.subbehavior_with_name('defender1')
         defender2 = self.subbehavior_with_name('defender2')
-        if (defender1.robot != None and defender2.robot != None):
-            if (not defender1.go_clear and not defender2.go_clear):
-                self.go_clear = False
+        defender1.go_clear = True
+        defender2.go_clear = False
 
-        #main.system_state().draw_circle(robocup.Point(0, 0), constants.Field.ArcRadius * 2,constants.Colors.Red, "Clear Ball")
+    def on_exit_clearing(self):
+        defender1 = self.subbehavior_with_name('defender1')
+        defender2 = self.subbehavior_with_name('defender2')
+        defender1.go_clear = False
+        defender2.go_clear = False
 
     def recalculate(self):
         goalie = self.subbehavior_with_name('goalie')
--- a/soccer/gameplay/tactics/defense.py
+++ b/soccer/gameplay/tactics/defense.py
@@ -80,6 +80,19 @@ class Defense(composite_behavior.CompositeBehavior):
         self._debug = value
 
     def execute_defending(self):
+        
+        defender1 = self.subbehavior_with_name('defender1')
+        defender2 = self.subbehavior_with_name('defender2')
+        close_defender=defender1
+        if(main.ball().pos.mag()<constants.Field.ArcRadius*2):
+            if(defender1!=None and defender2!= None):
+                if(defender1.robot!=None and defender2.robot!=None):
+                    if(not defender1.go_clear and not defender2.go_clear):
+                        #if(defender1.time_to_ball()>defender2.time_to_ball()): #removing this technically makes things work
+                            #close_defender=defender2
+                        if close_defender.should_clear_ball(close_defender.time_to_ball()):
+                            close_defender.go_clear=True
+                            self.go_clear=True
         self.recalculate()
 
         goalie = self.subbehavior_with_name("goalie")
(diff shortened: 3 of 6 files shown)


soccer/gameplay/role_assignment.py
```diff
@@ -18,6 +18,9 @@ def __init__(self):
         self.priority = 0
         self.require_kicking = False
 
+        # multiply this by the distance between two points to get the cost
+        self.PositionCostMultiplier = 1.0
```
justbuchanan: Go ahead and rename this to lowercase with underscores so it matches the rest of the instance variables.  The capital letter style is more for global constants.

soccer/gameplay/skills/pivot_kick.py
```diff
@@ -172,5 +172,6 @@ def role_requirements(self):
             if self.use_chipper:
                 r.chipper_preference_weight = role_assignment.PreferChipper
             r.require_kicking = True
+            r.PositionCostMultiplier = 1.5
```
justbuchanan: Can you put a comment here noting why the position cost multiplier is overridden?

soccer/gameplay/tactics/defense.py
```diff
@@ -20,6 +20,12 @@
 # The old defense strategy had a goalie and two defenders that didn't coordinate with eachother
 # and tended to overlap and not get an optimal positioning - this tactic handles the coordination.
 class Defense(composite_behavior.CompositeBehavior):
+    class State(Enum):
+        ## gets between a particular opponent and the goal.  stays closer to the goal
+        defending = 1
+        clearing = 2
```
justbuchanan: Add a comment here for what the "clearing" state is for.

soccer/gameplay/tactics/defense.py
```diff
@@ -61,7 +76,21 @@ def debug(self):
...
         self._debug = value
 
-    def execute_running(self):
+    def execute_defending(self):
+
+        defender1 = self.subbehavior_with_name('defender1')
+        defender2 = self.subbehavior_with_name('defender2')
+        close_defender = defender1
+        if (main.ball().pos.mag() < constants.Field.ArcRadius * 2):
+            if (defender1 != None and defender2 != None):
```
justbuchanan: I'm pretty sure `defender1` and `defender2` will always be non-null, so you shouldn't have to check that here.

soccer/gameplay/tactics/defense.py
```diff
@@ -71,6 +100,23 @@ def execute_running(self):
...
             # TODO: move a lot of this code into modules in the evaluation folder
+
+        main.system_state().draw_circle(
+            robocup.Point(0, 0), constants.Field.ArcRadius * 2,
+            constants.Colors.Red, "Clear Ball")
+
+    def execute_clearing(self):
+        defender1 = self.subbehavior_with_name('defender1')
+        defender2 = self.subbehavior_with_name('defender2')
+        if (defender1 != None and defender2 != None):
```
justbuchanan: Same as above - I don't think this check is needed.

soccer/gameplay/tactics/defense.py
```diff
@@ -61,7 +76,21 @@ def debug(self):
...
+    def execute_defending(self):
+
+        defender1 = self.subbehavior_with_name('defender1')
+        defender2 = self.subbehavior_with_name('defender2')
+        close_defender = defender1
+        if (main.ball().pos.mag() < constants.Field.ArcRadius * 2):
+            if (defender1 != None and defender2 != None):
+                if (defender1.robot != None and defender2.robot != None):
+                    if (not defender1.go_clear and not defender2.go_clear):
+                        #if(defender1.time_to_ball()>defender2.time_to_ball()): #removing this technically makes things work
```
justbuchanan: x

soccer/gameplay/tactics/defense.py
```diff
@@ -71,6 +100,23 @@ def execute_running(self):
...
+
+    def execute_clearing(self):
+        defender1 = self.subbehavior_with_name('defender1')
+        defender2 = self.subbehavior_with_name('defender2')
+        if (defender1 != None and defender2 != None):
+            if (defender1.robot != None and defender2.robot != None):
+                if (not defender1.go_clear and not defender2.go_clear):
+                    self.go_clear = False
+
+        main.system_state().draw_circle(
```
justbuchanan: It might be good to comment this out to avoid cluttering the field view.

soccer/gameplay/tactics/positions/submissive_defender.py
```diff
@@ -13,11 +13,13 @@
...
 class SubmissiveDefender(
-        single_robot_composite_behavior.SingleRobotCompositeBehavior):
+    single_robot_composite_behavior.SingleRobotCompositeBehavior):
     class State(Enum):
         ## gets between a particular opponent and the goal.  stays closer to the goal
         marking = 1
-        # TODO: add clear state to get and kick a free ball
+        clearing = 2
+
+    go_clear = False
```
justbuchanan: Putting this declaration here makes it a class variable - shared between all instances of SubmissiveDefender.  If you put it inside __init__ as `self.go_clear = False`, it will be an instance variable which is probably what you want.

soccer/gameplay/tactics/positions/submissive_defender.py
```diff
@@ -29,10 +31,56 @@ def __init__(self):
...
+                            SubmissiveDefender.State.clearing,
+                            lambda: self.go_clear,
+                            "when it is safe to clear the ball")
+        self.add_transition(
+            SubmissiveDefender.State.clearing,
+            SubmissiveDefender.State.marking,
+            lambda: self.subbehavior_with_name('kick-clear').state == behavior.Behavior.State.completed or not self.should_clear_ball(self.time_to_ball()),
+            "done clearing")
+
+    def time_to_ball(self):
```
justbuchanan: This would be a good function to put in the `evaluation` folder since it's pretty generic and could be useful to many different behaviors.

soccer/gameplay/tactics/positions/submissive_defender.py
```diff
@@ -29,10 +31,56 @@ def __init__(self):
...
+        if main.ball().pos.mag() < constants.Field.ArcRadius * 2:
+            safe_to_clear = True
+            #change this to use the config system when we figure out how to do that
+            max_vel = 3.5
+            max_accel = 1.8
+
+            for robot in main.system_state().their_robots:
+                their_dist_to_ball = robot.pos.dist_to(main.ball().pos)
+                #if their robot is moving faster than ours, assume it is at its maximum speed, otherwise assume its max speed is the same as ours
+                if robot.vel.mag() > max_vel:
```
justbuchanan: This if/else could also be written as `their_max_vel = max(max_vel, robot.vel.mag())`

soccer/gameplay/tactics/positions/submissive_defender.py
```diff
@@ -29,10 +31,56 @@ def __init__(self):
...
+        max_vel = 3.5
+        max_accel = 1.8
+        delay = .1  #tune this better
+        rpos = self.robot.pos
+        bpos = main.ball().pos
+        #calculate time for self to reach ball using max_vel + a slight delay for capture
+        dist_to_ball = self.robot.pos.dist_to(main.ball().pos)
+        return (dist_to_ball / max_vel) + delay
+
+    def should_clear_ball(self, our_time_to_ball):
```
justbuchanan: Add a comment above the method giving a brief description of what it's doing - mainly that it returns true if our bot is closer than the closest opponent.

soccer/gameplay/tactics/positions/submissive_defender.py
```diff
@@ -151,6 +199,29 @@ def execute_marking(self):
     def on_exit_marking(self):
         self.remove_subbehavior('move')
 
+    def on_enter_clearing(self):
+        print("CLEARING BALL")
```
justbuchanan: x

soccer/gameplay/tactics/positions/submissive_defender.py
```diff
@@ -151,6 +199,29 @@ def execute_marking(self):
...
+        kick.aim_params['error_threshold'] = 1.0
+        kick.aim_params['max_steady_ang_vel'] = 12
+
+        # chip
+        kick.chip_power = 1.0
+        kick.use_chipper = True
+
+        kick.target = robocup.Segment(
+            robocup.Point(-constants.Field.Width / 2, constants.Field.Length),
+            robocup.Point(constants.Field.Width / 2, constants.Field.Length))
```
justbuchanan: It might be good to make the target segment a little narrower so that it tends to kick the ball more length-wise down the field than sideways - this'll help make it more likely that the ball stays in play rather than getting kicked out towards the side.

soccer/gameplay/tactics/positions/submissive_defender.py
```diff
@@ -151,6 +199,29 @@ def execute_marking(self):
...
+
+        self.add_subbehavior(kick, 'kick-clear', required=True)
```
justbuchanan: Due to a limitation in our role assignment system (https://github.com/RoboJackets/robocup-software/issues/90), `required` should be set to False here.  Basically the issue is that the Defender behavior is optional (required=False), but it's optionalness doesn't carry down to its subbehaviors.  It's on the todo-list to make the `required` flag make more sense in the context of parent/child behaviors, but it hasn't been done yet.

soccer/gameplay/tactics/positions/submissive_defender.py
```diff
@@ -151,6 +199,29 @@ def execute_marking(self):
...
+
+    def on_exit_clearing(self):
+        print("DONE CLEARING BALL")
```
justbuchanan: x

soccer/gameplay/tactics/defense.py
jgkamat: not sure about this, but if we only have 1 robot on defense, this might make a difference. (eg: we take 2 robots off of defense for some reason while we only have 3 robots)

soccer/gameplay/tactics/positions/submissive_defender.py
```diff
@@ -29,10 +31,56 @@ def __init__(self):
...
+            for robot in main.system_state().their_robots:
+                their_dist_to_ball = robot.pos.dist_to(main.ball().pos)
+                #if their robot is moving faster than ours, assume it is at its maximum speed, otherwise assume its max speed is the same as ours
+                if robot.vel.mag() > max_vel:
+                    their_max_vel = robot.vel.mag()
+                else:
+                    their_max_vel = max_vel
+
+                #calculate time for the closest opponent to reach ball based on current /vel/pos data * .9 for safety
+                their_time_to_ball = (their_dist_to_ball / max_vel) * .9
```
jgkamat: Can you make the `0.9` a constant of this class? That way it's easier to tune at comp if there are issues.

soccer/gameplay/tactics/defense.py
```diff
@@ -20,6 +20,12 @@
 # The old defense strategy had a goalie and two defenders that didn't coordinate with eachother
 # and tended to overlap and not get an optimal positioning - this tactic handles the coordination.
 class Defense(composite_behavior.CompositeBehavior):
+    class State(Enum):
+        ## gets between a particular opponent and the goal.  stays closer to the goal
+        defending = 1
+        clearing = 2
+
+    go_clear = False
```
jgkamat: Same as below, this should be self.go_clear and in `__init__`

soccer/gameplay/tactics/defense.py
justbuchanan: The Defense behavior is setup so that it always creates two SubmissiveDefender behaviors.  They're marked as optional, so there are definitely times when a robot is not assigned to the behavior, but the behavior is still always there.

jgkamat: Ah, so the `if (defender1.robot != None and defender2.robot != None):` should take of what we want (logically). Thanks!

soccer/gameplay/tactics/positions/submissive_defender.py
```diff
@@ -29,10 +34,45 @@ def __init__(self):
...
+            SubmissiveDefender.State.clearing,
+            SubmissiveDefender.State.marking,
+            lambda: self.subbehavior_with_name('kick-clear').state == behavior.Behavior.State.completed or not self.should_clear_ball(evaluation.ball.time_to_ball(self.robot)),
+            "done clearing")
+
+    def should_clear_ball(self, our_time_to_ball):
+        #Returns true if our robot can reach the ball sooner than the closest opponent
+        if main.ball().pos.mag() < constants.Field.ArcRadius * 2:
+            safe_to_clear = True
+            #change this to use the config system when we figure out how to do that
```
justbuchanan: add TODO after the # sign

soccer/gameplay/tactics/positions/submissive_defender.py
```diff
@@ -29,10 +34,45 @@ def __init__(self):
...
+            max_vel = 3.5
+            max_accel = 1.8
+
+            for robot in main.system_state().their_robots:
+                their_dist_to_ball = robot.pos.dist_to(main.ball().pos)
+                #if their robot is moving faster than ours, assume it is at its maximum speed, otherwise assume its max speed is the same as ours
+                their_max_vel = max(max_vel, robot.vel.mag())
```
justbuchanan: It may be best to just always assume their robots can go as fast as ours.  Often times our defenders are stationary, so with this code, an enemy that's barely moving will be assumed to be at its max speed.

soccer/gameplay/tactics/positions/submissive_defender.py
```diff
@@ -7,20 +7,25 @@
...
         ## gets between a particular opponent and the goal.  stays closer to the goal
         marking = 1
-        # TODO: add clear state to get and kick a free ball
+        clearing = 2
 
     def __init__(self):
         super().__init__(continuous=True)
+
+        self.go_clear = False
+        self.safety_multiplier = 0.9
```
justbuchanan: It would be good to add a comment for `safety_multiplier` specifying what it changes.  It'll make it more straightforward if we need to tune the value later.

soccer/gameplay/tactics/positions/submissive_defender.py
```diff
@@ -7,20 +7,25 @@
...
     class State(Enum):
         ## gets between a particular opponent and the goal.  stays closer to the goal
         marking = 1
-        # TODO: add clear state to get and kick a free ball
+        clearing = 2
 
     def __init__(self):
         super().__init__(continuous=True)
+
+        self.go_clear = False
```
justbuchanan: Also add a comment here that the `go_clear` variable is used by exernal behaviors to tell the `SubmissiveDefender` to clear the ball.

soccer/gameplay/tactics/positions/submissive_defender.py
```diff
@@ -29,10 +34,45 @@ def __init__(self):
...
+                            "when it is safe to clear the ball")
+        self.add_transition(
+            SubmissiveDefender.State.clearing,
+            SubmissiveDefender.State.marking,
+            lambda: self.subbehavior_with_name('kick-clear').state == behavior.Behavior.State.completed or not self.should_clear_ball(evaluation.ball.time_to_ball(self.robot)),
+            "done clearing")
+
+    def should_clear_ball(self, our_time_to_ball):
+        #Returns true if our robot can reach the ball sooner than the closest opponent
+        if main.ball().pos.mag() < constants.Field.ArcRadius * 2:
```
justbuchanan: I noticed that the ball position is checked to be within radius both here and in the `Defense` behavior.  Is there a way this can be consolidated so that it's only checked in one place?

soccer/gameplay/tactics/positions/submissive_defender.py
chachmu: This is based on the opponent robots current velocity, not ours. That way, we assume that the opponent robot has the same max velocity as ours unless they are already moving faster.

soccer/gameplay/tactics/positions/submissive_defender.py
chachmu: @justbuchanan If I remove that call then it would still work but the rest of the if statements after it would get called a lot, if that's ok then I can remove it. It's necessary in the other spot because the robot needs to be able to escape the clearing behavior if the ball rolls out of the radius.

chachmu:
This commit just failed CircleCI due to the stylecheck on aim.py which I did not touch, is something going on with stylecheck?

jgkamat:
If you merge master into your branch that should go away. 
We check on the latest master, not the diverging commit, which would be better.

soccer/gameplay/tactics/positions/submissive_defender.py
justbuchanan: Ahh, you're right, I completely misread the code.  :+1:

soccer/gameplay/tactics/defense.py
```diff
@@ -61,7 +76,20 @@ def debug(self):
...
+                if (not defender1.go_clear and not defender2.go_clear):
+                    if close_defender.should_clear_ball(
+                            evaluation.ball.time_to_ball(
+                                close_defender.robot)):
+                        close_defender.go_clear = True
+                        self.go_clear = True
+
         self.recalculate()
 
         goalie = self.subbehavior_with_name("goalie")
```
justbuchanan: Now that `defending` isn't the only main state, the chunk of code here for checking the goalie should be put into `execute_running()` so that it runs during both `clearing` and `defending`.

soccer/gameplay/tactics/positions/submissive_defender.py
justbuchanan: How about something like this:
* remove any logic related to determining if the ball should be cleared from SubmissiveDefender and put it all in Defense.  You could have a `should_clear_ball()` in Defense that returns True if the ball is within radius and one of our two defending robots is closer (time-wise) than the closest opponent.  The SubmissiveDefender behavior will then unconditionally go for the ball when `go_clear` is set and all responsibility for deciding when the robot should clear will be solely up to the Defense behavior.
* Remove the Defense.go_clear variable and change the transitions to and from `clearing` to call the `should_clear_ball()` method.
* Remove most of `execute_clearing` - that logic should now be in `should_clear_ball()`
* Make the on_enter_clearing and on_exit_clearing methods set the go_clear variable for the defenders

I think this will simplify things a bit and will consolidate the logic for deciding when to clear the ball.  Thoughts?

soccer/gameplay/tactics/defense.py
```diff
@@ -71,6 +113,21 @@ def execute_running(self):
...
+        defender1 = self.subbehavior_with_name('defender1')
+        defender2 = self.subbehavior_with_name('defender2')
+        defender1.go_clear = True
+        defender2.go_clear = False
+
+    def on_exit_clearing(self):
+        defender1 = self.subbehavior_with_name('defender1')
+        defender2 = self.subbehavior_with_name('defender2')
+        defender1.go_clear = False
+        defender2.go_clear = False
```
justbuchanan: If we're always using the `defender1` behavior as the ball-clearer (which I think is a good idea), we shouldn't have to set `go_clear` to false for `defender2` - it'll always already be false.

soccer/gameplay/tactics/defense.py
```diff
@@ -71,6 +113,21 @@ def execute_running(self):
...
 
             # TODO: move a lot of this code into modules in the evaluation folder
+
+            #main.system_state().draw_circle(robocup.Point(0, 0), constants.Field.ArcRadius * 2,constants.Colors.Red, "Clear Ball")
+
+    def execute_clearing(self):
+        defender1 = self.subbehavior_with_name('defender1')
+        defender2 = self.subbehavior_with_name('defender2')
+        defender1.go_clear = True
+        defender2.go_clear = False
```
justbuchanan: Same here - we shouldn't have to set it to `defender2`'s flag to `False`

soccer/gameplay/tactics/defense.py
```diff
@@ -71,6 +113,21 @@ def execute_running(self):
             # raise RuntimeError("Defense tactic requires a goalie id to be set")
 
             # TODO: move a lot of this code into modules in the evaluation folder
+
+            #main.system_state().draw_circle(robocup.Point(0, 0), constants.Field.ArcRadius * 2,constants.Colors.Red, "Clear Ball")
+
+    def execute_clearing(self):
```
justbuchanan: This should be `on_enter_clearing()` since this code only needs to get called once - not continuously while we're in the `clearing` state.

justbuchanan:
Nice, it's looking good :+1: .  I made a couple more notes, but after that go ahead and merge it.

soccer/gameplay/tactics/defense.py
chachmu: For some reason if it isn't called continuously it doesn't work, the robots just enter clearing for one frame and then exit it. I think it might have something to do with the behavior changing over to the closer robot and somehow resetting the go_clear flag or something.

justbuchanan: Hmm, it may be an issue with the role assignments system or something.  I'll look at it and get back to you.

chachmu: @justbuchanan I figured that part out, now I just have one question about the goalie. If the ball goes within the white line, the goalie tries to clear it. Should I make the defenders only clear if the ball is outside that region so they don't fight with the goalie?

justbuchanan: Good call, I hadn't thought about the goalie.  There is an obstacle created in the GameplayModule that keeps all non-goalie robots out of the goal zone.  This will keep the defenders from actually being able to go into the goal zone, but they'll sort of waffle around on the edge of the goal zone unproductively.  It would be best if they leave the ball to the goalie if it's in the goal zone so that they can focus on defending.

justbuchanan: Take a look at the `Field_Dimensions` class - I think it's got a goal zone shape with a "contains point" method you can use here.

chachmu:
@justbuchanan I merged master into this branch like Jay had previously suggested but now the stylecheck still fails but it says artifact not found when I try to view the details. Any suggestions?

soccer/gameplay/tactics/defense.py
```diff
@@ -116,17 +118,15 @@ def execute_running(self):
...
-    def execute_clearing(self):
+    def on_enter_clearing(self):
         defender1 = self.subbehavior_with_name('defender1')
         defender2 = self.subbehavior_with_name('defender2')
         defender1.go_clear = True
-        defender2.go_clear = False
 
     def on_exit_clearing(self):
         defender1 = self.subbehavior_with_name('defender1')
         defender2 = self.subbehavior_with_name('defender2')
```
justbuchanan: x

soccer/gameplay/tactics/defense.py
```diff
@@ -116,17 +118,15 @@ def execute_running(self):
 
             #main.system_state().draw_circle(robocup.Point(0, 0), constants.Field.ArcRadius * 2,constants.Colors.Red, "Clear Ball")
 
-    def execute_clearing(self):
+    def on_enter_clearing(self):
         defender1 = self.subbehavior_with_name('defender1')
         defender2 = self.subbehavior_with_name('defender2')
```
justbuchanan: x

chachmu: whoops, thought I had deleted those

justbuchanan:
The artifacts directory that contains build logs doesn't go live until after the build is fully finished.  If you click on the "ci/circleci" Details link, it will take you to the main build page if you're interested in seeing it's current progress.

To fix the style issue, wait til the artifacts go live and run the command specified on the "circle/style" Details page.

chachmu:
@justbuchanan How long does that usually take? The ci/circleci details say that the build finished and the test metadata has been uploaded.

justbuchanan:
Yeah you're right. @jgkamat can you take a look at this? The build page says there are no artifacts

justbuchanan:
@chachmu you can disregard the style error and go ahead and merge this.

jgkamat:
Huh, thats weird. I'll definetly take a look at this branch in a bit

jgkamat:
I'm going to blame circle for this one, If I ssh into the machine, the correct dir is still populated with the artifacts. I could contact support to get them to fix this but I dont think its worth it at this point. Let me know if you see it anywhere else and I'll get on it.

PR #716, opened 2016-06-17
Add logging on Referee data protobuf

### LINKED ISSUE #710: Log referee data
Right now the [raw_referee](https://github.com/RoboJackets/robocup-software/blob/master/common/protobuf/LogFrame.proto#L91) field of the LogFrame is a `bytes` field, but it would be easier to use if it was an embedded protobuf mesage, similar to how we store vision data.

Also, the [simple_logger](https://github.com/RoboJackets/robocup-software/blob/09a5e6bddb1eb7aaa0a2bb8d920c972f3035f960/logging/simple_logger.cpp#L112) currently writes to this field, but soccer does not.  The place to add this to soccer would be in Processor.cpp.

If you could get this working on a new branch and send a pull request, that would be awesome!

ryanstrat:
Fixes #710 

Data from the refbox is now stored in a new raw_refbox field in a LogFrame Protobuf.

I derived this branch off of my bsv-export branch because I needed my in-progress BSV exporter to view the referee data, since we don't currently display it in any other place. I'm still adding export fields to the output on the ryan/bsv-export branch, which is why this is a separate pull request.

### DIFF
--- a/soccer/LogViewer.cpp
+++ b/soccer/LogViewer.cpp
@@ -182,7 +182,7 @@ bool LogViewer::exportBSV(char* logFilename, char* bsvFilename) {
         SSL_Referee_Stage stage;
         SSL_Referee_Command command;
 
-        if (currentFrame->raw_referee_size() > 0) {
+        if (currentFrame->raw_refbox_size() > 0) {
             SSL_Referee referee = currentFrame->raw_refbox(0);
 
             stage = referee.stage();
--- a/soccer/LogViewer.cpp
+++ b/soccer/LogViewer.cpp
@@ -12,8 +12,6 @@
 #include <algorithm>
 #include <fcntl.h>
 
-#include <iostream>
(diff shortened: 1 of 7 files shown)


soccer/LogViewer.cpp
```diff
@@ -1,37 +1,72 @@
...
     }
 
+    // Process args
+    for (int i = 1; i < argc; ++i) {
+        const char* var = argv[i];
+
+        if (strcmp(var, "-bsv") == 0) {
+            bsv = true;
+            ++i;
+            bsvFilename = argv[i];
```
jgkamat: This may segfault if '-bsv' is the last arg :P

ryanstrat:
I'm going to close this pull request and reopen one for my BSV export branch.

I've fixed the segfault Jay found in that branch, which has all of these changes merged as well.

PR #717, opened 2016-06-17
Export BSV File from RJ Protobuf Log & Log Referee data in Soccer

ryanstrat:
This is a combo of former pull request #716 

This branch exports a protobuf log into two human readable BSV files. The files can be inspected manually for checking log files or they can be imported into a spreadsheet app or data platform.

Also merged into this branch is the contents of pull request #716 
Referee packets are now stored as a protobuf in the log by soccer. compared with no referee data before.

The simple logger has not been updated to reflect this change, but probably should be.


soccer/Processor.cpp
```diff
@@ -386,6 +386,16 @@ void Processor::run() {
...
 
+        // Log referee data
+        vector<NewRefereePacket*> refereePackets;
+        _refereeModule.get()->getPackets(refereePackets);
+        fprintf(stderr, "Ref packets received this loop:%lu\n",
+                refereePackets.size());
+        for (NewRefereePacket* packet : refereePackets) {
+            SSL_Referee* log = _state.logFrame->add_raw_refbox();
+            log->CopyFrom(packet->wrapper);
+        }
```
justbuchanan: I think there's a memory leak here.  Packets are allocated with `new`, so they should be `delete`d here after you're done with them.

soccer/LogViewer.cpp
```diff
@@ -1,37 +1,82 @@
...
+        }
+    }
+
     LogViewer win;
 
-    win.readFrames(argv[1]);
-    win.showMaximized();
+    if (bsv) {
+        // Export BSV. Runs log_viewer in headless mode
+        if (win.exportBSV(logFilename, bsvFilename) == false) {
```
justbuchanan: What does the return value of `exportBSV()` mean?  Does `false` indicate success?

justbuchanan:
Nice, BSV export is a cool feature to have.  I might recommend making it its own program though - BSV export doesn't really fit with what the log viewer app is - it looks a little forced making them the same program.

justbuchanan:
Also, we might as well change simple logger to log referee data the same way as soccer - it would be best to have them log data in the same way to avoid future confusion.

ryanstrat:
I'll be working on the simple logger today.

I'm all for BSV conversion being a separate program, but I have no idea how to make that happen. CMake is still basically magic. Could you point me in the right direction for that?

soccer/Processor.cpp
ryanstrat: I don't see where I'm declaring anything with new in the above code. The buffer is also emptied when it is read in the getPackets() method.

Also, NewRefereePacket is a very unfortunate class name.

soccer/LogViewer.cpp
ryanstrat: I think at one point, I wanted it to return if reading the input file was successful, which I still think makes sense. I will go ahead and add that

soccer/Processor.cpp
ryanstrat: Actually, I think I see it now.

We make the NewRefPackets with new, then add them to the vector, and in getFrames, we just flush the pointers from the vector, but we should also delete the objects there as well. I'll fix that.

I suspect that vision frames have the same memory leak in this case.

ryanstrat: I have no idea how to fix this. I tried a few things, but couldn't get it to compile.

I would very much like some assistance with this tonight.

justbuchanan:
Haha CMake is a little magical.  The `logging` directory would probably be a good place for this.  All you would have to do is move the relevant code to a new .cpp file under /logging, maybe "log_converter.cpp" or something.  Then open up the [CMakeLists.txt](https://github.com/robojackets/robocup-software/blob/ryan/bsv-export/logging/CMakeLists.txt#L7) file in that directory and add a new executable.  It would look something like this:

~~~{.cpp}
add_executable(log_converter log_converter.cpp)
target_link_libraries(simple_logger common)
~~~

I think that should about do it, but let me know if you have any issues.

soccer/Processor.cpp
justbuchanan: The best thing would probably be to use `std::shared_ptr`.  It basically adds reference-counting so that when all of the `shared_ptr`s themselves go out of scope, the object that they point to gets `delete`d.

If you're not familiar with C++ smart pointers, I can help you out here.  The main thing would be to change the declaration of the packet vector in NewRefereeModule.hpp from `std::vector<NewRefereePacket*>` to `std::vector<std::shared_ptr<NewRefereePacket>>` and to change the allocation of the packets from `NewRefereePacket* packet = new NewRefereePacket` to `auto packet = make_shared<NewRefereePacket>()`.

Also, I agree... `NewRefereePacket` should really be `RefereePacket` - I think that naming is left over from when we converted from new to old.

justbuchanan: The nice thing with `shared_ptr` is that you never need to call `delete` or worry about memory leaks.  As long as you store it in a `shared_ptr`, it's all taken care of for you automatically.

justbuchanan: Btw, it looks like vision packets do get deleted after they're used: https://github.com/robojackets/robocup-software/blob/ryan/bsv-export/soccer/Processor.cpp#L386.

Smart pointers would be preferable there too, but it's not a big deal.

ryanstrat:
@justbuchanan 

I think I've covered everything and I'm ready for you to review again.

ryanstrat:
Also CheckStyle is failing on a file I didn't touch. Please advise

logging/log_converter.cpp
```diff
@@ -0,0 +1,218 @@
...
+#include <vector>
+
+#include <fcntl.h>
+#include <time.h>
+
+using namespace Packet;
+using namespace std;
+
+#define BAR "|"
+#define MATCH_ID_LENGTH 32
```
justbuchanan: This is fine as-is, but it's slightly preferred to avoid macros where possible, since they don't obey scope and type rules of c++.  BAR could be `const char` and MATCH_ID_LENGTH could be `const int`.

logging/log_converter.cpp
```diff
@@ -0,0 +1,218 @@
...
+bool readFrames(const char* filename) {
+    QFile file(filename);
+
+    if (!file.open(QFile::ReadOnly)) {
+        fprintf(stderr, "Can't open %s: %s\n", filename,
+                (const char*)file.errorString().toLatin1());
+        return false;
+    }
+
+    int n = 0;
```
justbuchanan: Is `n` used for anything?

justbuchanan:
Looks good to me, go ahead and merge when ready.

The checkstyle issue is sort of an edge case that pops up occasionally.  The reason is that  it's setup to run on files that *differ* from master, which includes the files that you've changed since branching off master as well as files that have changed on master since then.  I haven't gotten around to it yet, but I'm going to change it to only run on files that you've changed, so this shouldn't happen.

For now though, if this gets merged into master, it won't be an issue.  If you want to, you can click on "Details" next to the failed style check and it will give you a command to copy-paste into a terminal to fix the style issue.

justbuchanan:
Also add log_converter to `run/.gitignore` so git doesn't try to track it.

logging/log_converter.cpp
ryanstrat: I don't think so. That's legacy from log_viewer. I got rid of it and everything still works. I don't see where it is used at all either. I wonder if at one point it was used to place the frames into an array.

PR #720, opened 2016-06-18
Make indirect passes try to pass to another robot

### LINKED ISSUE #369: Don't attempt to score right away on indirect kicks
Right now we do, which would count against us if we were able to make it.  The ball must touch another robot before we can score.

jgkamat:
Fixes #369 

This makes our_free_kick take a line_kick to a second robot (which has its best receive point determined by the touchpass algorithm), and passes to it. I also added some docs! 

I'm kind of queasy about one of the changes though, I'll point it out in a inline comment below.

### DIFF
--- a/soccer/gameplay/plays/restarts/our_free_kick.py
+++ b/soccer/gameplay/plays/restarts/our_free_kick.py
@@ -11,7 +11,6 @@ import evaluation.touchpass_positioning
 
 class OurFreeKick(standard_play.StandardPlay):
 
-    tpass = evaluation.touchpass_positioning
     running = False
 
     def __init__(self):
@@ -42,7 +41,7 @@ class OurFreeKick(standard_play.StandardPlay):
 
         lastKicker = kicker
         if self.indirect:
-            receive_pt, target_point, probability = OurFreeKick.tpass.eval_best_receive_point(
+            receive_pt, target_point, probability = evaluation.touchpass_positioning.eval_best_receive_point(
                 main.ball().pos)
             pass_behavior = tactics.coordinated_pass.CoordinatedPass(
                 receive_pt, None, (kicker, lambda x: True))
--- a/soccer/gameplay/plays/restarts/our_free_kick.py
+++ b/soccer/gameplay/plays/restarts/our_free_kick.py
@@ -5,12 +5,14 @@ import skills.pivot_kick
 import constants
 import robocup
 import main
+import tactics.coordinated_pass
 import evaluation.touchpass_positioning
 
 
 class OurFreeKick(standard_play.StandardPlay):
 
     tpass = evaluation.touchpass_positioning
+    running = False
 
     def __init__(self):
         super().__init__(continuous=True)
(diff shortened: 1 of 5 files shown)


soccer/gameplay/plays/restarts/our_free_kick.py
```diff
@@ -21,25 +30,44 @@ def __init__(self):
...
 
     @classmethod
     def score(cls):
         gs = main.game_state()
-        return 0 if gs.is_ready_state() and gs.is_our_free_kick() else float(
+        return 0 if OurFreeKick.running or (gs.is_ready_state() and gs.is_our_free_kick()) else float(
             "inf")
 
+    def on_enter_running(self):
+        OurFreeKick.running = True
```
jgkamat: I was having issues with this play cutting off (so the receiver wouldnt get a chance to actually catch the ball), so I'm writing to a variable here which will force the score to stay at 0 (as long as an instance of this play is running). There is a race condition here, but we won't ever want to run more than one restart at a time, so I think its fine for now (?)

@justbuchanan

soccer/gameplay/plays/restarts/our_free_kick.py
```diff
@@ -2,16 +2,25 @@
...
 import constants
 import robocup
 import main
+import tactics.coordinated_pass
+import evaluation.touchpass_positioning
 
 
 class OurFreeKick(standard_play.StandardPlay):
+
+    tpass = evaluation.touchpass_positioning
```
justbuchanan: What's the reasoning for the `tpass` variable?  Any advantage over just calling `evaluation.touchpass_positioning` below?

soccer/gameplay/skills/_kick.py
```diff
@@ -24,6 +24,8 @@ def __init__(self):
 
         self.shot_obstacle_ignoring_robots = []
 
+        self.aim_params = {}
```
justbuchanan: Is this used anywhere?

jgkamat: I made some changes to _kick to allow `line_kick` to work with coordinated pass. Its a little bit hacky, but it should allow anything that subclasses kick to be used in coordinated_pass.

soccer/gameplay/plays/restarts/our_free_kick.py
jgkamat: This is a good point, I'll fix it in the next push

soccer/gameplay/tactics/coordinated_pass.py
```diff
@@ -30,18 +30,32 @@ class State(enum.Enum):
...
         receiving = 3  # the kicker has kicked and the receiver is trying to get the ball
 
-    ## Skillreceiver is a class that will handle the receiving robot. See pass_receive and angle_receive.
+    ## Init method for CoordinatedPass
+    # @param skillreceiver a class that will handle the receiving robot. See pass_receive and angle_receive for examples.
     # Using this, you can change what the receiving robot does (rather than just receiving the ball, it can pass or shoot it).
-    def __init__(self, receive_point=None, skillreceiver=None):
+    # Subclasses of pass_receive are preferred, but check the usage of this variable to be sure.
+    # @param receive_point The point that will be kicked too. (Target point)
+    # @param skillkicker A tuple of this form (kicking_class, ready_lambda). If none, it will use (pivot_kick lambda x: x == pivot_kick.State.aimed).
```
justbuchanan: Instead of kicking class, this should probably say **instance** of a kicking class, right?  Same for skillreceiver.

jgkamat: ah, yes :+1:

jgkamat:
This is in comp2016, Closing.

PR #722, opened 2016-06-20
Changed RRTPlanner to user Simple Beziers. Added debugging text to Paths.

ashaw596:
Change RRT to use simple bezier paths. 
Add debug text to paths. 
Changed capture to ignore ball as an obstacle

### DIFF
--- a/soccer/gameplay/skills/capture.py
+++ b/soccer/gameplay/skills/capture.py
@@ -99,7 +99,8 @@ class Capture(single_robot_behavior.SingleRobotBehavior):
 
     def execute_course_approach(self):
         # don't hit the ball on accident
-        self.robot.set_avoid_ball_radius(Capture.CourseApproachAvoidBall)
+        #self.robot.set_avoid_ball_radius(Capture.CourseApproachAvoidBall)
+        self.robot.disable_avoid_ball()
         pos = self.find_intercept_point()
         self.robot.face(main.ball().pos)
 
--- a/soccer/gameplay/skills/capture.py
+++ b/soccer/gameplay/skills/capture.py
@@ -105,7 +105,6 @@ class Capture(single_robot_behavior.SingleRobotBehavior):
         self.lastApproachTarget == None
 
     def execute_course_approach(self):
-        # don't hit the ball on accident
         pos = self.find_intercept_point()
 
         if (self.lastApproachTarget != None and
@@ -114,7 +113,8 @@ class Capture(single_robot_behavior.SingleRobotBehavior):
 
         self.lastApproachTarget = pos
 
-        if (pos-main.ball().pos).mag() < Capture.CourseApproachAvoidBall + constants.Robot.Radius:
+        # don't hit the ball on accident
+        if pos.dist_to(main.ball().pos) < Capture.CourseApproachAvoidBall + constants.Robot.Radius:
             self.robot.disable_avoid_ball()
         else:
             self.robot.set_avoid_ball_radius(Capture.CourseApproachAvoidBall)
--- a/soccer/gameplay/skills/capture.py
+++ b/soccer/gameplay/skills/capture.py
@@ -106,21 +106,23 @@ class Capture(single_robot_behavior.SingleRobotBehavior):
 
     def execute_course_approach(self):
         # don't hit the ball on accident
-        #self.robot.set_avoid_ball_radius(Capture.CourseApproachAvoidBall)
-        self.robot.disable_avoid_ball()
         pos = self.find_intercept_point()
 
         if (self.lastApproachTarget != None and
             (pos - self.lastApproachTarget).mag() < 0.1):
-            self.robot.move_to(self.lastApproachTarget)
-            main.system_state().draw_circle(self.lastApproachTarget,
-                                            constants.Ball.Radius,
-                                            constants.Colors.White, "Capture")
+            pos = self.lastApproachTarget
+
+        self.lastApproachTarget = pos
+
+        if (pos-main.ball().pos).mag() < Capture.CourseApproachAvoidBall + constants.Robot.Radius:
+            self.robot.disable_avoid_ball()
         else:
-            main.system_state().draw_circle(pos, constants.Ball.Radius,
-                                            constants.Colors.White, "Capture")
-            self.robot.move_to(pos)
-            self.lastApproachTarget = pos
+            self.robot.set_avoid_ball_radius(Capture.CourseApproachAvoidBall)
+
+        self.robot.move_to(pos)
+        main.system_state().draw_circle(self.lastApproachTarget,
+                                        constants.Ball.Radius,
+                                        constants.Colors.White, "Capture")
 
     def on_exit_course_approach(self):
         self.lastApproachTarget == None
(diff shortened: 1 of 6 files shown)


soccer/gameplay/skills/capture.py
```diff
@@ -99,7 +99,8 @@ def on_enter_course_approach(self):
 
     def execute_course_approach(self):
         # don't hit the ball on accident
-        self.robot.set_avoid_ball_radius(Capture.CourseApproachAvoidBall)
+        #self.robot.set_avoid_ball_radius(Capture.CourseApproachAvoidBall)
+        self.robot.disable_avoid_ball()
```
justbuchanan: I think we probably want to still avoid the ball here.  The reason it's there is because we've had problems bumping into the ball during the course approach state.  Keep in mind sim behavior may differ from reality a good bit here.

soccer/planning/Path.hpp
```diff
@@ -227,6 +239,14 @@ class AngleFunctionPath : public Path {
 
     virtual RJ::Time startTime() const override { return path->startTime(); }
     virtual void setStartTime(RJ::Time t) override { path->setStartTime(t); }
+    virtual void setDebugText(QString string) override {
+        path->setDebugText(std::move(string));
+    }
+
+    virtual void drawDebugText(SystemState* state, const QColor& color = Qt::darkCyan,
```
justbuchanan: Why the overrides for this and `setDebugText()`?  I might have missed something, but they don't look they change superclass behavior.

soccer/planning/Path.hpp
```diff
@@ -90,6 +90,17 @@ class Path {
      */
     virtual std::unique_ptr<Path> clone() const = 0;
 
+    virtual void setDebugText(QString string) {
+        _debugText = std::move(string);
+    }
+
+    virtual void drawDebugText(SystemState* state, const QColor& color = Qt::darkCyan,
```
justbuchanan: These probably don't need to be virtual - it seems like something that doesn't really need to be overridden by subclasses.

soccer/gameplay/skills/capture.py
ashaw596: I don't think it should be a problem since we switch to fine approach at a good distance from the ball or any distance given that its in front of the ball. Also the end velocity is set to zero so it shouldn't be at speed towards the ball.

soccer/planning/Path.hpp
ashaw596: It forwards it to the path set.

soccer/planning/Path.hpp
ashaw596: I need it in order to forward the function correctly in the angularpath.

soccer/gameplay/skills/capture.py
justbuchanan: What about the case when the ball is moving away from the robot, so the course approach target point is on the opposite side of the ball.  The avoid here ensures that the path planner generates a path *around* the ball rather than through it.

soccer/planning/Path.hpp
justbuchanan: Ahh gotcha, I missed that.  Should they also set the debug text for the angle path itself or just the inner x/y-path?

soccer/gameplay/skills/capture.py
ashaw596: Well, the problem is that the path around the ball is usually pretty crappy because the robot can usually either catch up the the ball and capture it or the balls going at high speed, its usually going to go out.

justbuchanan: I agree that our capture is pretty crappy for moving balls.  With out current strategy of getting on the other side of it to catch it, we definitely need the avoid ball line here.  I do think it would be worthwhile to look into going directly for it, but in the meantime we should keep the avoid.

ashaw596: Ok, how about that. I disable the ball only when we try to plan straight into it. So the ball isn't moving the anyways. This prevents the messed up constant pathplanning when we try to plan into the ball obstacle while still going around the ball when we need to catch it.

soccer/gameplay/skills/capture.py
```diff
@@ -106,20 +106,23 @@ def on_enter_course_approach(self):
...
             (pos - self.lastApproachTarget).mag() < 0.1):
-            self.robot.move_to(self.lastApproachTarget)
-            main.system_state().draw_circle(self.lastApproachTarget,
-                                            constants.Ball.Radius,
-                                            constants.Colors.White, "Capture")
+            pos = self.lastApproachTarget
+
+        self.lastApproachTarget = pos
+
+        if (pos-main.ball().pos).mag() < Capture.CourseApproachAvoidBall + constants.Robot.Radius:
```
justbuchanan: `-

soccer/gameplay/skills/capture.py
```diff
@@ -106,20 +106,23 @@ def on_enter_course_approach(self):
 
     def execute_course_approach(self):
         # don't hit the ball on accident
```
justbuchanan: x

soccer/gameplay/skills/capture.py
justbuchanan: If the ball obstacle causes constant replanning, isn't there something we can do on the planning side to fix it rather than here?

Consider the case where the ball is moving away from the robot.  Here's what your new code says to do:
* go directly towards a point on the otherside of the ball, ignoring the ball as an obstacle.
* if you get within the course approach radius of the ball, then count the ball as an obstacle and try go around it

It seems like we would get better paths if we just count it as an obstacle the whole time, right?  Am I misunderstanding what's going on?

ashaw596: Um, the code says if the target point is not the ball it self go to that point with the ball as an obstacle. Otherwise it says go straight to the ball. 
I can't really do anything when we keep telling the robot to go straight into an obstacle....
I mean I can add some heuristics to keep down replanning, but why do we keep telling our robots to plan paths running straight into obstacles?

justbuchanan: Ahh, I gotcha - I think I was reading "pos" as "robot's current position" on that line - my mistake.

It looks like `find_intercept()` will sometimes return a point less than the course approach dist since it samples increments of 5cm from the ball until it finds a good point, while the course approach dist is 10cm.

I think what you've got now is good :+1:

justbuchanan:
@jgkamat The "ci/circleci" status here doesn't look quite right.

ashaw596:
Weird.

external/DoCIF
```diff
@@ -1 +1 @@
-Subproject commit 59e96f6dbb269c8b9c4dfcf44bbdc60ff876927f
+Subproject commit e52def19124bec4c9b743a8f7276ef2b14f64b51
```
ashaw596: Um, why was this in the patch I applied?

justbuchanan: See https://github.com/RoboJackets/robocup-software/pull/733

justbuchanan: @ashaw596 59e96f6dbb is the new one - it looks you merged master to get the new one, then added a commit reverting back to the old one.

ashaw596: Welp whatevs. I reverted that commit.

PR #725, opened 2016-06-22
Add support for ubuntu 15.10

jgkamat:
This should fix ubuntu 15.10 and 15.04 support, at the very least for compiling soccer, if not the firmware. 

@ryanstrat try it out and make sure it works. When i tried it I had to add the arm repo for it to work, even on 15.10 for some reason (which I'm sure wasn't the case before).

hmm...


ryanstrat:
My laptop is on 16.04 as of yesterday. @guyfleeman You needed 15.10 support, right?

guyfleeman:
@ryanstrat if the plan is still to use my laptop as one of the two comp machines, then yes to be on the safe side

jgkamat:
@guyfleeman I'll be able to fix any problems that come up for you at comp, but to be on the safe side, don't run ubuntu-setup if nothing's broken right now.

jgkamat:
Merging this now, as it LGTM, and we'll have everyone on 16.04 anyway. I'll get testing on 14.04 soon.

PR #728, opened 2016-06-22
Tune ball placement values

chachmu:
Removed calls for robot.shield_from_teammate() because it no longer exists


justbuchanan:
Looks good.

These should have been removed in https://github.com/RoboJackets/robocup-software/pull/663, when the avoid radius stuff got removed.

PR #733, opened 2016-06-23
Fixes the artifact deployment on a failure

### LINKED ISSUE #726: Build not publishing artifacts when build fails.
I suspect this is because we fail during the dependency step, rather than the testing step (to take advantage of circleci caching). If they can't fix it for some reason, we'll carry the failure into the testing stage, and everything should work.

This is more of an announcement for anyone trying to get their build artifacts, than a true issue here.

jgkamat:
Closes #726 

This fixes that artifact deployment bug we saw earlier @justbuchanan

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -32,7 +32,7 @@ dependencies:
 # The test phase dosen't actually run tests. We need the caching from the deploy step.
 test:
   override:
-    - sh -c true
+    - ./external/DoCIF/commands/exit.sh
 
 deployment:
   github:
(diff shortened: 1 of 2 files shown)


justbuchanan:
Nice!  I was missing having our logs :)

LGTM :+1:

jgkamat:
I'm going to see if that last commit passes, if it does, I'll go ahead and merge.

jgkamat:
Also just wanted to let you know this is at the bottom of the build step now =)

screen shot 2016-06-22 at 9 33 34 pm

justbuchanan:
That's pretty slick, I like it!

PR #734, opened 2016-06-23
Albert/path planning fail elegant

### LINKED ISSUE #715: Goalie rampages if ball is placed near the edge of the goal.
Sometimes if the ball is placed behind the defenders but in front or to the side of the goalie pathplanning or something related freaks out and the robots start driving randomly in loops. It seems like the easiest way to recreate this is by placing the ball just in front of the corner of the goal. It doesn't seem to be related entirely to the normal path planning because it happens during both move and worldvelocity commands.
goalierampage

[logs.zip](https://github.com/RoboJackets/robocup-software/files/319738/logs.zip)

ashaw596:
Give up when RRTPlanning Fails.
Fixes #715


justbuchanan:
:+1:

PR #751, opened 2016-06-24
New plays

chachmu:
A few test plays that I want to mess around with and a couple of plays in case we have less than 6 robots at comp.

### DIFF
--- a/soccer/gameplay/plays/offense/bypass.py
+++ b/soccer/gameplay/plays/offense/bypass.py
@@ -1,0 +1,77 @@
+import robocup
+import play
+import behavior
+import constants
+import main
+import skills.move
+import skills.capture
+import enum
+import evaluation
+import tactics.coordinated_pass
+import skills.angle_receive
+
+
+class RollingPass(play.Play):
+    #INCOMPLETE
+    #Just an idea for a play I wanted to test
+
+    class State(enum.Enum):
+        ## Capture ball
+        capturing = 1
+        ##prep
+        setup = 2
+        ## Determine target
+        passing = 3
+        ## Shoot
+        kicking = 4
+
+    def __init__(self):
+        super().__init__(continuous=False)
+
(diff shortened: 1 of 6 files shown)


soccer/gameplay/plays/robot_offensive3.py
```diff
@@ -0,0 +1,49 @@
...
+        goalie=tactics.positions.goalie.Goalie()
+        goalie.shell_id = main.root_play().goalie_id
+        self.add_subbehavior(goalie,'goalie',required=True)
+
+        striker=skills.pivot_kick.PivotKick()
+        striker.target = constants.Field.TheirGoalSegment
+        striker.aim_params['error_threshold'] = 0.15
+        striker.aim_params['max_steady_ang_vel'] = 7
+        striker.aim_params['min_steady_duration'] = 0.1
+        striker.aim_params['desperate_timeout'] = 2.5
```
jgkamat: Can you make these global vars for this class instead of magic numbers? It makes it easier to find them when tuning at comp.

soccer/gameplay/plays/offense/bypass.py
```diff
@@ -0,0 +1,77 @@
...
+        self.add_transition(behavior.Behavior.State.start,
+                            RollingPass.State.capture, lambda: True,
+                            'immediately')
+        self.add_transition(
+            RollingPass.State.capturing, RollingPass.State.passing,
+            lambda: self.subbehavior_with_name("capture"),
+            'all subbehaviors completed')
+
+        self.add_transition(
+            RollingPass.State.passing, RollingPass.State.passing,
```
jgkamat: Shoud this be passing -> kicking?

soccer/gameplay/plays/robot_offensive3.py
chachmu: I will, right now I'm just kind of messing around with a few different ideas to see what works.

PR #752, opened 2016-06-24
fixed ball sensing

### LINKED ISSUE #750: Fix ball sense (again)
I thought I had it all working as of #698, but there's something wrong.  It may be hardware-related.

chachmu:
Closes #750. Mostly due to some detectors being soldered on backwards which cause them to not work. Also, the code determining if we had the ball or not was inverted.


chachmu:
Ball sense works provided the control board has been modified. R18 must be removed and pins 1 and 3 of j19 (the emit port) must be connected.

PR #774, opened 2016-07-12
Make soccer continue without goalie

### LINKED ISSUE #772: Make soccer continue to run if goalie loses vision or is deselected

chachmu:
This is one possible solution to the goalie issue. I personally feel it is a little too hacky. The other option would be to make role assignments not return a robot if it requests a robot with an invalid robot id. Closes #772.


chachmu:
Is there a way to hide all the old commits from comp2016 somehow since this isn't really related to that?

jgkamat:
you can change the 'base' of the PR to comp2016, but we won't do a merge until comp2016 is merged anyway so yah.

soccer/gameplay/tactics/defense.py
```diff
@@ -48,9 +48,11 @@ def __init__(self, defender_priorities=[20, 19]):
                             lambda: not self.should_clear_ball(),
                             "done clearing")
 
-        goalie = tactics.positions.submissive_goalie.SubmissiveGoalie()
-        goalie.shell_id = main.root_play().goalie_id
-        self.add_subbehavior(goalie, "goalie", required=False)
+        valid_shells = [bot.shell_id for bot in main.our_robots()]
+        if main.root_play().goalie_id in valid_shells:
```
justbuchanan: Keep in mind that since this happens at `__init__`, a goalie won't be assigned if the right robot shows up later on.

chachmu:
This pull request is replaced by #777 which is a better solution and also fixes several other issues.

PR #775, opened 2016-07-12
updated stylize to 0.3.0

justbuchanan:
The new version changes the way the `--diffbase` option works.  This should fix most of the formatting conflicts we see on pull requests.  See https://github.com/justbuchanan/stylize/pull/19 for more info.

### DIFF
--- a/util/requirements3.txt
+++ b/util/requirements3.txt
@@ -6,4 +6,4 @@ munkres	# provides the 'hungarian algorithm', which we use for robot role assign
 pylint #static checker for python
 pyserial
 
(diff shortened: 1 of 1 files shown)


jgkamat:
I was originally thinking that this should be done in our shell scripts, but this is a good solution too.

I'll test it out and if it's good, can I go ahead and merge it?

justbuchanan:
Yeah go for it.  It wouldn't hurt to test it out and make sure I didn't overlook anything, but it passes all the unit tests :).

jgkamat:
LGTM, seems to work with everything I threw at it. Merging!

:smile:

PR #777, opened 2016-07-14
Fixes most current issues with goalie

### LINKED ISSUE #761: When using external ref, goalie id should be pulled from ref packet

### LINKED ISSUE #765: soccer's --noref flag should take effect immediately, not after gui, etc has launched

### LINKED ISSUE #772: Make soccer continue to run if goalie loses vision or is deselected

chachmu:
Goalie and external ref revamp, goalie now set by external ref, Goalie id selector in soccer is greyed out if  an external ref is connected, external ref flag is passed fully through to newRefereeModule, unspecified goalie id in python is stored as -1 instead of None, Plays now continue to run without a goalie. 

Better solution that replaces #774. Closes  #772, Closes #761, Closes #765.

Todo:

- [x] Default to using external ref in -sim

- [x] Reenable goalie id selector if external ref is passing invalid id

- [x] Double check the logic in passing both the external ref flag and the selected  goalie id

- [x] Make pretty


jgkamat:
I'm going to look at this once comp2016 is merged (and test it for real) but it looks good so far.

chachmu:
I went ahead and traced through the logic for passing both the goalie id and the external ref flag from start to finish one more time to make sure it made sense. Make pretty seems to have picked up some changes in processor.cpp and root_play.py which might mess with the merge but everything else looks good.

chachmu:
I think this PR also fixes issue #765. I will need to check to see if its actually sending packets back but as of now it switches the ui over to external ref mode upon starting instead of waiting so I think its fixed.

chachmu:
It was still responding to refbox signals but I fixed it now.

jgkamat:
Can you change the base of this branch if you can (if not, open another PR to master).

jgkamat:
ah wait theres an issue with this branch (I changed comp2016 retroactively to try to save space). I'm going to move your changes over to a new branch (and open a PR on that.) Closing this now (as merging it would be really bad in the current state). I'll link the new one when it's ready.

jgkamat:
@chachmu New pr is at  #783. Can you verify if all the commits made it safely there (and test it out again?). 

If you are 100% sure the migration is fine, delete this branch. We can delete it later though, it should be fine.

chachmu:
I can't check right now, but I will try to do so by tomorrow night.

PR #779, opened 2016-07-14
Increase size of Name tab for Config Widget

### LINKED ISSUE #760: Config widget in soccer should have a wider first column
The default size is too small, so it has to be dragged wider to see the config variable's name.

chachmu:
Make minimum size of Name tab 175. Closes #760.


jgkamat:
LGTM, merging...

PR #783, opened 2016-08-07
Fixes most current issues with goalie

### LINKED ISSUE #761: When using external ref, goalie id should be pulled from ref packet

### LINKED ISSUE #765: soccer's --noref flag should take effect immediately, not after gui, etc has launched

### LINKED ISSUE #772: Make soccer continue to run if goalie loses vision or is deselected

jgkamat:
Continuation of #777. See there for more info.

Better solution that replaces #774. Closes #772, Closes #761, Closes #765.

### DIFF
--- a/soccer/MainWindow.cpp
+++ b/soccer/MainWindow.cpp
@@ -396,7 +396,7 @@ void MainWindow::updateViews() {
             QString::fromStdString(currentFrame->behavior_tree()));
     }
 
-    if (RJ::timestamp() - _processor->status().lastRefereeTime > 100000 and
+    if (RJ::timestamp() - _processor->status().lastRefereeTime > 100000 &&
         !_autoExternalReferee) {
         _ui.fastHalt->setEnabled(true);
         _ui.fastStop->setEnabled(true);
@@ -454,10 +454,10 @@ void MainWindow::updateViews() {
         _processor->refereeModule()->useExternalReferee());
 
     std::vector<int> validIds = _processor->state()->ourValidIds();
-    if (_autoExternalReferee and
+    if (_autoExternalReferee &&
         std::find(validIds.begin(), validIds.end(),
                   _processor->state()->gameState.getGoalieId()) !=
-            validIds.end() and
+            validIds.end() &&
         RJ::timestamp() - _processor->status().lastRefereeTime < 500 * 1000) {
         // External Ref is connected and transmitting a valid goalie id
         _ui.goalieID->setEnabled(false);
--- a/soccer/MainWindow.cpp
+++ b/soccer/MainWindow.cpp
@@ -396,9 +396,7 @@ void MainWindow::updateViews() {
             QString::fromStdString(currentFrame->behavior_tree()));
     }
 
-    if (std::time(nullptr) -
-            (_processor->refereeModule()->received_time / 1000000) >
-        1) {
+    if (RJ::timestamp() - _processor->status().lastRefereeTime > 100000 and !_autoExternalReferee){
         _ui.fastHalt->setEnabled(true);
         _ui.fastStop->setEnabled(true);
         _ui.fastReady->setEnabled(true);
--- a/soccer/MainWindow.cpp
+++ b/soccer/MainWindow.cpp
@@ -396,7 +396,8 @@ void MainWindow::updateViews() {
             QString::fromStdString(currentFrame->behavior_tree()));
     }
 
-    if (RJ::timestamp() - _processor->status().lastRefereeTime > 100000 and !_autoExternalReferee){
+    if (RJ::timestamp() - _processor->status().lastRefereeTime > 100000 and
+        !_autoExternalReferee) {
         _ui.fastHalt->setEnabled(true);
         _ui.fastStop->setEnabled(true);
         _ui.fastReady->setEnabled(true);
--- a/soccer/MainWindow.cpp
+++ b/soccer/MainWindow.cpp
@@ -454,8 +454,9 @@ void MainWindow::updateViews() {
     _ui.actionUse_External_Referee->setChecked(
         _processor->refereeModule()->useExternalReferee());
 
-    if(_autoExternalReferee and RJ::timestamp() - _processor->status().lastRefereeTime < 500 * 1000){
-        //Actually using external ref
+    std::vector<int> validIds = _processor->state()->ourValidIds();
+    if(_autoExternalReferee and std::find(validIds.begin(), validIds.end(), _processor->state()->gameState.getGoalieId())!=validIds.end() and RJ::timestamp() - _processor->status().lastRefereeTime < 500 * 1000){
+        //External Ref is connected and transmitting a valid goalie id
         _ui.goalieID->setEnabled(false);
         //this function changes the index which is 1 higher than the actual id
         if(_ui.goalieID->currentIndex() != _processor->state()->gameState.getGoalieId()+1)
--- a/soccer/MainWindow.cpp
+++ b/soccer/MainWindow.cpp
@@ -177,6 +177,9 @@ void MainWindow::processor(Processor* value) {
                                          _processor->logger().filename());
         _ui.actionStart_Logging->setEnabled(false);
     }
+
+    _autoExternalReferee = _processor->externalReferee();
+    _ui.goalieID->setEnabled(not _autoExternalReferee);
 }
 
 void MainWindow::logFileChanged() {
@@ -450,6 +453,17 @@ void MainWindow::updateViews() {
 
     _ui.actionUse_External_Referee->setChecked(
         _processor->refereeModule()->useExternalReferee());
+
+    if(_autoExternalReferee and RJ::timestamp() - _processor->status().lastRefereeTime < 500 * 1000){
+        //Actually using external ref
+        _ui.goalieID->setEnabled(false);
+        //this function changes the index which is 1 higher than the actual id
+        if(_ui.goalieID->currentIndex() != _processor->state()->gameState.getGoalieId()+1)
+        _ui.goalieID->setCurrentIndex(_processor->state()->gameState.getGoalieId()+1);
+    }
+    else{
+        _ui.goalieID->setEnabled(true);
+    }
 
     // update robot status list
     for (const OurRobot* robot : _processor->state()->self) {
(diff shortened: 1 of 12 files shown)


chachmu:
@jgkamat I checked over this and everything looks correct. It also compiles with all of the new changes. The only thing I did notice is that even though I just recloned the repository, running git status reports that an untracked file util/clang has been added.

jgkamat:
You can clean that file with `git clean -f` or by deleting it manually.

Nice! I'll look over this in a bit.

soccer/Processor.hpp
```diff
@@ -99,7 +99,10 @@ class Processor : public QThread {
     bool joystickValid();
     JoystickControlValues getJoystickControlValues();
 
-    void externalReferee(bool value) { _externalReferee = value; }
+    void externalReferee(bool value) {
+        _externalReferee = value;
```
jgkamat: `_externalReferee` dosen't seem used anywhere except the construtor of this class, maybe it could be removed (and anywhere it is needed could reference _refereeModule? That way we don't have multiple references to the same value.

soccer/gameplay/role_assignment.py
```diff
@@ -181,11 +181,13 @@ def assign_roles(robots, role_reqs):
     # To do our calculations though, we need to have a flat list of RoleRequirements to match to our list of robots
     # The flatten_tree() method does this and keeps track of how to rebuild the tree so we can do so at the end
     def flatten_tree(tree, path_prefix=[]):
+        valid_shell_ids = [bot.shell_id() for bot in robots]
         for key, subtree in tree.items():
             if isinstance(subtree, dict):
                 flatten_tree(subtree, path_prefix + [key])
             elif isinstance(subtree, RoleRequirements):
-                role_reqs_list.append(subtree)
+                if subtree.required_shell_id == None or subtree.required_shell_id in valid_shell_ids:
```
jgkamat: Kind of a nitpick but this should be `is None`. `== None` can be overrided by the thing calling it, but it's good python style to use is None unless you actually want to check for equality, rather than if an object is None. 

This is the opposite of java unfortunately, for example `a == b` -> `a is b` and `a.equals(b)` -> `a == b`.

soccer/gameplay/root_play.py
```diff
@@ -80,19 +80,22 @@ def execute_running(self):
             if self.play != None:
                 if self.play.__class__ not in map(lambda tup: tup[0],
                                                   enabled_plays_and_scores):
-                    logging.info("Current play '" + self.play.__class__.
-                                 __name__ + "' no longer enabled, aborting")
+                    logging.info("Current play '" +
```
jgkamat: While I prefer these changes, they might get overridden by checkstyle later, if it leaves them, thats great!

soccer/gameplay/root_play.py
```diff
@@ -192,7 +195,7 @@ def goalie_id(self):
 
     @goalie_id.setter
     def goalie_id(self, value):
-        self._goalie_id = None if value == -1 else value
+        self._goalie_id = value
```
jgkamat: Is there a reason these checks were removed? I can't find any use of this method, so I'm not sure if it matters or not.

jgkamat:
Overall, looks very nice, and it seems to get what we want! Nice job! :tada:

I want to try to make sure this dosen't break any edge cases, but I don't think it does from a first inspection.

chachmu:
@jgkamat I'm in France with my family right now so I won't be able to do any work on this until I get to GTL on Monday but I should be able to make the changes you mentioned pretty quickly once I'm there.

jgkamat:
no rush, this isn't too important right now (since we'll be focusing most of our effort on training first semester).

Thanks for checking in and letting me know though!

soccer/gameplay/role_assignment.py
chachmu: @jgkamat Hmm... This is actually a little confusing. The way the system was originally set up, the c++ side used None if the goalie wasn't set and the python side used -1. I switched the python side to match the c++ side so that even if the external ref sends a weird value like -1 (which it can technically send) it won't break anything. As such, the value of required_shell_id is actually set to None while the value should technically exist. Does this mean I should use == since I am checking if the int actually is set to None and not that it doesn't exist?

soccer/gameplay/root_play.py
chachmu: @jgkamat I believe this method gets used in the c++ side through the wrapper which makes it pretty hard to track. As for the reason I removed that check, it is because the c++ side now uses None instead of -1 to track if the goalie value is set, see my previous comment for more information.

chachmu: @jgkamat I think this function gets called through the c++ wrapper which makes it pretty hard to track. Previously, the c++ side represented the goalie id not being set with -1 while the python side used None. I have changed it so that the python side matches the c++ side which was necessary, I can't remember the exact reason but I think it made it possible for the role assignment system to handle the goalie going out of vision or something. Anyway, long story short the check is no longer necessary because python just mimics exactly what the c++ side sends it.

soccer/gameplay/role_assignment.py
chachmu: @jgkamat I will go ahead and change this. One thing I wanted to point out is that I noticed that the checks in the [setup_goalie_if_needed](https://github.com/RoboJackets/robocup-software/blob/209e1105b98de70713596a2f9a0054c0497bda8f/soccer/gameplay/root_play.py#L199) function of root_play use `== None`, I'm not sure if that really matters or not though.

jgkamat: since you are checking int's it dosen't matter. The difference between `==` and `is` matters only when comparing objects that have equals overrided, so if required_shell_id somehow became an object instead of an int, then it might matter. This is mostly for future cases (if we mess it up now, it can bite us in the future if we use it incorrectly).

soccer/gameplay/root_play.py
jgkamat: :+1: this seems good

jgkamat:
:+1: LGTM! I'll give others a chance to review for a couple of days before merging

soccer/gameplay/root_play.py
justbuchanan: I think the C++ side still uses -1 to indicate that no goalie is selected ("None" from the dropdown in the gui).  See [MainWindow.cpp](https://github.com/robojackets/robocup-software/blob/3ed28930ceb81f271784b00eb55f36c87e28d7d9/soccer/MainWindow.cpp#L1167)  and [GameplayModule.cpp](https://github.com/robojackets/robocup-software/blob/master/soccer/gameplay/GameplayModule.cpp#L265).

Doing a search for `goalie_id` in the python code shows several places where `None` is handled separately from normal values.

jgkamat: Whoops, I think my grep tool missed them for some reason
[one](https://github.com/RoboJackets/robocup-software/blob/3ed28930ceb81f271784b00eb55f36c87e28d7d9/soccer/gameplay/tactics/defense.py#L112) [two](https://github.com/RoboJackets/robocup-software/blob/master/soccer/gameplay/tactics/positions/defender.py#L199).

@chachmu can you restore the previous setter, in case there is an edge case somewhere?

chachmu: @justbuchanan @jgkamat will take another look later today to figure out whats going on with this.

chachmu: @justbuchanan Having looked over the code again, yes the c++ side still uses -1 which is intentional. The issue is that I forgot to go through the python side to change all of the checks to use -1 so I will do that now.

chachmu: @justbuchanan @jgkamat It may take me a little bit longer to remember how all this works, its been a month since I did this and the internet here is having some issues. I will do what I can but I might not be able to get back to you with 100% confidence until sometime on Monday.

jgkamat: Its fine, there's no rush, better to get this right than merge it quickly =)

I'll look over this issue more in detail later as well.

chachmu: @justbuchanan I went through and checked over this, I had forgotten to check for other places that goalie_id might have been using None instead of -1. I have fixed these.

@jgkamat If you are referring to the goalie_id setter I definitely remember that it was necessary to remove that, although I am not 100% certain of the reason now. Having looked through the code again and this time checked for places that goalie_id might be expected to be None, I believe I have taken care of all of the issues that might come up from changing the setter.

jgkamat: If you want to change the behavior I'm going to do another look over to try and find more places it could be used. I would rather we kept the old behavior though, I'll try and see why it needed to be changed.

chachmu: @jgkamat I will be travelling this weekend so I will have plenty of time on the train to look it over again. It would be nice if you would double check that everything runs properly though because my computer is pretty slow so it's hard to tell if the process is hanging due to the changes I made or just lag.

justbuchanan: C++ uses -1 to indicate no goalie is selected because it's stored as an int and -1 is the best "invalid" value to use.  Python has the advantage of having more flexible types so it may be nicer to just keep using `None` in python rather than -1.  Either way will work though.

soccer/MainWindow.cpp
```diff
@@ -451,6 +453,23 @@ void MainWindow::updateViews() {
     _ui.actionUse_External_Referee->setChecked(
         _processor->refereeModule()->useExternalReferee());
 
+    std::vector<int> validIds = _processor->state()->ourValidIds();
+    if (_autoExternalReferee and
```
ashaw596: Are we switching to use and for everything instead of &&?

jgkamat: seems people wanted to stick to a convention, so can you switch this to &&?

chachmu: Whoops, I was still in python mode. I will change this.

chachmu:
Ok, I will try to look over this one more time tomorrow just to be sure since its pretty important that the goalie runs properly.

ashaw596:
Uhh, this grays out all the play and stop etc buttons on my computer. Anyone else getting that?

chachmu:
@ashaw596 The controls are greyed out when autoref is on. This is what I was referring to when I said that it might be easier to default to autoref off when starting soccer for convenience.

jgkamat:
Closing in favor of #852

Once that's merged, this branch can probably be removed.

Issue #782, opened 2016-08-01
CircleCI Style command no longer works


jgkamat:
This is because the curl url is now being redirected, we need to change the displayed `curl` command to actually resolve properly.

jgkamat:
changing `curl` to `wget -qO -` fixes this. I think everyone has wget installed, so we can probably just swap it out (unless someone figures out why curl isn't working with the redirect).

jgkamat:
On another note, the status tags might become impossible if circleci stops the redirects. We might want to look into a way of getting these URL's programatically.

jgkamat:
Style seems to be somewhat broken in general now too (checking more files than it needs to?)

I'll look into that as well.

justbuchanan:
Hmm, that's odd.  Do you think it's related to the new version of stylize that changed the `--diffbase` option?  I can look at it when I'm back at my computer later.

jgkamat:
I messed around with it and your diffbase seems to work (on my machine, in this scenario). I think that this is an option with the setup,

Don't look into it, I'll let you know if its an issue with stylize.

jgkamat:
@justbuchanan that particular issue was that I merged an old version of master, everything was working correctly on your end :smile: 

I also looked at the `merge-base` docs, and it's EXACTLY what we want! =)

PR #784, opened 2016-08-07
Make curl to follow redirects

jgkamat:
Style was not working, this should tell curl to follow the new circleci redirect to the proper file.

#782


jgkamat:
Looking good, ping me if you see further issues!

PR #785, opened 2016-08-07
Fix 14.04 Compatibility (and add 14.04 CI)

jgkamat:
Add 14.04 compatiblity and add CI for testing it!

This is important to some people. The main issue is making sure gcc-4.9 is used instead of 4.8.


jgkamat:
semaphoreci is the new 14.04 status token. I don't think theres a way to make that more clear :cry:

PR #789, opened 2016-08-29
Fixed up Arch build.

### LINKED ISSUE #787: CMake adding both -std=gnu++11 and std=c++14 when generating build.ninja.
(This is only happening on my Arch system as far as I can tell.)

Building soccer was failing, complaining about std::make_unique calls changed in #306. Since these were added in c++14, it appeared that on my machine the build was using c++11.

I grepped through the repo and only found references in the build/build.ninja (not a source controlled file) generated by cmake which was specifying both -std=c++14 and -std=gnu++11:

`FLAGS = -std=c++14 -Werror=return-local-addr -g -Werror=return-type -Werror=delete-non-virtual-dtor -Werror=uninitialized -fPIC   -fPIC -std=gnu++11`

This doesn't happen in Ubuntu. I figured it might have been related to the CMake/Ninja version I was using (3.6.1 instead of 3.5.1 and 1.7.1 vs 1.5.1), however when manually installed 3.6.1/1.7.1 in Ubuntu, cleaned, and built, it worked fine and didn't add the extra -std=gnu++11 to the build.ninja file.

I fixed the issue on my arch machine by removing the CMAKE_CXX_FLAGS and replacing it by setting the CMAKE_CXX_STANDARD variable to 14 instead. I'm not sure if it's a more correct way to do it but it seems to work in both Ubuntu and Arch. The changes are in the fixed_up_build_arch branch right now.

jpfeltracco:
Added a missing sdl2 package and changed our method of specifying C++14.

I grabbed the CMakeLists snippet from [this website](https://crascit.com/2015/03/28/enabling-cxx11-in-cmake) which describes a few different approaches if you want to check those out.

closes  #787

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,7 +1,6 @@
 cmake_minimum_required(VERSION 3.0.0)
 project("GT RoboJackets RoboCup")
 
-
 # include cmake files in the 'cmake folder'
 set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_SOURCE_DIR}/cmake")
 
@@ -14,7 +13,8 @@ enable_testing()
 include(SetupGTest)
 
 # C++ version
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++14")
+set(CMAKE_CXX_STANDARD 14)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
 
 # Because we use ninja, we have to explicitly turn on color output for the compiler
 if ("${CMAKE_CXX_COMPILER_ID}" MATCHES "Clang")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,8 +13,16 @@ enable_testing()
 include(SetupGTest)
 
 # C++ version
-set(CMAKE_CXX_STANDARD 14)
-set(CMAKE_CXX_STANDARD_REQUIRED ON)
+if ("${CMAKE_VERSION}" LESS "3.1")
+    # This is to support the CMake version backported to Ubuntu 14.04
+    message("CMAKE_VERSION < 3.1, specifying c++14 flag.")
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++14")
+else()
+    # This is the more modern way of setting our c++ version
+    message("CMAKE_VERSION >= 3.1, setting CMAKE_CXX_STANDARD to 14.")
+    set(CMAKE_CXX_STANDARD 14)
+    set(CMAKE_CXX_STANDARD_REQUIRED ON) # Don't fall back to older versions
+endif()
 
 # Because we use ninja, we have to explicitly turn on color output for the compiler
 if ("${CMAKE_CXX_COMPILER_ID}" MATCHES "Clang")
--- a/util/arch-packages.txt
+++ b/util/arch-packages.txt
@@ -49,6 +49,7 @@ libspnav
 
 # joystick library
 sdl
+sdl2
 
 # debugger for MBED code
 arm-none-eabi-gdb


jgkamat:
This seems to break the 14.04 build (but it's fine on 16.04). Do you know if this was added in a recent version of cmake?

jgkamat:
CMAKE_CXX_STANDARD was added in cmake 3.1, but we are using 3.0.2 on ubuntu 14. I'm not sure whether to upgrade that version or try a different approach (since this is very clean)

jpfeltracco:
It's a little ugly but maybe we could just use an if statement on the CMake version and use the new way for CMake >= 3.1 and the old way for CMake < 3.1?

jgkamat:
That sounds like a good solution too. If statements are definitely there in the version of cmake currently on ubuntu14 (which we backported). If you could figure out how to do that it would be great!
https://cmake.org/cmake/help/v3.0/command/if.html
This is the docs for if statements, although I'm not sure how to compare the version numbers. Let me know if its really hard to compare version numbers or something...

jpfeltracco:
Okay, just added an attempt at using if statements. It seems to work fine in the testing I did, but we need to check that 14.04 works now.

jgkamat:
Yup, it seems to work now (semaphore ci is the CI system for 14.04). I'll review this tomorrow during the meeting. Thanks for the quick response!

jpfeltracco:
Sounds good!

jgkamat:
This LGTM, although, the [firmware cmakelists](https://github.com/RoboJackets/robocup-firmware/blob/master/CMakeLists.txt#L17) still has this problem.  I could move it over easily, but since you did the work I'd rather you get the contribution :P.

As for this PR, I'll merge it at the meeting today. Nice job!

jpfeltracco:
Cool, I'll fix that over in firmware as well.

We should probably discuss how we approach issues that inherently touch both repos, because as we continue our firmware changes we'll probably run into several of these situations and should have a standard way of handling it.

Duplicating issues doesn't seem like a great solution, it feels like having a single issue is cleaner, but that would mean putting the issue only in one repo which also feels wrong. Do you know of any good way of handling this? (We can discuss at the meeting tonight too)

jgkamat:
We should definetly discuss this at the meeting tonight

For most things, we won't have this probelm (since the software cmake and the firmware cmake are very different, even the root one). However, if we had made a mistake in a file that was duplicated (any file in the root of the repo) we'll need to change this twice. This issue is a perfect example of this. I think that this is the perfect way to handle such an issue, making a single issue/pr first (and mentioning it applies to firmware) and porting it over once it's reviewed properly.

This problem is solved for the shared code files by `robocup-common` which has all the radio stuff that needs to be shared as of now. I would be happy to take suggestions though (the more ideas we have, the better), but overall I think this is an issue that won't come up too often once we iron things out.

CMakeLists.txt
```diff
@@ -14,7 +13,16 @@ enable_testing()
 include(SetupGTest)
 
 # C++ version
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++14")
+if ("${CMAKE_VERSION}" LESS "3.1")
```
audiofanatic: You want to test the version like so:

```
if (CMAKE_VERSION VERSION_LESS 3.1)
```

Using `LESS` is a numeric test, but `CMAKE_VERSION` isn't generally going to be a real number since it would have three components, i.e. have the form `x.y.z`. The `VERSION_LESS` form is specifically meant for situations like this one where you need to test a version number and either side of the test could be one of `x`, `x.y` or `x.y.z` or even `x.y.z.t` forms. Note also that some of the version components can have values of 10 or more which would also not lead to correct behaviour with a simple `LESS` comparison. Have a read of the docs related to these things for full details:

https://cmake.org/cmake/help/latest/variable/CMAKE_VERSION.html

https://cmake.org/cmake/help/latest/command/if.html (search for VERSION_LESS)

jgkamat: Yup, that seems correct, I didn't realize this at the time! Thanks! That probably would have hit us in the future.

We probably didn't see this so far because we've only been looking for versions that are older. I'll fix this in a bit!

Thanks for taking a look at this! :smile:

jgkamat: We want to still support 14.04 until it's EOL (per @barulicm), which I think is a good idea. There are still a lot of 14.04 machines that might want to run our software :P.

jpfeltracco: :+1:

barulicm: Supporting 14.04 is a good idea for atleast another year. You are allowed to require newer versions of CMake if you want to. 
I would like to point out that, while I love the new CMake standard flags as much as anyone, you have managed to turn one "not as pretty as it could be" line into 10 "more complicated than it needs to be" lines.
I'd recommend either just using the old style cxx_flags version or setting your minimum cmake version higher.

Issue #674, opened 2016-05-20
Segfault in PyQt


ashaw596:
I get this seg fault randomly occasionally when running Basic122.
It takes a while. Not really sure whats causing it.

```
* thread #11: tid = 0x3d139, 0x0000000101b39148 QtCore`QObject::objectName() const + 8, name = 'QThread', stop reason = EXC_BAD_ACCESS (code=EXC_I386_GPFLT)
    frame #0: 0x0000000101b39148 QtCore`QObject::objectName() const + 8
QtCore`QObject::objectName:
->  0x101b39148 <+8>:  movq   0x30(%rax), %rax
    0x101b3914c <+12>: testq  %rax, %rax
    0x101b3914f <+15>: je     0x101b39166               ; <+38>
    0x101b39151 <+17>: movq   0x28(%rax), %rax
(lldb) bt
* thread #11: tid = 0x3d139, 0x0000000101b39148 QtCore`QObject::objectName() const + 8, name = 'QThread', stop reason = EXC_BAD_ACCESS (code=EXC_I386_GPFLT)
  * frame #0: 0x0000000101b39148 QtCore`QObject::objectName() const + 8
    frame #1: 0x0000000109f4ac71 QtCore.so`qtcore_do_find_child(QObject const*, _object*, QString const&, QFlags<Qt::FindChildOption>) + 161
    frame #2: 0x0000000109f4ad4c QtCore.so`qtcore_do_find_child(QObject const*, _object*, QString const&, QFlags<Qt::FindChildOption>) + 380
    frame #3: 0x0000000109f48967 QtCore.so`meth_QObject_findChild(_object*, _object*, _object*) + 343
    frame #4: 0x000000010269f900 Python`PyCFunction_Call + 60
    frame #5: 0x000000010270326b Python`PyEval_EvalFrameEx + 22399
    frame #6: 0x0000000102706faf Python`fast_function + 213
    frame #7: 0x0000000102703100 Python`PyEval_EvalFrameEx + 22036
    frame #8: 0x0000000102706faf Python`fast_function + 213
    frame #9: 0x0000000102703100 Python`PyEval_EvalFrameEx + 22036
    frame #10: 0x0000000102706faf Python`fast_function + 213
    frame #11: 0x0000000102703100 Python`PyEval_EvalFrameEx + 22036
    frame #12: 0x00000001027066f0 Python`_PyEval_EvalCodeWithName + 1884
[...]

PR #791, opened 2016-09-12
Save Defense Checkbox to not use FindChild every frame

ashaw596:
Save Defense Checkbox to not use FindChild every frame

### DIFF
--- a/soccer/gameplay/GameplayModule.cpp
+++ b/soccer/gameplay/GameplayModule.cpp
@@ -341,32 +341,27 @@ void Gameplay::GameplayModule::run() {
     {
         try {
             // vector of shared pointers to pass to python
-            vector<OurRobot*>* botVector = new vector<OurRobot*>();
-            for (auto itr = _playRobots.begin(); itr != _playRobots.end();
-                 itr++) {
-                OurRobot* ourBot = *itr;
+            vector<OurRobot*> botVector;
+            for (auto  ourBot : _playRobots) {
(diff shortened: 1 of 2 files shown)


jgkamat:
Related to #674. May fix one of the (multiple) segfaults there.

Looks good at a first glance, I'm going to do another pass later on.

@jpfeltracco

jpfeltracco:
I ran a couple of times on this branch and I haven't seen the find_child segfault and running out of memory didn't cause a segfault either so I'm cautiously optimistic about this.

I did get a new segfault which I posted to the original issue for now.

jgkamat:
LGTM, @ashaw596, go ahead and merge when ready.

soccer/gameplay/GameplayModule.cpp
```diff
@@ -341,32 +341,27 @@ void Gameplay::GameplayModule::run() {
...
                 if (bot && bot->visible) {
-                    theirBotVector->push_back(bot);
+                    theirBotVector.push_back(bot);
                 }
             }
             getMainModule().attr("set_their_robots")(theirBotVector);
 
             getMainModule().attr("set_game_state")(_state->gameState);
 
-            getMainModule().attr("set_system_state")(&_state);
```
jgkamat: Do you have any idea why this was a address before?

ashaw596: Really not sure since _state is an pointer already. sooooo....

PR #792, opened 2016-09-12
Possibly fix thread unsafe copying issue

ashaw596:
I think this should fix the RadioRx copying segfault.


jgkamat:
LGTM, I've never reproduced this particular segfault, but it seems others are (and it seems better to have the mutex before we read from radio anyway).

PR #793, opened 2016-09-14
Make ubuntu-setup set proper permissions on external

jgkamat:
There are some issues with the existing ubuntu-setup script that can cause your external to become owned by root. This patch should fix that. Just re-run ubuntu-setup to fix any issues you might have!


jgkamat:
Let me know if you run into issues ASAP.

PR #800, opened 2016-09-28
Only update behavior tree if it changes

### LINKED ISSUE #798: Can't copy debug behavior tree text.
It would be nice to be able to select and copy text from the debug behavior tree in soccer.

jgkamat:
There might be a better way to do this, but this seems the cleanest and
the simplest right now. Not too much complexity here...

@jpfeltracco could you give this a quick look over?

Fixes #798


jgkamat:
I'm an idiot ignore this

PR #808, opened 2016-10-24
Fix Joystick Control on Logitech Gamepads

### LINKED ISSUE #807: Fix Joystick Control on Soccer
Sometime during the merge I screwed up the joystick control drivers.

Last known good version is 1055fac355

jgkamat:
Closes #807

Still needs testing, but this seemed to work a few weeks ago.


jgkamat:
I've tested this (at least) on the logitech F310 and it seems to work. If we need to, we'll continue work on this before Rockwell.

PR #810, opened 2016-11-01
Arch linux: easily change which AUR helper to use

justbuchanan:
This allows you to set the aur helper via the AUR_HELPER environment variable.  For example, if you wanted to use pacaur, you could do `AUR_HELPER=pacaur util/arch-setup`.

### DIFF
--- a/util/arch-setup
+++ b/util/arch-setup
@@ -4,6 +4,8 @@
 # specified in the environment if available, otherwise fallback to a default. To
 # run arch-setup with a different aur helper, do:
 #   AUR_HELPER=myprogram ./util/arch-setup
+# note that this only works with 'pacman-like' aur helpers.  See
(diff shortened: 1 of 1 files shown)


jgkamat:
This would work for AUR_HELPER=yaourt, but other helpers may not use the same syntax as yaourt. 
see: https://wiki.archlinux.org/index.php/AUR_helpers#Comparison_table

Some of them, like yaourt are 'pacman like', but most are 'specific'.

I don't think there's a good way to get around that, so I think this will be good enough for now (since most people use yaourt or pacaur). :+1:

justbuchanan:
Good point, I just updated the comment to say that.

jgkamat:
I'll merge this since this is small and useful, and no one has said anything for a while :smile:

PR #811, opened 2016-11-02
Albert/fix Tons of UI Changes and Refactoring

ashaw596:
- Added DebugThrowIf function to Common submodule
- Add default playbook to run-sim which has basic123 and repeatedlineup
- Major Refactor of MainWindow.hpp working and ui. Changed it to not manually have radio menus and allow setting values from the ui file.
- Set default of Goalie to 0 instead of not set
- Change UI to add Frame numbers, pause button now plays and pauses
- Major Refactor of Logger.hpp. Changed to user boost::circular_buffer and make it actually work.
- Fixed slider so it actually works
- Added a ton of shortcuts to control soccer
q-quicksave(already existed)
e-quickload(already existed)
p-play
s-stop
h-halt
space-pause/play
left arrow - previous frame
right arrow - next frame
shift/left arrow - reverse speed increase
shift/right arrow - play faster

### DIFF
--- a/soccer/MainWindow.cpp
+++ b/soccer/MainWindow.cpp
@@ -46,14 +46,15 @@ void calcMinimumWidth(QWidget* widget, QString text) {
     widget->setMinimumWidth(rect.width());
 }
 
-MainWindow::MainWindow(Processor* processor, QWidget* parent) :
-        QMainWindow(parent), _updateCount(0),
-        _autoExternalReferee(true), _doubleFrameNumber(-1),
-        _lastUpdateTime(RJ::timestamp()), _history(2 * 60),
-        _processor(processor) {
+MainWindow::MainWindow(Processor* processor, QWidget* parent)
+    : QMainWindow(parent),
+      _updateCount(0),
+      _autoExternalReferee(true),
+      _doubleFrameNumber(-1),
+      _lastUpdateTime(RJ::timestamp()),
+      _history(2 * 60),
+      _processor(processor) {
     qRegisterMetaType<QVector<int>>("QVector<int>");
-
-
 
     _ui.setupUi(this);
     _ui.fieldView->history(&_history);
@@ -123,7 +124,6 @@ MainWindow::MainWindow(Processor* processor, QWidget* parent) :
     rotateGroup->addAction(_ui.action270);
     qActionGroups.push_back(rotateGroup);
 
-
     auto visionChannelGroup = new QActionGroup(this);
     visionChannelGroup->addAction(_ui.actionVisionPrimary_Half);
     visionChannelGroup->addAction(_ui.actionVisionSecondary_Half);
@@ -138,7 +138,7 @@ MainWindow::MainWindow(Processor* processor, QWidget* parent) :
     connect(_ui.manualID, SIGNAL(currentIndexChanged(int)), this,
             SLOT(on_manualID_currentIndexChanged(int)));
 
-//    channel(0);
+    //    channel(0);
 
     updateTimer.setSingleShot(true);
     connect(&updateTimer, SIGNAL(timeout()), SLOT(updateViews()));
@@ -164,14 +164,12 @@ void MainWindow::configuration(Configuration* config) {
 }
 
 void MainWindow::initialize() {
-
     // Team
     if (_processor->blueTeam()) {
         _ui.actionTeamBlue->trigger();
     } else {
         _ui.actionTeamYellow->trigger();
     }
-
 
     if (_processor->logger().recording()) {
         _ui.actionStart_Logging->setText(QString("Already Logging to: ") +
@@ -181,7 +179,7 @@ void MainWindow::initialize() {
 
     // Initialize to ui defaults
     on_goalieID_currentIndexChanged(_ui.goalieID->currentIndex());
-    for (const auto &qActionGroup : qActionGroups) {
+    for (const auto& qActionGroup : qActionGroups) {
         qActionGroup->checkedAction()->trigger();
     }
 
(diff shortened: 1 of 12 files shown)


jgkamat:
Thanks for the PR, I'll look at it later today!  (I'm so happy for shortcuts, we finally have them!)

The setting goalie ID to 0 will work while #783 is being finalized (which is taking much longer than expected), but that PR has a bunch of other nice things (such as not freezing when the goalie goes out of vision), which we want. If you added other shortcuts than the ones made above, could you add them to the above comment or docs? I'll and add the ones above to documentation before this is merged.

soccer/MainWindow.cpp
```diff
@@ -267,52 +262,67 @@ void MainWindow::updateViews() {
...
-                .arg(QString::number(_processor->logger().numFrames()),
-                     QString::number(_processor->logger().maxFrames()),
+                .arg(QString::number(_processor->logger().size()),
+                     QString::number(_processor->logger().capacity()),
                      QString::number((_processor->logger().spaceUsed() + 512) /
                                      1024)));
     }
 
+    auto value = _ui.logHistoryLocation->value();
+    if (static_cast<int>(value) != frameNumber()) {
```
jgkamat: Is this not needed? :confused:

soccer/MainWindow.cpp
```diff
@@ -267,52 +262,67 @@ void MainWindow::updateViews() {
...
+    int liveFrameNumber = _processor->logger().currentFrameNumber();
+    if (live()) {
         _doubleFrameNumber = liveFrameNumber;
     } else {
-        _doubleFrameNumber += _playbackRate;
+        _doubleFrameNumber += *_playbackRate;
 
-        int minFrame = _processor->logger().firstFrameNumber();
-        int maxFrame = _processor->logger().lastFrameNumber();
+        int minFrame = _processor->logger().firstFrameNumber() + 10;
```
jgkamat: Any idea why we need a +10? Magic numbers kill me :cry:

soccer/MainWindow.cpp
```diff
@@ -1082,47 +1089,56 @@ void MainWindow::on_actionSeed_triggered() {
...
-    int minFrame = _processor->logger().firstFrameNumber();
-    int maxFrame = _processor->logger().lastFrameNumber();
-    _doubleFrameNumber = value + minFrame;
-    _doubleFrameNumber = min<double>(maxFrame, _doubleFrameNumber);
-
-    emit historyLocationChanged(_doubleFrameNumber - minFrame);
+    // Sync frameNumber with logHistory slider
+    _doubleFrameNumber = value;
 
     // pause playback
```
jgkamat: This comment should probably be removed.

soccer/MainWindow.cpp
```diff
@@ -1082,47 +1089,56 @@ void MainWindow::on_actionSeed_triggered() {
...
+    on_logHistoryLocation_sliderPressed();
+}
+
 void MainWindow::on_logPlaybackRewind_clicked() {
-    live(false);
-    _playbackRate += -0.5;
+    if (live()) {
+        _playbackRate = -1;
+    } else {
+        *_playbackRate += -0.5;
```
jgkamat: I don't think this should be derefrenced right? (half of `_playbackRate` are dereferenced and half are used w/ the value :confused:)

soccer/VisionReceiver.cpp
```diff
@@ -9,15 +9,18 @@
...
+VisionReceiver::VisionReceiver(bool sim, int port)
+    : simulation(sim), _running(false), port(port) {}
 
 void VisionReceiver::stop() {
-    _running = false;
-    wait();
+    if (isRunning()) {
+        // Handle Case where start has been called, but _running has not been
+        // set to true yet.
+        while (!_running) {
```
jgkamat: I'm not 100% sure why this is needed... if isRunning() return's true won't _running also be true? or are they different variables. I really don't like spinlocks :/

soccer/ui/MainWindow.ui
```diff
@@ -82,6 +82,9 @@
              <height>24</height>
```
jgkamat: The ui jiggles for me when the numbers change, they probably need padding around them or something. I can work on that when I get some free time...

jgkamat: We might also want to leave this as a beginner issue.

ashaw596: I thought I set the font to monospace. Is that not a valid font on arcH?

soccer/MainWindow.cpp
ashaw596: Oh it just gives the processor some buffer to work with since the buffer will keep moving as we display them.

soccer/MainWindow.cpp
ashaw596: Why?

soccer/ui/MainWindow.ui
jgkamat: The font definitely seems to be monospace but it still seems to be jiggling. (And there's also the problem when you go from `99` -> `100`. This shouldn't be too hard to fix though, just needs a padding element (I think).

ashaw596: Hmm Thats weird. It should really only jiggle from changes in number of digits now.

soccer/MainWindow.cpp
ashaw596: oh, the equals method is overwritten for boost option. The += isn't. This is correct.

soccer/MainWindow.cpp
jgkamat: oh scratch this, I didn't understand what was going on here.

soccer/MainWindow.cpp
jgkamat: :cry: I guess there's not too much we can do about that...

soccer/VisionReceiver.cpp
jgkamat: Ah I think I get it now. I don't think theres a cleaner solution w/o mutexes so I guess this is ok...

ashaw596:
oh well. Should we merge it?

ashaw596:
? @jgkamat

jgkamat:
Whoops, I was held back by automation fair. I think this is good to go! :+1: 

Although I would prefer that we fix the jiggling before we merge, I think we can get someone else to do that later.

Issue #300, opened 2015-04-27
Investigate physics of ball and dribbler in simulator


justbuchanan:
Right now, the ball is always kicked straight from the robot, which doesn't match real life in the case of doing a touchpass.  The ball should be kicked at an angle depending on the direction that it comes to the robot at.

cc @jgkamat

jgkamat:
I tried decreasing the power of the receiver, and the kicker kicks at an angle, as expected.
![map18](https://cloud.githubusercontent.com/assets/4349709/7341370/4ca016c4-ec6f-11e4-9671-1d0acfcd336d.png)

Maybe this is only an issue when the kickpower is at max, or maybe when the kick power is at max, it's ratio is so high, it dosent matter. In any case, I'll work on finding the angle the robot should be, as the max kick power will be reduced at comp. It would be really helpful to have a expected kick speed function. Do you know how to get the current/predicted (at a point) speed of the ball?

ashaw596:
I think the predicted speed should be given by the BallFilter probably

justbuchanan:
Awesome!  I just assumed after seeing it the first time that the simulator was setup to always kick straight regardless of the ball's initial speed.

The kick speed at competition is limited to 8m/s, so when they do inspections, we tune a MaxKickPower config variable so that calling `robot.kick(1.0)` will kick slightly under 8m/s.  Ideally we'd have a `kickSpeed()` method where you could specify the desired kick velocity of the ball in m/s, but it doesn't exist yet (see issue #288).

The BallFilter calculates the estimated current position and velocity of the ball, which you can access in python with `main.ball().vel`.  We also have a [function](https://github.com/RoboJackets/robocup-software/blob/master/soccer/gameplay/evaluation/ball.py#L35) for predicting the position of the ball a fixed amount of time in the future given it's current state.  Right now it just predicts a future position, but we could add velocity as well.

I'll go ahead and close this issue for now if it's behaving as expected, but reopen it if you notice anything too weird.

PR #812, opened 2016-11-08
Adds custom failure notifications that don't break randomly

jgkamat:
Slack integrations suck.

This should keep us pacified for the time being. 

example of a trigger:
Myproj - Build #300 Failed on CircleCI!
Link to Failure

I still need to add semaphore triggers too.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -33,7 +33,7 @@ dependencies:
 test:
   override:
     # Also run webhooks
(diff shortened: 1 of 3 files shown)


jgkamat:
This does hardcode our default branch as master, but I doubt that will change anytime soon.

jgkamat:
Haven't tested the semaphore script but it _should_ work. Merging, and I'm going to kill official slack notifications.

Let me know if you notice a faiure w/o a slack notification.

PR #814, opened 2016-11-10
Add Python Version 3.5 to build system search path

guyfleeman:
CMake stops looking at V3.3 if you're on the version that ships with 16.04. The default python3 version is 3.5. I had to explicitly tell CMake to search for 3.5.

Not sure this is the best way to fix this, maybe this is better fixed in the packages list. Thought this would be the best way to share my problem and fix.


jgkamat:
Hmm, it's interesting that this is needed, since I set up python35 to work on ubuntu already here: https://github.com/RoboJackets/robocup-software/blob/cbe3b3665c2880944fa5613add771015e6d10749/soccer/CMakeLists.txt#L89

can you post the error that you're getting on master here? This seems to do the same thing as was done in the soccer cmake, and I'm not sure if one is better/worse than the other, but I'd rather not have both.

guyfleeman:
$ make clean && make run-sim
              
cd build && ninja clean || true
ninja: error: loading 'build.ninja': No such file or directory
rm -rf build
mkdir -p build
cd build && cmake -GNinja -Wno-dev --target  all  .. && ninja  all
-- The C compiler identification is GNU 5.4.0
-- The CXX compiler identification is GNU 5.4.0
-- Check for working C compiler using: Ninja
-- Check for working C compiler using: Ninja -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working CXX compiler using: Ninja
-- Check for working CXX compiler using: Ninja -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
CMAKE_VERSION < 3.1, specifying c++14 flag.
-- Found Qt5: /usr/lib/x86_64-linux-gnu/cmake/Qt5Widgets
-- Looking for include file pthread.h
-- Looking for include file pthread.h - found
-- Looking for pthread_create
-- Looking for pthread_create - not found
-- Looking for pthread_create in pthreads
-- Looking for pthread_create in pthreads - not found
-- Looking for pthread_create in pthread
-- Looking for pthread_create in pthread - found
-- Found Threads: TRUE  
-- Found PROTOBUF: /usr/lib/x86_64-linux-gnu/libprotobuf.so  
-- Found PythonInterp: /usr/bin/python3 (found suitable version "3.5.2", minimum required is "3.2") 
CMake Error at /usr/local/share/cmake-3.0/Modules/FindPackageHandleStandardArgs.cmake:136 (message):
  Could NOT find PythonLibs (missing: PYTHON_LIBRARIES PYTHON_INCLUDE_DIRS)
  (Required is at least version "3.2")
Call Stack (most recent call first):
  /usr/local/share/cmake-3.0/Modules/FindPackageHandleStandardArgs.cmake:343 (_FPHSA_FAILURE_MESSAGE)
  /usr/local/share/cmake-3.0/Modules/FindPythonLibs.cmake:197 (FIND_PACKAGE_HANDLE_STANDARD_ARGS)
  CMakeLists.txt:98 (find_package)


-- Configuring incomplete, errors occurred!
See also "/home/willstuckey/robotics/robocup-software/build/CMakeFiles/CMakeOutput.log".
See also "/home/willstuckey/robotics/robocup-software/build/CMakeFiles/CMakeError.log".
makefile:13: recipe for target 'all' failed
make: *** [all] Error 1

jgkamat:
Hmm, which version of cmake are you running? (`cmake --version`). I think this could be a problem if you are using a python version with an old cmake.

guyfleeman:
cmake --version
cmake version 3.0.2

CMake suite maintained and supported by Kitware (kitware.com/cmake).

guyfleeman:
Should I add a conditional check for older versions of cmake to apply this property?

jgkamat:
(assuming you are on ubuntu 16) I think you have the crappy PPA we set up earlier still in effect. Could you uninstall cmake (and try removing that PPA) and reinstalling it? http://packages.ubuntu.com/xenial/cmake

`sudo apt purge cmake && sudo apt install cmake`, and see if we get 3.5

jgkamat:
This issue was fixed, it was related to something with the old Cmake ppa we were using.

PR #819, opened 2016-11-13
Remove QuanternionDemo files

jgkamat:
This PR fixes a crash on manual select of real robots


jgkamat:
If anyone knows what this is for, let me know. We need to merge this soon...

justbuchanan:
The QuaternionDemo stuff has existed for a long time and hasn't been used at least since I joined RoboCup.  I don't know a lot about it, but it was used to visualize the rotation readings from the on-board IMU.  It was a good first step at actually using the IMU for useful things in motion control, although I don't think it made it much beyond that.

The code for quaternion demo would be good to look at later on when you guys work on using the IMU.

See this wikipedia article for a little more info on quaternions: https://en.wikipedia.org/wiki/Quaternions_and_spatial_rotation.

jgkamat:
I think if it could ever be useful, we should bring it back. We removed it because it was conflicting with the restructure of a few things in QT (just the menu entries, I think), so it shouldn't be too hard to bring back.

PR #820, opened 2016-11-14
Refactor RJ:Time to use std::chrono

ashaw596:
Refactored much of the code to use std::chrono. This forces time type checks and automatic unit conversions between second, microseconds, and timestamps. The old api is accessible using RJ::timestamp(), but the return type is now RJ::Timestamp instead of RJ::Time.
RJ::Timestamp is a int64 representing microseconds since Epoch as RJ:Time use to be.
RJ::Time is now a chrono::time_point of the system_clock. This about the same as the Timestamp, but has type safe checks. There is a convenience RJ::timestamp(RJ::Time) conversion function to convert from Time to Timestamp, but its usage should be sparingly used.

RJ::Seconds is a double representation of Duration. Durations can be added to RJ::Times and the difference between two Times a Duration. The chrono literal namespace has been include so 10s and 15ms will be used for chrono seconds and millisecond units for convenience. Be aware that if you do 1s /10, you will get 0s since the chrono seconds unit is integers. RJ::Seconds(1.0/10) can be used to get a double representation or 1000ms/10= 100ms.

RJ::Seconds::count() will return the double representation of the number of seconds though this should be used sparingly if ever.


jgkamat:
This is quite a large change, but this is definitely a good change. I might not be able to look at it until the break though, but I'll look at this eventually :P

soccer/NewRefereeModule.cpp
```diff
@@ -167,10 +167,10 @@ void NewRefereeModule::run() {
 
         stage = (Stage)packet->wrapper.stage();
         command = (Command)packet->wrapper.command();
-        sent_time = packet->wrapper.packet_timestamp();
+        // sent_time = packet->wrapper.packet_timestamp();
         stage_time_left = packet->wrapper.stage_time_left();
-        command_counter = packet->wrapper.command_counter();
-        command_timestamp = packet->wrapper.command_timestamp();
```
jgkamat: Even if we arent using these comammands anywhere, I think it's better to keep them in case we want to use them for later. The C Compiler should optimize them away unless we need them. Thoughts?

soccer/Processor.cpp
```diff
@@ -311,7 +315,11 @@ void Processor::run() {
             if (packet->wrapper.has_detection()) {
                 SSL_DetectionFrame* det = packet->wrapper.mutable_detection();
 
-                double rt = packet->receivedTime / 1000000.0;
+                double rt =
+                    RJ::numSeconds(packet->receivedTime.time_since_epoch());
+                //                double rt =
```
jgkamat: Probably should be deleted

soccer/modeling/BallFilter.cpp
```diff
@@ -11,44 +11,43 @@ static const float Velocity_Alpha = 0.05;
...
+    auto part = std::exp(-0.2913f * t.count());
+    auto speed = s0 * part;
+    auto distance = s0 * -3.43289f * (part - 1.0f);
+
+    prediction.time = time;
+    prediction.valid = true;
+    prediction.pos = pos + vel.normalized(distance);
+    prediction.vel = vel.normalized(speed);
+
+    // return MotionInstant(pos + vel.normalized(distance),
```
jgkamat: I know you didn't add this, but I think it could be removed..

soccer/planning/Path.hpp
```diff
@@ -164,12 +168,14 @@ class AngleFunctionPath : public Path {
...
+     * @param[in] 	startTimeIntoPath The time on the path to start checking
+     *from
      * @return 		true if it hits an obstacle, otherwise false
      */
-    virtual bool hit(const Geometry2d::ShapeSet& obstacles, float& hitTime,
-                     float startTime) const override {
-        return path->hit(obstacles, hitTime, startTime);
+    virtual bool hit(const Geometry2d::ShapeSet& obstacles,
+                     RJ::Seconds startTimeIntoPath,
+                     RJ::Seconds* hitTime) const override {
```
jgkamat: Is there any reason this can't be a reference instead of a pointer? It would reduce a bunch of null checking (unless we want to pass in null somewhere).

soccer/planning/Path.cpp
```diff
@@ -25,15 +25,15 @@ void Path::draw(SystemState* const state, const QColor& color,
 
     // Get the closest step size to a desired value that is divisible into the
     // duration
-    const float duration = getDuration();
-    const float desiredStep =
-        0.25;  // draw the path by interpolating every x seconds
+    const RJ::Seconds duration = getDuration();
+    const RJ::Seconds desiredStep(
```
jgkamat: Can the comment go above this line so the style checker is happy and dosent split it?

soccer/NewRefereeModule.cpp
ashaw596: I mean they won't optimize them out since its part of an class. I changed it to keep it in, but its sort of unnecessary right now.

soccer/Processor.cpp
ashaw596: deleted.

soccer/modeling/BallFilter.cpp
ashaw596: removed.

soccer/modeling/RobotFilter.cpp
```diff
@@ -18,18 +18,21 @@ void RobotFilter::update(const RobotObservation* obs) {
     }
 
     int s = obs->source;
-    double dtime = (obs->time - _estimate[s].time) / 1000000.0f;
-    bool reset = _estimate[s].time == 0 || (dtime > Coast_Time);
+    RJ::Seconds dtime = (obs->time - _estimate[s].time);
+    // bool reset = _currentEstimate[s].time == 0 || (dtime > Coast_Time);
 
+    bool reset = (dtime > Coast_Time);
+    ;
```
ashaw596: completely necesssary

soccer/planning/Path.hpp
ashaw596: Well its an output variable so you want to give the option to not pass it in. Also some people (google which much of our style happens to be based off) rather encourage the use of pointers for output and only use of const references instead of mutable references.

soccer/planning/Path.cpp
ashaw596: changed.

PR #821, opened 2016-11-15
Fix Ui Framerate issues.

ashaw596:
by some minor refactoring and removing a ridculous _loopmutex lock that was probably blocking for like .2 seconds at a time. It was replaced with a lock around each robot's radiorx.

framerates from 5 to around 40 under heavy load.....


soccer/Processor.cpp
```diff
@@ -362,7 +361,6 @@ void Processor::run() {
         // Read radio reverse packets
         _radio->receive();
 
-        _loopMutex.lock();
```
ashaw596: This mutex lock was ridiculous and I don't see any real reason for it other than the radiopacket.

jgkamat:
Since this PR could break a ton of things, I want to make sure it gets a lot of testing on real robots before we merge. I'll take a look at it once i get free time :cry:.

ashaw596:
Happen to get a chance to look at it?

PR #823, opened 2016-11-23
Tuning 11/23

jgkamat:
Bumped min velocity and acceleration, and re-tuned PID controller.

Robots should move a lot faster, but might be a bit out of control.

### DIFF
--- a/soccer/config/realGermany.xml
+++ b/soccer/config/realGermany.xml
@@ -201,7 +201,7 @@
   <Field_Edge_Obstacle value="0.33"/>
  </PathPlanner>
  <MotionConstraints>
-  <Max_Acceleration value="1.3"/>
+  <Max_Acceleration value="1.5"/>
   <Max_Velocity value="1.5"/>
   <Max_Angle_Speed value="0.436"/>
   <Replan_Threshold value="0.3"/>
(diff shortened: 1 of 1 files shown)


ashaw596:
Um,  I'd recommend not raising min velocity. It really suppose to be the
min velocity that prevents the robot from getting stuck. It should not
really be used to speed up the robots. Pid should really be used to
compensate paths being planned faster than the robots are accelerating and
maxacceleration and maxspeed can Increase to plan faster paths. If it's a
constant replanning issue like grabbing the ball, that should be fixed in
the path planning layer.

jgkamat:
Hmm, raising MaxVel was the only thing that seemed to help move the robots faster, anything else (raising p, or velMultipler) seemed to make the robots go out of control. This configuration kept the robots in control while going faster than before. Also, is MaxVel in m/s? 1m/s might be a little small, since the field has grown quite a bit from before. The current config follows the planned path exactly, its just that our planned paths are a bit slow, considering the size of the double field. This is a bit too much, but I'm working on finding a good balance...

ashaw596:
maxVel is in m/s. Yeah maxVel and maxAccel is what tunes our motion planning. I set it to 1 m/s at comp since our motion control was um not very good.

ashaw596:
It used to be like 2.5

ashaw596:
or more like 1.5-2

ashaw596:
My point was that minEffectiveVelocity only affects motion control(ie robots not moving) and only screws up our motion planning and control when therobots are moving or moving slowly..

jgkamat:
oooh you're talking about minEffective, yah I agree I shouldn't be touching that :P 

I'm going to bring that back down to ~0.01/0.02 ish when I'm done.

jgkamat:
Tuning on the shop computer is way better than this, closing.

PR #833, opened 2016-11-30
Format Team Labels in Soccer

### LINKED ISSUE #824: Team Labels in Soccer shouldn't have brackets.
Currently, team labels in soccer say `<Yellow Team>` or `<Blue Team>`.

These labels shouldn't have brackets, but should be bold or underlined.

[See this snippet for where to change teamnames](https://github.com/RoboJackets/robocup-software/blob/6ea6cf696dd5ecd7caca044bee93a308cea32e5c/soccer/MainWindow.cpp#L432). This field accepts HTML tags to make it bold/underlined. Tweak it until it looks nice! 

screenshot_20161127_174544

petschekr:
Fixes #824 by changing team labels to bold and setting their color to the color of the team that they represent.

### DIFF
--- a/soccer/MainWindow.cpp
+++ b/soccer/MainWindow.cpp
@@ -209,8 +209,7 @@ void MainWindow::addLayer(int i, QString name, bool checked) {
     on_debugLayers_itemChanged(item);
 }
 
-enum Side { Yellow, Blue };
-string formatLabelBold(Side side, string label) {
+string MainWindow::formatLabelBold(Side side, string label) {
     string color;
     // Colors match up with those statically defined in MainWindow.ui
     if (side == Side::Yellow) {
--- a/soccer/MainWindow.cpp
+++ b/soccer/MainWindow.cpp
@@ -22,6 +22,7 @@
 
 #include <iostream>
 #include <ctime>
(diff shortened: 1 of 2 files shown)


soccer/MainWindow.cpp
```diff
@@ -415,7 +429,9 @@ void MainWindow::updateViews() {
         _processor->refereeModule()->stage_time_left / 1000.0f, 'f', 2)));
 
     const char* blueName = _processor->refereeModule()->blue_info.name.c_str();
-    _ui.refBlueName->setText(strlen(blueName) == 0 ? "<Blue Team>" : blueName);
+    string blueFormatted = strlen(blueName) == 0 ? "Blue Team" : blueName;
```
jgkamat: This is perfect, I didn't notice before, but without this, the styling wouldn't apply to teams set by the refbox. :+1:

jgkamat:
@ashaw596 can you take a look at this too? Albert knows way more about C++ coding conventions.

petschekr:
Made those changes that you requested by moving the enum to the header file and defining the helper function as a private static method. Should be good to go now.

PR #834, opened 2016-11-30
Fixed the issue!

### LINKED ISSUE #826: Default Playbooks should include all restart plays
Right now, we have a playbook load/save sytem, which can be activated with the buttons above the plays dialouge.

We load [this file by default](https://github.com/RoboJackets/robocup-software/blob/master/soccer/gameplay/playbooks/example.pbk).

This file should be modified to add all the plays under restarts. You can do this by editing the file by hand, or by checking all the plays, and saving to this file.

screenshot_20161127_180112

hchittanuru3:
Added all the plays from the restarts directory so that they show in restarts in the simulator. Closes #826

### DIFF
--- a/soccer/gameplay/playbooks/example.pbk
+++ b/soccer/gameplay/playbooks/example.pbk
@@ -1,3 +1,12 @@
 offense/basic_122
 restarts/kick_penalty
-testing/two_side_attack
+testing/repeated_line_up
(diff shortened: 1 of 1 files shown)


jgkamat:
Perfect, I'm going to go ahead and manually merge this into master. Thanks for your contribution! :smile:

Issue #754, opened 2016-06-30
Soccer should look for the "logs" directory relative to the "run" directory


justbuchanan:
Right now it looks relative the current _working directory_.  This is an issue because if you launch it with `./run/soccer` from the root of the repo, it looks for `./logs` which probably doesn't exist.  This is why we failed to log our first match at comp.

chachmu:
This also seems to apply to the option to create a log folder under debug, I think it uses the same variable though so changing that should fix it.

jpfeltracco:
[Relevant Code](https://github.com/RoboJackets/robocup-software/blob/6d6947842789e90c67462b7fcbe8a21d8951764c/soccer/main.cpp#L157)

jgkamat:
Currently, the makefile cd's inside the run folder before running soccer when using `make run`, so anyone working on this, just change the logs dir in the two places in the referenced code, and test it out by manually running `./run/soccer`. I'll add the makefile changes once a PR is made, unless the assignee wants to stab at that too.

rachitk:
Just made a PR #835 to make the change requested. Not positive if it's the correct change or if it's fully complete yet, as `make run-sim` does not log properly with this change, likely due to makefile changes still needing to be made.

Issue #831, opened 2016-11-28
Establish convention for referencing files


jgkamat:
Right now referencing files in soccer is a mixed bag, and I want to try to standardize it so we don't have issues.

Currently, the convention is to assume we are running from the `./run` directory, so to get perfect behavior you need to: `cd run && ./soccer`. The makefile does this automatically with `make run`.

Here are different schemes I can come up with. Feel free to add some more

1. Keep the current scheme. Reference everything from the ./run/ folder, and force people to run from the ./run folder.
2. Switch to the root of the repo. this will force people to run our binary with ./run/soccer, and have their current directory be the root of our repo. We will need to change the makefile to reference this
3. Use relative paths everywhere, and establish a file which we won't move as our reference point. This is my favorite. In python/bash this is trivial, with `dirname` or `os.path.dirname(os.path.realpath(__file__))`, but in c theres no standard cross platform way to do this. I think we can use the python bridge to take advantage of the python binding to dirname, which should be cross platform. When we do this, all file references must be relative to the directory root which we define. This will allow people to run our binary from any current directory.

Let me know what you guys think!

Related: #754 #764

rachitk:
Referenced this in #835. Made changes in `makefile` that has implications for this topic, and essentially makes all of the changes that option 2 requires. 

If option 2 is rejected in favor of another option, then #835 should NOT be merged.

jgkamat:
https://github.com/RoboJackets/robocup-common/blob/950f393193f3cc8211b41ab6d2ccaafb407789ee/Utils.cpp#L51

This is the correct way to do it, I think this is the way we'll do it going forward. This issue will now track the conversion. I'll update your pr @rachitk

jgkamat:
This issue is pretty complicated/diverse so I'm going to make the final few changes to it. Your commits will be part of #848, which will get merged once it gets tested/reviewed. Thanks for the contribution! :smile:

Issue #764, opened 2016-06-30
Playbook load in soccer should show file chooser in the correct folder


justbuchanan:
Right now it shows the root of the project, it should launch to `soccer/gameplay/playbooks`.

jpfeltracco:
[Relevant function](https://github.com/RoboJackets/robocup-software/blob/a10654f07fdf73f2951ae9a0081b865a23102880/soccer/gameplay/playbook.py#L7)

[Relevant call to function](https://github.com/RoboJackets/robocup-software/blob/a10654f07fdf73f2951ae9a0081b865a23102880/soccer/gameplay/main.py#L147)

jgkamat:
So the reference above is for the actual loading of the file, for the file chooser you will need to look at [this bit of code](https://github.com/RoboJackets/robocup-software/blob/6ea6cf696dd5ecd7caca044bee93a308cea32e5c/soccer/MainWindow.cpp#L1203). However, this already works when running soccer from the run directory. See #754 as well.

I think all paths should be relative to the root of robocup-software, so I'm in favor of changing the `../soccer/gameplay/playbooks/` to `./soccer/gameplay/playbooks/`, and swapping it over in the makefile. If you want to tackle this issue, go ahead and do the above for now.

This relative/absolute path is starting to get kind of annoying, so I'm going to open another issue for it.

rachitk:
Made the change in #835 as requested; no changes have been made to makefile yet, so `make run` and the like now do not work while `./soccer/run` does.

PR #835, opened 2016-11-30
Changing check for logs directory, playbook default location, and corresponding makefile changes

rachitk:
Makes the change that was requested in #754. More changes will need to be made to makefile as this breaks logging when using `make run`. Logging will work properly when using `./run/soccer` with this change.

### DIFF
--- a/makefile
+++ b/makefile
@@ -13,16 +13,16 @@ all:
 	$(call cmake_build_target, all)
 
 run: all
-	cd run; ./soccer
+	./run/soccer
 rs: run-sim
 run-sim: all
 	-pkill -f './simulator --headless'
-	cd run; ./simulator --headless &
-	cd run; ./soccer -sim -pbk example.pbk
+	./run/simulator --headless &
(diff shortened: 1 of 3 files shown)


rachitk:
Added another change which makes the change requested in #764; it has the same issue as the first one in terms of breaking `make run` until the makefile changes are made.

rachitk:
Went ahead and made the makefile changes as requested; simply removed all instances of `cd run` and edited the respective filepaths after accordingly. All logging and playbook functionality made in the first two commits works properly now. This PR should close #754 and #764, as well as address concerns made in #831, though this change needs to be made known for future reference.

jgkamat:
Currently this is blocked on #831, I'll work on that once I don't have a pile of work to do :smile: 

Hang tight!

jgkamat:
This issue is pretty complicated/diverse so I'm going to make the final few changes to it. Your commits will be part of #848, which will get merged once it gets tested/reviewed. Thanks for the contribution!

PR #838, opened 2016-11-30
Created global variable to limit how far robot's travel

### LINKED ISSUE #830: Repeated line up shoud provide an easy way to traverse only part of the field
Repeated line up currently traverses the entire field, whcih is what we want by default.

However, during tuning, we may want to insert a buffer, to limit how far we command the robots to go, so if there is a problem, we won't bump into the walls.

[This should be easily achevable by changing the calls to gen_line here to decimal values](https://github.com/RoboJackets/robocup-software/blob/8aa13ff4334af3d8bba1f5e64febd36c311c8d38/soccer/gameplay/plays/testing/repeated_line_up.py#L54). However, it would be nice to have a single constant value at the top of the file to tune both of these gen line calls at once, so it's easy to do this in a rush.

rk1998:
I created a variable called BUFFER that can be set to any value and it will change how far the robot moves across the field during repeated lineup. BUFFER should be a decimal value.

Closes #830

### DIFF
--- a/soccer/gameplay/plays/testing/repeated_line_up.py
+++ b/soccer/gameplay/plays/testing/repeated_line_up.py
@@ -11,6 +11,7 @@ import time
 class RepeatedLineUp(play.Play):
 
     Pause = 2.0
+    BUFFER = .5
 
     class State(enum.Enum):
         left = 0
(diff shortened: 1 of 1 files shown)


soccer/gameplay/plays/testing/repeated_line_up.py
```diff
@@ -11,6 +11,7 @@
 class RepeatedLineUp(play.Play):
 
     Pause = 2.0
+    BUFFER = .5
```
jgkamat: Could we have this default to 1, We'll manually override it if needed.

jgkamat:
I'm going to go ahead and apply another commit on here, and merge it in. Thanks for the contribution!

😄

PR #848, opened 2016-12-21
Refactor refrences to be relative to soccer (in the run directory)

### LINKED ISSUE #754: Soccer should look for the "logs" directory relative to the "run" directory

### LINKED ISSUE #764: Playbook load in soccer should show file chooser in the correct folder

### LINKED ISSUE #831: Establish convention for referencing files

jgkamat:
This fixes the various hardcoded file paths (relative to the current directory) to instead be relative to the directory where soccer is located, which makes it consistent. 

Closes #831, closes #754, closes #764. Replaces #835

Let me know if you find any other file paths referring to the current directory! Feedback is appreciated. :smile:

### DIFF
--- a/soccer/joystick/GamepadController.cpp
+++ b/soccer/joystick/GamepadController.cpp
@@ -27,8 +27,11 @@ dpdown:h0.4,dpright:h0.0,dpright:h0.2,dpup:h0.0,dpup:h0.1,leftshoulder:h0.0,\
 dpup:h0.1,leftshoulder:h0.0,leftshoulder:b4,lefttrigger:b6,rightshoulder:b5,\
 righttrigger:b7,leftstick:b10,rightstick:b11,leftx:a0,lefty:a1,rightx:a2,righty:a3,");
     // Attempt to add additional mappings (relative to run)
+    ApplicationRunDirectory();
     if (SDL_GameControllerAddMappingsFromFile(
-            "../external/sdlcontrollerdb/gamecontrollerdb.txt") == -1) {
+            ApplicationRunDirectory()
+            .filePath("../external/sdlcontrollerdb/gamecontrollerdb.txt")
+            .toStdString().c_str()) == -1) {
         cout << "Failed adding additional SDL Gamecontroller Mappings: "
              << SDL_GetError() << endl;
     }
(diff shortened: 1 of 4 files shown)


jgkamat:
I _think_ this shoulld be done content wise. I'll merge this in if someone trys this out and dosen't find any problems. (at the very minimum, test out the playbook load/save default dir, and the logging dir).

soccer/joystick/GamepadController.cpp
```diff
@@ -27,8 +27,12 @@ dpdown:h0.4,dpright:h0.0,dpright:h0.2,dpup:h0.0,dpup:h0.1,leftshoulder:h0.0,\
 dpup:h0.1,leftshoulder:h0.0,leftshoulder:b4,lefttrigger:b6,rightshoulder:b5,\
 righttrigger:b7,leftstick:b10,rightstick:b11,leftx:a0,lefty:a1,rightx:a2,righty:a3,");
     // Attempt to add additional mappings (relative to run)
+    ApplicationRunDirectory();
```
jgkamat: This can probably be removed.

jgkamat:
@chachmu could you give this a quick look over? I just want to make sure this PR dosen't break any of the playbook load/save stuff and logging (and some stuff is blocked on it so I would like to get this merged).

chachmu:
@jgkamat LGTM

jgkamat:
Let me know if anyone has issues relating to logging, joysticks, or playbook load/save. :fork_and_knife: :cake:

PR #849, opened 2016-12-24
Clean up boost python cmake config

jgkamat:
This should remove all the 'fake failures' of finding Boost Python in the cmake build log, which were pretty annoying.

~This could have a problem finding boost python, but I don't think it should. Could someone try this out on ubuntu (just launch soccer, and run a play), and get back to me if it works? Thanks!~

EDIT: Never mind, this is completely broken, I'll try and find a way to fix this.


jgkamat:
Ubuntu packages boost python in a really weird way so this has to stay... :cry:

PR #850, opened 2016-12-28
Improve the way that boost python is checked for on ubuntu

jgkamat:
Better version of #849 which should actually work..

It would be nice for someone to test if soccer opens up fine with this on ubuntu.

In addition to reducing spam in the cmake build, it also will work for future python versions, instead of us having to keep adding python-3#+1 every time ubuntu upgrades.


jgkamat:
Not too big of a change, so I'm going ahead and merging this. Let me know if you see errors pertaining to boost python after this commit.

PR #852, opened 2017-01-11
Final Goalie Fixes

### LINKED ISSUE #761: When using external ref, goalie id should be pulled from ref packet

### LINKED ISSUE #765: soccer's --noref flag should take effect immediately, not after gui, etc has launched

### LINKED ISSUE #772: Make soccer continue to run if goalie loses vision or is deselected

chachmu:
I am recreating this pull request because the old one has more lines of merge errors than I am actually adding code. This pull request:

- [x] Closes #772
- [x] Closes #765
- [x] Closes #761 
- [x] Keeps the previous goalie selected if the external ref goalie id changes to be invalid
- [x] Disables the Ref Shortuct bar IFF an external ref is connected
- [x] Disables the goalie selector IFF the external ref is transmitting a valid goalie id
- [x] Add option for No Goalie
- [x] Grey out Invalid Goalie IDs in the dropdown menu

Continuation of #777 and #774 and #783

### DIFF
--- a/soccer/MainWindow.cpp
+++ b/soccer/MainWindow.cpp
@@ -721,6 +721,10 @@ void MainWindow::updateStatus() {
     bool haveExternalReferee =
         (curTime - ps.lastRefereeTime) < RJ::Seconds(0.5);
 
+    if(_ui.goalieID->currentIndex() != _processor->state()->gameState.getGoalieId()+1){
+        _ui.goalieID->setCurrentIndex(_processor->state()->gameState.getGoalieId()+1);
+    }
+
     /*if (_autoExternalReferee && haveExternalReferee &&
     !_ui.externalReferee->isChecked())
     {
--- a/soccer/MainWindow.cpp
+++ b/soccer/MainWindow.cpp
@@ -707,17 +707,17 @@ void MainWindow::updateStatus() {
 
     std::vector<int> validIds = _processor->state()->ourValidIds();
 
-    for(int i=1;i<17;i++){
-        //This will need to change if the max number of robots changes
+    for (int i = 1; i < 17; i++) {
+        // This will need to change if the max number of robots changes
         //.stdString()
-        if(std::find(validIds.begin(), validIds.end(), i-1)!=validIds.end()){
-            //The list starts with None so i is 1 higher than the shell id
-            _ui.goalieID->setItemData(i,true,Qt::UserRole);
+        if (std::find(validIds.begin(), validIds.end(), i - 1) !=
+            validIds.end()) {
+            // The list starts with None so i is 1 higher than the shell id
+            _ui.goalieID->setItemData(i, true, Qt::UserRole);
+        } else {
+            _ui.goalieID->setItemData(i, false, Qt::UserRole - 1);
         }
-        else{
-            _ui.goalieID->setItemData(i,false,Qt::UserRole -1);
-        }
-        //std::cout<<_ui.goalieID->itemText(i).toStdString();
+        // std::cout<<_ui.goalieID->itemText(i).toStdString();
     }
 
     if (haveExternalReferee && _autoExternalReferee) {
--- a/soccer/MainWindow.cpp
+++ b/soccer/MainWindow.cpp
@@ -709,7 +709,6 @@ void MainWindow::updateStatus() {
 
     for (int i = 1; i < 17; i++) {
         // This will need to change if the max number of robots changes
-        //.stdString()
         if (std::find(validIds.begin(), validIds.end(), i - 1) !=
             validIds.end()) {
             // The list starts with None so i is 1 higher than the shell id
@@ -717,7 +716,6 @@ void MainWindow::updateStatus() {
         } else {
             _ui.goalieID->setItemData(i, false, Qt::UserRole - 1);
         }
-        // std::cout<<_ui.goalieID->itemText(i).toStdString();
     }
 
     if (haveExternalReferee && _autoExternalReferee) {
--- a/soccer/MainWindow.cpp
+++ b/soccer/MainWindow.cpp
@@ -707,8 +707,7 @@ void MainWindow::updateStatus() {
 
     std::vector<int> validIds = _processor->state()->ourValidIds();
 
-    for (int i = 1; i < 17; i++) {
-        // This will need to change if the max number of robots changes
+    for (int i = 1; i <= Num_Shells; i++) {
         if (std::find(validIds.begin(), validIds.end(), i - 1) !=
             validIds.end()) {
             // The list starts with None so i is 1 higher than the shell id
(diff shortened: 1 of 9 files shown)


chachmu:
I misinterpreted what setting a reviewer does, my bad, his isn't quite ready yet.

 For some reason the most recent commit bad fixed the invalid goalie id issue but then stopped working immediately after I pushed it, I should have time to fix it in about within the next few hours.

jgkamat:
No rush, we've still got a lot of time to work on this. Thanks for going through and redoing it, I'm gonna make sure this gets merged in this time! :smile:

chachmu:
@jgkamat Ok, I have tested everything and I think it's ready. I did want to ask one question though. When not using the external ref, what should happen if the user selects an invalid goalie id from the dropdown menu in soccer? Currently it just makes it so that there is no goalie. Would it be better to make it not change to the invalid option? Or maybe gray out the invalid id #'s in the dropdown?

jgkamat:
Hmm, greying out the invalid ID's in soccer would be nice (but we should also have the option to explicitly have no goalie, for testing purposes). I'm fine with just having no goalie for now, unless you want to work on that a little bit more.

I'll take a look at this as soon as I can, thanks again for putting so much time on this! :smile:

soccer/MainWindow.cpp
```diff
@@ -718,8 +703,54 @@ void MainWindow::updateStatus() {
...
+            _processor->state()->gameState.getGoalieId() + 1) {
+            _ui.goalieID->setCurrentIndex(
+                _processor->state()->gameState.getGoalieId() + 1);
+        }
+    } else {
+        _ui.goalieID->setEnabled(true);
+    }
 
     /*if (_autoExternalReferee && haveExternalReferee &&
     !_ui.externalReferee->isChecked())
```
chachmu: This was commented out before I began working on it. Should I just remove it since it is replaced by the new code?

jgkamat: :+1:, yup, I was gonna ask you that when I finished the reiew, that would be perfect! :smile:

soccer/MainWindow.cpp
```diff
@@ -718,14 +703,54 @@ void MainWindow::updateStatus() {
...
+    bool haveExternalReferee = (curTime - ps.lastRefereeTime) < RJ::Seconds(1);
+
+    std::vector<int> validIds = _processor->state()->ourValidIds();
 
-    /*if (_autoExternalReferee && haveExternalReferee &&
-    !_ui.externalReferee->isChecked())
-    {
-        _ui.externalReferee->setChecked(true);
-    }*/
+    for (int i = 1; i < 17; i++) {
```
jgkamat: Can you remove 17 from this loop? In Constants.hpp, we have Num_Shells which you can use (which is 16, so you'll need to do NumShells+1 or use i+1 everywhere and start i at 0).

soccer/Processor.cpp
```diff
@@ -402,6 +401,7 @@ void Processor::run() {
         for (NewRefereePacket* packet : refereePackets) {
             SSL_Referee* log = _state.logFrame->add_raw_refbox();
             log->CopyFrom(packet->wrapper);
+            curStatus.lastRefereeTime = packet->receivedTime;
```
jgkamat: Just checking, the refereePackets list is sorted? If not we should put this in a max along with the old lastRefTime

soccer/gameplay/role_assignment.py
```diff
@@ -185,11 +185,13 @@ def assign_roles(robots, role_reqs):
     # To do our calculations though, we need to have a flat list of RoleRequirements to match to our list of robots
     # The flatten_tree() method does this and keeps track of how to rebuild the tree so we can do so at the end
     def flatten_tree(tree, path_prefix=[]):
+        valid_shell_ids = [bot.shell_id() for bot in robots]
         for key, subtree in tree.items():
             if isinstance(subtree, dict):
                 flatten_tree(subtree, path_prefix + [key])
             elif isinstance(subtree, RoleRequirements):
-                role_reqs_list.append(subtree)
+                if subtree.required_shell_id == None or subtree.required_shell_id in valid_shell_ids:
```
jgkamat: This should be `is None`. It's not really used anywhere else but I would prefer to start sticking to it in the future.

soccer/MainWindow.cpp
```diff
@@ -718,14 +703,54 @@ void MainWindow::updateStatus() {
...
+        _ui.fastStop->setEnabled(true);
+        _ui.fastReady->setEnabled(true);
+        _ui.fastForceStart->setEnabled(true);
+        _ui.fastKickoffBlue->setEnabled(true);
+        _ui.fastKickoffYellow->setEnabled(true);
+    }
+
+    if (std::find(validIds.begin(), validIds.end(),
+                  _processor->state()->gameState.getGoalieId()) !=
+            validIds.end() &&
```
jgkamat: I forgot to mention one other thing:

Right now, if the refbox sends us a invalid ID (not on the field), we use the default (0 in our case). I think that we should set it anyway, because if we set 0 when we aren't supposed to, thats a rule violation. 

A case where this can happen is when we pull the goalie off the field (I'm not sure if this code would reset the ID to 0, which I don't think would happen). But in any case, if we pull the goalie off the field, soccer crashes, and we have to restart, we end up with a goailie ID of 0. 

I think this if statement should be relaxed to be only `(haveExternalRefereee)`. Thoughts?

soccer/Processor.cpp
chachmu: It appears that the packets are pushed into a vector in the order that they are received but I have no clue if that actually guarantees that the list is always properly sorted.

soccer/MainWindow.cpp
chachmu: I don't think that would do what you are thinking. Removing the find would make it so that the selector is disabled at any time you have an external ref, valid ID or not. I think that might be better solved with a competition mode that saves the current goalie somehow each time it is updated and then uses that if the program crashes and restarts.

soccer/MainWindow.cpp
```diff
@@ -718,14 +703,50 @@ void MainWindow::updateStatus() {
...
-        (curTime - ps.lastRefereeTime) < RJ::Seconds(0.5);
+    bool haveExternalReferee = (curTime - ps.lastRefereeTime) < RJ::Seconds(1);
+
+    std::vector<int> validIds = _processor->state()->ourValidIds();
+
+    for (int i = 1; i <= Num_Shells; i++) {
+        if (std::find(validIds.begin(), validIds.end(), i - 1) !=
+            validIds.end()) {
+            // The list starts with None so i is 1 higher than the shell id
+            _ui.goalieID->setItemData(i, true, Qt::UserRole);
```
jgkamat: This line seems to be broken.

When I bring a robot off the field, and back onto the field, the dropdown stays disabled. 

The logic is correct (it's running this line at the correct times), but this line does not actually re-enable that dropdown. I'm looking at it now.

jgkamat: Setting this to 
`setItemData(i, 33, Qt::UserRole - 1);` fixes this, but I want to try to find the proper way of disabling entries to a qcombobox.

PR #853, opened 2017-01-12
Update sdl controller DB

jgkamat:
This PR updates the SDL controller db, as that repo added my PR which means we can remove this dirty hack.

### DIFF
--- a/soccer/joystick/GamepadController.cpp
+++ b/soccer/joystick/GamepadController.cpp
@@ -19,13 +19,6 @@ GamepadController::GamepadController()
         return;
     }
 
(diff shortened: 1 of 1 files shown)


jgkamat:
Needs actual testing on all the controllers we have, currently blocked on that.

jgkamat:
If a new member has nothing to do, they can head on down to the field, and test out this PR with the controllers in the back closet. Let me know if you're interested in working on that.

jgkamat:
Tested this out and found no regressions on the F310. Let me know if you see joystick bugs (make sure you update your submodules though!)

Issue #781, opened 2016-07-16
Robots veer off course when turning and moving


jgkamat:
For an example, see this:
![snapshot4](https://files.gitter.im/RoboJackets/robocup-software/23Ag/snapshot4.png)

I'm not sure if this is a problem with the soccer PID or the per wheel PID, but I think its on the per wheel PID.

Soccer seems to send the correct velocities, but I'm not 100% sure.

justbuchanan:
Tuning the pid in soccer should fix this.  There's no per-wheel pid in the simulator.

jgkamat:
So then the issue that we saw at comp was just improperly tuned PID, I'm guessing. Hopefully this isn't a bigger issue than that.

jgkamat:
I don't think this is an issue with the PID. 

I think the issue is that as we rotate at a higher speed, the rotation alters our sent commands (as our commands are sent in the robot's frame of reference), Since our rotation speed is increasing, we continue to accumulate errors even though the PID will try to compensate. I think we could fix this by predicting where the robot will face when the command is actually received (maybe length of a frame / 2) and using that angle instead of the current robot's angle to send velocities on. Do we do anything like this already?

ashaw596:
I tried that a bit and it didn't really help much (in Sim at least). I
think it's more a problem that in turning  cases the robot acts a bit more
like a car than the particle we model it as.

ashaw596:
Could also just be a Sim thing. We could test the robot itself by driving
forward and then switch to turning in place. Ideally it should slow down in
a straight line even if it's turning(though wheel momentum might be an a
issue). If it slows down while turning in a circle, it could be pid not
controlling the wheels well,  pid overcompensating, or our particle model
is weird with the wheels(would lqr help?). With our motors, the robot might
also just stop and fall over with really fast wheel deceleration.

jgkamat:
@jpfeltracco found out that this is an issue with us limiting our speed. We don't take rotation into account when driving, so when we rotate and drive, wheels at the top can't drive fast enough and we swerve.

jgkamat:
@jpfeltracco does #856 fix this? I thought it did, but I wanted to be sure.

jpfeltracco:
Yep, I'll go ahead and close this, I thought I mentioned in the PR that it closes this but I guess not.

Issue #770, opened 2016-07-07
Implement full angle paths with motion/rotation constraints


ashaw596:
I've noticed that fast turning causes Pathing errors even in simulator, and this should be implemented eventually anyways to have accurate shooting and stuff.

Currently, PID is used to control angle on the soccer side with a target Angle or Location in MotionControl.cpp. I've already modified Paths and Pathplanner to have the capability to have angle information, but haven't implemented actual angle path planning. InterpolatedPath has angles in their Waypoints.

We would use trapezoidal motion planning for rotation and stuff.

jpfeltracco:
These pathing errors show up in our simulator when running the `RepeatedTurningLineUp` play where the robots completely diverge from their paths because they are attempting to rotate while at full velocity.

This divergence can be minimized by capping the angular velocity by 
```cpp
rotationConstraints.maxSpeed - C * target.vel.mag()
```
Where C is some constant (0.9 seemed to work okay). This essentially prevents the robots from rotating while at max velocity and seems to mostly remove the pathing errors.

This branch has this hack in it: [jeremy/pathingHack](https://github.com/RoboJackets/robocup-software/tree/jeremy/pathingHack)

Full angle path planning that takes into consideration the full constraints of the robot is in the works.

ashaw596:
Nice!

jpfeltracco:
[This Issue is related](https://github.com/RoboJackets/robocup-software/issues/781).

So my rough idea for solving this is:
1. Add a max wheel speed constant.
2. Run the paths through an angle planning function after [here](https://github.com/RoboJackets/robocup-software/blob/master/soccer/planning/IndependentMultiRobotPathPlanner.cpp#L78)
3. The function would first pick an ideal angle rotation speed, then calculate the wheel speeds required to achieve both that rotation and velocity using the translation/rotation to wheel velocity matrix.
    * If any wheel speed exceeds the max wheel speed, we just cut the angle rotation speed in half and check again (or find a way to calculate the rotation required to bring us up to max wheel speed)
    * Otherwise, simply accept that angle speed.

To implement this I believe I would have to add a rotationCommand to the PlanRequest struct, but I'm a little confused with whether or not I should add the rotationCommand to the SinglePlanRequest as well, and why exactly we have these two similar structs.

Also, do you think it would be a better idea to do angle planning at the same time we're fitting the velocity curve? That way I guess we could scale back the translational velocity if we can't finish the rotation in time or something.

[...]

PR #856, opened 2017-01-15
Clamp World Acceleration instead of robot acceleration

ashaw596:
Clamp World Acceleration instead of robot acceleration.
Fixes like half our problems.


jpfeltracco:
Closes #781, related to #770.

Basically, in motion control we were clamping acceleration in robot coordinates, not field coordinates. This meant that when the robot rotated, we would see massive acceleration just due to the change in the robot coordinate system. The fix was to do clamping in field coordinates, not robot coordinates.

PR #862, opened 2017-01-16
First attempt at fixing goalie disable code

jgkamat:
Fixes a small regression in #852

@chachmu could you review this? (since you worked on this before)

I think this shouldn't cause any problems but let me know if it does

### DIFF
--- a/soccer/MainWindow.cpp
+++ b/soccer/MainWindow.cpp
@@ -144,8 +144,10 @@ MainWindow::MainWindow(Processor* processor, QWidget* parent)
     _logPlaybackButtons.push_back(_ui.logPlaybackPlay);
     _logPlaybackButtons.push_back(_ui.logPlaybackLive);
 
-    // Get the item model from the goalieID boxes so we can disable them properly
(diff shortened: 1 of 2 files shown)


chachmu:
I won't have time to test this until tomorrow, but I'm a little unclear about what it is supposed to be changing.

chachmu:
I won't have time to test this until tomorrow but it all looks correct.

chachmu:
Is there a way to test this in simulator? I can see that it is correctly disabling the items which would suggest that since this method directly sets flags it also enables them but I can't find a way to remove a robot from vision while running sim. Everything seems fine but if this needs to be tested with real robots I wouldn't be able to test this until Thursday because I won't be able to make it to todays meeting.

jgkamat:
As far as I can tell, there isn't. We would need to remove/add a robot from the sidebar in order for this to be tested properly. I've tested this a bit on the field, but we can wait a little bit more before merging, it's not too important.

You don't need robots to test this, you should be able to use an empty shell to try this out.

PR #867, opened 2017-01-17
Make a Competition Mode

### LINKED ISSUE #762: Allow all game-relevant parameters to be specified in soccer command line arguments
This would allow us to auto-restart soccer if it crashes during a match.  Right now this isn't possible because some important things have to be configured in the gui.  Relevant to #761.

### LINKED ISSUE #778: Create a Competition Mode
Would it be a good idea to make a mode for competition (maybe "make run-comp") that restarts itself if it crashes and maybe defaults to logging somewhere to keep us from forgetting to log? This would require the completion of #762 but it would help solve a lot of our issues with seg faults during competition or other problems.

chachmu:
This will make it possible to automatically restart soccer if it segfaults during competition as well as set the important parameters automatically. Closes #778.

TODO:

- [x] Set Yellow/Blue Team from refpacket
- [x] Closes #762 
  - [x] Add argument for Defend Vision
  - [x] Add argument for Vision Channel 
- [x] Automate restart of soccer

### DIFF
--- a/runcomp.sh
+++ b/runcomp.sh
@@ -1,16 +1,14 @@
 #!/bin/bash
 
-read -p "Defend Plus X (0 or 1): " defendPlusX
+read -p "Defend Plus X (plus or minus): " defendPlusX
 read -p "Vision Channel (1,2 or Full): " visionChannel
 read -p "Specify Playbook file (default comp.pbk): " playbook
 
 make
 
-until ./run/soccer -pbk ${playbook:-"comp.pbk"} -defend ${defendPlusX:-"0"} -vision ${visionChannel:-"full"}; do
+until ./run/soccer -pbk "${playbook:-"comp.pbk"}" -defend "${defendPlusX:-"minus"}" -vision "${visionChannel:-"full"}"; do
 	echo "Soccer crashed with exit code $?. Restarting" >&2
 	sleep 1
 done
 
 echo "Soccer exited normally."
-
-
--- a/runcomp.sh
+++ b/runcomp.sh
@@ -2,11 +2,11 @@
 
 read -p "Defend Plus X (0 or 1): " defendPlusX
 read -p "Vision Channel (1,2 or Full): " visionChannel
-read -p "Specify Playbook file: " playbook
+read -p "Specify Playbook file (default comp.pbk): " playbook
 
 make
 
-until ./run/soccer -pbk $playbook -defend $defendPlusX -vision $visionChannel; do
+until ./run/soccer -pbk ${playbook:-"comp.pbk"} -defend ${defendPlusX:-"0"} -vision ${visionChannel:-"full"}; do
 	echo "Soccer crashed with exit code $?. Restarting" >&2
 	sleep 1
 done
--- a/runcomp.sh
+++ b/runcomp.sh
@@ -1,0 +1,16 @@
+#!/bin/bash
+
+read -p "Defend Plus X (0 or 1): " defendPlusX
+read -p "Vision Channel (1,2 or Full): " visionChannel
+read -p "Specify Playbook file: " playbook
+
+make
+
+until ./run/soccer -pbk $playbook -defend $defendPlusX -vision $visionChannel; do
+	echo "Soccer crashed with exit code $?. Restarting" >&2
+	sleep 1
+done
+
+echo "Soccer exited normally."
+
+
--- a/soccer/MainWindow.cpp
+++ b/soccer/MainWindow.cpp
@@ -102,19 +102,19 @@ MainWindow::MainWindow(Processor* processor, QWidget* parent)
 
     _ui.debugLayers->setContextMenuPolicy(Qt::CustomContextMenu);
 
-    //Action Group 0
+    // Action Group 0
     QActionGroup* teamGroup = new QActionGroup(this);
     teamGroup->addAction(_ui.actionTeamBlue);
     teamGroup->addAction(_ui.actionTeamYellow);
     qActionGroups.push_back(teamGroup);
 
-    //Action Group 1
+    // Action Group 1
     QActionGroup* goalGroup = new QActionGroup(this);
     goalGroup->addAction(_ui.actionDefendMinusX);
     goalGroup->addAction(_ui.actionDefendPlusX);
     qActionGroups.push_back(goalGroup);
 
-    //Action Group 2
+    // Action Group 2
     QActionGroup* rotateGroup = new QActionGroup(this);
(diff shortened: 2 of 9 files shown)


soccer/NewRefereeModule.cpp
```diff
@@ -179,6 +179,13 @@ void NewRefereeModule::run() {
         ballPlacementx = packet->wrapper.designated_position().x();
         ballPlacementy = packet->wrapper.designated_position().y();
 
+        std::string blue_name = blue_info.name;
+        for(char& letter : blue_name){
+            letter=tolower(letter);
+        }
+
+        blueTeam(blue_name == "robojackets");
```
chachmu: An important thing to note about checking this way is that if the name is not set in the refbox or is incorrect we will always be Yellow. I am not actually sure if that is a good thing or not.

chachmu:
The most recent commit doesn't seem to properly select the correct DefendX in the menu, it sets the value properly but the menu doesn't reflect that, I'll look into that.

chachmu:
To run competition mode:
run ./runcomp.sh
Follow the prompts to fill in DefendX, Vision Channel, and the Playbook file.

jgkamat:
Wow, that was really fast! I'll try to take a look at it soon but I think it might be until the weekend. (If someone else wants to do code review on this as well, that would be :+1:)

Thanks so much! :smile:

chachmu:
I would like to make it so you can just hit enter on each value when running the bash file to set it to a default but I was having some issues with strings so I figured I would put up what I had and talk to someone who actually knows bash tomorrow.

runcomp.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+read -p "Defend Plus X (0 or 1): " defendPlusX
+read -p "Vision Channel (1,2 or Full): " visionChannel
+read -p "Specify Playbook file: " playbook
+
+make
+
+until ./run/soccer -pbk $playbook -defend $defendPlusX -vision $visionChannel; do
```
jgkamat: Not a full review, but if you want to add default values to this (ex: user just hits enter), you can change your variable references to:

${playbook:-"DefaultValue"} instead of $playbook.

Keep in mind, if playbook has any value, including an invalid one (like a single letter), this won't trigger.

chachmu: Awesome, that worked perfectly.

soccer/MainWindow.cpp
```diff
@@ -219,6 +247,36 @@ string MainWindow::formatLabelBold(Side side, string label) {
...
+            blueTeam ? _ui.actionTeamBlue->trigger()
+                     : _ui.actionTeamYellow->trigger();
+        }
+    } else {
+        _ui.goalieID->setEnabled(true);
+        // enable Blue/Yellow team
+        qActionGroups[0]->setEnabled(true);
+    }
+
+    // update Yellow/Blue team from packet
```
jgkamat: Isn't this implemented above? It looks like its working fine. If so, please remove this :smile:

soccer/main.cpp
```diff
@@ -114,6 +120,27 @@ int main(int argc, char* argv[]) {
...
+                printf("No vision channel specified after -vision\n");
+                usage(argv[0]);
+            }
+            i++;
+            // USE STRCMP
+            if (strcmp(argv[i], "1") == 0) {
+                visionChannel = 1;
+            } else if (strcmp(argv[i], "2") == 0) {
+                visionChannel = 2;
+            }
```
jgkamat: I think that you want to check if full is present here (and if it isn't we should error out, and not fail silently and assume full).

soccer/main.cpp
```diff
@@ -114,6 +120,27 @@ int main(int argc, char* argv[]) {
             playbookFile = argv[++i];
         } else if (strcmp(var, "-noref") == 0) {
             noref = true;
+        } else if (strcmp(var, "-defend") == 0) {
+            if (i + 1 >= argc) {
+                printf("Field half not specified after -defend\n");
+                usage(argv[0]);
+            }
+            i++;
+            if (strcmp(argv[i], "plus") == 0) {
```
jgkamat: Can you check for the alternate case here as well (and fail when we get an invalid argument). 

Failing is much better than assuming defaults silently, imo.

soccer/Processor.hpp
```diff
@@ -273,5 +276,10 @@ class Processor : public QThread {
 
     VisionReceiver vision;
 
+    // 1 for Primary Half
+    // 2 for Secondary Half
+    // 3 for Full channel
+    int _visionChannel;
```
jgkamat: Can this be made an enum instead? It would be a LOT cleaner :smile:

soccer/MainWindow.cpp
```diff
@@ -171,8 +175,9 @@ void MainWindow::initialize() {
 
     // Initialize to ui defaults
     on_goalieID_currentIndexChanged(_ui.goalieID->currentIndex());
-    for (const auto& qActionGroup : qActionGroups) {
-        qActionGroup->checkedAction()->trigger();
+    // Do not initialize DefendPlusX(Group 1) or Vision(Group 3)
+    for (int i : {0, 2, 4}) {
```
jgkamat: I really don't like these magic numbers.

Is it possible to refer to these by name (maybe `radioGroup`?). Or maybe we could figure out what numbers correspond to these somehow? I'm not sure how QActionGroups work though...

runcomp.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+read -p "Defend Plus X (0 or 1): " defendPlusX
+read -p "Vision Channel (1,2 or Full): " visionChannel
+read -p "Specify Playbook file (default comp.pbk): " playbook
+
+make
+
+until ./run/soccer -pbk ${playbook:-"comp.pbk"} -defend ${defendPlusX:-"0"} -vision ${visionChannel:-"full"}; do
```
jgkamat: Can you quote the variable dereferences here? For example:

`"${playbook:-"comp.pbk"}"`

It's good style to quote variable expansion in bash when possible (one reason is that it preserves whitespace in the argument). In this case it dosen't make too much of a difference though.

soccer/MainWindow.cpp
chachmu: This is really just a vector that we make, I can just turn it into a map.

chachmu:
Sure, sounds good to me.

jgkamat:
:+1: Merging now! Thanks, this really helps us (although I wish it woudln't :smile:)

PR #869, opened 2017-01-18
Only reset simulator scene when placing ball

### LINKED ISSUE #865: Right Click kick at speed seems to be broken
This seems to be a recent thing, it was working at comp last year. Probably needs a bisect.

Not sure if this isn't a duplicate, let me know if it is. 

When the right click kick ball occurs, the ball is first moved to a weird location, then kicked. This is most likely an issue with the simulator. The location where the ball moves to seems to depend on the initial placemnet of the ball and/or the strength/direction of the kick.

petschekr:
Fixes #865


jgkamat:
@jpfeltracco can you sign off on this before we merge it though?

If you think it's good, go ahead and merge!

jpfeltracco:
LGTM too, nice job @petschekr!

PR #870, opened 2017-01-18
Hug goal

### LINKED ISSUE #864: Goalie should 'hug' the goal when the ball is sufficently to the right/left of the goal
Right now, when the ball is in a corner on our side of the goal, the goalie brushes up against the goal wall, but leaves a small gap. This is most likely obstacle avoidance (as we don't want to hit the goal wall during normal play), but we should override this in specific cases.

This is reproducable in the simulator

krispreza:
fixed issue to have goalie hug post

closes #864


jgkamat:
I'm going to be a little bit busy, so I don't think I can review this fully until the weekend. It's pretty important that we get this right, since we don't want to break the goalie code!

Thanks for the contribution though! This will be a huge improvement to the goalie in my opinion (in the past, we've had a lot of goals go by due to this).

jgkamat:
Ok, I think this dosen't quite solve the issue.

My original intent with the hugging goal issue was to set the `y` point, not the `x` point. This /might/ be a good idea though (setting the x point to +-max when the ball is far enough back), but I think we need to discuss this with some more members before we do that. Currently the y point is a little too high, and (I think) can be decreased, especially when we are close to the edges. 

Also, `goalie.py` is an old goalie, so this change dosen't actually change the goalie's behavior at all. (goalie.py is a standalone goalie, but we now have a bit smarter placement of defense and goalie to block all threats). The new goalie is `submissive_goalie.py`.

Submissive goalie actually dosen't set its locations, and instead defers to defense.py (which organizes all the robots to cover the goal as best as possible). It's pretty complicated, but reading through it would give you a really great overview to how our defense works. It's a little dense, so let me know if I can help you out to give you an overview (the best I can).

One option that you could do is to simply override the Y value of the defender robot in submissiveGoalie itself (so we would use defense.py coordinates, except when we're at maxX or minX, or something like that. That would probably be the simplest (and easiset) way to solve this issue.

One more thing to keep in mind is that the goalie has a little bit of obstacle avoidance to avoid hitting the goal, so we may not be able to get exactly up against the goal without problems. I think we'll be able to get fairly close without that though, so let me know if the robot isn't getting close enough.

Let me know if you have any questions! Just add more commits to your branch, and it will automatically edit this PR.

soccer/gameplay/tactics/positions/submissive_goalie.py
```diff
@@ -91,7 +91,8 @@ def block_line(self, value):
 
         self._move_target.x = min(
             max(self._move_target.x,
-                -SubmissiveGoalie.MaxX), SubmissiveGoalie.MaxX)
+                -SubmissiveGoalie.MaxX + 0.01), SubmissiveGoalie.MaxX - 0.01)
```
jgkamat: This file looks good, although I'm not sure why we need the +0.01. Could you explain that to me again?  I didn't understand it too well during the meeting.

soccer/gameplay/tactics/positions/goalie.py
```diff
@@ -195,6 +195,11 @@ def execute_block(self):
 
     def execute_defend(self):
         dest_x = main.ball().pos.x / constants.Field.Width * Goalie.MaxX
+        if main.ball().pos.y <= constants.Field.Width / 8:
```
jgkamat: This file's changes can be removed. You can do this with:

`git revert 7bdec2b`

soccer/gameplay/tactics/positions/submissive_goalie.py
```diff
@@ -37,6 +29,15 @@ class State(enum.Enum):
...
         super().__init__(continuous=True)
 
+        self.MaxX = constants.Field.GoalWidth / 2.0
+        self.SegmentY = constants.Robot.Radius + 0.01
+
+        # The segment we stay on during the 'block' state
+        # It's right in front of the goal
+        self.RobotSegment = robocup.Segment(
+        robocup.Point(-self.MaxX, self.SegmentY), robocup.Point(self.MaxX, self.SegmentY))
+        print("test",self.RobotSegment)
```
jgkamat: Can this print statement be removed?

PR #872, opened 2017-01-18
Timeout

### LINKED ISSUE #320: Add --timeout flag to simulator
This way we can optionally have the simulator exit after a few seconds of not receiving any messages from soccer.  See PR #319.

kevinatorchen:
Closes #320

### DIFF
--- a/simulator/SimulatorGLUTThread.cpp
+++ b/simulator/SimulatorGLUTThread.cpp
@@ -46,8 +46,7 @@ static void glutDisplayCallback(void) {
 
 SimulatorGLUTThread::SimulatorGLUTThread(int argc, char* argv[],
                                          const QString& configFile,
-                                         bool sendShared,
-                                         bool showWindow,
+                                         bool sendShared, bool showWindow,
                                          RJ::Seconds timeoutsimulator)
     : _argv(argv),
       _argc(argc),
--- a/simulator/SimulatorGLUTThread.cpp
+++ b/simulator/SimulatorGLUTThread.cpp
@@ -46,7 +46,7 @@ static void glutDisplayCallback(void) {
 
 SimulatorGLUTThread::SimulatorGLUTThread(int argc, char* argv[],
                                          const QString& configFile,
-                                         bool sendShared, bool showWindow)
+                                         bool sendShared, RJ::Seconds timeoutsimulator, bool showWindow)
     : _argv(argv),
       _argc(argc),
       _env(nullptr),
@@ -58,7 +58,7 @@ SimulatorGLUTThread::SimulatorGLUTThread(int argc, char* argv[],
       _maxCameraDistance(10.f),
       _showWindow(showWindow),
       _stopped(false) {
-    initialize(configFile, sendShared);
+    initialize(configFile, sendShared, timeoutsimulator);
 }
 
 SimulatorGLUTThread::~SimulatorGLUTThread() {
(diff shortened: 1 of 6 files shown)


jgkamat:
Thanks! I'm slightly bogged down by classes, so it might take me up to the weekend to look at it, sorry.

Thanks for the contribution, though! I'm ***REALLY*** looking forward to no more simulator processes hogging my cpu.

simulator/physics/Environment.cpp
```diff
@@ -44,6 +45,12 @@ Environment::Environment(const QString& configFile, bool sendShared_,
 
 Environment::~Environment() { delete _field; }
 
+/*
```
jgkamat: Can this be removed?

simulator/physics/Environment.hpp
```diff
@@ -76,6 +81,8 @@ class Environment : public QObject {
 
     void dropFrame() { _dropFrame = true; }
 
+    // RJ::Time timeFromUpdate();
```
jgkamat: Can this be deleted too?

simulator/simulator.cpp
```diff
@@ -64,6 +63,9 @@ int main(int argc, char* argv[]) {
         } else if (strcmp(argv[i], "--smallfield") == 0) {
             Field_Dimensions::Current_Dimensions =
                 Field_Dimensions::Single_Field_Dimensions * scaling;
+        } else if (strcmp(argv[i], "--timeout") == 0) {
+            timeoutsimulator = RJ::Seconds(stoi(argv[i + 1]));
```
jgkamat: You will need to do a check here to see if `argv[i+1]` is out of range or not. 

Right now this will cause the simulator to segfault if you run it like this: `./run/simulator --timeout` with no args after timeout.

simulator/SimulatorGLUTThread.cpp
```diff
@@ -46,7 +46,9 @@ static void glutDisplayCallback(void) {
 
 SimulatorGLUTThread::SimulatorGLUTThread(int argc, char* argv[],
                                          const QString& configFile,
-                                         bool sendShared, bool showWindow)
+                                         bool sendShared,
+                                         RJ::Seconds timeoutsimulator,
```
jgkamat: A slight nitpick, but could you make `timeoutSimulator` the last argument?

simulator/simulator.cpp
```diff
@@ -20,7 +20,6 @@ void quit(int signal) {
 void usage(const char* prog) {
     fprintf(stderr, "usage: %s [-c <config file>] [--sv]\n", prog);
     fprintf(stderr, "\t--help       Show usage message\n");
-    fprintf(stderr, "\t--sv         Use shared vision multicast port\n");
```
jgkamat: This line was deleted by mistake, could you add it back?

simulator/simulator.cpp
```diff
@@ -64,8 +65,10 @@ int main(int argc, char* argv[]) {
             Field_Dimensions::Current_Dimensions =
                 Field_Dimensions::Single_Field_Dimensions * scaling;
         } else if (strcmp(argv[i], "--timeout") == 0) {
-            timeoutsimulator = RJ::Seconds(stoi(argv[i + 1]));
-            i++;
+            if (i < argc - 1) {
+                timeoutsimulator = RJ::Seconds(stoi(argv[i + 1]));
+                i++;
+            }
```
jgkamat: In the case that this if statement dosen't run, we should fail and print the usage. There's a usage function above that should do thiss for us!

simulator/simulator.cpp
```diff
@@ -64,6 +65,15 @@ int main(int argc, char* argv[]) {
         } else if (strcmp(argv[i], "--smallfield") == 0) {
             Field_Dimensions::Current_Dimensions =
                 Field_Dimensions::Single_Field_Dimensions * scaling;
+        } else if (strcmp(argv[i], "--timeout") == 0) {
+            if (i < argc - 1) {
+                timeoutsimulator = RJ::Seconds(stoi(argv[i + 1]));
```
jgkamat: So sorry for not catching this earlier (this is kind of my bad), but stoi can throw an exception if we run 
`./run/simulator --timeout badinput`. 

Can you add a try catch around this, to print out an error message and print usage (or at least exit cleanly).

It seems like the error is `std::invalid_argument`

kevinatorchen:
Modified it. I created a method that would print the error message since I realized this is done in two circumstances.

PR #873, opened 2017-01-18
Add min iterations to rrt

justbuchanan:
As discussed in #697, the old rrt implementation always ran for a fixed number of iterations, while the new one takes the first path it finds. This PR adds a `minIterations` parameter to the rrt so it will now (optionally) search longer in order to try to find a better path.

This value is hardcoded in SingleRobotPathPlanner.cpp next to the old maxIterations parameter. It would probably be good to make these configurable via the gui so you can tune them more easily.

### DIFF
--- a/soccer/planning/LineKickPlanner.hpp
+++ b/soccer/planning/LineKickPlanner.hpp
@@ -19,7 +19,7 @@ namespace Planning {
  */
 class LineKickPlanner : public SingleRobotPathPlanner {
 public:
-    LineKickPlanner() : SingleRobotPathPlanner(false), rrtPlanner(250){};
+    LineKickPlanner() : SingleRobotPathPlanner(false), rrtPlanner(0, 250){};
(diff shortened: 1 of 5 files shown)


jgkamat:
Yah, I think it would be a good idea to add this to the existing config system if possible. We can leave that for later though. 

I'll take a look at this later this week! :smile:

soccer/planning/LineKickPlanner.hpp
```diff
@@ -19,7 +19,7 @@ namespace Planning {
  */
 class LineKickPlanner : public SingleRobotPathPlanner {
 public:
-    LineKickPlanner() : SingleRobotPathPlanner(false), rrtPlanner(250){};
+    LineKickPlanner() : SingleRobotPathPlanner(false), rrtPlanner(0, 250){};
```
jgkamat: This minIterations probably will have to be tuned later on..

PR #874, opened 2017-01-19
Change Kicking back to on Break Beam

ashaw596:
Any idea why it was changed in this commit? 
https://github.com/RoboJackets/robocup-software/commit/c609aa93c52077045937e0b5d8049a5353ba6764

I think its actually implemented in firmware. Can someone test this if they get a chance.

### DIFF
--- a/soccer/Robot.cpp
+++ b/soccer/Robot.cpp
@@ -292,14 +292,14 @@ void OurRobot::_kick(uint8_t strength) {
     uint8_t max = *config->kicker.maxKick;
     control->set_kcstrength(strength > max ? max : strength);
     control->set_shootmode(Packet::Control::KICK);
-    control->set_triggermode(Packet::Control::IMMEDIATE);
+    control->set_triggermode(Packet::Control::ON_BREAK_BEAM);
 }
 
 void OurRobot::_chip(uint8_t strength) {
     uint8_t max = *config->kicker.maxChip;
     control->set_kcstrength(strength > max ? max : strength);
     control->set_shootmode(Packet::Control::CHIP);
-    control->set_triggermode(Packet::Control::IMMEDIATE);
+    control->set_triggermode(Packet::Control::ON_BREAK_BEAM);
 }
 
 void OurRobot::_unkick() {


jgkamat:
@petersonev You might want to look at this

The original value (before I started messing around with it) was IMMEDIATE. I changed it because it didn't seem to be working (on break beam), but I think justin said that it wasn't implemented in firmware at the time. I /think/ this is the commit that was before I touched anything https://github.com/RoboJackets/robocup-software/commit/ba6ea678e5916418ea604476b5f822164b92e72f

I'm not sure what those two values do, but I think that this does need to be set to on_break_beam to get it to work (but it also needs to be implemented in firmware, which is why I reverted that at comp).

ashaw596:
I check firmware master and it looks like it was implemented I think.

chachmu:
Is the break beam currently working? We have had issues with them before.

petersonev:
Well we're not really running on the master branch of firmware right now (we're on decawave) not just because of change in radios but master is still using the actual kicker board and not the hacked kicker board firmware. But on the decawave branch kick on break beam should be working again and with this change soccer should work correctly with it too.

Also somewhat unrelated and not a big deal but it seems like some of the stuff defined in control (like the possible modes to send to the kicker) in software should be moved over to `rtp.hpp` so it can be shared in firmware because they use the same radio protocol.

And yes, break beam should be working. We've had some problems with them (we're looking into a better solution) but most of the problems on not working break beam is just because of broken flat flex cables.

ashaw596:
@petersonev Aren't all the protocol things already shared since they're in the common/protobuf?

petersonev:
Not really, like look at `enum Control_TriggerMode` in `Contol.pb.h` which isn't shared in common. In firmware we just have a comment saying what 0, 1, and 2 means for the value being received in the packet and use the constants 0, 1, or 2 whenever we're referring to them. Same thing for the song and as we add more stuff to packets it seems like it should all be shared in `rtp.hpp`

ashaw596:
Control.pb.h is generated by https://github.com/RoboJackets/robocup-common/blob/master/protobuf/Control.proto in common. Firmware should be using the protobuf definitions to read the radio packets.

petersonev:
@jgkamat @ashaw596 What do you guys need in the kicker status stuff I'm adding to the radio packets? @jgkamat mentioned that it would be helpful to have some info on the status of the kicker for kicking on break beam.

jpfeltracco:
So that we can get tlthe decawave stuff into master I'm going to add a flag to allow us to use a spoofed kickerboard

ashaw596:
@petersonev  Not 100% sure yet. Need to take a look at the justkicked function for a bit and see how it use to work. One sec. Btw, I'm rather confused. Whats the purpose/difference between RTP and our Protobuf files? It says its a "// binary-packed version of Control.proto"? Do we need another copy? Aren't they the same?

ashaw596:
@petersonev  So I think we use to send the voltage reading on the kicker in either kicker_status  or kicker_voltage in https://github.com/RoboJackets/robocup-common/blob/899ac43050e221767296b6d6f4d6e387a45a7462/protobuf/RadioRx.proto . I heard its impossible for us to read that anymore or something. Either way, we just need a way to receive when we just kicked the ball. Maybe set kicker status to 0 for a couple packets right after a kick and to 1 when charged/charging or something.

petersonev:
Gotcha so the main thing you need is to have a way to know when you kicked the ball. I'll try to figure out a little more reliable way of doing that. May be easiest just to send back time since last kick (which is somewhat proportional to voltage of the kicker)

jgkamat:
'Time of last kick' might be better than 'time since last kick' since that would be a bit more stable, but I'm not sure how time is structured in firmware.

(Another alternative is to just list the # of kicks made).

ashaw596:
Well for the kick on breakbeam. I tested on robots(I assume on decawave and it works.

Gonna merge.

Issue #875, opened 2017-01-19
Make shotObstacle a dynamic obstacle so robots can still go through before and after the shot.


ashaw596:
Currently we just block off giant parts of the screen and almost crash our pathplanner due to RRT failures.

jgkamat:
To make this dynamic we need to know when we kick, which might be a problem figuring out. We could fudge this, but idk.

ashaw596:
Yeah, the idea would be that we should know when we're going to kick. I'm halfway there on the line kick function and its easy to modify pivot kick to do this.

PR #876, opened 2017-01-19
Temporary workaround for shotObstacle

ashaw596:
Only enables the shotObstacle during the kicking and aiming phase of pivot kick instead of the entire time for kicking. Just prevents the constant impossible paths when most of the screen is blocked off while the robot isn't even near the ball while capturing.

Should be replaced with #875

### DIFF
--- a/soccer/gameplay/skills/pivot_kick.py
+++ b/soccer/gameplay/skills/pivot_kick.py
@@ -111,7 +111,10 @@ class PivotKick(single_robot_composite_behavior.SingleRobotCompositeBehavior,
             self.remove_subbehavior('aim')
 
     def on_enter_capturing(self):
+        # We disable the shot obstacle in this step just in case we had a
+        # failure and we're restarting this play
         self.enable_shot_obstacle = False
+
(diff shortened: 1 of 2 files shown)


jgkamat:
I'm interested in why the diff is so big, but I'll figure this out when I take a look at this.

I'll try to get to this this weekend, I'm slightly backed up right now, sorry! :cry:

ashaw596:
Haha woops. One sec

ashaw596:
Ok there you go. 3 lines.

soccer/gameplay/skills/pivot_kick.py
```diff
@@ -111,6 +111,7 @@ def remove_aim_behavior(self):
             self.remove_subbehavior('aim')
 
     def on_enter_capturing(self):
+        self.enable_shot_obstacle = False
```
jgkamat: This is slightly vauge (why its being done).
Could you add a comment saying something like:
"If kick fails, disable shot obstacle for second try"

PR #880, opened 2017-01-22
Dark ui

Jason27chan:
Allows user to switch between no theme and dark theme. 
User can also set default theme using environmental value "SOCCER_THEME"

### DIFF
--- a/soccer/ui/StyleSheetManager.cpp
+++ b/soccer/ui/StyleSheetManager.cpp
@@ -2,19 +2,18 @@
 #include "StyleSheetManager.hpp"
 #include <Utils.hpp>
 
-StyleSheetManager::StyleSheetManager(QMainWindow* window) {
-	win =  window;
-}
+QString StyleSheetManager::DARK = "../soccer/ui/QTDark.stylesheet";
+QString StyleSheetManager::NONE = "";
 
-void StyleSheetManager::changeStyleSheet(QString name) {
+void StyleSheetManager::changeStyleSheet(QMainWindow* window, QString name) {
 	if (name.compare("NONE") == 0) {
-		win->setStyleSheet(NONE);
+		window->setStyleSheet(NONE);
 	}
 	else if (name.compare("DARK") == 0) {
 		QFile file(ApplicationRunDirectory().filePath(DARK));
     	file.open(QFile::ReadOnly);
   		QString styleSheet = file.readAll();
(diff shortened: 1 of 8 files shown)


jgkamat:
I think there was a small mistake in this PR (it seems you committed some referee changes), which I've cleaned up in the last commit. Once we're done, I'll 'rebase' this branch onto master, so that those Merge commits are cleaned up!

I'll review this as normal until then. Thanks so much! This looks really great! :smile:

soccer/ui/StyleSheetManager.cpp
```diff
@@ -0,0 +1,26 @@
+
+#include "StyleSheetManager.hpp"
+#include <Utils.hpp>
+
+// To add a new style sheet, declare the static variable 
+// in the header file and initialize the path here
+QString StyleSheetManager::DARK = "../soccer/ui/QTDark.stylesheet";
+QString StyleSheetManager::NONE = "";
```
jgkamat: Right now, you are storing each stylesheet in its own variable. It's better to put them into a hash table, which makes more logical sense, is cleaner, and is faster.

[Here's an example of using a c++ hashtable](https://en.wikipedia.org/wiki/Unordered_associative_containers_(C%2B%2B)#Usage_example). It might be a little bit unfamiliar, but I can help you understand what's going on!

soccer/ui/StyleSheetManager.cpp
```diff
@@ -0,0 +1,26 @@
...
+
+void StyleSheetManager::changeStyleSheet(QMainWindow* window, QString name) {
+	// Add new cases here
+
+	if (name.compare("NONE") == 0) {
```
jgkamat: You should be able to replace this with `name == "NONE"` (which is cleaner).

soccer/ui/StyleSheetManager.cpp
```diff
@@ -0,0 +1,26 @@
...
+		window->setStyleSheet(NONE);
+	} else if (name.compare("DARK") == 0) {
```
jgkamat: Same as above.

jgkamat:
Also, one more thing. 

It would be really nice to add a docs page for style changes (where you could add instructions on adding a theme). Let me know if you're willing to write that, otherwise I'll do it later on.

jgkamat:
@Jason27chan I need to explain a little to you about git, don't touch this until tomorrow :P

PR #882, opened 2017-01-25
Simulator dropdown menu is now disabled when not in simulation

### LINKED ISSUE #858: The simulator bar should be disabled when running with real robots
The simulator menu contains several buttons that really only make sense when running make run-sim. The whole bar should be disabled when running soccer for real. The name of the bar can be found [here](https://github.com/RoboJackets/robocup-software/blob/d18e9b7fd56f42e1888e494064e0c264a9cceae8/soccer/ui/MainWindow.ui#L1946). QT items can be disabled with .setEnabled(false), like [this](https://github.com/RoboJackets/robocup-software/blob/d18e9b7fd56f42e1888e494064e0c264a9cceae8/soccer/MainWindow.cpp#L169) which should be done in the MainWindow constructer [here](https://github.com/RoboJackets/robocup-software/blob/2a5fed376542c2dcf5117fd26db1509e56e769fc/soccer/MainWindow.cpp#L45). Finally, the flag to check if the program is running in simulation is _processor->simluation().
simulatormenu

mmurley3:
Closes #858

### DIFF
--- a/soccer/MainWindow.cpp
+++ b/soccer/MainWindow.cpp
@@ -153,6 +153,10 @@ MainWindow::MainWindow(Processor* processor, QWidget* parent)
     // current Git index is dirty
     setWindowTitle(windowTitle() + " @ " + git_version_short_hash +
                    (git_version_dirty ? "*" : ""));
+
+    if (!_processor->simulation()) {
+        _ui.menu_Simulator->setEnabled(false);
+    }
 }
 
 void MainWindow::configuration(Configuration* config) {


soccer/MainWindow.cpp
```diff
@@ -153,6 +153,10 @@ MainWindow::MainWindow(Processor* processor, QWidget* parent)
     // current Git index is dirty
     setWindowTitle(windowTitle() + " @ " + git_version_short_hash +
                    (git_version_dirty ? "*" : ""));
+
+    if (!_processor->simulation()) {
```
jgkamat: It would be nice if you could add a small comment here :stuck_out_tongue:

PR #887, opened 2017-01-29
Tune PID automatically

### LINKED ISSUE #291: Use genetic algorithms for tuning motion control parameters
This would go well as a tuning 'play'.  It can start with an initial set of values for the parameters and run a robot along a test trajectory.  As the robot runs, we'll integrate the robot's tracking error over time, which will be the 'fitness function' for the GA.

[scikit-learn](http://scikit-learn.org/stable/) would probably be a good candidate for the genetic algorithm library.

A couple things that need to happen before we can do this:
- [ ] be able to set robot tuning parameters from soccer - this will require an addition to the radio protocol (#304)
- [ ] be able to query and/or set a robot's path from python
- [ ] add a set 'model parameters' and 'tuning parameters' to each robot in soccer.
- [ ] per-robot configurations

chachmu:
This will allow us to automatically tune per robot pid using a play. Closes #291.

- [x]  Make play to move back and forth along line
- [x]  Function to analyze the effectiveness of a certain set of pid values
- [x]  System to determine and test new pid values
- [x]  Find a way to have consistent robot movement for testing


jgkamat:
For the function to analyze pid values and determine new pid values, Try to keep as much of it in common as possible (so that we can reuse it later in firmware as well). Thanks!

chachmu:
@jgkamat Right now I am just storing the position of the robot each frame in the python side. Is there an easy way to get that information directly from the robot? If not this gets around 220 data points per trip horizontally across the field so its not too bad, do you think thats enough?

chachmu:
Testing setting the PID from python, if you want to see what our robots look like when drunk, run the pid play once and then run something else, it's pretty funny

jgkamat:
Hmm, I'm not sure, but I think that would be enough. 

Another thing is that I think this is the wrong approach to getting data (I think). We want to get a integration of the error of the PID controller in soccer, not neccescarily the error of movement. This might work (I'm not sure), but I think a cleaner way of doing this would be do log the error in the PID controller itself, while you were running a play like this.

I also wouldn't reccomend doing a full lineup,  maybe we could use a quarter of the field?

We should talk more about this on tuesday.

chachmu:
My current plan was to just have the other robots get out of the way and then run 1 at a time to tune, so we could put them anywhere on the field. I will look into getting error straight from the PID controller. Presumably we would just want to minimize the error from the PID controller if we do that?

jgkamat:
Yup, that was exactly what I was thinking. This play seems good, but the actual values you want to minimize should be the ones in the PID controller itself (not time to destination) since the path planner could just decide to be slower at times. The PID controller is in common  (and thats where any pid decisions should happen) so I think it shouldn't be too hard. We can tie the optimizer to our PID implementation, so that anything using our PID implementation (including firmware) can benifit from this.

chachmu:
Should I just write the PID tuning stuff direcly into the Pid.hpp and Pid.cpp files or should I make a seperate set of files for the tuning functions?

chachmu:
@jgkamat Would we be trying to minimize _errSum (the error used on the i component) since it would be the one that would take into account the error across the whole path?

jgkamat:
I think that it would be best to seperate the tuning parts into seperate files (maybe classes that take in PID objects?)
I think that's the correct error. You'll want to sum up the error [passed into here](https://github.com/RoboJackets/robocup-common/blob/master/Pid.cpp#L35), and that would become the value you're trying to minimize. 

I also would reccomend starting this outside of soccer completely (maybe with some sample data?) but I'm not sure if that's a good idea or not.

jgkamat:
this PR is linked to robojackets/robocup-common#8. I'll review this one after the common one (since this depends on the common changes).

soccer/gameplay/plays/tuning/pid.py
```diff
@@ -0,0 +1,80 @@
...
+        self.add_transition(
+            pid.State.prep, pid.State.testing,
+            lambda: self.has_subbehavior_with_name('move2') and self.subbehavior_with_name('move2').state == behavior.Behavior.State.completed and self.subbehavior_with_name('move2').robot.vel.mag() < .05,
+            'finished moving')
+
+        self.add_transition(pid.State.testing,
+                            behavior.Behavior.State.completed,
+                            lambda: self.check_continue(), 'command changed')
+
+        self.positions = []
```
JNeiger: This is never used

soccer/gameplay/plays/tuning/pid.py
```diff
@@ -0,0 +1,80 @@
...
+
+    def create_lineup(self):
+        xsize = constants.Field.Width / 2 - .5
+
+        return robocup.Segment(
+            robocup.Point(xsize, .25), robocup.Point(xsize, 1.5))
+
+    def check_continue(self):
+        return False
```
JNeiger: Is this going to be used for when multiple robots are added?

soccer/gameplay/plays/tuning/pid.py
```diff
@@ -0,0 +1,80 @@
...
+        self.add_subbehavior(line_up, 'line_up', required=True, priority=80)
+
+    def execute_prep(self):
+        if self.has_subbehavior_with_name(
+                'move1') and self.subbehavior_with_name(
+                    'move1').state == behavior.Behavior.State.completed:
+            self.remove_subbehavior('move1')
+            xsize = constants.Field.Width / 2
+            move2 = skills.move.Move(robocup.Point(-xsize + .1, 2))
+            self.add_subbehavior(move2, 'move2', required=True, priority=100)
```
JNeiger: Can these two move statements be combined into a single one?

soccer/gameplay/tactics/tune_pid.py
```diff
@@ -0,0 +1,76 @@
...
+
+class Tune_pid(single_robot_composite_behavior.SingleRobotCompositeBehavior):
+    class State(enum.Enum):
+        tune = 1
+        process = 2
+
+    def __init__(self):
+        super().__init__(continuous=False)
+
+        self.pause_time = 0
```
JNeiger: This is never used

soccer/gameplay/plays/tuning/pid.py
```diff
@@ -0,0 +1,80 @@
...
+
+        for state in pid.State:
+            self.add_state(state, behavior.Behavior.State.running)
+
+        self.add_transition(behavior.Behavior.State.start, pid.State.prep,
+                            lambda: True, 'immediately')
+
+        self.add_transition(
+            pid.State.prep, pid.State.testing,
+            lambda: self.has_subbehavior_with_name('move2') and self.subbehavior_with_name('move2').state == behavior.Behavior.State.completed and self.subbehavior_with_name('move2').robot.vel.mag() < .05,
```
JNeiger: Is there a way to format this so the line isn't as long? Either split it up into multiple lines or into a function?

soccer/gameplay/tactics/tune_pid.py
```diff
@@ -0,0 +1,76 @@
...
+
+        self.add_transition(Tune_pid.State.process, Tune_pid.State.tune,
+                            lambda: self.tune and time.time() - self.pause > 1,
+                            'continue tuning')
+
+        self.add_transition(Tune_pid.State.process,
+                            behavior.Behavior.State.completed,
+                            lambda: not self.tune, 'done tuning')
+
+        self.positions = []
```
JNeiger: This is never used

soccer/motion/MotionControl.hpp
```diff
@@ -32,6 +32,11 @@ class MotionControl {
      */
     void run();
 
+    /**
+     *This returns the PID controller denoted by the char passed in
+     */
+    Pid* getPid(char controller);
```
JNeiger: I would add comments about the possible controller options

soccer/gameplay/plays/tuning/pid.py
chachmu: :O I completely forgot to update that part... Good catch. I definitely need to go back through this whole play and recheck it, I was pretty focused on getting some good test data from a single robot in time for the TDP. I will be going through and updating this to cycle through all of the robots.

soccer/gameplay/plays/tuning/pid.py
chachmu: Given more recent code changes I think so. Originally when it was using the normal path planner if the robot didn't approach the point perfectly horizontal, it would plan an arc instead going in the right direction immediately which threw off the error detection.

soccer/gameplay/plays/tuning/pid.py
chachmu: This should be fixed once I remove the second move statement

jgkamat:
Copying from email:

> When we tune PID's, we first set the velMultiplier/accelMultipler so that we basically get 0 error (or as close to 0 as possible), so for you, even if you have terrible PID values, it dosen't matter, since the PID isn't being used at all. To actually test PID, we need to introduce error into the system. I think this means we have to alter our approach to this. This approach is perfect for tuning the accelMultipler and velMultipler parameters (since we want to have 0 error when possible), but when we tune PID, we have to intentonally introduce error. In my example, I set accelMultipler/velMultipler to 1 (actually, I screwed up and set accel to 0 but whatever) which introduced error into the system, allowing the PID to work. This is a pretty hacky way to do this though. The best way to do it would be to send NO PATH AT ALL, and let the PID do the movement all by itself. I'll explain this more when we meet next. For now, I would try finding the best possible velMultipler/accelMultipler values instead (since this is perfect for that). I suspect they would be the same as what we have, since we did a LOT of tuning in the simulator. 

In order to 'introduce error' into the system, we can simply not let a path plan and let PID do 100% of our movement. Optimizing this will probably optimize our general PID as well.

To do this (super hackily) you can apply this patch:


[pidonly.txt](https://github.com/RoboJackets/robocup-software/files/870384/pidonly.txt)

We can probably create a new command in robot.hpp (along with a bunch of other places) which is a 'stronger' version of movedirect, which creates no path at all, and instead just assumes the robot is at the target position (which is effectively what this does). We'll need to make the movement distances much less though. 
 
However, this method of 'introducing error' might not produce a good set of PID values for us just moving around, because this is a much more extreme case than normal. We'll have to test it out and see...

jgkamat:
@chachmu should we target this to be merged before comp? I think we shoud, even if it's not finished, because it would make merging this into master eventually a lot easier.

Thoughts?

chachmu:
I completely agree, I was hoping to get it to the point that it would at least somewhat reliably pick a value but at this rate even that probably won't happen. I will probably end up merging what is done soon and start working on the ball placement since that will actually be used in comp.

chachmu:
I will probably put this back to the normal straight line testing and then merge it tonight unless anyone has any objections.

chachmu:
@jgkamat Does this look good to merge?

soccer/gameplay/skills/move_direct.py
```diff
@@ -42,6 +43,12 @@ def threshold(self):
     def threshold(self, value):
         self._threshold = value
 
+    def velocity_check(self):
+        if (self.check_velocity):
+            return self.robot.vel.mag() < .05
```
jgkamat: This `0.5` is in a ton of places, can we make it  a constant instead (maybe in constants.py) so we can tune it from one place?

soccer/motion/MotionControl.cpp
```diff
@@ -240,3 +240,17 @@ void MotionControl::_targetBodyVel(Point targetVel) {
     _robot->control->set_xvelocity(targetVel.x());
     _robot->control->set_yvelocity(targetVel.y());
 }
+
+Pid* MotionControl::getPid(char controller) {
+    // just in case there is confusion
+    switch (tolower(controller)) {
+        case 'a':
```
jgkamat: This is super hacky (passing chars in here). I'm not sure how the python bridge would handle enums though, so I'll allow it for now. If you have some spare time could you look into that?

chachmu: It would probably be easier to fix after comp especially since I also have no clue how enums would work across the bridge.

soccer/gameplay/skills/move_direct.py
chachmu: I feel like that might be a larger issue since there is a threshold to check if the robot's velocity is close to zero in like half of our plays

jgkamat: agreed, that should be a bigger issue. We'll target to have it fixed before or during fall 2017 then.

PR #890, opened 2017-02-01
Fix passing issues in two side attack

### LINKED ISSUE #888: Two side attack play doesn't consider passing
When the center robot has the ball at the beginning of the play, it should consider whether to pass to the left robot up the field, the right robot up the field, or take a direct shot on goal, if available. Right now, the center robot always takes a shot on goal even if an opposing robot is directly between it and the goal and passing to the outside would be a much better option.

petschekr:
Fixes a bug where the center robot would always take a direct shot rather than passing to the side, even when the direct shot was infeasible. This was caused by a faulty state transition condition that always returned `True`. Therefore, the play never spent any time in the `passing` state and transitioned immediately from `setup` to `kicking`.

Closes #888

### DIFF
--- a/soccer/gameplay/plays/offense/two_side_attack.py
+++ b/soccer/gameplay/plays/offense/two_side_attack.py
@@ -46,7 +46,8 @@ class TwoSideAttack(standard_play.StandardPlay):
 
         self.add_transition(
             TwoSideAttack.State.passing, TwoSideAttack.State.kicking,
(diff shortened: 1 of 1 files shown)


jgkamat:
The style checker seems to be stupid and wants that statement on one line (Even though we have a minimum line set much shorter than that). I'm merging this in anyway :P.

Issue #883, opened 2017-01-25
Make behavior-diagrams throws a bunch of errors


Jason27chan:

jgkamat:
An example of the errors being thrown:

```
ERROR:root:Error generating fsm diagram for behavior 'Basic122':'NoneType' object has no attribute 'isChecked'
Traceback (most recent call last):
  File "generate_fsm_diagrams.py", line 37, in <module>
    klass().write_diagram_png(filepath)
  File "/home/jay/Code/robocup-software/soccer/gameplay/plays/offense/basic_122.py", line 27, in __init__
    super().__init__(continuous=False)  # FIXME: continuous?
  File "/home/jay/Code/robocup-software/soccer/gameplay/standard_play.py", line 17, in __init__
    self.use_standard_defense()
  File "/home/jay/Code/robocup-software/soccer/gameplay/standard_play.py", line 24, in use_standard_defense
    if ui.main.defenseEnabled() and not self.has_subbehavior_with_name(
  File "/home/jay/Code/robocup-software/soccer/gameplay/ui/main.py", line 24, in defenseEnabled
    return _defense_checkbox.isChecked()
AttributeError: 'NoneType' object has no attribute 'isChecked'
```

jgkamat:
Fixed in #895

PR #895, opened 2017-02-05
Kil behavior errors

Jason27chan:
the "make behavior-diagrams" command was throwing a series of errors because some classes required the GUI to be instantiated before the classes could be instantiated
To fix this, we created some mock objects that could be substituted. Ex. mock ball at position 0,0.
One error still remains because it is very difficult to make a mock game_state


soccer/gameplay/plays/restarts/their_kickoff.py
```diff
@@ -31,7 +31,7 @@ def __init__(self):
         self.add_subbehavior(
             tactics.stopped.circle_on_center.CircleOnCenter(
                 # TODO find a way to do this without hard coding 3 defense/goalie robots (or make those constants)
-                min_robots=1 if len(main.our_robots()) > 3 else 0),
+                min_robots = 3 if (main.our_robots() is not None) and len(main.our_robots()) > 3 else 0),
```
jgkamat: Should this be`1 if (main.our_robots() is not None)` instead of `3`?

chachmu:
Don't forget to add "closes #883" so it links the issue with the pull request.

soccer/gameplay/generate_fsm_diagrams.py
```diff
@@ -21,8 +20,20 @@ def mkdir_p(path):
 logging.getLogger().setLevel(logging.INFO)
 
 import main
+import ui.main
 main.init()
 
+class moc_Point():
```
justbuchanan: If you `import robocup` at the top of the file, you can use `robocup.Point` rather than creating a mock class here. Probably the same with the Ball mock too.

jgkamat: Unfortunately we can't do this (at least with our current setup, which is the root of this problem in the first place). The generate-behavior-diagrams script runs python on these files directly (which means it's not stated via boost python, and the entire python-cpp bridge is unavailable). 

I was thinking about trying to get a cpp target that we could call (and start python and run the behhavior diagram scripts) which would allow us to get everything we need, but I don't think thats feasable in the short term (although it would be a good intermediate project for those who want to learn about the python bridge). I could be overlooking something, so let me know if I'm just being stupid :P

justbuchanan: Hmm, I think you can use it since the `behavior-diagrams` makefile target depends on `all`, which ensures that the robocup python module is compiled beforehand. Also most of the plays/behaviors `import robocup` already. I think the catch here is that the robocup module has to be imported after the `sys.path.append('../../run')` line. Let me know if that doesn't work though.

jgkamat: Oh, I think I get it now, that lets python import the cpp bindings from the `robocup.so` file in run? That actually makes a lot of sense now that I think of it (so python isn't started specially from cpp at all, but just loads that so file). 

Adding:
```
sys.path.append('../../run')
import robocup
```
Works fine! @Jason27chan You can probably remove all these moc classes and replace them with real classes, as suggested above.

justbuchanan: Yep, that's pretty much it. Most of the python code can run completely independently of soccer just by importing the robocup.so file as a python module, which contains every class, function, etc that we setup with boost python bindings. At some point, I should probably write up a little more documentation on how that stuff works since it's pretty tough to figure out just from the code.

One thing you can do if you want to play around with it is to run python from the robocup-software directory with the appropriate paths added. It can be useful for testing code out quickly or viewing documentation.
```shell
make # ensure robocup.so python module is compiled and put in ./run
PYTHONPATH=run:soccer/gameplay python
```

Then you can use run robocup code in the python console:
```python
import robocup
p = robocup.Point(0.5, 0.5)
help(p)
help(robocup.SystemState)

import behavior
help(behavior)
```

jgkamat:
This PR was cherry picked into master as of 07ba292992bbc7516579a361e4cb2cad0fc63311. 

Thanks for the contribution! :smile: 

Also, all the diagrams are finally building for me :tada:

Issue #298, opened 2015-04-26
Improve evaluation of shots and passes


justbuchanan:

justbuchanan:
One thing that we could use here that we don't do currently is model the % chance of kicking where we aim as a gaussian as described [here](https://books.google.com/books?id=oO6YNGtlzbcC&pg=PA298&lpg=PA298&dq=robocup+shot+gaussian&source=bl&ots=9xo_iq6A8I&sig=WKDCuM5wph57wJdpNzo_l1Z3vPA&hl=en&sa=X&ei=A6o9VZvjDZLOgwTi3oCYBA&ved=0CB4Q6AEwAA#v=onepage&q=robocup%20shot%20gaussian&f=false).  That's an old link, but the math is pretty timeless.  The basic idea is that when we aim for a certain shot angle, the angle the ball actual leaves the robot at will follow a normal distribution centered at the aimed-at angle.  The key here is to figure out the std dev of this distribution, which is something that we'll have to measure.  Once we know that, we can do a normalcdf using the angular width of the segment available to shot at, which gives us the chance that we'll make the shot.

ashaw596:
Once we figure out the variance and mean (I mean I hope it kicks straight
on avarage), boost seems to have a good implementation for this (assuming
it follows a normal distribution).
http://www.boost.org/doc/libs/1_42_0/libs/math/doc/sf_and_dist/html/math_toolkit/dist/dist_ref/dists/normal_dist.html

PR #896, opened 2017-02-07
Implemented Kick Eval

### LINKED ISSUE #892: Switching from Win_Eval to a new method
I think I've found a better way to calculate whether we are clear for a shot / pass than win_eval. Additionally, it has a few more options to change sensitivity in comparison and we can deal with the case of a moving robot much more easily. The only difference is that this would only return the chance, it would not return the most optimal windowed area.

At least in the matlab implementation of each, the new method preforms significantly faster. In the order of .75 seconds while the win_eval takes about 34 seconds for all ~5000 points in the graph. If this speed difference transfers over to the soccer, I think the added complexity of two different functions might be worthwhile. 

In the image below, the dark parts in the "Space" graph are where the robots are. Yellow is 100% while blue is 0%. The pass target is the bottom center and the color at each point represent the pass chance.

image

Thoughts?

[Here is a link](https://github.com/JNeiger/robocup-software/blob/adaptive_formation/soccer/gameplay/evaluation/defensive_positioning.py#L78) to the current iteration. It's a part of #822 right now. I'll get around to cleaning it up eventually.

JNeiger:
This is the window_evaluator alternative that calculates the probability of a pass succeeding to a specific segment based upon the standard deviation of the kick as well as the possible movement of other robots.

- [x] Implement actual test cases
- [x] Double check correctness

Closes #892 and maybe #298


justbuchanan:
FWIW, this reminds me of [CMDragons' ETDP](http://wiki.robocup.org/images/4/4a/Small_Size_League_-_RoboCup_2014_-_ETDP_CMDragons.pdf) from a couple years ago. Check out section 3/3.1. Theirs isn't exactly the same, but the idea of viewing passing/shooting as a statistical problem instead of just a geometric one is fairly similar. It's a neat approach!

JNeiger:
Funny enough, I think I actually read that paper when I was first starting and used the general statistical approach as a starting point.

jgkamat:
This PR is pretty big, so it might be a while before I can review this... Maybe @joshhting or @jpfeltracco could give it a look over as well (to speed it up)?.

Are you using this to replace the window evaluator, or are you just using this to evaluate shots/passes? (ie: were you able to find a good solution to the issues we discussed).

JNeiger:
I'm just using it to evaluate shots/passes right now. I'm still looking into a way of combining everything together. I have a couple ideas but no promises yet.

JNeiger:
Closing in favor of #915

PR #897, opened 2017-02-08
Stopped CoordinatedPass from restarting behaviors when robots are switched mid-pass

### LINKED ISSUE #755: Restarting single robot behaviors isn't always the best thing to do
In coordinated pass, if the receiver is temporarily assigned to a different robot (which happens occasionally), the receiver gets restarted, which messes up coordinated pass transitions, causing the pass to hang indefinitely.

Restarting a single robot behavior when a new robot is re-assigned is often the correct thing to do, but not always.  This needs some thought.

jcarn:
Closes #755

### DIFF
--- a/soccer/gameplay/single_robot_composite_behavior.py
+++ b/soccer/gameplay/single_robot_composite_behavior.py
@@ -10,6 +10,27 @@ import role_assignment
 # Note: The method-resolution-order in python dictates that as we have it listed below, SingleRobotBehavior methods take precedence over CompositeBehavior methods
 class SingleRobotCompositeBehavior(single_robot_behavior.SingleRobotBehavior,
                                    composite_behavior.CompositeBehavior):
+    ## Constructor.
+    # If the behavior should not be restarted when it is assigned to a different robot, then autorestart must be false
+    # @param continuous should the behavior runs until it is manually stopped
+    # @param autorestart function governing the behavior's restarting itself when its robot is switched
+    def __init__(self,
+                 continuous=False,
+                 autorestart= lambda: True):
+        single_robot_behavior.SingleRobotBehavior.__init__(
+            self,
+            continuous=continuous)
+        composite_behavior.CompositeBehavior.__init__(self,
+                                                      continuous=continuous)
+        self.autorestart = autorestart
+
+    @property
+    def autorestart(self):
+        return self._autorestart
+
+    @autorestart.setter
+    def autorestart(self, value):
+        self._autorestart = value
 
     ## we over-ride this to enforce the rule that there can't be more than one subbehavior
     def add_subbehavior(self, bhvr, name, required=True, priority=100):
@@ -27,7 +48,7 @@ class SingleRobotCompositeBehavior(single_robot_behavior.SingleRobotBehavior,
             reqs = composite_behavior.CompositeBehavior.role_requirements(self)
             if self.robot != None:
                 for req in role_assignment.iterate_role_requirements_tree_leaves(
-                        reqs):
+                    reqs):
                     req.previous_shell_id = self.robot.shell_id()
 
             return reqs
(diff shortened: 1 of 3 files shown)


soccer/gameplay/single_robot_composite_behavior.py
```diff
@@ -10,6 +10,24 @@
 # Note: The method-resolution-order in python dictates that as we have it listed below, SingleRobotBehavior methods take precedence over CompositeBehavior methods
 class SingleRobotCompositeBehavior(single_robot_behavior.SingleRobotBehavior,
                                    composite_behavior.CompositeBehavior):
+    #If the behavior should not be restarted when it is assigned to a different robot, then restart_behavior_on_switch_robot must be false
+    def __init__(self,
+                 continuous=False,
+                 restart_behavior_on_switch_robot=True):
```
jgkamat: eeeeh can this be renamed? Maybe `restart_on_switch` or maybe `autorestart`?

This is /really/ long. I would also reccomend adding documentation [see this guide](https://robojackets.github.io/robocup-software/md_doc_Documentation.html) for this method too.

soccer/gameplay/single_robot_composite_behavior.py
```diff
@@ -56,7 +74,7 @@ def assign_roles(self, assignments):
         # was different from the one that just ran Capture. Here we just restart the behavior
         # If it gets assigned a new robot.
```
jgkamat: Can you modify this comment a little bit to reflect the new contents?

soccer/gameplay/single_robot_composite_behavior.py
jgkamat: Actually, on further thought, it might be better to take in a function here, so that plays can fine tune if they want to be reset or not. I'm not sure how this would work though (run the function every tick, or should we run it whenever we have a robot change). This needs to be thought out more, but I think it would be great to have it in here.

jcarn:
Changed the autorestart parm to a function. We don't need to do any hacky stuff regarding it being a boolean or a function because I just made its default value a lamda function that always returns True.

jgkamat:
I think that your master might have some commits that aren't on rj/master (since there are more commits/changes appearing in this pr). Can you give me the last commit you worked on, and I'll rework this branch to exclude those changes.

jcarn:
Yeah it looks like I accidentally did a git add .
I'll save you some effort and fix it myself, hopefully I'll learn something about git in the process

jcarn:
Didn't mean to close, things should be fixed

jgkamat:
I'm not exactly sure what you did, I'll need to hear it from you in person :P 
The two git commands I would have used would be `git revert --hard` to throw away bad commits and `git cherry-pick` to copy commits onto this branch (if I had needed to). If you have a single commit with both bad and good changes, you can split them in an interactive rebase `git rebase -i` but that is a little harder to do. Antother way to split commits is to first do a `git reset --soft` to throw away the commits, then only stage the hunks for the first commit (I think `git add -p`?) and then repeat for the second commit. I'll take a look at this soon! :smile:

soccer/gameplay/single_robot_composite_behavior.py
jgkamat: @jcarn did you decide against the lambda option (and renaming this variable?) I think that maybe some of your commits got dropped when you pushed the new version, I can help you recover them if you flag me down the next time you see me. 

I'm fine with leaving out the lambda for now, but I'd really like this name to be shorter :P

jcarn:
Dunno why it is closing like that but whatever.Anyway this is 100% right now. My local and remote repos were in a bit of a state, but I think I've resolved all of that. The files here look good, everything should work now

soccer/gameplay/skills/pass_receive.py
```diff
@@ -88,6 +89,11 @@ def __init__(self, captureFunction=(lambda: skills.capture.Capture())):
             lambda: self.subbehavior_with_name('capture').state == behavior.Behavior.State.failed or self.check_failure() or time.time() - self.kicked_time > PassReceive.DesperateTimeout,
             'ball missed :(')
 
+    ## This function will detirmine how this behavior reacts to its robot changing. If the ball is kicked and the robot switches, restarting causes all sorts of hitches, so we prevent this.
+    # Note: False means the behavior will not restart
+    def autorestart():
```
jgkamat: This needs to have a self in the argument list!

jgkamat:
I'm going to cherry pick this into master after fixing the self issue. This is looking good!

PR #904, opened 2017-02-12
Switch to grSim

### LINKED ISSUE #105: Switch to grSim
Our simulator works pretty well, but the tuning isn't quite right and it's a little buggy.  There's an open source simulator called 'grSim' that's used by several teams and is pretty good.  Switching to it would make it easier to maintain our project and give us some better results.

### LINKED ISSUE #246: Add 2015 bot to simulator
Our simulator looks at a config file at startup that lists what robots to put where on the field.  It currently differentiates between 2008 and 2011 bots, so we should add an option and a 3d model for the 2015 bot.

jpfeltracco:
Closes #105 

Todo:
- [x] Get kick strength conversion correct
- [x] Add a RJ2015 model to grSim?
     - [x] Tune whichever model we end up using
- [x] Build grSim as submodule
    - [x] Add vartypes as grSim submodule

[edit by jay]: closes #246. We will need to do a pass over all the other sim issues and close them when this is merged as well.


jgkamat:
I didn't notice this before, but because [grSim is licensed under gpl3](https://github.com/mani-monaj/grSim/blob/master/LICENSE.md), we'll have to relicense robocup-software under gpl3, which is allowed (since apache2 is gpl3 compatible). We'll probably want to talk to will and the rest of the team about this, but grSim is definetly worth it, in my opinion.

jpfeltracco:
We aren't linking to grSim though, just building it during our build process. Isn't linking the critical step?

jgkamat:
Oh, I thought we were linking to it (I'm pretty sure we link to our simulator too). I think it should be fine if we use grSim as a submodule and build it independently (since a submodule let's users clone the repo themselves).

CMakeLists.txt
```diff
@@ -6,7 +6,7 @@ set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_SOURCE_DIR}/cmake")
 
 # put executables in the 'run' folder
 set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_SOURCE_DIR}/run)
-set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_SOURCE_DIR}/run)
+set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_SOURCE_DIR}/run/lib)
```
jgkamat: @jpfeltracco reverting this line will fix the test-python build problem. The reason that it was working for us is that the robocup.so file (which is placed in `run` was stale, and not removed via a make clean (which should be fixed), and the new file was going to `./run/lib`. I do agree that this is a better idea, but we need to be 100% sure that everything changes the path correctly when we do.

jpfeltracco:
I just rebased this on top of master so it should be good to go.
It's certainly not perfect, it's missing some of the integration we had with our old simulator, but I think we can add that stuff pack in as the need arises. Hopefully we'll work out the quirks as time goes on.

If someone else could test this out on a machine with more power than mine, that would be great, my laptop struggles running soccer.

soccer/MainWindow.cpp
```diff
@@ -162,6 +165,11 @@ MainWindow::MainWindow(Processor* processor, QWidget* parent)
     if (!_processor->simulation()) {
         _ui.menu_Simulator->setEnabled(false);
     }
+
+    _ui.actionQuickloadRobotLocations->setEnabled(false);
```
jgkamat: Can you add a comment (maybe 'disabled due to lack of support in grsim')?

soccer/Processor.cpp
```diff
@@ -85,14 +86,16 @@ Processor::Processor(bool sim, bool defendPlus, VisionChannel visionChannel)
     _gameplayModule = std::make_shared<Gameplay::GameplayModule>(&_state);
     _pathPlanner = std::unique_ptr<Planning::MultiRobotPathPlanner>(
         new Planning::IndependentMultiRobotPathPlanner());
-    vision.simulation = _simulation;
-
+    vision.simulation = true;
```
jgkamat: I'm not sure, but won't this break soccer on real robots? I'm not sure exactly what this variable does though...

soccer/MainWindow.cpp
```diff
@@ -974,30 +982,44 @@ void MainWindow::channel(int n) {
...
+    Geometry2d::Point ballPos =
+        _ui.fieldView->getTeamToWorld() * state()->ball.pos;
+    ball_replace->set_x(ballPos.x());
+    ball_replace->set_y(ballPos.y());
+    ball_replace->set_vx(0);
+    ball_replace->set_vy(0);
+    _ui.fieldView->sendSimCommand(simPacket);
 }
 
 void MainWindow::on_actionResetField_triggered() {
```
jgkamat: Maybe we can keep the reset button working (the reset button is very useful, so I would like to keep it if possible). In the right click menu, there is 'yellow robots->put all inside with formation x' which we could call instead of a real reset function when we click this.

soccer/SimFieldView.cpp
```diff
@@ -98,9 +101,13 @@ void SimFieldView::mouseMoveEvent(QMouseEvent* me) {
 
 void SimFieldView::mouseReleaseEvent(QMouseEvent* me) {
     if (_dragMode == DRAG_SHOOT) {
-        SimCommand cmd;
-        *cmd.mutable_ball_vel() = _teamToWorld.transformDirection(_shot);
-        sendSimCommand(cmd);
+        grSim_Packet simPacket;
```
jgkamat: Dragging and letting go of the ball seems to be crashing soccer for me: 
```
[libprotobuf FATAL /var/tmp/portage/dev-libs/protobuf-3.1.0/work/protobuf-3.1.0/src/google/protobuf/message_lite.cc:289] CHECK failed: IsInitialized(): Can't serialize message of type "grSim_Packet" because it is missing required fields: replacement.ball.x, replacement.ball.y
terminate called after throwing an instance of 'google::protobuf::FatalException'
  what():  CHECK failed: IsInitialized(): Can't serialize message of type "grSim_Packet" because it is missing required fields: replacement.ball.x, replacement.ball.y
make: *** [makefile:32: run-sim] Aborted
```

soccer/VisionReceiver.cpp
```diff
@@ -36,15 +36,21 @@ void VisionReceiver::run() {
...
     if (simulation) {
+        if (!socket.bind(SimVisionPort, QUdpSocket::ShareAddress)) {
+            throw runtime_error("Can't bind to shared vision port");
+        }
+        multicast_add(&socket, SharedVisionAddress);
         // The simulator doesn't multicast its vision.  Instead, it sends to two
         // different ports.
         // Try to bind to the first one and, if that fails, use the second one.
+        /*
         if (!socket.bind(SimVisionPort)) {
```
jgkamat: Maybe this can be completely removed, we aren't planning on using the old simulator again...

soccer/radio/SimRadio.hpp
```diff
@@ -9,14 +10,18 @@
...
+    SimRadio(SystemState& system_state, bool blueTeam = false);
 
     virtual bool isOpen() const override;
     virtual void send(Packet::RadioTx& packet) override;
     virtual void receive() override;
     virtual void switchTeam(bool blueTeam) override;
 
 private:
+    SystemState& _state;
+    ;
```
jgkamat: This extra semicolon can be removed right? Or am I missing something... :c

soccer/radio/SimRadio.cpp
```diff
@@ -17,32 +25,90 @@ SimRadio::SimRadio(bool blueTeam) {
...
-        _reversePackets.push_back(RadioRx());
-        RadioRx& packet = _reversePackets.back();
-
-        if (!packet.ParseFromString(buf)) {
-            printf("Bad radio packet of %d bytes\n", n);
-            continue;
+    for (int i = 0; i < Robots_Per_Team; i++) {
+        RadioRx rx;
+        rx.set_robot_id(i);
+        rx.set_hardware_version(RJ2015);
```
jgkamat: Finally :P

makefile
```diff
@@ -38,9 +38,10 @@ run-sim2play: all
 run-release: all-release
 	./run/soccer
 run-sim-release: all-release
-	-pkill -f './simulator --headless'
-	./run/simulator --headless &
-	./run/soccer -sim -pbk example.pbk
+	-pkill -f './grsim'
+	./run/grsim &
+	./run/soccer -sim
```
jgkamat: Any reason this is repeated twice? :/

jpfeltracco: Because copy pasting is hard.

Is the "-pbk example.pbk" still even used by the way?

jgkamat: I was intending it to be used during comp (so when we do a release run, it would set up playbooks automatically for us), but I'm fine with this being removed.

soccer/MainWindow.cpp
jpfeltracco: The API provided by grSim doesn't give us the ability to just "call" that function. I see two options here:
1. We add a new grSim protobuf packet that gives us the option for reset which makes us diverge even further from grSim.
2. We just hardcode robot positions and send a series of RobotReplacement messages.
```protobuf
message grSim_RobotReplacement {
required double x=1;
required double y=2;
required double dir=3;
required uint32 id=4;
required bool yellowteam=5;
}
```

Thoughts?

jgkamat: hmmmm, I like option 1 more, but I'd rather not diverge as much as possible. I think we should do 2 for now, and see if we can't get 1 merged into grSim at some point.

jpfeltracco:
I've just pushed up a new commit that allows us to get a kicker status back from grSim. Turns out they do have some limited robot status coming back, but they have to protobuf format for it. They just send back simple bytes.

@jgkamat I just tried out coordinated pass and it seems to be somewhat working. I adjusted the kick strength formula to try to help out a bit too. Could you try it out and let me know what you think?

jgkamat:
Nice, great! I'll test it out later today! :smile: :tada:

jgkamat:
@jpfeltracco I'm getting the same error as CI when I try to compile this: 

```
FAILED: soccer/CMakeFiles/robocup.dir/MainWindow.cpp.o 
ccache /usr/bin/c++   -DQT_CORE_LIB -DQT_GUI_LIB -DQT_NETWORK_LIB -DQT_NO_DEBUG -DQT_OPENGL_LIB -DQT_SVG_LIB -DQT_WIDGETS_LIB -DQT_XML_LIB -Drobocup_EXPORTS -Isrc/googletest/include -I../common -Icommon -isystem /usr/include/eigen3 -I../
soccer -Isoccer -isystem /usr/include/libusb-1.0 -isystem /usr/include/SDL2 -isystem /usr/include/qt5 -isystem /usr/include/qt5/QtWidgets -isystem /usr/include/qt5/QtGui -isystem /usr/include/qt5/QtCore -isystem /usr/lib64/qt5/mkspecs/li
nux-g++ -isystem /usr/include/qt5/QtXml -isystem /usr/include/qt5/QtOpenGL -isystem /usr/include/qt5/QtNetwork -isystem /usr/include/qt5/QtSvg -I../external/rrt/src -isystem /usr/include/python3.4m -Werror=return-local-addr -fPIC   -fPIC
 -std=gnu++14 -MD -MT soccer/CMakeFiles/robocup.dir/MainWindow.cpp.o -MF soccer/CMakeFiles/robocup.dir/MainWindow.cpp.o.d -o soccer/CMakeFiles/robocup.dir/MainWindow.cpp.o -c /home/jay/Code/robocup-software/soccer/MainWindow.cpp        
In file included from /usr/include/python3.4m/pytime.h:6:0,
                 from /usr/include/python3.4m/Python.h:65,
                 from /usr/include/boost/python/detail/wrap_python.hpp:151,
                 from /usr/include/boost/python/detail/prefix.hpp:13,
                 from /usr/include/boost/python/args.hpp:8,
                 from /usr/include/boost/python.hpp:11,
                 from ../soccer/gameplay/GameplayModule.hpp:6,
                 from /home/jay/Code/robocup-software/soccer/MainWindow.cpp:5:
/usr/include/python3.4m/object.h:435:18: error: expected unqualified-id before ‘;’ token
     PyType_Slot *slots; /* terminated by slot==0. */
                  ^
ninja: build stopped: subcommand failed.
make: *** [makefile:18: all] Error 1
```

jpfeltracco:
I hadn't done a clean build in a while so my version was compiling fine until I cleaned.

Seems that QT defines some macros that annoyingly conflict with that python header. From what I've pieced together, best practice is to not use "slots, emit, signal" in our code, add a CMake flag to undefine those macros, then change our code to use "Q_SLOTS, Q_EMIT, Q_SIGNAL".

I had to add a commit to a branch on the RRT as well to make these changes, but everything seems to be compiling fine again!

jgkamat:
I'm getting this error in grsim while it's running:
```
Sending UDP datagram failed (maybe too large?). Size was: 162 byte(s).
```
Are you seeing this too (and is it a problem, or just my network being bad). 

Otherwise, AFAIK, the behavior of everything looks good! We _still_ need someone on the 'official' setup to test this though. @joshhting, @chachmu, @jcarn, @JNeiger, @ashaw596  can you try out grSim on your own machines and see how it performs? I would like as many people to try it out as possible before we commit. @ashaw596, can you test this out on macos? I would prefer this not to break macos support, so it would be great if you could report errors/things early on. 

I'm going to do another code review of this soon, hopefully we can get this merged asap! :smile: :tada:

jpfeltracco:
Hm, has this only been an issue recently or was this happening on earlier commits too?

jgkamat:
After a bit moe investigation, this is definetly related to my internet connection. That error dosen't show up until I go offline (and I seem to loose robots then too). Could you try linking up grsim to broadcast over localhost instead of on the actual network? This would mean we would all share one grSim instance, and if we were offline, we can't use the simulator :P.

JNeiger:
I just tried it on Ubuntu 16.04 VirtualBox VM.

Within the build process:
There are a few warnings about narrowing conversion from double to float in file graphics.cpp:521.
RefereeTab has a "No relevevant classes found. No output generated."
Also, "XML Parsing error inside file '/home/vagrant/.grsim.xml'. Error: File not found"

While running:
The field display on grSim is forced on top of everything on my desktop. See picture below.
Everything seems to be in slow motion when I run TwoSideAttack.

image

jpfeltracco:
Alright, here are my thoughts:

@jgkamat 
I'll look into those networking issues. grSim should definitely already be sending everything over localhost, but I'll verify. Could you check if you get the same issues on earlier commits of this branch? I changed how the robot status is sent back and potentially I'm flooding the network or something, although the rate looked fine when I last checked. I'll see if I can duplicate this on my end asap.

@JNeiger 
1. Those compiler warnings are really things that need to be fixed on grSim's side. We can submit a pull request to them when we have the time.
2. I'm not sure what that missing grsim.xml is referring to, there might be some config file it expects, but I'll look into it.
3. I'm assuming you've got "Enable 3D Acceleration" checked in your VM settings? I've seen this issue several times with VMs, it seems that VirtualBox doesn't support any recent OpenGL when 3D acceleration is active. It's part of the reason I moved to linux as my host :/. I'm not sure there will be much we can do about it.

jgkamat:
I can repro this as early as 2696178, which is the earliest one I can build... I'll take a closer look at this tomorrow though. Can you run this without network access on your machine?

jpfeltracco:
I can't duplicate it on my machine unfortunately :/

jpfeltracco:
So @jgkamat, what should we do about these issues? I can't duplicate the networking problems and didn't see anything unreasonable in grSim networking code. Everything works over localhost just fine for me :/.

jgkamat:
Yah I took a look at it and I couldnt find anything fishy, if it works for you and it works on the stock vm I wouldn't worry about it (I'll figure out what's wrong on my own :P).

As for the rest of this, I think it's almost ready, but I really would like a few more people to try it out before it gets merged....

joshhting:
When I compile on my gnome partition, I get the following error:

```
In file included from soccer/ui_MainWindow.h:38:0,
                 from ../soccer/MainWindow.hpp:11,
                 from ../soccer/MainWindow.cpp:1:
../soccer/SimFieldView.hpp:6:40: fatal error: protobuf/grSim_Commands.pb.h: No such file or directory
compilation terminated.
```

joshhting:
I'm not sure if somebody already tested it on the field computer or not but grSim works fine on it

joshhting:
Also @jpfeltracco, the reset button does indeed fix a fallen over robot

jgkamat:
On my issue, it seems like grsim is sending on `224.5.23.2`, which is why I'm not getting it when offline. I can fix this by running `./util/route` which sets up routing that to localhost. 

If this works for you that's fine, but we might want to change that to broadcasting on localhost directly. 

Also, I want to try to push this through soon if possible (because I don't want to change something this big right before comp). (please test this out if you're free). 

@joshhting are you getting this error only on your ubuntu gnome machine (and is it working fine in the VM)? later comments imply you got it working at some point...

joshhting:
It seems to work fine now for ubuntu gnome for whatever reason. ¯\_(ツ)_/¯

jgkamat:
@ashaw596 could you test this out and report any issues? I want to run this at comp and if it breaks your build that has to be fixed.

jgkamat:
Also, for installing on MacOS, see the [grSim installation guide](https://github.com/mani-monaj/grSim/blob/master/INSTALL.md#mac-os-x-installation) for help in installing dependencies. I don't think there's anything that should stop it from running, but we'll see.

ashaw596:
Whooops, just saw this. I'll try this on my mac tomorrow! Was gonna do it tonight, but it looks like its gonna take a couple hours to update my git on china internet 🙃

ashaw596:
Hmm, it is broken on mac right now, but I think I just need to fix some dependencies. but I think master is too. I'm getting this error again 
`TypeError: No to_python (by-value) converter found for C++ type: SystemState*
libc++abi.dylib: terminating with uncaught exception of type std::runtime_error*`

If this is the last issue though, I think its fine. I can just use my alienware laptop during comp. That has ubuntu 16.04 (Since I don't think anyone else uses native mac now anyways =/)

ashaw596:
I can fix the build later for mac if it comes to that.

jgkamat:
@ashaw596 if you think it's worth it, then I could try to fix that (a bisect would be helpful though :3). 

For now though, I'll target everything to ubuntu16 (since you have a laptop with that), and we can have patches to support mac/other systems later if we need.

ashaw596:
@jgkamat Sounds good. Yeah I think it's best to just focus on 16.04 for now. I can fix it for my mac later. I think the master are pretty simple issues and it shouldn't be tooo bad to fix the dependencies for grsim. I've had it running before (a long time ago). I can do it later in a different patch though.

ashaw596:
Seems to build and work on my ubuntu laptop I think! (except for an uninitialized debug bool in the rrt source, but thats simple)

soccer/config/sim.xml
```diff
@@ -1,28 +1,28 @@
 <config>
  <Rev2008>
   <translation>
```
jgkamat: This tuning might need to be revisited, but from basic checks they seem fine.

jgkamat:
Going to ignore style for now, the diff is too big! :stuck_out_tongue_winking_eye:

jgkamat:
Planning on merging this later today, this is the last chance to voice any grSim issues before It's on master. Are you :+1: on it, @jpfeltracco?

jpfeltracco:
Yup, LGTM.

jgkamat:
In order to keep something like #320, I'm adding a clause to kill the simulator when soccer quits (when we run it from make). If you want to keep the simlator up, run it manually (or if people want, I'll add more make targets).

jgkamat:
For anyone experiencing build errors as a result of this, please try:

1. `make clean`
2. `./util/ubuntu-setup.sh`.

If those don't help, please open a new issue with your problem ASAP and I'll take a look at it.

PR #913, opened 2017-02-22
Make the Pathplanner only consider current velocity in target direction when calculating the start speed

ashaw596:


jgkamat:
Aaaaah theres so much code and so little time! I'll look at this ASAP (but it would definetly be faster to find someone else to review this). @jpfeltracco or @joshhting maybe? I'll get to this eventually if no one does it before then :cry: 

Thanks for the PR Though!

ashaw596:
Its k. I added them as reviewers if they wanted to look at it, but its ok if this takes awhile. This fix isn't that important since its used very rarely.

ashaw596:
Yeah, it causes some really bad behavior if you don't.

PR #914, opened 2017-02-23
Fix rrt build failing when checking out a new version or pulling a new version of rrt without a clean.

jgkamat:
I think I fixed the build issues with the rrt. This needs cleanup and verifications though. 

There's an rrt update, included in this pr too.


jgkamat:
Merging now. It is likeley that something broke, but I hope not! :crossed_fingers: 

Let me know if you see any issues and I'll follow up.

Issue #902, opened 2017-02-10
Combine the functionality of Kick_Eval and Win_Eval


JNeiger:
It would be nice to combine these two to simplify the process of estimating a shot / pass.

Kick eval uses probabilities to estimate whether something will be blocked to a specific point.
Win eval uses "windows" to find the largest segment in which we are not blocked.

The overall goal would be to find the largest segment to pass to while producing a probability associated with that segment in a single function call. 

There are 3 main probability functions that we have to keep in mind when doing this. 

- First is the probability of the opponents robot to block a kick at a specific angle.
- The second is the target. The best example is shooting on goal. It will be 1 when the kick will go through the goal and 0 when it is outside the goal.
- The last is the distribution of the kick from our robot. In theory, this should just be a normal distribution.
---
I think I'm going to try to represent the probabilities in a 2D space where X is the angle in reference to the robot (Think first person POV) and Y will be the probability in that direction. This should simplify the degrees of freedom later on.

The 3 probabilities will then be...
- Normal distributions centered around the angle the opponents robots are at and scaled horizontally  by distance from the kicker.
- A band-pass filter looking distribution centered around the target with the correct width.
- A normal distribution

The first two will be static while the third will be able to move along the X axis. We can use an optimization function (like a hill-climber) to find angle of the third function at the global max of the combination of the 3 functions. 

To find the combination of the 3 functions, I think we should be able to sample the 3rd function and use the angle at each sample point to find the value at the other two functions. We can then divide the sum (Of all 3 multiplied together) at every sample, by the sum of the third function at the specific sample points. We did this with Kick_Eval and it was linearly related with the actual combination of the 3 functions so it /should/ be fine.

This should return the angle at which to kick as well as the probability at that kick angle.

JNeiger:
There is actually a better way to do this. In short, we are trying to convolve the product of equation 1 and 2 with equation 3. If we represent everything as either normal distributions or unit step functions, we can get a closed form solution. I have a feeling that we can also determine where the peaks are based on the location the blocking robots.

Here is the closed form solution for the two robot case. A bunch of this can be predefined and cached.
![image](https://cloud.githubusercontent.com/assets/5797304/22904543/d6cf28a6-f20a-11e6-9ce2-aef423387f73.png)

Here's the graph of the robots position distribution.
![image](https://cloud.githubusercontent.com/assets/5797304/22904689/4f3fddda-f20b-11e6-8bf0-964765327290.png)

Here's the bounds.
[...]

PR #915, opened 2017-02-26
Combine Window Eval and Kick Eval

### LINKED ISSUE #298: Improve evaluation of shots and passes

JNeiger:
Closes #298 #892 #902 
Would close PR #896 too

I found a closed form solution that should combine the two.
The idea is that there are 3 main probability / score distributions that need to be used to evaluate this.
1. Where the robot might be
2. What consists of a good kick
3. Our kick distribution

Distribution 1 is a set of normal distributions, 
Distribution 2 is a band pass filter looking distribution created using unit step functions
Distribution 3 is just a normal distribution.

1 and 2 are combined with addition to create a new set.
A convolution of 1/2 and 3 yields a set of distributions where Y is the probability of success at that angle for that specific item in the set..
The min of the set at each angle is the worst case chance of success.

I use multiple Gradient Ascent optimizers to find the max segment.

- [x] Fix bug when there are no robots
- [x] Add tests for Gradient Ascent optimizer (single)
- [x] Ditto for parallel
- [x] Clean up variable names and function names
- [x] Add more comments for optimizer operation
- [x] Check execution speed
- [x] Clean up code style


jgkamat:
Does this contain the changes in #896? Also, this PR seems to have a cmake error when building it (saying it can't find a file). https://3444-15662105-gh.circle-artifacts.com/0/tmp/circle-artifacts.Akg8HXA/compile.txt

JNeiger:
I would hold off at least another day or so. I want to make sure there's not glaring issues when I finish porting everything to soccer. I've only done major tests in the Mathematica and Matlab version so far

The cmake error is because I forgot to add the cpp file to the git commit. I'll add it once I get to the next commit point.

JNeiger:
All in all, it's about ~3 times slower than Win_Eval. If that's ok, this should be all good to go.

JNeiger:
I think I'm good to go. I added in all the changes we talked about Tuesday

soccer/optimization/Gradient1DConfig.hpp
```diff
@@ -71,8 +68,7 @@ class Gradient1DConfig {
           maxValue(maxValue),
           maxThresh(maxThresh) {}
 
-    std::tuple<float, float>(*f)(float, FunctionArgs*);
-    FunctionArgs* args;
+    std::function<std::tuple<float, float>(float)>* f;
```
ashaw596: Um, I don't think this needs to be pointer. Right now its a pointer to a function pointer I think.

soccer/KickEvaluator.cpp
```diff
@@ -338,44 +335,43 @@ vector<tuple<float, float> > KickEvaluator::convert_robots_to_polar(
     return botLocations;
 }
 
-void KickEvaluator::init_gradient_configs(ParallelGradient1DConfig* pConfig,
-                                          KickEvaluatorArgs* keArgs) {
-    pConfig->GA1DConfig.reserve(keArgs->robotStDevs.size());
+void KickEvaluator::init_gradient_configs(
+    ParallelGradient1DConfig* pConfig,
+    function<tuple<float, float>(float)>* func, const vector<float>* robotMeans,
+    const vector<float>* robotStDevs, const float boundaryLower,
```
ashaw596: I'd recommend you change those const pointers to const references. Its a good coding guideline we're been following in the rest of soccer (trying to at least) and prevents null from being passed in (unless you really want to be able ot pass in null). Pointers for non const stuff is fine.

soccer/optimization/Gradient1DConfig.hpp
JNeiger: Yeah, It's a pointer to a function pointer container. I might be over thinking this, but I was worried about the amount of data it may have to copy when being passed around.

When std::function wraps just a normal function pointer, it doesn't use that much space, like ~12 bytes on the stack. When you wrap the results of a std::bind or a lambda function, you have to worry about all the passed in arguments. For both, all the arguments are copied into the std::function. 

Additionally, std::function only has ~4-8 bytes of storage for this data. If you have anymore, it transfers everything to the heap.

For my use-case, it's anywhere from 4 - 26 copies depending on the amount of robots (It can even be more in some cases). My std::function size is around ~40 bytes (Assuming 32 bit system).

This works fine for simple arguments, but I was worried that if someone decides to throw in the robots vector or something bigger, it may cause a significant overhead that would be a pain to refactor later on.

ashaw596:
Std function allocates everything it needs on the heap and handles its own
memory management. If you use move semantics it'll be the same as copying a
pointer.

On Mar 2, 2017 10:43 AM, "Joseph Neiger" <notifications@github.com> wrote:

*@JNeiger* commented on this pull request.
------------------------------

In soccer/optimization/Gradient1DConfig.hpp
<https://github.com/RoboJackets/robocup-software/pull/915#discussion_r103954243>
:

> @@ -71,8 +68,7 @@ class Gradient1DConfig {
           maxValue(maxValue),
           maxThresh(maxThresh) {}

-    std::tuple<float, float>(*f)(float, FunctionArgs*);
-    FunctionArgs* args;
+    std::function<std::tuple<float, float>(float)>* f;

Yeah, It's a pointer to a function pointer container. I might be over
thinking this, but I was worried about the amount of data it may have to
copy when being passed around.

When std::function wraps just a normal function pointer, it doesn't use
that much space, like ~12 bytes on the stack. When you wrap the results of
a std::bind or a lambda function, you have to worry about all the passed in
arguments. For both, all the arguments are copied into the std::function.

Additionally, std::function only has ~4-8 bytes of storage for this data.
If you have anymore, it transfers everything to the heap.

For my use-case, it's anywhere from 4 - 26 copies depending on the amount
of robots (It can even be more in some cases). My std::function size is
around ~40 bytes (Assuming 32 bit system).

This works fine for simple arguments, but I was worried that if someone
decides to throw in the robots vector or something bigger, it may cause a
significant overhead that would be a pain to refactor later on.

—
You are receiving this because you commented.

Reply to this email directly, view it on GitHub
<https://github.com/RoboJackets/robocup-software/pull/915#discussion_r103954243>,
or mute the thread

JNeiger:
Is there a something similar to std::move but the original object still has ownership?

For example, right now I have a single std::function being passed into all the optimizer configs. If I'm reading the std::move stuff right, I would have to have a std::function declaration for each optimizer config which would kinda defeat the purpose.

ashaw596:
Oh, if its like that. Pointer makes sense. You could store a const reference too. Who actually has ownership? You could give everyone a shared_pointer to the std::function without that much overhead.

ashaw596:
Hmm. are you actually sharing a function with multiple configs? It only looks like one in the current code. You might as well just move the function (unless you intend to reuse functions in future code).

soccer/KickEvaluator.cpp
```diff
@@ -355,24 +355,24 @@ void KickEvaluator::init_gradient_configs(
 
     // <PrevStart, Start>
     vector<tuple<float, float> > xStarts;
-    xStarts.reserve(robotStDevs->size());
+    xStarts.reserve(robotStDevs.size());
 
     // Add left boundary
     float startX = boundaryLower + *start_x_offset * *kick_std_dev;
-    xStarts.push_back(forward_as_tuple(boundaryLower, startX));
+    xStarts.emplace_back(forward_as_tuple(boundaryLower, startX));
```
ashaw596: Since forward_as_tuple creates a tuple. You won't really get speedup from emplace. I think emplace_back(boundaryLower, startX) should foward correctly and cosntruct the tuple once. Or everything might be optimized to be the same anyways. Not sure. without the forward_as_tuple shouldn't be slower though.

soccer/KickEvaluator.cpp
```diff
@@ -315,7 +315,7 @@ tuple<float, float> KickEvaluator::rect_to_polar(const Point origin,
 
 vector<tuple<float, float> > KickEvaluator::convert_robots_to_polar(
     const Point origin, const Point target) {
-    vector<Robot*> bots = get_valid_robots();
+    vector<Robot*> bots = move(get_valid_robots());
```
ashaw596: function return values are already temporary values so the move won't really do anything here.  move just casts things to an rvalue and function returns are rvalue.

soccer/KickEvaluator.cpp
```diff
@@ -0,0 +1,396 @@
...
+}
+
+vector<tuple<float, float> > KickEvaluator::convert_robots_to_polar(
+    const Point origin, const Point target) {
+    vector<Robot*> bots = move(get_valid_robots());
+    vector<tuple<float, float> > botLocations;
+    botLocations.reserve(bots.size() + botLocations.size());
+
+    // Convert each bot position to polar
+    for_each(bots.begin(), bots.end(), [&botLocations, target, origin, this](
```
ashaw596: Since you're doing fancy stl stuff!! Yay! You might as well use transform if you want. http://en.cppreference.com/w/cpp/algorithm/transform
Would look like  std::transform(bots.begin(), bots.end(), std::back_inserter(botLocations),
                  [target, origin, this](Robot* bot) { return rect_to_polar(origin, target, bot->pos)});

soccer/KickEvaluator.cpp
```diff
@@ -0,0 +1,396 @@
...
+                                           Robot* bot) {
+        botLocations.push_back(rect_to_polar(origin, target, bot->pos));
+    });
+
+    // Convert imaginary obstacles to polar
+    for_each(
```
ashaw596: transform! if you want.

soccer/KickEvaluator.hpp
```diff
@@ -0,0 +1,148 @@
...
+    /**
+     * @brief Initializes configurable fields
+     * @note See configuration documentation for details
+     */
+    static void createConfiguration(Configuration* cfg);
+
+    /**
+     * @brief Robots that should not be consider obstacles
+     */
+    std::vector<Robot*> excluded_robots;
```
ashaw596: Just wondering. Can these be const Robot*? Are you editing them anywhere?

soccer/KickEvaluatorTest.cpp
```diff
@@ -0,0 +1,98 @@
...
+        0, 0, 0.1, std::ref(robotMeans), std::ref(robotStdevs),
+        std::ref(robotVertScores), -2, 2);
+
+    std::tuple<double, double> expected = std::make_tuple(1, 0);
+
+    EXPECT_NEAR(std::get<0>(res), std::get<0>(res), 0.01);  // Value
+    EXPECT_NEAR(std::get<1>(res), std::get<1>(res), 0.01);  // Derivative
+}
+
+// TEST(KickEvaluator, speed) {
```
ashaw596: Should these be uncommented(or deleted?)

soccer/WindowEvaluatorTest.cpp
```diff
@@ -36,3 +36,23 @@ TEST(WindowEvaluator, eval_pt_to_seg) {
     // the window should be our goal segment
     EXPECT_EQ(ourGoalSegment, windows[0].segment);
 }
+
+// TEST(WindowEvaluator, speed) {
```
ashaw596: Uncommented (or deleted?)

soccer/optimization/GradientAscent1D.cpp
```diff
@@ -0,0 +1,82 @@
...
+    float x_offset = gamma * currentdx;
+
+    // Force within bounds
+    x_offset = std::min(config->maxXMovement, x_offset);
+    x_offset = std::max(config->maxXMovement * -1, x_offset);
+
+    return currentx + x_offset;
+}
+
+int GradientAscent1D::sign(float val) { return (0.0 < val) - (val < 0.0); }
```
ashaw596: Just wondering. Does this need to be a instance function? Can't this just be defined in the header(Or just defined in the cpp file)

soccer/optimization/ParallelGradientAscent1D.cpp
```diff
@@ -0,0 +1,74 @@
+#include "ParallelGradientAscent1D.hpp"
+#include <math.h>
+#include <iostream>
+
+ParallelGradientAscent1D::ParallelGradientAscent1D(
+    ParallelGradient1DConfig* config)
+    : config(config) {
+    // Create list of GA1Ds
+    for (int i = 0; i < (config->GA1DConfig).size(); i++) {
```
ashaw596: Not sure where you pushing for efficiency. But you could use reserve then transform here too if you want.

JNeiger:
Let me try and explain my "ideal" implementation.
Here is the rough object relationship graph for Parallel Gradient Config.
```
    Parallel Gradient Config
     |         ...        |
    \/                   \/
Gradient Config     Gradient Config
```

- Each of the "Child" Gradient Configs have to be able to store the function. This is because each "Child" Gradient Config is tied to an instance of a standalone Gradient Optimizer.
- I would prefer for the function to be changeable between separate "Child" Gradient Configs to allow for constant arguments to change.
- I would also prefer for either the Parallel Gradient Config or the class which Parallel Gradient Config is used to own the std::function. As it is right now, the class which Parallel Gradient Config is used owns them. The other way would be a bit more complex to implement I think.
- The implementation should be generic enough for any other uses
- And I would like to minimize the std::function copy / declaration amount

Also, What do you mean by "move the function"?

soccer/optimization/ParallelGradientAscent1D.cpp
```diff
@@ -0,0 +1,74 @@
...
+        // Assume ascending order for xStart
+        for (int i = 0; i < GA1Ds.size() - 1; i++) {
+            GradientAscent1D lower = GA1Ds.at(i);
+            GradientAscent1D upper = GA1Ds.at(i + 1);
+
+            // Erase elements if they get too close
+            // This helps kill any GA1Ds that are going up the same hill
+            if (fabs(lower.getXValue() - upper.getXValue()) <
+                config->xCombineThresh) {
+                GA1Ds.erase(GA1Ds.begin() + i + 1);
```
ashaw596: Not sure if you want efficiency, but removeif should be a good bit faster here (and a bit more readable)
If you need hyperefficiency, the swap delete trick would work if order doesn't matter.
std::swap(vec.begin() + index, vec.end() - 1);
    vec.pop_back();

soccer/optimization/ParallelGradientAscent1D.cpp
```diff
@@ -0,0 +1,74 @@
...
+}
+
+/**
+ * Returns a list of all X values for each max in ascending order
+ */
+std::vector<float> ParallelGradientAscent1D::getMaxXValues() {
+    std::vector<float> xVals;
+    xVals.reserve(GA1Ds.size());
+
+    for (auto& GA1D : GA1Ds) {
```
ashaw596: transform if you want. Really this is really readible anyways. Doesn't matter at all.

soccer/optimization/ParallelGradientAscent1D.cpp
```diff
@@ -0,0 +1,74 @@
...
+}
+
+/**
+ * Returns a list of all X values for each max in ascending order
+ */
+std::vector<float> ParallelGradientAscent1D::getMaxValues() {
+    std::vector<float> vals;
+    vals.reserve(GA1Ds.size());
+
+    for (auto& GA1D : GA1Ds) {
```
ashaw596: transform if you want. Really this is really readible anyways. Doesn't matter at all.

ashaw596:
Ok. That makes sense I think. The pointer makes sense (if looks a bit weird).

soccer/KickEvaluator.cpp
```diff
@@ -0,0 +1,396 @@
...
+
+    static float rmean;
+    static float rstdev;
+    static float rstdev2;
+    static float robotV;
+
+    static float kx;
+    kx = kmean - x;
+
+    static float fterm;  // First term,  Robot normal distribution
```
JNeiger: It was a noticeable change if I recall correctly. I want to say 5-8%.

soccer/KickEvaluator.hpp
JNeiger: They get edited through python side. There should be some sort of method in the robocup-py.cpp file that provides the python accessor for it.

soccer/KickEvaluator.cpp
```diff
@@ -0,0 +1,396 @@
...
+#else
+        int i, j;
+#endif
+    } n;
+} _eco;
+
+// Floating point magic
+#define EXP_A (1048576 / 0.69314718055994530942)
+#define EXP_C 60801
+#define fast_exp(y) (_eco.n.i = EXP_A * (y) + (1072693248 - EXP_C), _eco.d)
```
ashaw596: Oh woops I'm going insane. That implementation will only be better for calculating exp for literals(like e^100) Not for calculations. 
The benchmarks for that macro implementation don't seem much faster though (which is weird).
Did you see this one? https://codingforspeed.com/using-faster-exponential-approximation/
360x seems alot better than 
Slow exp time=306
vs
Fast exp time=286

ashaw596: Eh seems like all the speedups aren't very much actually after you throw in optimization flags.
Found this one too, but it looks like these are all similar.  Without macros tho.
http://www.smipple.net/snippet/Martin%20Ankerl/A%20Fast,%20Compact%20Approximation%20of%20the%20Exponential%20Function

soccer/KickEvaluator.cpp
ashaw596: Hmm. Did you do this with optimzed compile flags? Idk. It seems like cache locality should really be more important than the time it takes to push a stack pointer.

ashaw596: But whatever works!

soccer/KickEvaluator.cpp
JNeiger: Surprisingly, that's almost twice as slow compared to the union.

ashaw596: Weird even with -O3 in the compile flags?

soccer/KickEvaluator.cpp
```diff
@@ -0,0 +1,396 @@
...
+}
+
+tuple<float, float> KickEvaluator::rect_to_polar(const Point origin,
+                                                 const Point target,
+                                                 const Point obstacle) {
+    Point obstacleDir = obstacle - origin;
+    Point targetDir = target - origin;
+    float angle = obstacleDir.angle() - targetDir.angle();
+
+    return forward_as_tuple(obstacleDir.mag(), fixAngleRadians(angle));
```
JNeiger: It returns a tuple<float &&, float&&>. I played around with it a bit and they're about the same speed. I'm gonna switch it to make_tuple since that makes more sense in context

soccer/KickEvaluator.cpp
JNeiger: Oh god... It turns out none of this has been with any optimization...

JNeiger: Bad news, at -O2, the Taylors Series is still twice as slow. Good news, I can now go through half a million executions in only 2.3 seconds.

ashaw596: Yeah taylor series will be alot slower. Thats why i made the comment that I was being stupid =D. That was optimized for running in compile time on literals. This one should have the same or better performance since it looks basically the same as the macro.
http://www.smipple.net/snippet/Martin%20Ankerl/A%20Fast,%20Compact%20Approximation%20of%20the%20Exponential%20Function

ashaw596: lol 2.3 seconds? It was 20 before right? lol

soccer/KickEvaluator.cpp
JNeiger: With -O2, they are the same basically. I'm gonna switch back to just using normal variables as it would make the code cleaner

soccer/KickEvaluator.cpp
JNeiger: 20 ms improvement with that inline function. Not bad lol

soccer/KickEvaluator.cpp
jgkamat: if you want to test out the code with optimizations turned on, you can try running `make run-release` or `make run-release-sim`. It's helpful to keep the code as fast as possible though, even when we have these off, since it's nice to have optimizations turned off when debugging (due to faster compile times)

jgkamat: The release builds also add CPU specific optimizations, which may or may not be a big diffference (especially with math computations). If you wanted O3, you can [add that to this line](https://github.com/RoboJackets/robocup-software/blob/master/CMakeLists.txt#L37).

soccer/optimization/ParallelGradientAscent1D.cpp
JNeiger: How would you do a removeif here? I use both the current and next value in the vector.

ashaw596: Oh woops. I didn't realize what was happening exactly. Um it would be a bit more complicated and sort of dumb, but you could do like.
auto i=1;
GA1Ds.erase(remove_if(GA1Ds.begin(), GA1Ds.end(), [&GA1Ds] (const auto &value) {
auto&next = GA1Ds[i];
i++;
return fabs(value.getXValue() - next .getXValue()) < config->xCombineThresh);
}
Might technically be faster since not everything gets copied each iteration. Not actually sure if it matters though since idk what the size of GAID is.

JNeiger: The size is 2*Num_Of_Robots + 2 at the largest. It shrinks down pretty quick though

ashaw596: Your choice. if GA1Ds is smallish. memcopies of that size aren't thatttt expensive.

JNeiger: Would there still be a copy if I just did fabs(GA1Ds.at(I).getXValue() - GA1Ds.at(I+1).getXValue()) ?

ashaw596: I mean you should be using references there. but the main problem is erasing from the middle of a vector is an o(n) copy of everthing after that index.

JNeiger: I'm having trouble keeping this inside the bounds of the vector so I'm just going to leave it as is.

ashaw596: Oh right the off by one error at the end. Hmm, yeah not really worth it i guess unless it was actually speeding it up. You'd have to iterate to the end -1 instead, then copy the last one before you erase everything after. Something like.
int i = 1;
auto end = GA1Ds.end()-1;
auto last = std::remove_if(GA1Ds.begin(), GA1Ds.end()-1, [this, &i](auto& value) {
                auto& next = GA1Ds[i];
                i++;
                return fabs(value.getXValue() - next.getXValue()) <
                       config->xCombineThresh;
            });
*last = *end;
GA1Ds.erase(last + 1);

PR #916, opened 2017-03-01
Export Graph Data

### LINKED ISSUE #912: Add option to get data from StripChart

mmurley3:
Adds option to export current graph's data points to a .csv file.  Currently units are not defined well.
Closes #912

### DIFF
--- a/soccer/MainWindow.cpp
+++ b/soccer/MainWindow.cpp
@@ -369,7 +369,8 @@ void MainWindow::updateViews() {
                                    _longHistory.begin());
 
     // Set the original history vector
-    _history.assign(_longHistory.begin(), _longHistory.begin() + _history.size());
+    _history.assign(_longHistory.begin(),
+                    _longHistory.begin() + _history.size());
 
     // Update field view
     _ui.fieldView->update();
--- a/soccer/MainWindow.cpp
+++ b/soccer/MainWindow.cpp
@@ -50,12 +50,13 @@ MainWindow::MainWindow(Processor* processor, QWidget* parent)
       _doubleFrameNumber(-1),
       _lastUpdateTime(RJ::now()),
       _history(2 * 60),
+      _longHistory(10000),
       _processor(processor) {
     qRegisterMetaType<QVector<int>>("QVector<int>");
     _ui.setupUi(this);
     _ui.fieldView->history(&_history);
 
-    _ui.logTree->history(&_history);
+    _ui.logTree->history(&_longHistory);
     _ui.logTree->mainWindow = this;
     _ui.logTree->updateTimer = &updateTimer;
 
@@ -366,6 +367,9 @@ void MainWindow::updateViews() {
     // Read recent history from the log
     _processor->logger().getFrames(frameNumber(), _history.size(),
                                    _history.begin());
+
+    _processor->logger().getFrames(frameNumber(), _longHistory.size(),
+                                   _longHistory.begin());
 
     // Update field view
     _ui.fieldView->update();
--- a/soccer/MainWindow.cpp
+++ b/soccer/MainWindow.cpp
@@ -365,11 +365,11 @@ void MainWindow::updateViews() {
     // update history slider in ui
 
     // Read recent history from the log
-    _processor->logger().getFrames(frameNumber(), _history.size(),
-                                   _history.begin());
-
     _processor->logger().getFrames(frameNumber(), _longHistory.size(),
                                    _longHistory.begin());
+
+    // Set the original history vector
+    _history.assign(_longHistory.begin(), _longHistory.begin() + _history.size());
 
     // Update field view
     _ui.fieldView->update();
--- a/soccer/ProtobufTree.cpp
+++ b/soccer/ProtobufTree.cpp
@@ -489,27 +489,10 @@ void ProtobufTree::contextMenuEvent(QContextMenuEvent* e) {
             connect(updateTimer, SIGNAL(timeout()), chart, SLOT(update()));
         }
     } else if (exportAction && act == exportAction) {
-        QVector<int> path;
-        QStringList names;
-        for (QTreeWidgetItem* i = item; i; i = i->parent()) {
-            int tag = i->data(Column_Tag, Qt::DisplayRole).toInt();
-            path.push_back(tag);
-            names.append(i->text(Column_Field));
-        }
-        reverse(path.begin(), path.end());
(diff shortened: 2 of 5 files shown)


jgkamat:
Sorry this is taking so long, just bumping this to let you know I am working on the review for this! I'll definetly have this done during break, if not before then.

ashaw596:
Whoops didn't see  this.

ashaw596:
So I'm a bit confused about how to use the feature right now? I'm assuming theres some button or option added somewhere? Do you think you could add some documentation?

ashaw596:
I'm not really sure what data is being outputed right now, since the numbers don't seem to match? Could you check the units? Are you maybe outputing the x and y coordinates on the graph instead of the actual time and values?

jgkamat:
I guess you figured out how to export it, but for anyone else, it should show up in a drop down _after_ creating a new chart. I think that this does not display data from logs, but only the current data on the screen (but I have not looked at this enough to be sure). Ideally we would grab this data straight from logs, as you said before.

soccer/StripChart.cpp
```diff
@@ -47,6 +49,30 @@ void StripChart::function(Chart::Function* function) {
...
+        outfile << "Time,Value\n";
+
+        for (unsigned int i = 0; i < _history->size(); ++i) {
+            float v = 0;
+            if (_history->at(i) && function->value(*_history->at(i).get(), v)) {
+                if (autoRange) {
+                    newMin = min(newMin, v);
+                    newMax = max(newMax, v);
+                }
+                QPointF pt = dataPoint(i, v);
```
jgkamat: The segment just above this seems like we're re-creating the exact values that are printed to the stripchart. This has the limitations of (maybe) being inaccurate, and being limited to the size of the stripchart currently. The CSV import really is only useful if it captures all of the data, so it would probably be a better idea to grab this data from the logs. 

I _think_ this can be done by simply not using the datapoint function above, and instead using `(framenumber, v)` as your point. We would need to be careful about this though, since we would probably want to convert `framenumber` into seconds if possible. I think there's a way to get that from the frame somehow.

soccer/StripChart.cpp
```diff
@@ -47,6 +49,30 @@ void StripChart::function(Chart::Function* function) {
...
+    for (unsigned int x = 0; x < _functions.size(); x++) {
+        auto function = _functions[x];
+        
+        std::ofstream outfile("chartData.csv");
+        outfile << "Time,Value\n";
+
+        for (unsigned int i = 0; i < _history->size(); ++i) {
+            float v = 0;
+            if (_history->at(i) && function->value(*_history->at(i).get(), v)) {
+                if (autoRange) {
```
jgkamat: Is the stuff in this if statement needed? You don't seem to be using newMin or newMax anywhere?

soccer/ProtobufTree.cpp
```diff
@@ -485,6 +488,28 @@ void ProtobufTree::contextMenuEvent(QContextMenuEvent* e) {
         if (updateTimer) {
             connect(updateTimer, SIGNAL(timeout()), chart, SLOT(update()));
         }
+    } else if (exportAction && act == exportAction) {
```
jgkamat: Could this section get some comments please :P

soccer/ProtobufTree.cpp
```diff
@@ -485,6 +488,28 @@ void ProtobufTree::contextMenuEvent(QContextMenuEvent* e) {
...
+        QVector<int> path;
+        QStringList names;
+        for (QTreeWidgetItem* i = item; i; i = i->parent()) {
+            int tag = i->data(Column_Tag, Qt::DisplayRole).toInt();
+            path.push_back(tag);
+            names.append(i->text(Column_Field));
+        }
+        reverse(path.begin(), path.end());
+        reverse(names.begin(), names.end());
```
jgkamat: Is names ever used again? I can't see it being used anywhere.

soccer/ProtobufTree.cpp
```diff
@@ -485,6 +488,28 @@ void ProtobufTree::contextMenuEvent(QContextMenuEvent* e) {
...
         }
+    } else if (exportAction && act == exportAction) {
+        QVector<int> path;
+        QStringList names;
+        for (QTreeWidgetItem* i = item; i; i = i->parent()) {
+            int tag = i->data(Column_Tag, Qt::DisplayRole).toInt();
+            path.push_back(tag);
+            names.append(i->text(Column_Field));
+        }
+        reverse(path.begin(), path.end());
```
jgkamat: I'm not sure why this is being reversed, could you explain this?

chachmu:
@mmurley3 I got figured out how to export multiple graphs and add the names of each column to the top of the csv. I think Jay mentioned that instead of exporting to a set folder and filename it should prompt the user by popping up a save dialog. QT already has a built in way to do this so if you want we can talk about it at the next meeting.

jgkamat:
This seems like it might be the think we are looking for: http://doc.qt.io/qt-5/qfiledialog.html

Although, I can't verify it right now, it seems right.

chachmu:
That's it. I've used it before so it shouldn't be too hard to get working.

soccer/StripChart.cpp
```diff
@@ -47,15 +52,54 @@ void StripChart::function(Chart::Function* function) {
...
+            outfile << std::endl;
+        }
+
+    }
+    outfile.close();
+}
+
 QPointF StripChart::dataPoint(int i, float value) {
-    float x = width() - (i * width() / _history->size());
+    //float x = width() - (i * width() / _history->size());
```
jgkamat: The commented versions of these functions can probably be removed, right?

soccer/MainWindow.cpp
```diff
@@ -367,6 +368,9 @@ void MainWindow::updateViews() {
     _processor->logger().getFrames(frameNumber(), _history.size(),
                                    _history.begin());
 
+    _processor->logger().getFrames(frameNumber(), _longHistory.size(),
```
jgkamat: How often is this function called? Why are we running getFrames twice (once on the smaller log size, and once on the bigger one) Can these be merged?

jgkamat:
So overall it LGTM code wise, so :+1: on that! However, the chart exported seems to have time inverted, for example, when time is 0, we actually get the most recent time value instead of the earliest time value. Once this is fixed, I'm :+1: on merging! :D

PR #918, opened 2017-03-05
Straight pp

kevinatorchen:
This pull request would close issue #66 in RRT.

[edit by @jgkamat] closes robojackets/rrt#66


soccer/planning/RRTPlanner.cpp
```diff
@@ -88,6 +91,9 @@ std::unique_ptr<Path> RRTPlanner::run(PlanRequest& planRequest) {
     vector<DynamicObstacle> actualDynamic;
     splitDynamic(obstacles, actualDynamic, dynamicObstacles);
 
+    //Straight line path planning
```
jgkamat: This comment can probably be removed.

soccer/planning/RRTPlanner.cpp
```diff
@@ -192,7 +215,11 @@ vector<Point> RRTPlanner::runRRT(MotionInstant start, MotionInstant goal,
     biRRT.setStepSize(*RRTConfig::StepSize);
     biRRT.setMinIterations(_minIterations);
     biRRT.setMaxIterations(_maxIterations);
```
jgkamat: Right now, this runs the RRT twice for every path plan, can we bump down minIterations to 0, and decrease the max iterations, and maybe even bump up the stepsize to get the first iteration as fast as possible? I think we might even be able to get it to only take one step (which would be much faster than running it twice). 

Make sure to set these dynamically, for example, don't hard code the stepsize higher, you could just set it to the distance between the goal and the target (and maybe you could set minIterations to 2).

soccer/planning/RRTPlanner.cpp
```diff
@@ -177,13 +184,29 @@ std::unique_ptr<InterpolatedPath> RRTPlanner::generateRRTPath(
...
+    vector<Point> straight = runRRTHelper(start, goal, motionConstraints, obstacles, state, shellID, true);
+    if (straight.size() != 0) {
+        cout << "Straight path!" << endl;
+        return straight;
+    }
+    cout << "Regular RRT Planning" << endl;
+    return runRRTHelper(start, goal, motionConstraints, obstacles, state, shellID, false);
+}
+
+vector<Point> RRTPlanner::runRRTHelper(MotionInstant start, MotionInstant goal,
```
jgkamat: Can you add a documentation header to this method?

kevinatorchen: There is already a Javadoc for this method in the header file. It reads:

/**
     * Helper method for runRRT(), which creates a vector of points representing 
     * the RRT path.
     */

jgkamat: oh, that slipped my mind, sorry, your good! :smile:

soccer/planning/RRTPlanner.cpp
```diff
@@ -6,12 +6,14 @@
...
 #include "RoboCupStateSpace.hpp"
 #include "motion/TrapezoidalMotion.hpp"
+#include "DirectTargetPathPlanner.hpp"
 
 #include <stdio.h>
 #include <stdlib.h>
 #include <Eigen/Dense>
 #include <algorithm>
 #include <iostream>
+#include <sys/time.h>
```
jgkamat: Can this be removed :P

jgkamat:
I think the latest commit reverted some updated submodules, I'll bump them back for now, so builds can pass again. Make sure to either run `git submodule update` or avoid commiting submodule changes when pulling new changes!

soccer/planning/RRTPlanner.cpp
```diff
@@ -6,6 +6,7 @@
 #include "RRTUtil.hpp"
 #include "RoboCupStateSpace.hpp"
 #include "motion/TrapezoidalMotion.hpp"
+#include "DirectTargetPathPlanner.hpp"
```
jgkamat: Is this include being used anywhere?

kevinatorchen:
Hey Jay, I have removed the unnecessary import statement, and I have modified the call to biRRT's getPath() method because the way the method returns the value is now different. (Instead of directly returning the value, it now uses an out parameter.)

jgkamat:
Thanks! This looks ready to merge, thanks!

I'll pull this in a bit.

PR #920, opened 2017-03-08
Fix rotation bug when halting during stop play

### LINKED ISSUE #866: When stopped, halting and re-stopping play causes robots to shift positions
This seems kind of weird, and might be a red flag of some bigger problems in our role assignments.

When we stop play, halt, and restart play, no robots should move, because everything is in their correct places. However, robots often switch places for no apparent reason. This may be due to the ball obstacle for some reason, but this needs to be triaged to figure out the underlying cause.

jcarn:
Closes #866 . So the following was happening:

1. Stop play starts and selects a tactic (circle around ball vs. circle around center)
2. Tactic doesn't know how many robots it will receive so it creates 6 move behaviors with destinations evenly spaced around the circle
3. Role assignment occurs and the tactic might not receive 6 robots, usually it gets 3
4. The (usually 3) robots are assigned to 3 of the 6 points around the circle based on their proximity to those original points
5. The move behaviors' points are updated now that the tactic knows it only has 3 robots
6. The robots stick to their current behavior because of the high robot change cost, and move to the new destination

I basically made the original assignment points impossible to reach so the tactic doesn't make stupid assignments, and once it knows how many robots it gets it can make actual assignments.

As a side note, do you think it might be a good idea to let behaviors assign their own robot change cost? This would let some behaviors that have a lower actual cost of switch robots, like the move play, switch easier than other with more overhead, like aiming and kicking.

### DIFF
--- a/soccer/gameplay/tactics/stopped/circle_near_ball.py
+++ b/soccer/gameplay/tactics/stopped/circle_near_ball.py
@@ -143,23 +143,18 @@ class CircleNearBall(composite_behavior.CompositeBehavior):
         if (self.num_robots != num_robots):
             self.num_robots = num_robots
             self.remove_all_subbehaviors()
-            i = 0
             for pt in range(6):
                 self.add_subbehavior(skills.move.Move(),
-                                     name="robot" + str(i),
+                                     name="robot" + str(pt),
                                      required=False,
-                                     priority=6 - i)
-                i = i + 1
-
-        i = 0
+                                     priority=6 - pt)
         #assign destinations for the number of robots we have
-        for pt in self.get_circle_points(num_robots):
+        for i, pt in enumerate(self.get_circle_points(num_robots)):
             self.subbehavior_with_name("robot" + str(i)).pos = pt
-            i = i + 1
+
         #unassign destinations from behaviors without robots
         for i in range(num_robots, 6):
             self.subbehavior_with_name("robot" + str(i)).pos = None
-            i = i + 1
 
         # set robot attributes
         for b in self.all_subbehaviors():
--- a/soccer/gameplay/tactics/stopped/circle_near_ball.py
+++ b/soccer/gameplay/tactics/stopped/circle_near_ball.py
@@ -29,14 +29,12 @@ class CircleNearBall(composite_behavior.CompositeBehavior):
                             lambda: not self.all_subbehaviors_completed(),
                             "robots aren't lined up")
 
-        i = 0
         #create move behaviors with no position (we can't assign position because we don't know how many bots we have)
-        for pt in range(6):
+        for i in range(6):
             self.add_subbehavior(skills.move.Move(),
                                  name="robot" + str(i),
                                  required=False,
                                  priority=6 - i)
-            i = i + 1
 
     def get_circle_points(self, num_of_points):
         radius = constants.Field.CenterRadius + constants.Robot.Radius + 0.01
@@ -143,11 +141,11 @@ class CircleNearBall(composite_behavior.CompositeBehavior):
         if (self.num_robots != num_robots):
             self.num_robots = num_robots
             self.remove_all_subbehaviors()
-            for pt in range(6):
+            for i in range(6):
                 self.add_subbehavior(skills.move.Move(),
-                                     name="robot" + str(pt),
+                                     name="robot" + str(i),
                                      required=False,
-                                     priority=6 - pt)
+                                     priority=6 - i)
         #assign destinations for the number of robots we have
         for i, pt in enumerate(self.get_circle_points(num_robots)):
             self.subbehavior_with_name("robot" + str(i)).pos = pt
--- a/soccer/gameplay/tactics/stopped/circle_near_ball.py
+++ b/soccer/gameplay/tactics/stopped/circle_near_ball.py
@@ -16,7 +16,7 @@ class CircleNearBall(composite_behavior.CompositeBehavior):
     def __init__(self):
         super().__init__(continuous=True)
 
-        self.num_robots = 0;
+        self.num_robots = 0
         self.add_transition(behavior.Behavior.State.start,
                             behavior.Behavior.State.running, lambda: True,
                             'immediately')
@@ -32,11 +32,10 @@ class CircleNearBall(composite_behavior.CompositeBehavior):
         i = 0
         #create move behaviors with no position (we can't assign position because we don't know how many bots we have)
         for pt in range(6):
-            self.add_subbehavior(
-                skills.move.Move(),
-                name="robot" + str(i),
-                required=False,
-                priority=6 - i)
+            self.add_subbehavior(skills.move.Move(),
+                                 name="robot" + str(i),
+                                 required=False,
+                                 priority=6 - i)
             i = i + 1
 
     def get_circle_points(self, num_of_points):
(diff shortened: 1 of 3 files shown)


soccer/gameplay/tactics/stopped/circle_near_ball.py
```diff
@@ -131,7 +131,10 @@ def execute_completed(self):
                 num_robots += 1
 
         i = 0
-        for pt in self.get_circle_points(num_robots):
+        for pt in self.get_circle_points(6 if num_robots == 0 else num_robots):
```
jgkamat: I'm kind of confused why the `6 if num_robots == 0 else num_robots` is needed? Could you explain why?

soccer/gameplay/tactics/stopped/circle_on_center.py
```diff
@@ -65,7 +65,10 @@ def goto_center(self):
         dirvec.rotate(robocup.Point(0, 0), -perRobot * ((num_robots - 1) / 2))
 
         for i in range(6):
-            pt = ball_pos + dirvec
+            pt = ball_pos + dirvec if num_robots > 0 else robocup.Point(
+                -constants.Field.Width, -constants.Field.Length)
+            if num_robots < 1:
+                pt = robocup.Point(-5, -5)
```
jgkamat: I'm not sure where this -5,-5 came from? What exactly is this?

soccer/gameplay/tactics/stopped/circle_near_ball.py
jcarn: If we don't know how many robots the tactic will have it defaults to 0, hence the "if num_robots == 0". When this is the case, we don't want the tactic to assign any robots so we give it 6 impossible points so that robots are automatically reassigned to the best point when they are actually added to the tactic and proper points are calculated.

soccer/gameplay/tactics/stopped/circle_on_center.py
jcarn: So -5, -5 is just a really far away point. I guess I put that mostly for testing purposes, ideally it would be like (infinity, infinity)

soccer/gameplay/tactics/stopped/circle_near_ball.py
jgkamat: (in response to both comments)

I think it would be better to properly handle the case of having no robots in the play. We could either set a variable to tell that no robots are assigned, transition the state into 'waiting for robot' or just set none here. This feels wayy too hacky, and can easily break in the future :P

jcarn: Yeah that makes a lot of sense. I'll whip that up

soccer/gameplay/tactics/stopped/circle_near_ball.py
```diff
@@ -155,11 +139,27 @@ def execute_running(self):
...
+        #if the number of robots has changed, recreate move behaviors to match new number of robots
+        if (self.num_robots != num_robots):
+            self.num_robots = num_robots
+            self.remove_all_subbehaviors()
+            i = 0
+            for pt in range(6):
+                self.add_subbehavior(skills.move.Move(),
+                                     name="robot" + str(i),
+                                     required=False,
+                                     priority=6 - i)
```
jgkamat: This can be `6-pt`, since pt increases from 0->5. It would clean up the `i=0` and `i+1` bits, they are kinda 'unpythonic'. I'm not sure what that means though.

soccer/gameplay/tactics/stopped/circle_near_ball.py
```diff
@@ -155,11 +139,27 @@ def execute_running(self):
...
+                i = i + 1
 
         i = 0
+        #assign destinations for the number of robots we have
         for pt in self.get_circle_points(num_robots):
             self.subbehavior_with_name("robot" + str(i)).pos = pt
             i = i + 1
+        #unassign destinations from behaviors without robots
+        for i in range(num_robots, 6):
+            self.subbehavior_with_name("robot" + str(i)).pos = None
```
jgkamat: Can we just remove the subbehavior here? It might cause problems with trying to reference it now, but it should work now that the play reloads itself when the numRobots changes, right?

soccer/gameplay/tactics/stopped/circle_on_center.py
```diff
@@ -26,33 +26,34 @@ def __init__(self, min_robots=0):
...
+        #if the number of robots has changed, recreate move behaviors to match new number of robots
+        if (self.num_robots != num_robots):
+            self.num_robots = num_robots
+            self.remove_all_subbehaviors()
+            i = 0
+            for pt in range(6):
+                self.add_subbehavior(skills.move.Move(),
+                                     name="robot" + str(i),
+                                     required=False,
+                                     priority=6 - i)
```
jgkamat: Same comment as before, `6 - i` can be `6 - pt` and we can remove the i stuff.

soccer/gameplay/tactics/stopped/circle_near_ball.py
```diff
@@ -155,11 +139,27 @@ def execute_running(self):
         for b in self.all_subbehaviors():
             if b.robot is not None:
                 num_robots += 1
+        #if the number of robots has changed, recreate move behaviors to match new number of robots
+        if (self.num_robots != num_robots):
+            self.num_robots = num_robots
+            self.remove_all_subbehaviors()
+            i = 0
+            for pt in range(6):
```
jgkamat: Do you know if this `6` can be replaced with a constant? If not, we should open a new issue to refactor all these 6s away into references to Constants.hpp

jcarn: There is one in Constants.hpp, I made an issue to port that into python

soccer/gameplay/tactics/stopped/circle_near_ball.py
jcarn: Yup changed. I think unpythonic is anything that doesn't make you say "huh, python is an obviously superior language"

soccer/gameplay/tactics/stopped/circle_near_ball.py
jcarn: Off the top of my head I don't know why but when I did this it caused the play to not work at all.

jcarn: Oh probably because the play starts with no robots, so if it never adds any behaviors (because it keeps remove them) it will never get any robots and nothing will happen. It needs behaviors with no position to get robots

soccer/gameplay/tactics/stopped/circle_on_center.py
jcarn: Yuh

soccer/gameplay/tactics/stopped/circle_near_ball.py
```diff
@@ -143,23 +141,18 @@ def execute_running(self):
...
                 self.add_subbehavior(skills.move.Move(),
                                      name="robot" + str(i),
                                      required=False,
                                      priority=6 - i)
-                i = i + 1
-
-        i = 0
         #assign destinations for the number of robots we have
-        for pt in self.get_circle_points(num_robots):
+        for i, pt in enumerate(self.get_circle_points(num_robots)):
```
jgkamat: :heart:

Issue #894, opened 2017-02-04
Deprecate 14.04


jgkamat:
This is a (possibly) long term issue. 

If we decide that no one is using 14.04, then we might do this right now. 

This issue will track things that need to be done when dropping 14.04 support. This dosen't apply to robojackets/robocup-firmware, as that already requires 16.04. See comments on #789 for more information.

Dropping 14.04 will mean we forever loose the PPA madness that we dealt with in previous years (I will never let that happen again). If you have questions, or objections, note them here. We aren't doing anything just yet.

- [x] See if anyone in RJ is using 14.04 :+1:
  - [x] IGVC :+1:
  - [x] RoboRacing :+1: 
  - [x] Outreach/TE Laptops :+1: 
  - [x] Any RC Members :+1:
  - [x] See if RJ has any demo laptops with 14.04 :+1: 
- [ ] See if anyone in the community needs 14.04 support
- [x] Determine if it's reasonable to drop 14.04 support (EOL for 14.04 is in 2019)
- [x] Set a target date for dropping 14.04
- [ ] Investigate arch, fedora, or debian8/9 build systems (we can do this, since we're dropping 14.04).

jgkamat:
I'm going to go ahead with this now. Since 18.04 is in alpha now, we'll have a lot more freedom on adapting 18.04 if we drop 14.04 now. 

I'm expecting to do this sometime in the next couple of weeks.

PR #955, opened 2017-03-30
Bump rrt version to master

jgkamat:
Needed to make a small change because of a minor API change

This is just bumping the RRT to master, hopefully nothing breaks too badly :P


jgkamat:
Forgot we need qtdeclarative now, I guess this is blocked on #894

jgkamat:
or maybe not! =)

PR #958, opened 2017-04-04
Adaptive formation

### LINKED ISSUE #822: New Offensive Play - Adaptive Formation
Adaptive Formation allows for multiple robots on offense to move from defense into a more offensive setup.

On the offensive side, passing will find the best position to receive in based on "space", location on the field, shot chance, and a few other things.

On defense, I'm thinking of adding more a zone style defense for robots further away from the ball while producing a heavier marking style defense near the ball narrowing down pass possibilities.

Non-play features being added:
- [x] Number of robots on their offense
- [x] Draw heuristics on top of field as a density graph to make tuning coefficients easier
- [x] Space heuristic (Think Kernel Density Estimation but for opponent robots)
- [x] Field position heuristic (Center-ness, dist from goal, and angle on goal weights)
- [x] Shot chance method

Offense:
- [x] Find best position to pass to
- [x] Update shooting to take opponent movement into consideration (Can improve with more motion prediction)

Defense:
- [x] Risk score for opponent robots
- [x] Estimate direction of pass / shot based off
- [x] Find best position for the 3 offensive robots
- [x] Chip clear possibility when deep in our side of the field

JNeiger:
Will close #822 
I'm still working on this, but I'm getting close to the point of being ready to merge.

- [x] Put the final touches on the defense
- [x] Clean up a few bugs
- [x] Lots of code cleanup


JNeiger:
@jgkamat I think it's about good. Let me know how performance is, I'm getting a few lag spikes here and there, but I'm running on a VM so it's hard to say just how bad it is.

jgkamat:
I'll try to take a look at this soon, but with finals approaching I'm not sure how much I can do. I'll definetely get this before a few weeks into the summer though :O

soccer/gameplay/evaluation/passing_positioning.py
```diff
@@ -0,0 +1,175 @@
...
+
+    points = get_points_from_rect(evaluation_zone, kick_point, 0.6, 0.5)
+
+    if points is None or len(points) == 0:
+        # Nothing can be done
+        return None, 0
+
+    # TODO: Setup to be a list of the top X points
+    bestScore = None
+    bestPointt = None
```
jgkamat: This should be `bestPoint` right? Python is annoying ;)

soccer/gameplay/evaluation/passing_positioning.py
```diff
@@ -0,0 +1,175 @@
...
+# @param kick_point: Point that we are passing from
+# @param evaluation_zone: Area to evaluate passes to
+# @param ignore_robots: Robots to ignore when calculating scores
+# @param field_weights: A tuple of the 3 difference weights to apply to field position 
+#               (Centerness, Distance to their goal, Angle off their goal)
+# @param weights: A tuple of the 4 different weights to apply to the evaulations overall (Weights are normalized)
+#               (space, field_position, shot_chance, kick_proximity)
+# @param debug: Displays pass lines and scores onto the field when true
+# @return bestPoint and bestScore in that order
+def eval_best_receive_point(kick_point,
```
jgkamat: Once this is merged in, we should probably replace the touchpass receive point in favor of this one so we don't have duplicates. We'll think about that later though.

soccer/gameplay/tactics/defensive_forward.py
```diff
@@ -0,0 +1,216 @@
...
+        return None  # pos * self.block_dist
+
+    def we_have_ball(self):
+        has_ball = False
+        for bot in main.our_robots():
+            has_ball = has_ball or robot_has_ball(bot)
+        return has_ball
+
+    # Wether any robot can collect the ball before the opponent
+    def within_range(self):
```
JNeiger: I'll have to double check, but placing the robot on the other side of an opponent robot should make it return false.

Overall, I was looking for a way to see whether we should contest the ball or not. I've had trouble with just linear distance in the past so I tried to used a poor-man's path planner to get that distance value. It's basically STOx's algorithm

soccer/gameplay/tactics/defensive_forward.py
```diff
@@ -0,0 +1,216 @@
...
+
+        # Find closest robot on our team
+        for bot in main.our_robots():
+            dist = self.estimate_path_length(bot.pos, target_pos,
+                                             main.their_robots())
+            if (dist < shortest_our_dist):
+                shortest_our_dist = dist
+
+        # Greater than 1 when we are further away
+        return shortest_our_dist / shortest_opp_dist < 1.05
```
JNeiger: That would be shortest_our_dist is less than shotest_opp_dist + 0.05 meters. Mine right now is testing if we are either closer or within 5% of the opponent. 

I can make it `shortest_our_dist < shortest_opp_dist * 1.05` if that would be cleaner

soccer/gameplay/plays/offense/adaptive_formation.py
```diff
@@ -0,0 +1,276 @@
...
+        closest_distance = (evaluation.opponent.get_closest_opponent(
+            main.ball().pos, 1).pos - main.ball().pos).mag()
+        if (closest_distance > self.clear_distance_cutoff):
+            return False
+
+        return True
+
+    def on_enter_collecting(self):
+        # 2 man to man defenders and 1 zone defender
+        defensive_forward = tactics.defensive_forward.DefensiveForward()
```
JNeiger: Yeah, that was one thing I was running into. In the best case, I would like to have the robots already moving towards the next pass location or opening up positions on the field by exploiting the defense but I couldn't think of a good score function to use. There are a bunch of possibilities here.

For now, I think I'll just make another play to do something really simple like hold the robots at 50% of the ball Y coordinate or something until they are needed as a pass.

soccer/gameplay/tactics/defensive_forward.py
jgkamat: Actually, I was able to later get this working, I think it's just a little hard to activate. Maybe this needs a bit of tweaking.

soccer/gameplay/tactics/defensive_forward.py
jgkamat: yah, I think that would be easier to read (and the intent is much more direct).

Why is there a *1.05 there? is that just a small buffer?

soccer/gameplay/tactics/defensive_forward.py
```diff
@@ -0,0 +1,216 @@
...
+
+        # Reset time since call
+        self.time_since_call = 0
+
+    # Continue updating the mark positions
+    def execute_blocking(self):
+        self.time_since_call = (self.time_since_call + 1) % self.cache_amnt
+
+        # You don't want to test if its 0 because it would execute instantly
+        if (self.time_since_call % self.cache_amnt == self.cache_amnt - 1):
```
JNeiger: Yep! Forgot to change that once I fixed the overflow problem on line 87

soccer/gameplay/evaluation/passing_positioning.py
JNeiger: There should be a warning for that or something...

soccer/gameplay/tactics/defensive_forward.py
JNeiger: Yeah, My estimate isn't the best so I figured you might as well be overaggressive. It's tunable so we can change it if it gets bad.

soccer/gameplay/evaluation/passing_positioning.py
jgkamat: There actually is a tool for this kind of thing, `flake8`, but no one uses it because it spits out wayy too many warnings. We can probably put a toned down version of flake8 into our CI system to catch python errors.

soccer/gameplay/tactics/defensive_forward.py
```diff
@@ -0,0 +1,216 @@
...
+                            lambda: self.within_range(), 'Collecting')
+
+        self.add_transition(
+            DefensiveForward.State.collecting, DefensiveForward.State.blocking,
+            lambda: not self.within_range(), 'Back to blocking')
+
+        self.add_transition(
+            DefensiveForward.State.collecting,
+            behavior.Behavior.State.completed,
+            lambda: self.collector is not None and evaluation.ball.robot_has_ball(self.collector.robot),
```
JNeiger: Is *.robot left as None in that case?

jgkamat: I'm fairly sure it is, so this statement can be fixed by adding a check for `*.robot is not None`

soccer/gameplay/plays/testing/test_adaptive_formation_weights.py
```diff
@@ -0,0 +1,126 @@
...
+import evaluation.passing_positioning
+import evaluation.shooting
+import evaluation.defensive_positioning
+import constants
+import robocup
+import math
+import enum
+
+
+class TestAdaptiveFormationWeights(play.Play):
```
JNeiger: It would be nice if we could make this a normal functionality of the simulator or something. Pass it function pointers and it will display their values over the entire field. I can't think of a good place to put it though. 

If we did that, all of the commented evaluation functions could be their own separate play and it would be really simple to implement.

jgkamat: We could even do something like parallelizing the execution if we did that (instead of just looping over the entire field, spawn 10 threads, and have each one of them scan over a section, and then return the result), and that would make such visualizations much faster! I think we might even be able to implement something like this on the python side.

soccer/gameplay/plays/testing/test_adaptive_formation_weights.py
```diff
@@ -0,0 +1,126 @@
...
+
+    def __init__(self):
+        super().__init__(continuous=True)
+        self.add_state(TestAdaptiveFormationWeights.State.testPointCoeff,
+                       behavior.Behavior.State.running)
+        self.add_state(TestAdaptiveFormationWeights.State.testBestPass,
+                       behavior.Behavior.State.running)
+
+        # Enable which portion we want to test
+        mode = 1
```
JNeiger: It does not unless there are some super secret settings somewhere.

soccer/gameplay/plays/testing/test_adaptive_formation_weights.py
JNeiger: I'm going to leave this for another issue so this one doesn't get /too/ complicated. I should be able to implemented it later this week / next week depending on my schedule.

soccer/gameplay/tactics/defensive_forward.py
JNeiger: Are you saying the function is too restricted and doesn't return true often enough?

soccer/gameplay/plays/testing/test_adaptive_formation_weights.py
jgkamat: Hmm, we should probably figure out how to add that (if qt will let us), we'll switch that over when that happens I guess...

JNeiger:
Was there anything else you wanted changed @jgkamat?

jgkamat:
It looks like you've fixed everything I commented on, I need to give this another pass though, once I do that I think this will be ready to merge in! :smile: 

I don't think I can get around do doing that until after finals though (maybe I'll get a bit of time this weekend, we'll see). I'll try and get to it ASAP! :P

soccer/gameplay/tactics/defensive_forward.py
```diff
@@ -0,0 +1,215 @@
...
+        for i in range(0, 2):
+            self.defenders[i] = skills.mark.Mark()
+            self.add_subbehavior(self.defenders[i],
+                                 self.names[i],
+                                 required=False,
+                                 priority=20 - 5 * i)
+            self.defenders[i].mark_robot = self.mark_bots[i]
+
+    # Uses their predicted kick direction to block
+    def get_block_pos(self, bot):
```
JNeiger: Oh yeah, I completely forgot about this. It was going to be a ghetto version of `mark` before I found out `mark` exist.

soccer/gameplay/plays/offense/adaptive_formation.py
```diff
@@ -0,0 +1,290 @@
...
+
+    def on_enter_collecting(self):
+        self.remove_all_subbehaviors()
+
+        # 2 man to man defenders and 1 zone defender
+        defensive_forward = tactics.defensive_forward.DefensiveForward()
+        self.add_subbehavior(defensive_forward, 'defend', required=True)
+
+    def on_exit_collecting(self):
+        self.remove_all_subbehaviors()
```
JNeiger: I accidentally disabled the zone midfielder thing in another commit. It should be added back. Let me know if you would rather a more defensive thing or if this works. It's really easy to switch them

jgkamat: The new behavior looks pretty good for a first pass, nice! :smile:

soccer/gameplay/plays/offense/adaptive_formation.py
```diff
@@ -0,0 +1,295 @@
...
+        self.add_transition(
+            AdaptiveFormation.State.passing, AdaptiveFormation.State.dribbling,
+            lambda: self.subbehavior_with_name('pass').state == behavior.Behavior.State.completed,
+            'Passed')
+
+        # Reset to collecting when ball is lost at any stage
+        self.add_transition(
+            AdaptiveFormation.State.dribbling,
+            AdaptiveFormation.State.collecting,
+            lambda: self.dribbler is not None and self.dribbler.robot is not None and evaluation.ball.robot_has_ball(self.dribbler.robot),
```
jgkamat: These transitions `dribbling->{collecting,clearing,shooting}` have the possibility of happening at the same time. Right now it prefers collecting since that's the lowest Enum number, but it would be preferable to explicitly make sure there's only one option at a time. Could you change the logic to depend on calls to the other states (for example, maybe prefer collecting to shooting and clearing).

soccer/gameplay/plays/offense/adaptive_formation.py
```diff
@@ -0,0 +1,295 @@
...
+        clear.aim_params['desperate_timeout'] = 3
+        clear.use_chipper = True
+        self.add_subbehavior(clear, 'clear', required=False)
+
+    def on_exit_clearing(self):
+        self.remove_subbehavior('clear')
+
+    def on_enter_passing(self):
+        # TODO: Use the moving recieve when finished
+        self.add_subbehavior(
```
jgkamat: In the current configuration, the pass receiver (and actually the kicker too) has a higher priority than defenders, causing a defender to be pulled off of defense. We can solve this by lowering the priority of this play (I think), or by simply making this play have a score of `inf` when we have less than 5 robots.

You can test out priorities by deleting our robots in `simulator/config/two_teams.xml`

JNeiger: I'm not sure I understand what your end goal is. Do you want this play to always try and steal 3 robots even if it means having no defense?

jgkamat: I would like the play to have lower priority than defense at all times, basically, making the play always prefer to give robots up to defense, rather than taking them from the defense.

PR #968, opened 2017-04-12
Implement behavior sequence

### LINKED ISSUE #438: Implement BehaviorSequence
We already have a class for this, but it isn't fully implemented and isn't used anywhere.  It needs to be rethought a bit and used in a couple places.  The main idea is to give a list of behaviors and have it run one after another.

jcarn:
I added some basic implementations of the sequence, looking into other more advanced ways to use it. I might consider going through and refactoring some of the state machine based plays, mostly things whose transitions look like "self.subbehavior_with_name("x").completed()"
Closes #438


soccer/gameplay/plays/sequences/move_sequence.py
```diff
@@ -0,0 +1,22 @@
...
+    def __init__(self, positions = None):
+        super().__init__()
+        self._positions = [] if positions == None else positions
+
+    def on_enter_start(self):
+        super().on_enter_start()
+        for pos in self.positions:
+            self.behaviors.append(skills.move.Move(pos))
+
+    @property
```
jcarn: this is what I am looking at, isn't working. It says I have no attribute "_positions" at line 16

jgkamat: Not sure, but I think this will solve your problem. Can you change line 7 above to use `self.positions` instead of `self._positions`? Let me know if that dosen't work.

jcarn: Yeah I've tried a few permutations, it it doesn't work

jgkamat: Ok, I've pinned down the problem. In your super call in the constructor, we enter the behaivor_sequence constructor, and the behavior_sequence constructor sets this play to the start state, which calls our overloaded function, which reads the value before we set it. We can fix this by not having an `on_enter_start` and instead making a transition when we start the play up (maybe on_enter_running?) I don't think we should ever have an `on_enter_start` in a play.

jcarn:
I'm going to go through and clean the code up, but I'm not really sure where to go from here. I could refactor some plays into sequences, but if you have any ideas on where else i could take this that'd be great.

jgkamat:
Sorry for not getting to this sooner, I think that the next step is to write (at least a bit) of documentation about behavior plays/sequences, and maybe write an example behavior sequence file (to be used in the docs page). Once that is in place, we can begin writing new plays with sequences (probably next year, for stability)!

soccer/gameplay/plays/sequences/move_sequence.py
```diff
@@ -0,0 +1,25 @@
+import single_robot_sequence
+import role_assignment
+import skills.move
+import behavior
+
+#Moves a SINGLE robot in a sequence of points
+class MoveSequence(single_robot_sequence.SingleRobotSequence):
```
justbuchanan: You may be able to just use SingleRobotSequence directly instead of having a subclass specialized for Move. Then you could use it like:

~~~{.py}
seq = SingleRobotSequence(behaviors = [skills.move.Move(pos) for pos in my_positions])
~~~

soccer/gameplay/single_robot_sequence.py
```diff
@@ -0,0 +1,65 @@
+import behavior_sequence
+import single_robot_behavior
+import role_assignment
+
+#A sequence to be carried out by a single robot, attempting to add non-single robot behaviors will raise exceptons
+class SingleRobotSequence(behavior_sequence.BehaviorSequence):
```
justbuchanan: Python lets you inherit from multiple classes at once, so you can make SingleRobotSequence a subclass of both BehaviorSequence and SingleRobotCompositeBehavior.  If you do that, I think you can replace this whole class definition with:

```{.py}
class SingleRobotSequence(
    behavior_sequence.BehaviorSequence,
    single_robot_composite_behavior.SingleRobotCompositeBehavior):
    def __init__(self, behaviors=[], autorestart=lambda: True):
        single_robot_composite_behavior.SingleRobotCompositeBehavior.__init__(
            self,
            autorestart=autorestart)
        behavior_sequence.BehaviorSequence.__init__(self, behaviors=behaviors)

```

That way you get role_requirements(), assign_roles(), autorestart, etc for free. It looks like those methods are mostly the same as SingleRobotCompositeBehavior.

soccer/gameplay/plays/sequences/move_sequence.py
jcarn: I was mostly using the MoveSequence as an example of how to implement a behavior sequence, but that is a clever way of doing it. Are there any cases that would make a good example of when to use behavior sequence?

soccer/gameplay/single_robot_sequence.py
jcarn: I did a lot of work in single_robot_composite_behavior, so I considered using the same multiple inheritance but honestly I was kind of scared to implement it. I'll give it a go now that I already have a working version

soccer/gameplay/plays/sequences/move_sequence.py
justbuchanan: Gotcha, makes sense. Just wanted to point out that the behavior sequence should be able to be used without creating a subclass for each type of behavior contained (i.e. Move). It would be great if BehaviorSequence could be used directly for things like `BehaviorSequence([LineUp(), PivotKick(), VictoryDance()])` without having to create a subclass.

From a brief skim of our plays/behaviors, I see a few places that could benefit from using sequences. One that you could probably do is RepeatedLineUp. It could look something like this:

~~~
sequence = BehaviorSequence([
  tactics.LineUp(generate_line(left)),
  Delay(1),
  tactics.lineUp(generate_line(right)),
  Delay(1)])
sequence.repeat = True
add_subbehavior(sequence)
~~~

The Delay behavior doesn't actually exist yet, but it would be easy to create a behavior that takes a timeout value and enters the completed state once the time elapses.

Another play that would actually be a great candidate for sequences is StressTest. It's basically equivalent to your MoveTest. Also take a look at TestCapture.

One (semi-related) thought I have from browsing through plays is that you could create another behavior similar to sequence, but that runs behaviors in parallel (maybe call it "BehaviorGroup"?). This would be a fairly simple wrapper class that just adds all of the given behaviors as subbehaviors then changes its state based on the subbehaviors' states. When all subbehaviors complete, it transitions to completed. If a subbehavior "un completes", it transitions to the running state. This could replace a lot of the duplicated logic we have throughout our gameplay code for doing this. A few behaviors that do this include: tactics.LineUp, stopped.CircleNearBall, and stopped.CircleOnCenter. Just a thought - it may be more trouble than it's worth.

I could see it being more useful for silly demos such as:
~~~
BehaviorSequence([
  BehaviorGroup([Move(pos) for pos in POINTS_FORMING_LETTER_G]),
  BehaviorGroup([Move(pos) for pos in POINTS_FORMING_LETTER_T])])
~~~

soccer/gameplay/single_robot_sequence.py
```diff
@@ -0,0 +1,51 @@
+import behavior_sequence
+import single_robot_behavior
+import role_assignment
+
+#A sequence to be carried out by a single robot, attempting to add non-single robot behaviors will raise exceptons
+class SingleRobotSequence(behavior_sequence.BehaviorSequence, single_robot_behavior.SingleRobotBehavior):
```
justbuchanan: If you subclass from SingleRobotCompositeBehavior instead of SingleRobotBehavior, you can inherit assign_roles() for free and avoid having to copy that code here.

soccer/gameplay/single_robot_sequence.py
```diff
@@ -0,0 +1,51 @@
...
+    @property
+    def behaviors(self):
+        return behavior_sequence.BehaviorSequence.behaviors.fget(self)
+
+    @behaviors.setter
+    def behaviors(self, value):
+        #Insures that all behaviors are
+        for b in value:
+            if not isinstance(b, single_robot_behavior.SingleRobotBehavior):
+                raise TypeError("SingleRobotSequence requires single robot behaviors")
```
justbuchanan: If you subclass SingleRobotCompositeBehavior, I think you can add this type check to it's implementation of add_subbehavior() and avoid having to override BehaviorSequence's `behaviors` property.

jgkamat:
@jcarn is this ready for another round of reviews? I'm not sure if you're working on it right now or not...

jgkamat:
In the interest of finishing this before comp, I'm going to hijack this PR. Let me know if you object @jcarn :smile:

soccer/gameplay/tactics/line_up.py
```diff
@@ -8,7 +8,8 @@
 
 class LineUp(composite_behavior.CompositeBehavior):
 
-    y_start = 1.0  # sometimes we have issues if we're right in the corner, so we move it up a bit
+    DEBOUNCE_MAX = 3
```
jgkamat: I added some debouncing to this play, since it was previously in the completed state when it first starts up (which makes it hard to use). It shouldn't break anything existing, I hope!

jgkamat:
I would appreciate a review on this, but I would like to merge this quickly because comp is coming up. We can patch anything that is broken once comp is over (or during comp).

jgkamat:
Scheduling merge for end of day, please get any objections in before then :).

This doesn't touch any 'important' files, other than lineup, which isn't for comp, so this shouldn't be too big of a problem if it breaks anything.

jgkamat:
Merging now, let me know if you see any issues :)

PR #977, opened 2017-04-28
Better way of finding the best point in passing / one touch

JNeiger:
This is an addition to the adaptive formation stuff. I split it up since this is could cause some major stability problems since it bridges the python / cpp in a weird way and the setup in adaptive formation works ok for now.

- [x] Finish up python interface
- [x] Clean up code / comments
- [x] Implement for passing_positioning
- [x] Add doc page comparing and contrasting the optimization algorithms

Edited files:
gameplay/plays/testing/test_adaptive_formation_weights.py
gameplay/robocup-py.cpp
soccer/optimization/NelderMead2D*


jgkamat:
Let's wait until #958 is merged until we look at this

soccer/gameplay/robocup-py.cpp
```diff
@@ -543,6 +546,52 @@ void Point_set_x(Geometry2d::Point* self, float x) { self->x() = x; }
...
+
+boost::shared_ptr<std::function<float(Geometry2d::Point)>>
+stdfunction_constructor(PyObject* function) {
+    // Create aliased function to hid python function args
+    std::function<float(Geometry2d::Point)> f =
+        std::bind(&point_python_callback, std::placeholders::_1, function);
+
+    return boost::shared_ptr<std::function<float(Geometry2d::Point)>>(
+        new std::function<float(Geometry2d::Point)>(f));
+}
```
JNeiger: Do I need to increment the reference counter for python functions if I'm going to be calling them in the "future"?

jgkamat: I'm not sure but I would lean towards yes, that's a tricky boost python question :/

jgkamat:
I'll try and take a look at this sometime this week if I get some time, otherwise I'll have an initial review in this weekend! Thanks again for all this hard work! :smile: :tada:

soccer/optimization/NelderMead2DTest.cpp
```diff
@@ -0,0 +1,56 @@
...
+
+    EXPECT_NEAR(nm.getValue(), 0, 0.001);
+    EXPECT_NEAR(nm.getPoint().x(), 0, 0.001);
+    EXPECT_NEAR(nm.getPoint().y(), 0, 0.001);
+}
+
+TEST(NelderMead2D, iteration_limit) {
+    std::function<float(Geometry2d::Point)> f = &evalFunction2;
+    NelderMead2DConfig config(
+        &f, Geometry2d::Point(1, 1), Geometry2d::Point(.0001, .0001),
```
jgkamat: can't this `f` variable be replaced with `&evalFunction2` to make this a bit simpler? Not entirely sure though.

soccer/gameplay/evaluation/passing_positioning.py
```diff
@@ -118,57 +80,25 @@ def eval_singl_point(kick_point, receive_point, ignore_robots, field_weights,
...
-    return bestPoint, bestScore
+                            weights=(1, 4, 15, 1)):
+    pythfunc = functools.partial(eval_single_point, kick_point, ignore_robots, \
+                field_weights, weights)
+    cppfunc = robocup.stdfunction(pythfunc)
+    nmConfig = robocup.NelderMead2DConfig(cppfunc, \
+                                        kick_point, \
+                                        robocup.Point(0.5, 2), \
+                                        robocup.Point(0.01, 0.01), \
+                                        1, 2, 0.75, 0.5, 50, 1, 0.1)
```
JNeiger: I should probably change these to arguments of `eval_best_recieve_point` because they are dependent on the function being optimized. The biggest change that I did was messing with the starting points of the triangle. Most of the other arguments are the default values I took from wikipedia.

Nelder Mead can be visualized as a triangle being flipped and resized as it moves around trying to find the best point. [Here](https://en.wikipedia.org/wiki/Nelder%E2%80%93Mead_method#/media/File:Nelder-Mead_Simionescu.gif) is a decent visualization of it.

As for the arguments themselves...
`kick_point` is the starting point of the triangle. 
`Point(0.5, 2)` describes where the other two points are. I want to say the X axis is flipped so the two points are +-0.5 , 2 in reference to the staring point. 
`Point(0.01, 0.01)` is the exit condition. Once the bounding box of the triangle is this size, the algorithm exits. 
`1` is the scalar value for a reflection of the triangle.
`2` is the scalar value for an expansion of a point in the triangle. 
`.75` is the contraction scalar value for a point in the triangle. 
`.5` is the shrink value for all points in the triangle. 
`50` is the max number of iterations. 
`1` and `0.1` specify the max value of the function and a range so we can exit the optimization early if we are close to the max.

To test these, I've been running `test_adaptive_formation_weights` and just checking where the white square is in comparison to where I think it should be.

soccer/optimization/NelderMead2DTest.cpp
JNeiger: I don't think so. Since we need the address of the std::function, we would need the address of the address of the function pointer `evalFunction2`. This was the cleanest way I could think of.

Do you remember that video that you showed everyone at the beginning of the year? It was like 4-5 minutes long and how a function pointer could be deferenced multiple times without changing the meaning to make the code look super complicated, eg: `****evalFunction2`. Does std::function work the same way?

jgkamat: oh, I get this now, I just confused myself...

I think this is the cleanest way to do this, I don't think that `std::function` has the same behavior of de-referencing function pointers as standard function pointers have. 

Does NelderMead2D need to take a function pointer, instead of a function reference? That might not be possible, since we're passing some python functions to it above, but that would make this a little bit cleaner.

JNeiger: I don't see why not. I'll try it and find out.

soccer/gameplay/evaluation/passing_positioning.py
```diff
@@ -118,57 +81,31 @@ def eval_singl_point(kick_point, receive_point, ignore_robots, field_weights,
...
-    if bestPoint is None:
-        return None, 0
-
-    if (debug):
-        main.system_state().draw_line(
-            robocup.Segment(kick_point, bestPoint), constants.Colors.Red,
-            "Best Point")
-
-    return bestPoint, bestScore
+                            nelder_mead_args=(robocup.Point(0.5, 2), \
```
jgkamat: nitpick: these backslashes aren't needed in `()`.

jgkamat:
@JNeiger are you 👍 on merging this? I just want to get confirmation from you before we do.

JNeiger:
Once this passes the checks, it's all good!

jgkamat:
Merging now, let me know if you see any issues :tada:

PR #981, opened 2017-05-07
Simplify and expand myRestart marking behaivor

### LINKED ISSUE #976: During theirRestart, marking is not assigning bots correctly to mark enemies.
Seems to work correctly for 1 bot however the other two bots running marking are inactive.

jgkamat:
Closes #976

This makes the theirRestart play consider robots that are far away from the ball, and does proper role assignment rather than manually sorting the robots somehow.

If someone can review/sanity check this, that would be great! :smile: 

Also, the 'cost' function I'm sorting on can probably be replaced with some of the adaptiveformation stuff (if it can predict threat levels of robots).

### DIFF
--- a/soccer/gameplay/plays/restarts/their_restart.py
+++ b/soccer/gameplay/plays/restarts/their_restart.py
@@ -46,41 +46,16 @@ class TheirRestart(standard_play.StandardPlay):
         their_kicker = min(main.their_robots(),
                            key=lambda opp: opp.pos.dist_to(ball_pos))
 
-        # we build an array of (OpponentRobot, float distToClosestOfOurBots) tuples
-        # we'll use these in a sec to assign our marking robots
-        open_opps_and_dists = []
-        for opp in main.their_robots():
-            # ignore their kicker
-            if opp == their_kicker: continue
-
-            ball_dist = opp.pos.dist_to(ball_pos)
-
-            # see if the opponent is close enough to the ball for us to care
-            # if it is, we record the closest distance from one of our robots to it
-            if ball_dist < 3.0:
-                # which of our robots is closest to this opponent
-                closest_self_dist = min([bot.pos.dist_to(opp.pos)
-                                         for bot in main.our_robots()])
-                open_opps_and_dists.append((opp, closest_self_dist))
+        # we build an array of OpponentRobots sorted rudimentarily by their threat
+        # Right now, this is (inverse of) distance to the ball * 2 + y position.
+        # Needs tuning/improvement. Right now this is excessively defensive
+        sorted_opponents = sorted(
+            filter(lambda x: x != their_kicker, main.their_robots()),
+            key=lambda x: x.pos.dist_to(ball_pos) * 2 + x.pos.y)
 
         # Decide what each marking robot should do
-        # @open_opps contains the robots we want to mark (if any)
+        # @sorted_opponents contains the robots we want to mark by priority
         # any robot that isn't assigned a mark_robot will move towards the ball
         for i, mark_i in enumerate(self.marks):
-            if mark_i.robot != None:
-                if i < len(open_opps_and_dists):
-                    # mark the opponent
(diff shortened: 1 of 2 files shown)


soccer/gameplay/plays/restarts/their_restart.py
```diff
@@ -46,41 +46,16 @@ def execute_running(self):
...
-                # which of our robots is closest to this opponent
-                closest_self_dist = min([bot.pos.dist_to(opp.pos)
-                                         for bot in main.our_robots()])
-                open_opps_and_dists.append((opp, closest_self_dist))
+        # we build an array of OpponentRobots sorted rudimentarily by their threat
+        # Right now, this is (inverse of) distance to the ball * 2 + y position.
+        # Needs tuning/improvement. Right now this is excessively defensive
+        sorted_opponents = sorted(
+            filter(lambda x: x != their_kicker, main.their_robots()),
+            key=lambda x: x.pos.dist_to(ball_pos) * 2 + x.pos.y)
```
JNeiger: I would almost rename `x` to something else, maybe `r`? Its a little awkward having `x.pos.y`

jgkamat:
I would like to merge this soon if no one has a :-1:, I'm not sure exactly what the old behavior was doing (with blacklisting robots that are far away), and this one makes more sense imo. This behavior could use some improvement, but we don't want idle robots at comp! :)

joshhting:
So as someone who spends most of their time in the rrt and not robocup-software, how would one go about testing this?

jgkamat:
You'll need to start soccer and the simulator, and then start the SSL refbox (which you can get from their GitHub) once the refbox has been started in a game (do a kickoff to started the game), put the ball on a sideline and click indirect/direct for the opposite team to initiate the kickoff. You'll need to make sure that all the restarts are checked as well before doing this. Let me know in the gitter if you run into problems!

joshhting:
Is optimally picking which of our robots to assign to block which enemy robot in the scope of this pull request? If so, this is definitely not the optimal way to assign our robots to their robots.
screenshot from 2017-05-24 20-02-20

jgkamat:
The main goal is to make sure everything is assigned. The sorting works like this: add every opponents y distance from our goal with the distance from the ball, and minimize the result. The reason the closest robot is excluded is because we assume that's the kicker robot. 

I think that assignment is pretty reasonable, but I would like to test out many more enemy configurations and see if anything stands out if possible before calling this good.

jgkamat:
It seems like role assignment might be borked from that picture though, so I'll need to take a look at that...

jgkamat:
@joshhting are you :+1: on merging this? I want to get this through so I can work on other things without worrying about conflicts :P

joshhting:
Everything else is fine, my only concern with the above image was that it seemed like our robot 2 should have been assigned to their robot 3 instead of 2, and our robot 1 should have been assigned to their robot 2 instead of 3.

jgkamat:
I'll take a look at that, hopefully I can figure that out :/

jgkamat:
After a quick look at this, I can't figure out exactly what's going on. The role_requirement logic seems to be fine, I'm probably going to have to step through with a debugger and see exactly what's tipping it, so this might take a while :cry:.

The related code is here: https://github.com/RoboJackets/robocup-software/pull/981/files#diff-1948f7515d94fea3db557c4697170b55R93

jgkamat:
So there dosen't seem to actually be a bug here, this is more of a weird consequence of our role assignment system (I think). Our role assignment tries to minimize overall cost (I think), and this solution actually ends up being lower cost, since the actual total distance that the robots travel overall is slightly less in this scenario (assuming straight lines). This is an edge case though, and shouldn't affect us in most senarios (when there is a clear advantage, that role assigment is taken). This could be solved by computing the distance from our robot->blocking robot as the cost instead of the distance from our robot to the marking point, but I feel that would introduce more problems...

I'm going to merge this now since I've spent a bunch of time testing this and I think this is good. If anyone sees any issues I'll follow up with patches! :)

PR #987, opened 2017-06-19
Update ball placement to work with new rules

### LINKED ISSUE #676: Tune ball placement behaviors
Ball placement works fairly well in the simulator, but will need some tuning in order to work with actual robots.  It could also use a little tuning to be more effective in the simulator.  See PR https://github.com/RoboJackets/robocup-software/pull/651

### LINKED ISSUE #985: Verify/Implement ball placement both when placing and when standing by
Our current stopped play will conflict with this rule:

capture

We should probably:

1) go into line up when a ball placment is issued to the opposing team
2) Implement #676 

@chachmu do you think you could take this?

chachmu:
- [x] Move out of the way of ball placement for other team
  - [x]  Move out of the way of own ball placement?
- [ ] Fix coordinate system to match SSL standard
- [ ] Increase speed of placement
- [x] Fix issue where robot goes to random point outside of circle for a second after placing, sometimes running into the ball

Closes #985 and Closes #676.


soccer/gameplay/plays/restarts/placement.py
```diff
@@ -69,9 +69,11 @@ def execute_placing(self):
                                         "Avoid")
 
     def on_exit_placing(self):
-        if self._our_restart:
+        if self.has_subbehavior_with_name('placer'):
```
jgkamat: This can be `self.remove_all_subbehaviors()` for simplicity.

chachmu: Whoops, forgot that existed, :+1:

soccer/gameplay/tactics/our_placement.py
```diff
@@ -59,11 +59,20 @@ def execute_dribble(self):
...
-        self.remove_subbehavior('dribble')
+        self.remove_all_subbehaviors()
+        print("EXIT DRIBBLE")
 
     def on_enter_pause(self):
+        self.robot.is_ball_placer = True
         self.pause_time = time.time()
 
+        #without this the robot tries to leave the circle in a random direction
+        self.robot.set_max_speed(0.01)
```
jgkamat: This sounds like an issue with the ball obstacle? If so, we should proably remove the obstacle and set the max speed low.

soccer/gameplay/tactics/our_placement.py
```diff
@@ -59,11 +59,20 @@ def execute_dribble(self):
...
+
+
+    def on_exit_pause(self):
+        self.robot.set_max_speed(5)
```
jgkamat: Maybe these max_speed values should go in their own const class somewhere? I think these would be nice to reuse for future plays.

joshhting:
What's the status of this? It'd be great if we could have this done by comp

jgkamat:
Yah, this is pretty critical. I'm not 100% sure how to send placement commands, but I might be able to take it over to push it through.

joshhting:
@chachmu If you don't respond by tonight I think we should just give it to @jgkamat, so you can help with motion control instead. We do need this done by comp

soccer/gameplay/plays/restarts/placement.py
```diff
@@ -59,8 +59,8 @@ def on_enter_placing(self):
                                  required=True,
                                  priority=90)
 
-        #line_up = tactics.line_up.LineUp(self.create_lineup())
-        #self.add_subbehavior(line_up, 'line_up', required=True, priority=80)
+        line_up = tactics.line_up.LineUp(self.create_lineup())
+        self.add_subbehavior(line_up, 'line_up', required=True, priority=80)
```
joshhting: Is this necessary? This puts us out of position to resume play once ball placement is finished.

jgkamat: We need to think carefully about this. By the rules, we need to stay 'out of the way' of the ball placement, so this is probably safer. I would assume we would have a bit of free 'stopped' time to get ready, so I don't think this would be a problem. This is definitely something that needs to be looked into closer though, thanks for bringing it up! 👍

joshhting: Also what if the referee orders the ball to be placed where we go to line up? LineUp is even less helpful in that instance.

jgkamat: ^ Yup agreed. We should either make a 'clearlineup' which lines up far away from the ball/target position, or make something more fancy. This is definitely a case we need to catch though, since placement will happen on the sidelines a lot. Thanks! 😄

jgkamat: Yup, this is a really important point, since we will be placing balls near the edges of the field usually, so this will become a problem. Maybe we can make a more dynamic line up that avoids the ball->target. This definitely needs to be considered though, thanks! :)

chachmu:
Sorry about that, I'll be in every night starting tonight so I can definitely finish this up.

soccer/gameplay/plays/restarts/placement.py
chachmu: Currently this checks which side of the field the placement is on and lines up on the other side.

soccer/gameplay/tactics/our_placement.py
chachmu: I'm not even sure if they will be necessary any more, they were mostly to deal with our issues with motion control last year.

soccer/gameplay/plays/restarts/placement.py
joshhting: From how I am interpreting the rules, it seems like we don't get any free time to reposition after ball placement. Is this your guys' interpretation as well?

jgkamat: The manual says

>When the ball goes out of play, robots should remain 500mm from the ball as the ball is placed until the restart signal is given by the referee.

I would assume the 'restart' signal would be the signal that is given during normal restarts (ie: when the ref places the ball themselves). That means we would have a little bit of time since the opposing team needs to get set up for a kick, but we won't have more than 5s or so, tops. 

This is a very new rule, so I would hope the refs would be lenient and give teams time to adjust.

I would reccomend placing the robots in a general defensive line (in relation to the target of the move command), perhaps, so they are close to their intended target places when resuming.

chachmu: We could just make a large line shaped obstacle from the ball to the placement position. This would cause our robots to move out of the way of any reasonable method of placing the ball while still getting as close to where we want them as they could.

jgkamat:
@chachmu any update on this? We gotta have this for comp. Probably we'll merge it during comp since it's getting pretty late.

chachmu:
@jgkamat Yeah, put it on the comp branch and I will finish up the tuning the parameters when we have robots on the field in Japan.

jgkamat:
Landing onto comp now.

PR #990, opened 2017-07-01
Fix coordinated pass becoming stuck in preparing

jgkamat:
Fixes a bug where coordinated_pass would get stuck in preparing.

Also does a bunch of style fixes which shouldn't affect anything :crossed_fingers:.

A review would be great :)

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -10,4 +10,6 @@ ignored-classes =
 [MESSAGES CONTROL]
 disable=locally-disabled,
         unused-import,
-        fixme
+        fixme,
+        missing-docstring,
+        C,R,W
--- a/makefile
+++ b/makefile
@@ -80,7 +80,7 @@ test-soccer:
 test-python: all
 	cd soccer/gameplay && ./run_tests.sh
 pylint:
-	pylint -j8 --disable=C,R,W --reports=n soccer/gameplay
+	pylint -j8 --reports=n soccer/gameplay
 mypy:
 	mypy soccer/gameplay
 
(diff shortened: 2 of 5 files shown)


jgkamat:
@chachmu @joshhting if either of you could give this a quick look over I would be :heart:.

Shouldn't be too big of a change though, just want a sanity check.

jgkamat:
bump? If I don't get a review I'm going to merge this tomorrow.

joshhting:
Style fixes didn't break anything, and the problem seems to be solved. Though sometimes when coordinated_pass is in the receiving state it occasionally takes a long pause before moving to the next state. Is having this sort of delay acceptable?

jgkamat:
@joshhting hmm, I can't seem to reproduce that on either the new or the old simulator. From a gut feeling though, that feels like a threshold problem (the kicker is trying to align too much, and it gets stuck). That probably would be solved though config changes. If you could get me a snapshot of the play call stack and a screenshot of the field, I might be able to take a look at that though (although I feel that should be another hi pri issue). I don't think this PR would affect that (but I could be wrong).

joshhting:
coordinated_pass_pause
So the robots get stuck in this state for a couple seconds before continuing, and it actually happens on master as well so it's not caused by these changes. It'd probably be better to make another issue for it

jgkamat:
That seems like tuning needed in capture. the new simulator made capture a little weird, but I hope that fixing it in the sim will make it better at comp too (since the new simulator is more realistic). We should make a new issue for that, and I'll try and take a look at it.

Issue #74, opened 2014-07-12
Improve OurRobot.hasBall()


justbuchanan:
Right now it just returns true if the ball sensor sees the ball.  It'd be nice to have another method that tracks the last time the ball was sensed, so noisiness in the ball sensor value doesn't throw us off.  The Aim behavior currently does this, but it'd be nice to have something more global.

PR #992, opened 2017-07-13
OurRobot::hasBall returns true more consistently

joshhting:
Addresses #74. I set the duration to 0.1 seconds because that seemed to work well, not sure if we'd prefer to shorten the duration or not.


soccer/Robot.hpp
```diff
@@ -530,6 +532,9 @@ class OurRobot : public Robot {
     friend class MotionControl;
 
 private:
+    mutable RJ::Time _lastBallSense;
+    const long _lostBallDuration = 200000000;
```
jgkamat: I feel like there's a way to express this in seconds, although I don't know how the time library works too well. 

Try RJ::seconds(0.2)? I'll try to figure it out if it's not clear....

soccer/Robot.cpp
```diff
@@ -235,6 +236,10 @@ const Geometry2d::Segment OurRobot::kickerBar() const {
     return Segment(pose * L, pose * R);
 }
 
+Geometry2d::Point OurRobot::mouthPos() const {
```
jgkamat: Just making sure, this is just a utility method right (you're not using it anywhere here). If so, can you add it to the python bridge too (that'd be useful).

soccer/Robot.hpp
jgkamat: Also, 0.2 seconds seems a little high for a first pass. Could we make this <=0.1 seconds and increase it if it helps?

I would also love for this to be tested on real robots if you get the chance, the ball dynamics are very different, especially up close.

joshhting:
Yeah this is still in status / developing because there is a known bug with this implementation that I'm trying to figure out

soccer/Robot.cpp
joshhting: Yeah I added it because I thought I was going to use it but ended up not, but figured it'd be good to leave in anyways

soccer/Robot.cpp
```diff
@@ -473,8 +478,14 @@ bool OurRobot::charged() const {
 }
 
 bool OurRobot::hasBall() const {
-    return _radioRx.has_ball_sense_status() &&
-           _radioRx.ball_sense_status() == Packet::HasBall && rxIsFresh();
+    if (_radioRx.has_ball_sense_status() &&
+           _radioRx.ball_sense_status() == Packet::HasBall && rxIsFresh()) {
+        OurRobot::_lastBallSense = RJ::now();
+        return true;
+    } else if ((RJ::now() - OurRobot::_lastBallSense).count() < OurRobot::_lostBallDuration) {
```
ashaw596: Should be RJ::now() - OurRobot::_lastBallSense <OurRobot::_lostBallDuration

soccer/Robot.hpp
ashaw596: Yeah, it would be best for _lostBallDuration  to be RJ::Seconds.

soccer/Robot.hpp
```diff
@@ -530,6 +532,9 @@ class OurRobot : public Robot {
     friend class MotionControl;
 
 private:
+    mutable RJ::Time _lastBallSense;
```
ashaw596: Is there a reason this has a mutable type modifier?

joshhting: _hasBall is const, but modifies this variable to know if it has recently seen the ball.

jgkamat: Hmm, it might be better just to make hasBall non-const now. I'm not sure if this or the const behavior is better style though......

joshhting: I decided on mutable instead of removing const because hasBall intuitively seems like a const function. But yeah I'm unsure of what's better coding style

soccer/Robot.cpp
joshhting: How to add something to python bridge? I'm not sure how the bridge works

jgkamat: It's a ton of boost python voodo magic in robocup-py. Here's the place you would add it (follow the general structure for adding methods there).

https://github.com/RoboJackets/robocup-software/blob/master/soccer/gameplay/robocup-py.cpp#L745-L781

If it dosen't work, don't sweat it, and I'll add it later (probably after comp though), just remind me :smile:

soccer/Robot.hpp
ashaw596: Oh, I see now. Um, it would be best to move the place where you update it to setRadioRx or somewhere in processor. hasBall really shouldn't be relied on to be called in order to set it and it should really be based on when we receive the information instead of when we call hasBall which I don't think is guarrenteed 100% to be called every frame

ashaw596:
Hmm. Ideally this shouldn't really be handled here cause some code needs to know if we have the ball this moment. We've been handling it in the plays right? Is this neccesary or could it be another function like bool recentlyHadBall(RJ::seconds duration);

jgkamat:
Maybe we could leave this as is, and make a 'smoothedHasBall' or similar. I don't like 'recentlyHasBall' because it implies it would return false when we actually have the ball. Either that, or we could make a 'rawHasBall' which returns raw sensor data for plays that must have it.

soccer/Robot.hpp
joshhting: Is there a good way to check for ball sense information without having to call hasball every frame?

soccer/gameplay/evaluation/ball.py
```diff
@@ -90,6 +90,8 @@ def angle_btw_three_pts(a, b, vertex):
         VA = math.sqrt((vertex.x - a.x)**2 + (vertex.y - a.y)**2)
         VB = math.sqrt((vertex.x - b.x)**2 + (vertex.y - b.y)**2)
         AB = math.sqrt((a.x - b.x)**2 + (a.y - b.y)**2)
+        if (VA * VA + VB * VB - AB * AB) / (2 * VA * VB) > 1:
```
jgkamat: I don't know too much about this function, but if `VA` or `VB` are 0 this results in a division by 0. Are we 100% sure these will never be 0? (the same behavior is there below, I just want to check and bring it up).

soccer/Robot.hpp
jgkamat: @joshhting this is pretty hacky (per before) since we can't rely on having hasBall being called reliably all the time. Maybe we could move it into `setRadioRx` (https://github.com/RoboJackets/robocup-software/blob/master/soccer/Robot.hpp#L401-L404), which should be called on every incoming packet (I think). It's even in the same class, so it should be easy to move it over.

soccer/gameplay/evaluation/ball.py
jgkamat: Yes this can be zero, but I'm not entirely sure when. I got:

```
ERROR:root:Play 'AdaptiveFormation' encountered an exception: <class 'ValueError'>.  Dropping and temp. blacklisting current play...
  File "/home/jay/Code/robocup-software/soccer/gameplay/composite_behavior.py", line 85, in spin
    bhvr.spin()
  File "/home/jay/Code/robocup-software/soccer/gameplay/composite_behavior.py", line 88, in spin
    self.handle_subbehavior_exception(name, exc)
  File "/home/jay/Code/robocup-software/soccer/gameplay/composite_behavior.py", line 85, in spin
    bhvr.spin()
  File "/home/jay/Code/robocup-software/soccer/gameplay/composite_behavior.py", line 67, in spin
    super().spin()
  File "/home/jay/Code/robocup-software/soccer/gameplay/fsm.py", line 66, in spin
    if transition['condition']():
  File "/home/jay/Code/robocup-software/soccer/gameplay/tactics/defensive_forward.py", line 64, in <lambda>
    evaluation.ball.robot_has_ball(self.collector.robot),
  File "/home/jay/Code/robocup-software/soccer/gameplay/evaluation/ball.py", line 97, in robot_has_ball
    math.cos(angle), math.sin(angle)), main.ball().pos, robot.pos)
  File "/home/jay/Code/robocup-software/soccer/gameplay/evaluation/ball.py", line 93, in angle_btw_three_pts
    return math.acos((VA * VA + VB * VB - AB * AB) / (2 * VA * VB))
ValueError: math domain error
```

while testing (not your change, but below). Since you're touching it, could you take a look?

joshhting: Actually I believe the ValueError is only when the expression results in a number greater than 1, not division by 0, which is what this line of code is meant to handle. It happens when the numerator and denominator are supposed to compute to equal values, but because of programming math the numerator is bigger than the denominator by a miniscule amount and passes in a number slightly larger than 1 to math.acos, which throws the ValueError.

joshhting: For reference, acos is only supposed to handle values in [0,1].

jgkamat: oh, my bad, I get it now. This LGTM :)

joshhting:
Would it make more sense to have this logic in `OurRobot::setRadioRx()` or the Processor.cpp loop that calls it?

jgkamat:
I personally think it's better if it's in `OurRobot::setRadioRx()`. Processor is kind of full...

joshhting:
Bump

ashaw596:
Hmm. I still don't really see the benefit of deleting the raw sensing data. The radioRx should represent the data that was received from the radio. Why not a seperate smooth has ball and raw has ball again? I saw you had that in an earlier version.

joshhting:
I guess the reasoning against smoothHasBall was that we would have no guarantees that smoothHasBall would be called every frame unless it was being called in the main processor loop, which I guess is an option we could do.

ashaw596:
You can update the _last_has_ball time variable in the Precessor loop when the radio Data is imported. The SmoothHasBall and RawHasBall should be just read functions(this would also let you get rid of the mutable on that variable).

joshhting:
You'd prefer that to be set outside of `setRadioRx` rather than inside the function?

ashaw596:
I think setRadioRx should just get the raw info from the radio as its name implies. Since you're grabbing that info there you can keep setting _lastBallSense there, but you shouldn't change the radio packet. Then you add the if statement in getBall(), and change the current getBall to getBallRaw.

soccer/Robot.cpp
```diff
@@ -67,6 +67,7 @@ OurRobot::OurRobot(int shell, SystemState* state)
     //_lastChargedTime = 0;
     _lastKickerStatus = 0;
     //_lastKickTime = 0;
+    _lastBallSense = RJ::Time();
```
ashaw596: Change this to zero in case it causes some random bug when all robots sense the ball at the start.

soccer/gameplay/evaluation/ball.py
ashaw596: Could this be changed to use point.angle_between? This is confusing.

ashaw596: It returns the angle between two directional vectors.

ashaw596:
Also, could you export has_ball_raw to robocup-py

soccer/Robot.cpp
joshhting: RJ::Time() is 0, it is not the same as RJ::now()

joshhting:
This has been tested on real robots with no issues caused by this branch. Is this good to merge?

soccer/Robot.hpp
```diff
@@ -386,6 +386,8 @@ class OurRobot : public Robot {
 
     void avoidOpponentRadius(unsigned shell_id, float radius);
 
+    Geometry2d::Point mouthPos() const;
```
jgkamat: I don't like the name `mouthPos`, since it's the center of the mouth. Perhaps `mouthCenterPos` or similar? I think a better name can be thought up though...

joshhting:
@ashaw596 I'll merge this if I don't get a review from you by Sunday's meeting

PR #993, opened 2017-07-17
Fix defense crowding when a robot is clearing

jgkamat:
When a robot is clearing the ball, they would no longer be counted in the threat detection algorithm, which would cause two robots to 'defend' the ball, which leads to no robots actually managing to do it (both in each other's way).

This is a little bit of a hack, all I do is filter away threats close to the ball when clearing, if someone else had a more elegant idea, please let me know.

Looking for a review, if anyone has time, since this is pretty important.


ashaw596:
The change lgtm, but I'm a bit confused about the algorithm. Are our own robots being considered as "threats" ? Or are we just removing the threats from the ball so the defenders stop getting to close to it when we're clearing it.

jgkamat:
I'm not to familiar with this logic myself, but anything that shows up in soccer with a red triangle to the goal should be a 'threat'. The ball is always a threat. Previously, when a defensive robot is clearing the ball, it would stop being 'assigned' to that threat, and the second defender would run over to mark the ball, interfering with the progress. Now, I simply filter away the ball and any defensive robots near it to prevent the 'ball' from being though of as a threat when we're clearing. I'm trying to think of a better way to represent this though, but I can't think of one. This will probably get the job done though :P.

joshhting:
Is it possible to have the robot clearing the ball be assigned to it while still carrying out the clearing action?

jgkamat:
I could try to implement it that way, but it would be a bigger change. It's way less hacky and less likely to break though, so I'll look at it.

jgkamat:
After staring at this code for way too long, I've found a way cleaner way to do this, which will be incoming soon. 

This code is a little too complicated :smile:

jgkamat:
This does a little bit of tweaking, and I wanted to ask for feedback if anyone thinks this is a bad idea.

This change only affects our behavior if we have exactly 2 robots on defense.

If we have 2 robots on defense, previously, the goalie would defend the (1) threat, while the defender would defend the (2) threat. Now, in that case, both the defender and the goalie will defend the (1) threat. I'm not sure if this is a good or a bad idea (good: the highest priority threat gets more robots, bad: a quick pass can probably destroy us).

Previously, 'defender1' was getting assigned to the (2) highest task, while being overridden to clear the ball. Now, 'defender1' always gets the highest priority, so it will be properly assigned to clear the ball and be tasked with the highest priority threat.

joshhting:
Is the ball always ranked as the highest threat? Also, I don't believe that assigning both defenders to the biggest threat is better than assigning them to different threats. Adding additional robots to the same threat yields diminishing returns on protection compared to assigning each of them an entire threat to guard.

jgkamat:
The first threat is always the ball, or a robot near it, unless the ball is moving towards an enemy robot. If it is, the enemy robot it's targeting is the main threat.

The reason that I think this could be a good idea, is that the 'primary' defender is the goalie, which isn't in a great position to block shots. If a defender also helped out (which happens normally) we would get a lot better coverage from direct shots.

I think I can actually come up with a method that shouldn't change the existing behavior though, except during clearing (while clearing, both the clearer and the goalie would be assigned to the ball).

jgkamat:
I think this version is the cleanest. It simply assigns the clearing robot the ball and removes it from the threat handlers. 

The only improvement I can think of now is removing the ball threat altogether (so if we have hypothetically 4 defenders, they would not crowd the ball), but I want to keep this pretty small :)

joshhting:
In this situation, the bots don't really move much for about 5-15 seconds, I think it's flickering between having a robot assigned to clearing and not having a robot assigned to clearing defpause

jgkamat:
Nice catch, I think that's the capture subbehavior being added and removed repeately (weird...). I'll check that out.

jgkamat:
Aha, it's role assignments, the two robots have the same cost, so it flickers between them. This is out of scope for this PR, but this should be a high priority issue, since this could happen in any child issue. I'll try and take a look at that tomorrow.

jgkamat:
This is the problem: https://github.com/RoboJackets/robocup-software/blob/master/soccer/gameplay/tactics/defense.py#L510. I'm trying to come up with a better solution now.

jgkamat:
Can you test this again @joshhting? I think it should be better now, but its now a tuning problem, so it could be off.

joshhting:
That issue is fixed, but now the same issue is happening to the Marking behavior 
markingpanic

joshhting:
It also flickers between "simulation" and "processing hung'

jgkamat:
The flickering in simulation->processing hung is probably a performance issue, try running with `make rrs` to compile with optimizations on (which is a LOT faster).

I can (rarely) reproduce the new flickering, I'll bump the robot change cost a little bit more (it's very small right now). Can you try again once I push a new commit (maybe 5 min)?

joshhting:
Still happens. At least it's able to recover after a second though screenshot from 2017-07-22 14-22-23

jgkamat:
Hmm, I tried for a couple of minutes and I still can't reproduce it. Can you try raising the value a little bit more (until you no longer see it) and add a commit fixing that? I don't want to raise it too high since the defense should be a little more dynamic than normal robots (which is why it was removed in the first place).

jgkamat:
@joshhting are you good on merging this? I'm :+1:, it looks good to me.

joshhting:
Give me like 10 minutes to do some double checking

jgkamat:
Ignoring style for now, the changes make it look worse. We have to look at that later....

PR #996, opened 2017-07-22
Add a default playbook file

### LINKED ISSUE #967: Add one playbook to gitignore
A lot of testing involves running one play repeatedly, and often requires restarting every time. It is quicker to set up a playbook with that play selected and designate as the playbook to use. However, it is annoying to deal with the git part of changing a file you don't want to commit. Could we create a playbook like testing_playbook that we add the the ignore?

jgkamat:
Closes #967

This is a little different from what we discussed previously, it's a file that's added (that I'm trusting people not to commit) instead of just an addition to gitignore. I think this will be more helpful because it includes usage instructions.

Let me know if you think it should be implmeented differently, I'm going to merge this soon since it's a tinychange and I want it in :smile:

### DIFF
--- a/makefile
+++ b/makefile
@@ -52,7 +52,7 @@ rr: run-release
 backend-simulator-soccer:
 	-pkill -f './grsim'
 	./run/grsim &
-	./run/soccer -sim
+	./run/soccer -sim -pbk testing.pbk
 # Kill grSim once we unblock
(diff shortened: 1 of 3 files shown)


jgkamat:
Merging now, please let me know if you have any feedback to here or to #967

Issue #956, opened 2017-04-03
Weird dips observed in observed robotVelocity


jgkamat:
When plotting the robot velocity, I see these weird dips occuring. this is via the simulator, but this is much more pronounced on the field computer. This is probably messing with our motion control, so we should figure out why this is happening. It seems to happen fairly regular intervals, and happens even when we control the robot manualy (so I doubt it's path planning lag or something else). 

On the field computer it jumps down to 0 (maybe not going fast enough?), but on my computer in the simulator it's visible, but much less exaggerated:

![2017-04-02-222325_1920x1080_scrot](https://cloud.githubusercontent.com/assets/4349709/24593596/0fd0052c-17f3-11e7-9f03-cfb403f59035.png)

jgkamat:
This seems to be related to a situation where we receive multiple vel packets in one processor frame. In this case, we do some weird stuff in RobotFilter.cpp. That should be a good place to track this down.

Ideally, we would handle it so that we have a smooth graph with little to no dips. The dips might be caused by the apparent 'lack of time' between the bunch of received packets.

jgkamat:
Optimizing the path planner will also fix this (as will compiling with optimizations), but this still should be fixed since this situation can be caused by a bunch of cases.

JNeiger:
It might be [this](https://github.com/RoboJackets/robocup-software/blob/master/soccer/modeling/RobotFilter.cpp#L25). If `obs->time` isn't being sent right, the second prediction will be too close and be cut off.

[Here](https://github.com/RoboJackets/robocup-software/blob/cea1fc10eef0261e6d933350e391b301280c5f10/soccer/Processor.cpp#L170) is the caller of update.

jgkamat:
Yes, albert said the same thing earlier.

It looks like (currently) we're setting the velocity to 0 when packets come in too fast (which seems very wrong). I'm not sure what the best thing to do would be though, I think we should take in all the packets, set the average velocity for the time we dropped out for, and keep going, but I'm not sure if that makes sense. We need to try out different things and see what works best.

jgkamat:
From what I can tell so far, this is caused by us getting two packets with the same exact position one after each other, causing velocity to drop to zero for a frame. The velocity_alpha is the reason we don't see a full drop all the time. 

The initial Min_Frame_Time thing seems to only run (for me) when the robots initialize, so I don't think thats the problem (on my machine) at this point. 

[...]

PR #997, opened 2017-07-22
Remove dips in observed robot velocity

joshhting:
Addresses #956. So I'm unsure what the purpose of this line of code that I changed is, but getting rid of it dramatically improves the smoothness of the observed velocity. I'd like to know what the purpose of what I deleted is.

### DIFF
--- a/soccer/modeling/RobotFilter.cpp
+++ b/soccer/modeling/RobotFilter.cpp
@@ -22,7 +22,7 @@ void RobotFilter::update(const RobotObservation* obs) {
     // bool reset = _currentEstimate[s].time == 0 || (dtime > Coast_Time);
 
     bool reset = (dtime > Coast_Time);
-    if (reset || dtime < Min_Frame_Time) {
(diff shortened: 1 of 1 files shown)


jgkamat:
Huhhhh, I definetly saw that before, but I have no idea why it was added. Maybe it's trying to reset the speed if we get double packets (two packets with the same ID?). I don't know why we would be resetting it if the time < a value (which is what it's doing right now). 

I definetly see that it's getting better with this though, I want to take a closer look at this though, give me a little bit.

jgkamat:
dTime is the reported difference in time between frames that are sent to us from ssl-vision (I think). Maybe the rate used to be lower, and it was increased at some point, and we never realized. Maybe this is to prevent us from handling duplicate frames somehow? 

I think @ashaw596 had an idea before what this was for, do you remember what it was? 

I don't see a problem with using it, but I'm hesitant, since this is pretty important :smile:

PR #1002, opened 2017-07-29
Fix yellow robot status in sim

jgkamat:
Authored by @jpfeltracco

### DIFF
--- a/soccer/radio/SimRadio.cpp
+++ b/soccer/radio/SimRadio.cpp
@@ -124,7 +124,35 @@ void SimRadio::receive() {
     }
 }
 
+void SimRadio::stopRobots() {
+    grSim_Packet simPacket;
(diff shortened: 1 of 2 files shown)


soccer/radio/SimRadio.cpp
```diff
@@ -134,7 +162,6 @@ void SimRadio::switchTeam(bool blueTeam) {
                                 .arg(blueTeam ? "blue" : "yellow")
                                 .toStdString());
     }
-    std::cout << "Bound to " << SimBlueStatusPort << std::endl;
```
jgkamat: Any particular reason you removed this?

jpfeltracco: It was inconsistent with the below block that didn't have a printout and the value it was printing out should have been "RadioRxPort + _channel" instead of just the SimBlueStatusPort. I figured we didn't really need the printouts anyway.

PR #1005, opened 2017-08-23
Chachmu/multi joystick

### LINKED ISSUE #1004: Multiple Robot Joystick Control
In order to help generate interest in robocup we want to try making an option to have people play matches together in

chachmu:
Add support to manual control multiple robots simultaneously.

- [ ] Connect Multiple Joysticks
- [ ] Send Manual Control to Multiple Robots

Closes #1004.


chachmu:
@jgkamat It looks to me like GamepadJoystick has more or less been replaced by GamepadController. Is this correct or am I just confused?

jgkamat:
@chachmu GamepadController is the new joystick, I left the GamepadJoystick around just in case the new controller broke anything. You can find it added [here](https://github.com/RoboJackets/robocup-software/blob/e009f4211815e9421b06e1a1d81ab9914803dd15/soccer/Processor.cpp#L70-L71). I think the gamepadjoystick class can be removed at this point.

chachmu:
Theoretically this will work although I highly doubt I did this right straight off the bat. I can't test until I have more than one controller so I might try to borrow a few while we're at core or something.

chachmu:
@jgkamat @joshhting Afaik this is ready to merge. Thoughts?

soccer/Processor.cpp
```diff
@@ -713,24 +748,57 @@ void Processor::sendRadioData() {
     }
 
     // Add RadioTx commands for visible robots and apply joystick input
-    for (OurRobot* r : _state.self) {
-        if (r->visible || _manualID == r->shell()) {
+
+    if (!_multipleManual) {
```
joshhting: Same thing as before. Can we generalize the if case and else case into a single generalized case?

chachmu:
@joshhting give this a shot. It worked as best I could tell when I tested with one joystick on a vm but it may have some issues with multiple joysticks or the 3d mouse. I can test those cases more extensively next meeting.

soccer/MainWindow.cpp
```diff
@@ -1461,3 +1512,43 @@ void MainWindow::on_actionVisionFull_Field_triggered() {
 }
 
 bool MainWindow::live() { return !_playbackRate; }
+
+void MainWindow::on_robotConfigButton_clicked() {
+    std::vector<std::pair<DebugCommunication::ConfigCommunication, float>>
+        configs;
+    // Loop through debug settings and read thier value
```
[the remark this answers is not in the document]
joshhting: Actually don't change this, we'll make it an intro issue

joshhting:
So I'm getting this error when running on grSim:
`ODE INTERNAL ERROR 1: assertion "bNormalizationResult" failed in _dNormalize4() [odemath.h:42]`
which causes grSim to crash. Unfortunately I'm not able to run it on real robots right now because Ri1W has the field

joshhting:
Also, didn't we decide that the joystick tab would show debug information for the robot currently selected under the manual drop down? Right now the joystick tab is just showing the first robot's debug info.

jgkamat:
The `ODE INTERNAL ERROR 1: assertion "bNormalizationResult" failed in _dNormalize4() [odemath.h:42]` seems like an ode error, which we can't fix without significant effort (and might be fixed in later versions already). I'm not sure why this PR would be triggering a bug in ODE though, that seems very weird. I can reproduce it on my machine as well (libode-0.14).

chachmu:
Before I made the most recent set of changes that wasn't happening so I have a few theories as to what it is. I will take a look and try to get this finalized during the Sunday meeting.

chachmu:
If we end up using this it should probably be as a branch for right now because this breaks single joystick hotswap for normal manual mode. Also, I just discovered a weird issue where changing the manual selector before activating manual mode makes multiple robots get assigned to controller 0.

chachmu:
I think this fixes everything except for an issue where soccer segfaults when you unplug the last joystick. I left a bunch of debug statements in in case there is anything else wrong with it but other than those 2 things this should be pretty similar to the final version

soccer/Processor.cpp
```diff
@@ -733,21 +745,53 @@ void Processor::sendRadioData() {
         }
     }
 
+    //------------------------------------------------------------------------------------------------------------------------------
+    // I BROKE STUFF
```
jgkamat: x

soccer/Processor.cpp
```diff
@@ -733,21 +745,53 @@ void Processor::sendRadioData() {
...
+    // Segfault after plug and unplug one
+
     // Add RadioTx commands for visible robots and apply joystick input
+    std::vector<int> manualIds = getJoystickRobotIds();  // BOOKMARK
```
jgkamat: Remove this comment please :)

soccer/joystick/GamepadJoystick.hpp
```diff
@@ -8,6 +8,8 @@
  * @brief Logitecch Gamepad/Joystick used to control robots
  *
  * Used to control the Logitech F310 and XBOX controllers
+ *
+ * Appears to have been mostly replaced by GamepadController
```
jgkamat: You should be more assertive here, something like "this has been replaced by gamepad controller"

Issue #923, opened 2017-03-16
Add lambda cost property to role requirements


gtrosawa:
# TODO arbitrary cost lambda property

jgkamat:
This is actually a good issue, despite the lack of documentation. 

Essentially, we would want to have plays be able to pass in a lambda expression to define costs, so that they can tweak the cost function however they like! This is a nice issue for people interested in learning more about role reqs.

chachmu:
I thought John was already working on that

jgkamat:
I'm pretty sure john was working on one of these things:

1. Behavior trees
2. stopped plays schenanigans
3. Lambda expression to prevent restarting of state machine when robot changes.

Although if he is, we should close this as a dupe of the issue he is working on. @jcarn

jcarn:
I kind of want this, if @mmurley3 is done with the chart soon I'd like him to work on this. Or anyone really, or I could do it

chachmu:
I think the stripchart issue can be finished up this week or next.

jgkamat:
moved to https://github.com/RoboJackets/robocup-software/issues/1018#issuecomment-328639620

Issue #1018, opened 2017-09-08
Lambda functions for subbehavior priority


JNeiger:
This will replace the static int arguments currently being used.

jgkamat:
To keep backwards compatibility with all the other plays, let's accept both integers or functions for this argument. It shouldn't be too hard, just a simple case statement.

JNeiger:
This issue is a duplicate of #923

PR #1022, opened 2017-09-24
Added cost function parameter to role_assignment

mmurley3:
addresses #923

### DIFF
--- a/soccer/gameplay/role_assignment.py
+++ b/soccer/gameplay/role_assignment.py
@@ -22,6 +22,10 @@ class RoleRequirements:
         # multiply this by the distance between two points to get the cost
         self.position_cost_multiplier = 1.0
 
+        # A robot assigned to the cost function
+        # If not defined, cost function applies to all robots
+        self.cost_func_shell_id = None
+        
         # A lambda function property that allows customization of cost
         self.cost_func = None
 
@@ -149,6 +153,14 @@ class RoleRequirements:
     def cost_func(self, value):
         self._cost_func = value
 
+    @property
+    def cost_func_shell_id(self):
+        return self._cost_func_shell_id
+
+    @cost_func_shell_id.setter
+    def cost_func_shell_id(self, value):
+        self._cost_func_shell_id = value
+
 
 # given a role requirements tree (with RoleRequirements or assignment tuples as leaves),
 # yields all of the RoleRequiements objects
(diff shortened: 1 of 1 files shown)


mmurley3:
Not sure if this overlaps with #1019 and/or #1018

JNeiger:
I think having both would be a good idea. #1018 deals with how many robots a play should be assigned while this cost function deals with which specific robot is assigned to that role.

On that note, we may want to pass in the specific robot the cost is being calculated for into the lambda function

mmurley3:
Not sure if this is exactly what you were talking about Joe, but give it a glance. A bit late, I know.

JNeiger:
That's actually pretty close considering how vague I was. 

I was thinking that for the cost function itself, having the specific robot the cost is being evaluated for, may be useful.

Eg:
```
if req.cost_func != none:
     cost += req.cost_func(robot)
```

One example of a use case for this is that when we want to find a receiver for a pass, we may want to try and get a receiver on our side of the ball so it is easier for the them pass or shoot directly afterwards. If we pass in the robot as an argument, it becomes a pretty simple problem to solve.

On top of that, the cost function will still continue to have the same functionality as described in the original issue since you can just ignore that argument.

Thoughts?

mmurley3:
Ah, I see what you mean.  I'll get it up in a bit.

mmurley3:
Ok, it should be good now.  I tested a few cases and it should work as long as people declare the robot parameter.  Let me know if there's anything that I missed.

soccer/gameplay/role_assignment.py
```diff
@@ -254,6 +265,8 @@ def fail(errStr):
                     cost += req.robot_change_cost
                 if not robot.has_chipper():
                     cost += req.chipper_preference_weight
+                if req.cost_func != None:
```
jgkamat: `const_func is not None`

soccer/gameplay/role_assignment.py
```diff
@@ -21,6 +21,10 @@ def __init__(self):
 
         # multiply this by the distance between two points to get the cost
         self.position_cost_multiplier = 1.0
+        
+        # A lambda function property that allows customization of cost
+        # Has exactly one parameter, which is a robot
+        self.cost_func = None
```
jcarn: What if we did self.cost_func = lambda r: 0

soccer/gameplay/role_assignment.py
jcarn: Then we don't need this check

soccer/gameplay/role_assignment.py
jcarn: This also gives a bit of an idea about what a function should look like, good for readability

PR #1025, opened 2017-09-24
Make a few tweaks to week 2 traning examples

jgkamat:
don't tell any of the new members `__init__.py` was renamed :wink:


jgkamat:
Since this is only comments, merging :)

PR #1026, opened 2017-09-26
Add ubuntu 17.04 support

EpicPi:
Clang packages need to not be specified - and then it works!


jgkamat:
Unfortunately, we need the `-3.6` at the end to keep ubuntu 14.04 support. 

If we confirm no one is currently needing 14.04 support, I'm fine with dropping it though, since it was a bit of a hack to begin with. See https://github.com/RoboJackets/robocup-software/issues/894.

If we still need it, I can add a bit of logic to add the `-3.6` only on 14.04.

As for running this now, if you run ubuntu-setup with your changes, you should be able to run everything just fine (but you'll need to make sure you have this patch whenever you re-run ubuntu-setup, of course).

EpicPi:
Ah, I'll see if I can work the logic and maybe learn a bit of shell script along the way.
Thanks!

jgkamat:
Feel free to use this PR as a base, it looks good so far!

I think a good starting place would be to try to modify [this variable](https://github.com/RoboJackets/robocup-software/blob/master/util/ubuntu-setup#L144), perhaps with sed or awk, to add the -3.6 only if we're using 14.04. 

If you get stuck let me know, I'd much rather have you learn a bunch and solve the issue than fix it myself :smile:

ashaw596:
Wait. @jgkamat what are we using clang for? Don't we use g++ to compile?

jgkamat:
That's a good point, I think we're just using it for `clang_format`, in which case, we can probably just drop the dependency entirely for 14.04. I think it might be used on mac, but that's not related. I'll see if I can figure out why we added it in the first place (besides clang_format)...

PR #1030, opened 2017-10-01
Fixes Kick Eval calculating the wrong pass when target is across -PI +PI

### LINKED ISSUE #1029: Kick Evaluator calculates the wrong pass chance to another robot sometimes
I think this only happens when the target and source position Y value is close to the same and the X value of the target is less than the source.

Here's a picture of the output of the pass chance:
image

Code for that picture:
`pass_pt, pass_chance = self.kick_eval.eval_pt_to_robot(main.ball().pos, pos)`

JNeiger:
Fixes #1029 

Here's the new results:
image

### DIFF
--- a/soccer/KickEvaluator.cpp
+++ b/soccer/KickEvaluator.cpp
@@ -6,7 +6,6 @@
 #include <vector>
 #include <math.h>
 #include <cmath>
-#include <iostream>
 
 REGISTER_CONFIGURABLE(KickEvaluator)
 
--- a/soccer/KickEvaluator.cpp
+++ b/soccer/KickEvaluator.cpp
@@ -265,7 +265,7 @@ float KickEvaluator::get_target_angle(const Point origin,
     Point left = target.pt[0] - origin;
     Point right = target.pt[1] - origin;
 
-    float angle = left.dot(right) / (left.mag()*right.mag());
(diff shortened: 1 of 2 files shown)


soccer/KickEvaluator.cpp
```diff
@@ -114,10 +114,10 @@ KickResults KickEvaluator::eval_pt_to_seg(Point origin, Segment target) {
     // No opponent robots on the field
     if (botMeans.size() == 0) {
         // Push it off to the side
-        botMeans.push_back(3);
+        botMeans.push_back(4);
```
joshhting: What is this line for?

JNeiger: I didn't really need to change this, but the further away from 0, the better. We just can push it too far away otherwise the exp estimation will return a wrong value messing up the Gradient Ascent nodes.

To calculate the derivative of the kicking function, you need the smallest robot index or something. So if there are no opponent robots on the field, we need to just create a fake robot to place there. This just puts it off to the side further. So its 4 radians instead of 3 away from the center of the target. This also helps when we try and find the largest window. I find the largest difference between neighboring local maximum. So if the robot was in the middle of the target segment, it would require something a bit more complicated.

PR #1031, opened 2017-10-01
replaced line intersect test with dot product test

Jason27chan:
Previously, the ball needed to be explicitly moving in a straight line towards the goal for this to return true.
This greatly broadens our definition of moving towards the goal. As long as some component of the ball is moving towards the goal, this function will now return true.


soccer/gameplay/evaluation/ball.py
```diff
@@ -7,20 +7,11 @@
...
-                              0),
-                robocup.Point(-constants.Field.GoalWidth / 2.0 - fudge_factor,
-                              0))
-
-            pt = ball_path.segment_intersection(WiderGoalSegment)
-            return pt != None
+        
+        # Checks the dot product between the vector between the ball's position and the goal
+        # and the velocity vector
+        vector_to_goal = robocup.Point(0 - main.ball().pos.x, 0 - main.ball().pos.y)
```
joshhting: This is a nitpick, but could you remove the 0s and just pass in `-main.ball().pos.n for` instead

ashaw596:
Doesn't this return true if the ball has any speed in the x direction. Just wondering, what is the reason for this change? This method is suppose to return currently  if the ball is actually going at our goal. The goalie shouldn't hit out balls that out going out.

joshhting:
What's the problem? We're trying to simplify this function to do what the function's name implies.

jgkamat:
I think that we're changing this function's behavior without looking at the consequences fully. 

For example, I think the goalie uses this function to determine if we plan to block a goal. See:

out

In both of these shots, the ball isn't going towareds the goal, so we don't want to have the goalie block it. Especially in the second case, the goalie would fly out of the goal to catch the ball.

I think that renaming the function to be better representitive of the behavior would be a better idea (or a deeper analysis at where this function is used and how).

soccer/gameplay/evaluation/ball.py
jgkamat: I think (but I'm not sure) that `-robocup.Point(main.ball().pos.x, main.ball().pos.y)` might work too.

JNeiger:
The function is only being used in the goalie and defense. In both cases, to intercept an incoming shot. That functionality is still needed somewhere for sure.

A better descriptor of this might be something along the lines of `is_advancing_on_our_goal`

jgkamat:
>That functionality is still needed somewhere for sure.

I think you're talking about this PR, correct? If so, I agree :) 

I think that this function should be named similar to `is_advancing_on_our_goal` (as suggested) and this function named to `shot_on_goal` or similar (I'm terrible with names). It think that both of these functions are useful for different purposes (the existing one for shots, the new one for high level defensive strategy). 

Perhaps we could return a score from 0-1 as well? I don't know if that's a good idea or not though...

joshhting:
Or `ball_moving_at_goal`. `Regardless, `is_moving_towards_our_goal` is a misleading name for the original function.

JNeiger:
I think returning a score for the new version could be helpful. It could represent a sort of urgency when assigning defenders to different threats.

ashaw596:
I don't really see the point of the new function. What are we trying to use it for? To me, is_moving_towards_our_goal  means it is the ball is literally moving towards our goal. Which means it should hit it. A dotscore doesn't really mean much practically. If you doing the check if its positive, its basically, are you moving in a direction within 180 degrees of our goal. What practically are we trying to achieve. Cause if its trying to see how close the ball is going towards our goal. Something more explicit would be vel_line.dist_to(goal). Or vel_line.angle_between(goal - ball). I don't know what the use case is so maybe I just don't see the point.

joshhting:
Valid. We originally decided to open this PR because the function name seemed inconsistent with its definition here, so we figured we should change the definition to match the name. I think we should revert this change and rename the function to match its definition instead.

joshhting:
Note that there is a commit on master that is not on here that uses `is_moving_towards_our_goal` assuming it's the old version of the function

soccer/gameplay/tests/test_circle_near_ball.py
```diff
@@ -0,0 +1,67 @@
...
+		test_point(self, width * -1, 0)
+		test_point(self, 0, 0)
+		test_point(self, width, 0)
+		test_point(self, width, length / 2)
+		test_point(self, width - constants.Robot.Radius, constants.Robot.Radius)
+		test_point(self, width + constants.Robot.Radius, constants.Robot.Radius * -1)
+
+	def test_normalize_angle(self):
+		print("test_normalize_angle ran")
+		# angle = -math.pi
```
jgkamat: This comment needs to be removed or uncommented. If this dosen't work but should, you can skip the test using an annotation.

chachmu:
Is everyone ok with me closing this pull request so we can open a new pull request to rename the function something more descriptive?

PR #1032, opened 2017-10-02
Fix SDL include path and add std namespace prefix.

jpfeltracco:
These issues were preventing a build with more recent versions of boost and sdl in Arch.


jpfeltracco:
Merging since this is trivial and we're passing tests.

PR #1036, opened 2017-10-07
Enable C++ 17 (for ubuntu 16.04) 

ashaw596:
Enables C++ 17 by updating to gcc-7.
Adds back the ppa, but its a pretty official one, mostly kept up by Ubuntu dev.
Also changed all boost::optional to std::optional as a proof of concept test, but C++17 actually has a lot of actual cool stuff.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -14,14 +14,14 @@ include(SetupGTest)
 
 message("${TARGET}")
 # C++ version
-if ("${CMAKE_VERSION}" VERSION_LESS "3.1")
+if ("${CMAKE_VERSION}" VERSION_LESS "3.8")
     # This is to support the CMake version backported to Ubuntu 14.04
-    message("CMAKE_VERSION < 3.1, specifying c++14 flag.")
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++14")
+    message("CMAKE_VERSION < 3.8, specifying c++17 flag.")
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++1z")
 else()
     # This is the more modern way of setting our c++ version
-    message("CMAKE_VERSION >= 3.1, setting CMAKE_CXX_STANDARD to 14.")
-    set(CMAKE_CXX_STANDARD 14)
+    message("CMAKE_VERSION >= 3.8, setting CMAKE_CXX_STANDARD to 17.")
+    set(CMAKE_CXX_STANDARD 17)
     set(CMAKE_CXX_STANDARD_REQUIRED ON) # Don't fall back to older versions
 endif()
 add_subdirectory(external/grSim)
--- a/soccer/MainWindow.hpp
+++ b/soccer/MainWindow.hpp
@@ -52,7 +52,7 @@ public:
         if (!live()) {
             _ui.logTree->setStyleSheet(
                 QString("QTreeWidget{%1}").arg(NonLiveStyle));
-            _playbackRate = boost::none;
+            _playbackRate = std::nullopt;
         }
     }
 
@@ -223,7 +223,7 @@ private:
     QTreeWidgetItem* _elapsedTimeItem;
 
     /// playback rate of the viewer - a value of 1 means realtime
-    boost::optional<double> _playbackRate;
+    std::optional<double> _playbackRate;
 
     // This is used to update some status items less frequently than the full
     // field view
--- a/soccer/Processor.hpp
+++ b/soccer/Processor.hpp
@@ -201,7 +201,7 @@ public:
     ////////
 
     // Time of the first LogFrame
-    boost::optional<RJ::Time> firstLogTime;
+    std::optional<RJ::Time> firstLogTime;
 
 protected:
     void run() override;
--- a/soccer/Robot.cpp
+++ b/soccer/Robot.cpp
@@ -544,14 +544,14 @@ Packet::HardwareVersion OurRobot::hardwareVersion() const {
     }
 }
 
-boost::optional<Eigen::Quaternionf> OurRobot::quaternion() const {
+std::optional<Eigen::Quaternionf> OurRobot::quaternion() const {
     if (_radioRx.has_quaternion() && rxIsFresh(RJ::Seconds(0.05))) {
         return Eigen::Quaternionf(_radioRx.quaternion().q0() / 16384.0,
                                   _radioRx.quaternion().q1() / 16384.0,
                                   _radioRx.quaternion().q2() / 16384.0,
                                   _radioRx.quaternion().q3() / 16384.0);
     } else {
-        return boost::none;
+        return std::nullopt;
     }
 }
 
--- a/soccer/Robot.hpp
+++ b/soccer/Robot.hpp
@@ -176,7 +176,7 @@ public:
 
     // Gets the robot quaternion.  Returns false (and does not change q) if not
     // available.
-    boost::optional<Eigen::Quaternionf> quaternion() const;
+    std::optional<Eigen::Quaternionf> quaternion() const;
 
     // Constraints
     const RobotConstraints& robotConstraints() const {
--- a/soccer/SystemState.cpp
+++ b/soccer/SystemState.cpp
@@ -16,7 +16,7 @@ using Planning::MotionInstant;
 class BallPath : public Planning::Path {
 public:
     BallPath(const Ball& ball) : ball(ball){};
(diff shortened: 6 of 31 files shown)


jgkamat:
I'd really prefer we don't do this, this really really sucked the last time we pushed PPA's like this and lead to a bunch of broken installs. There's a reason software is frozen, and upgrading compilers like this can cause a lot of problems in the longer runs (which is what leads to broken installs). Not to mention we would like to support other distributions besides just ubuntu, and upgrading this aggressively is not playing nice with that. 

While c++17 has a lot of features, so does python 3.6, newer versions of boost, and basically every other peice of software, and with new features comes new bugs/problems. It's usually a good idea to let things settle down for a bit (until they get into stable distibutions) before using them in something as big as robocup.

ashaw596:
Hmm. C++17 is bascially finished. They're working on 20 specification now. And its not like a boost update. They argue about this stuff in committee for years and its backwards compatible. Also our version of compiler we are using is now 2 major version behind in gcc which isn't great. 17.1 has gcc 7 by default now. I don't really remember breaking many builds when we switched gcc. The main thing was when we switched to clang for awhile and then switched back... The switch to 14 was mostly painless and the same ppa is still used in 14.04 build. The main one we had problems with was when we were manully compiling gcc for the firmware. I mean this pull request isn't ready by any case, and it definitely needs to be tested on all the systems. But I think we should definitely keep our code base updated.

jgkamat:
>The switch to 14 was mostly painless and the same ppa is still used in 14.04 build.

This is definetly not the case, we had *many* broken builds due to this particular hack for a long time. I think his is the PPA that is currently breaking the 14.04 upgrades to 16.04, and is one of the main reasons I want to drop 14.04 support (so we don't need to maintain ppa hacks anymore). IMHO we shoud have been targeting c++11 at this point, and only now should have been thinking about rolling out c++14, since it's begun to make it's way into all the stable distributions (but still not entirely). Pushing c++17 is frankly rediculous until perhaps 2019, at earliest (at which point the compiler bugs and inconsitencies will have been found, resolved, and shipped, so we don't need to beta test them).

> But I think we should definitely keep our code base updated.

Why not add a PPA for gcc8 development, and begin using c++20 features then? (gcc7 is "out of date" right now!) There's a reason why stable distributions are designed the way they are, if we would like unstable features we should be using unstable distributions. We can't claim to "support 16.04" if we are installing PPA's over it. I don't want to compromise the stability (and test/debug) of all of my systems with a newly released compiler just for one project, there are much better things I can be doing. I doubt you'll find a single C/C++ project anyone *actually uses* that will switch to C++17 until next year at earliest (or no one will be able to compile it).

This also makes it incredibly hard to install this on anything but the system we hack, which we suffered a lot from last year (with new members having opensuse installed). I really, *really*, never want to deal with PPAs ever again.

At best, this is probably a change that can be made earliest next year, so we can possibly build new (non ubuntu) starter images with rolling distributions to get gcc7.

ashaw596:
I don't really care enough if no one wants to. Yes the main issue is incompatibility with other systems we don't support to begin with. So it's probs not worth it if more than a couple people don't use Ubuntu. 

But think straw man arguments a bit annoying... No of course were not supporting C++20. It's basically nonexistent right now. It's not like this is a crazy new hack and we've gotten rid of ppas. We still use this ppa for 14.04 support. And it's not like gcc 7 isn't stable... They've been developing 8 since may. I don't think we'll maintained ppas are the worst thing in the world. 

Exactly which bugs did upgrading gcc or that ppa actually cause. I took a look back and I think the bugs were caused because mostly cause cmake not being configured correctly. Gcc isn't exactly releasing non working or non backwards.
 compatibility stuff that often.

And the some of other ppa stuff we've added in the past have been very sketchy.

jgkamat:
> Exactly which bugs did upgrading gcc or that ppa actually cause.

The ppa caused the upgrade to 16.04 to fail, since newer versions couldn't safely upgrade from the hacked in versions. That upgrade might have been worse than other gcc upgrades since I think the ABI changed in gcc 5.

> Gcc isn't exactly releasing non working or non backwards compatibility stuff that often.

Using gcc6 as a scapegoat, since it's been around a while.
- [bugs fixed in 6.2](https://gcc.gnu.org/bugzilla/buglist.cgi?bug_status=RESOLVED&resolution=FIXED&target_milestone=6.2)
- [bugs fixed in 6.3](https://gcc.gnu.org/bugzilla/buglist.cgi?bug_status=RESOLVED&resolution=FIXED&target_milestone=6.3)
- [bugs fixed in 6.4 (firefox compilation segfaulting!)](https://gcc.gnu.org/bugzilla/buglist.cgi?bug_status=RESOLVED&resolution=FIXED&target_milestone=6.4)

Writing compilers is hard, I guess, esp with all these new language features. I don't think clang or other compilers are free from this as well.

I ran the 6.0 branch before it stabilized a while back and it was not fun. The bugs that affected me were fixed in 6.3, I think.

- [bugs to be fixed in 7.3 (unreleased)](https://gcc.gnu.org/bugzilla/buglist.cgi?bug_status=RESOLVED&resolution=FIXED&target_milestone=7.3)

ashaw596:
Lol what did it break about the upgrade? It couldn't compile it? Lol gcc... Eh fine. I was thinking more of breaking our stuff. Yeah compilers bugs are pretty dumb. But I really don't think 7.2 is that bad. Bugs seem to mooostly be performance and c++17 implementation details. It's the default for 17.10 so I doubt it's that untested. But yeah having random builds break is not good.

ashaw596:
Also I really wanted structured bindings (and stuff) =(. You could right sorta python style.
auto& [x,y] = some_Point;

jgkamat:
For point, I'm not sure if that would work or not, since all the examples I can find are destructuring fields, and not values from getter methods as we use in point (we could get `_x`, but that seems hacky).

If it turns out to be something that we use a LOT, we could:

`#define DSTR_PT(var1, var2, object) auto& var1 = object.x(); auto& var2 = object.y();`

but I really doubt it's that useful. I don't see destructuring used much even in python.

ashaw596:
I mean its pretty ridiculously easy to creater a pair getter. return std::pair<reference_wrapper<int>, reference_wrapper<int>>(getX(), getY()). And its ridiculously useful for iterating over maps and stuff.
for (const auto &[key, value] : map){
vs
for (const auto &pair : map){
const auto key = pair.first;
const auto value = pair.second.

Theres no like killer change I can point to, but its like alot of minor things that add up. Its nice to replace 
namespace A{
namespace B{
}
}
namespace A::B::C {
}

All the parrellel algorithm implementations can speed up some hotspots.

The File System stuff is nice to have instead of using random stuffs.

variant could fix alot of design problems. Like aloooottt.

constexpr if, iterations, lambdas finally makes constexpr usable for lots of stuff.

idk, its just alot of nice stuff to have available to make our codebase cleaner.

JNeiger:
Can you go more in depth about that parallel algorithm stuff?
That sounds pretty cool


On Oct 7, 2017 10:56 AM, "ashaw596" <notifications@github.com> wrote:

I mean its pretty ridiculously easy to creater a pair getter. return
std::pair<reference_wrapper, reference_wrapper>(getX(), getY()). And its
ridiculously useful for iterating over maps and stuff.
for (const auto &[key, value] : map){
vs
for (const auto &pair : map){
const auto key = pair.first;
const auto value = pair.second.

Theres no like killer change I can point to, but its like alot of minor
things that add up. Its nice to replace
namespace A{
namespace B{
}
}
namespace A::B::C {
}

All the parrellel algorithm implementations can speed up some hotspots.

The File System stuff is nice to have instead of using random stuffs.

variant could fix alot of design problems. Like aloooottt.

constexpr if, iterations, lambdas finally makes constexpr usable for lots
of stuff.

idk, its just alot of nice stuff to have available to make our codebase
cleaner.

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/RoboJackets/robocup-software/pull/1036#issuecomment-334945218>,
or mute the thread

ashaw596:
Whoops. Looking into it more. It turns out they haven't actually implemented it in libstd yet. =( When the compiler is C++17, but the library hasn't still isn't done... Basically it adds a "Execution Policy" to alot of std algorithms. http://www.bfilipek.com/2017/08/cpp17-details-parallel.html Then if you run them in parrallel mode, then the function uses whatever to vectorized, and/or parrallelize it. Intel has an implementation based on OpenMPI https://software.intel.com/en-us/get-started-with-pstl And there's an hpx standard library that uses alot of methods. It seems to even support some CUDA implementations. HPX is a bigger project for concurrency and multithreading on everything tho. https://github.com/STEllAR-GROUP/hpx

Issue #1041, opened 2017-10-22
Add a test point for test_circle_near_ball.py


Jason27chan:
Test the point (0, 0) for the function circle_near_ball

use the method test_point(x, y) and place it after the line 44, which says "# Checks (0, 0). Should be right in front of our goal" 

This is what it should look after youre done 
\# Checks (0, 0). Should be right in front of our goal
test_point(self, 0, 0)

PR #1042, opened 2017-10-22
Add 3 unit tests 

Jason27chan:
Made unit tests for 
-ball.py
-circle_near_ball.py
-field.py


soccer/gameplay/tests/test_field.py
```diff
@@ -0,0 +1,151 @@
...
+	#
+	def set_robot_pos(self, x, y):
+		for robot in self.robots:
+			robot.set_pos(x, y)
+		main.set_their_robots(self.robots)
+
+	# Positions 6 robots around a point but not on it
+	#
+	# @param x: x coordinate of point
+	# @param y: y coordinate of point
```
JNeiger: Missing the `rad` param

soccer/gameplay/tests/test_field.py
```diff
@@ -0,0 +1,151 @@
...
+	# 
+	def set_robots_around_pos(self, x, y, rad):
+		self.robots[0].set_pos(x - constants.Robot.Radius * rad, y - constants.Robot.Radius * rad)
+		self.robots[1].set_pos(x - constants.Robot.Radius * rad, y + constants.Robot.Radius * rad)
+		self.robots[2].set_pos(x, y + constants.Robot.Radius * rad)
+		self.robots[3].set_pos(x + constants.Robot.Radius * rad, y - constants.Robot.Radius * rad)
+		self.robots[4].set_pos(x + constants.Robot.Radius * rad, y)
+		self.robots[5].set_pos(x, y - constants.Robot.Radius * rad)
```
JNeiger: This might be cleaner if you create a variable with the value `constants.Robot.Radius * rad`

soccer/gameplay/tests/test_field.py
```diff
@@ -0,0 +1,151 @@
...
+	def test_field_pos_coeff_at_pos(self):
+
+		# Easier than writing out whole function
+		# Finds field_pos_coeff_at_pos for a point
+		#
+		# @param x: x coordinate of point
+		# @param y: y coordinate of point
+		# @param center: How much to weight being close to the 'center of the field'
+		# @param dist: How much to weight being close to the opponents goal
+		# @param angl: How much to weight the angle between the robot and the goal (In turn, how small the goal is)
```
JNeiger: Missing `attack` param

soccer/gameplay/tests/test_ball.py
```diff
@@ -0,0 +1,82 @@
...
+
+	def set_pos(self, x, y):
+		self.pos = robocup.Point(x, y)
+
+	def set_vel(self, x, y):
+		self.vel = robocup.Point(x, y)
+
+class TestBall(unittest.TestCase):
+	
+
```
JNeiger: I'm not sure if `make pretty` will fix this, but to keep the style the same, the extra blank lines here to be removed.

soccer/gameplay/tests/test_circle_near_ball.py
```diff
@@ -0,0 +1,76 @@
...
+		test_point(self, constants.Field.CenterPoint.x, constants.Field.CenterPoint.y)
+		
+		# Checks a point that should be clearly outside the field
+		test_point(self, width * -1, length * -1)
+		
+		# Checks bottom left corner of field
+		test_point(self, width * -1 / 2, 0)
+
+		# Checks (0, 0). Should be right in front of our goal
+
```
JNeiger: Extra comment?

soccer/gameplay/tests/test_field.py
```diff
@@ -0,0 +1,151 @@
...
+		# Easier than writing out whole function
+		# Finds space_coeff_at_pos for a point
+		#
+		# @param x: x coordinate of point
+		# @param y: y coordinate of point
+		def run_function(x, y):
+			return evaluation.field.space_coeff_at_pos(robocup.Point(x, y))
+		
+		# both robots and position are at (0, 0)
+		self.assertTrue(run_function(0, 0) == 1, msg = msg + "Position occupied by robots is considered open")
```
JNeiger: You shouldn't need to add the `msg = msg +` part. The output of a failed test already tells us which function isn't working. The `msg` argument should only be there to describe why that test failed. Here is the output for just a `assertTrue(False, 'failure message goes here')`

```
test_fail (__main__.FailureMessageTest) ... FAIL

======================================================================
FAIL: test_fail (__main__.FailureMessageTest)
----------------------------------------------------------------------
  File "test_failwithmessage.py", line 6, in test_fail
    self.assertTrue(False, 'failure message goes here')
AssertionError: failure message goes here

----------------------------------------------------------------------
Ran 1 test in 0.000s

FAILED (failures=1)
```

The `"Position occupied by robots is considered open"` should be more than enough to explain the failure.

soccer/gameplay/tests/test_circle_near_ball.py
```diff
@@ -0,0 +1,76 @@
...
+		# Checks a point just inside the field
+		test_point(self, width / 2 - constants.Robot.Radius, constants.Robot.Radius)
+		
+		# Checks a point just outside of the field
+		test_point(self, width / 2 + constants.Robot.Radius, constants.Robot.Radius * -1)
+
+	def test_normalize_angle(self):
+		angle = -math.pi
+		while (angle < 0):
+			self.assertEqual(self.circle_near_ball.normalize_angle(angle),
```
JNeiger: Since these are floats/doubles, `self.assertAlmostEqual( [first], [second] )` should be used due to rounding errors.

soccer/gameplay/tests/test_field.py
JNeiger: `assertAlmostEqual` should be used here too instead of `assertTrue`

soccer/gameplay/tests/test_field.py
```diff
@@ -0,0 +1,151 @@
...
+		self.set_robot_pos(0, length)
+		self.assertTrue(run_function(0, 0) == 0, msg = msg + "Position devoid of robots is not considered open") 
+
+		# Robots are bottom right corner, check bottom left position 
+		self.set_robot_pos(width / 2, 0) 
+		self.assertTrue(run_function(width * -1 / 2, 0) == 0, msg = msg + "Postion devoid of robots is not considered open")
+
+		# 6 Robots are located around position, distance around ~.9 from position
+		self.set_robots_around_pos(0, length / 2, 5)
+		self.assertTrue(run_function(0, length / 2) >= 0, msg="Robots located around position considered to be open. Check constant inside function")
```
JNeiger: This should probably be replaced with two assert statements. First, `assertGreater( [first], [second] )`, and then `assertAlmostEqual`

soccer/gameplay/tests/test_field.py
```diff
@@ -0,0 +1,151 @@
...
+
+		msg = "evaluation.field.space_coeff_at_pos not functioning as intended: "
+
+		#########################################
+		## Test distance to center line factor ##
+		#########################################
+
+		# Test center of the field returns 1
+		num = run_function(0, length / 2, 1 ,0 ,0)
+		self.assertTrue(num == 1, msg + "center of the field doesn't return 1 for distance to center factor")
```
JNeiger: `assertAlmostEqual`

soccer/gameplay/tests/test_field.py
```diff
@@ -0,0 +1,151 @@
...
+
+		# Test halfway between center and edge returns not 0 or 1
+		num = run_function(width / 4, length / 2, 1, 0, 0)
+		self.assertTrue(num > 0 and num < 1, msg + "position between edge and center returns 1 or 0 for distance to center factor")
```
JNeiger: I would split this into `assertGreater` and `assertLess`

soccer/gameplay/tests/test_circle_near_ball.py
```diff
@@ -0,0 +1,76 @@
...
+	def test_get_circle_points(self):
+		width = constants.Field.Width
+		length = constants.Field.Length
+		
+		# Checks that the points around the ball are within the bounds of the field
+		def assert_circle_points(self):
+			array = self.circle_near_ball.get_circle_points(6)
+			for point in array:		
+				self.assertTrue(point.x >= width * -1 / 2 and point.x <= width / 2, msg="failure of get_circle_points. X position of points are outside the field. Point's x position is %d" %point.x)
+				self.assertTrue(point.y >= 0 and point.y <= length, msg="failure of get_circle_points. Y positions of points are outside the field. Point's y position is %d" %point.y )
```
JNeiger: These should be split up into `assertLess` and `assertGreater`. When trying to debug this and it fails later on, it will tell us what the actual values of the variables instead of `Error: Expected True, Received False`

JNeiger:
I didn't leave a comment on all the `assertTrue`'s that need to be changed, but I left a comment on the first one in each file to make it easier.

soccer/gameplay/tests/test_circle_near_ball.py
JNeiger: Oh, this corresponds to #1041

soccer/gameplay/tests/test_field.py
Jason27chan: why assertAlmostEqual instead of assertEqual?

JNeiger: `space_coeff_at_position` returns a floating point number, if the number isn't exactly 1 it will return false. `assertAlmostEqual` only checks equality down to like 6(?) decimal places

Jason27chan:
@JNeiger I made the changes, can you check it out?

JNeiger:
@jgkamat What is the best practice for unit testing? Is it 1 test class for each functionality with a test case for each reasonable test? Or is it 1 test class for each python file where each test case represents a different functionality and all the reasonable tests are within that one test case?

jgkamat:
I think that this is personal preference and/or style, I don't think it matters too much. What I've seen is usually one file per 'high level idea', and many test cases in that file. In our case, it makes sense to have one test class per play, which is what's done here AFAIK. I think the layout that's in this PR is pretty good, FWIW.

soccer/gameplay/tests/test_circle_near_ball.py
```diff
@@ -0,0 +1,79 @@
...
+
+	def test_get_circle_points(self):
+		width = constants.Field.Width
+		length = constants.Field.Length
+		
+		# Checks that the points around the ball are within the bounds of the field
+		def assert_circle_points(self):
+			array = self.circle_near_ball.get_circle_points(6)
+			for point in array:		
+				self.assertLessEqual(point.x, width, "X position of points are outside the field.")
```
JNeiger: I think this should be `width * 0.5`

PR #1043, opened 2017-10-24
Adaptive defense

JNeiger:
A new defensive play where the marking priority and location is a function of many different factors. Specifically, opponent robots are grouped based on their location in reference to the ball to be either strikers or wingers. Strikers are the robots near the ball and able to be grouped into a single blob while the wingers are possible pass targets and need to be blocked individually. Ideally, the marking location will also take into account probable receive locations in the direction of movement of the receiver robot.

Finally, the robots on offense and defense will take a specific formation that will be stretched and squished depending on the locations of these calculated opponent robots or marking locations.


JNeiger:
Closing in favor of #1225 which is merging from the correct branch

PR #1052, opened 2017-11-05
Fix capture picking bad robots

### LINKED ISSUE #375: Make capture choose a robot close to the intercept point
Capture moves the robot to an intercept point along the ball's trajectory if it is moving, so the role_requirements() method should specify that point as the preferred bot location, not the ball's current location.  This may be best as a custom cost function since it's robot-dependent.

jcarn:
Closes #375

### DIFF
--- a/soccer/gameplay/skills/capture.py
+++ b/soccer/gameplay/skills/capture.py
@@ -42,18 +42,21 @@ class Capture(single_robot_behavior.SingleRobotBehavior):
                             'immediately')
 
         self.add_transition(
-            Capture.State.course_approach, Capture.State.fine_approach,
-            lambda: (self.bot_in_front_of_ball() or self.bot_near_ball(Capture.CourseApproachDist)) and main.ball().valid,
-            'dist to ball < threshold')
+            Capture.State.course_approach,
+            Capture.State.fine_approach, lambda: (self.bot_in_front_of_ball(
+            ) or self.bot_near_ball(Capture.CourseApproachDist)) and main.ball(
+            ).valid, 'dist to ball < threshold')
 
         self.add_transition(
-            Capture.State.fine_approach, behavior.Behavior.State.completed,
-            lambda: self.bot_near_ball(constants.Robot.Radius + constants.Ball.Radius),
-            'has ball')
+            Capture.State.fine_approach,
+            behavior.Behavior.State.completed, lambda: self.bot_near_ball(
+                constants.Robot.Radius + constants.Ball.Radius), 'has ball')
 
         self.add_transition(
-            Capture.State.fine_approach, Capture.State.course_approach,
-            lambda: not (self.bot_in_front_of_ball() or self.bot_near_ball(Capture.CourseApproachDist)) and (not self.bot_near_ball(Capture.CourseApproachDist * 1.5) or not main.ball().pos),
+            Capture.State.fine_approach, Capture.State.course_approach, lambda:
+            not (self.bot_in_front_of_ball() or self.bot_near_ball(
+                Capture.CourseApproachDist)) and (not self.bot_near_ball(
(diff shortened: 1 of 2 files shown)


jcarn:
The role req stuff is @mmurley3 PR that hasn't merged yet, but it was useful

soccer/gameplay/skills/capture.py
```diff
@@ -152,5 +132,37 @@ def role_requirements(self):
...
         # try to be near the ball
         if main.ball().valid:
-            reqs.destination_shape = main.ball().pos
+            reqs.cost_func = lambda r: reqs.position_cost_multiplier * find_robot_intercept_ball(
+                r).dist_to(r.pos)
+
         return reqs
+
+
+def find_robot_intercept_ball(robot):
```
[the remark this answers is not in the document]
jcarn: It specifically uses the ball, but in the future having a function that could find an intercept with both a robot and a ball (and anything else that moves?) could be useful so I renamed it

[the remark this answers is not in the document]
JNeiger: That's a good point. I think I was comparing it to the previous version of that function.

soccer/gameplay/skills/capture.py
```diff
@@ -73,32 +76,9 @@ def bot_in_front_of_ball(self):
...
-            ball_time = evaluation.ball.rev_predict(main.ball().vel, dist)
-            robotDist = (pos - self.robot.pos).mag() * 0.6
-            bot_time = robocup.get_trapezoidal_time(robotDist, robotDist, 2.2,
-                                                    1, self.robot.vel.mag(), 0)
-
-            if bot_time < ball_time:
-                break
-
-        return pos
+        return find_robot_intercept_ball(self.robot)
```
JNeiger: Don't forget this one either

soccer/gameplay/skills/capture.py
```diff
@@ -152,5 +132,37 @@ def role_requirements(self):
...
+
+
+def find_robot_intercept_point(robot):
+    if robot is None:
+        return main.ball().pos
+    approach_vec = approach_vector(robot)
+    # sample every 5 cm in the -approach_vector direction from the ball
+    pos = None
+    for i in range(50):
+        dist = i * 0.05
```
jgkamat: I /really/ don't like magic numbers, can you make some global constants for these

soccer/gameplay/skills/capture.py
```diff
@@ -152,5 +132,37 @@ def role_requirements(self):
...
         # try to be near the ball
         if main.ball().valid:
-            reqs.destination_shape = main.ball().pos
+            reqs.cost_func = lambda r: reqs.position_cost_multiplier * find_robot_intercept_point(
+                r).dist_to(r.pos)
+
         return reqs
+
+
+def find_robot_intercept_point(robot):
```
jgkamat: This function is giving me a ton of deja vu, does it exist somewhere else already? I could just be misthinking though.

soccer/gameplay/skills/capture.py
```diff
@@ -152,5 +132,37 @@ def role_requirements(self):
...
+    approach_vec = approach_vector(robot)
+    # sample every 5 cm in the -approach_vector direction from the ball
+    pos = None
+    for i in range(50):
+        dist = i * 0.05
+        pos = main.ball().pos + approach_vec * dist
+        # how long will it take the ball to get there
+        ball_time = evaluation.ball.rev_predict(main.ball().vel, dist)
+        robotDist = (pos - robot.pos).mag() * 0.6
+        bot_time = robocup.get_trapezoidal_time(robotDist, robotDist, 2.2, 1,
```
jgkamat: The 2.2 constant seems pretty important (max speed?). We should pull this from motion constratints at the very least

PR #1059, opened 2017-11-19
added status indicator for plays

krispreza:


jgkamat:
Overall, looks amazing, I need to do a full review though.

soccer/gameplay/main.py
```diff
@@ -164,6 +164,13 @@ def save_playbook(file_name, isAbsolute=False):
         (PLAYBOOKS_DIR + '/' if not isAbsolute else '') + file_name,
         _play_registry.get_enabled_plays_paths())
 
+def clear():
+    global _play_registry
+    _play_registry.clear();
+
+def checkPlayStatus():
```
jgkamat: I think this name is misleading, can we name it to something like `numEnabledPlays` or similar?

soccer/MainWindow.cpp
```diff
@@ -949,6 +953,15 @@ void MainWindow::status(QString text, MainWindow::StatusType status) {
     }
 }
 
+void MainWindow::playStatus(bool color) {
```
jgkamat: I don't like the name `playStatus` either, it feels a bit too generic. 

Can we name it something like `playChangedStatus` or `playIndicatorStatus` or similar?

soccer/gameplay/GameplayModule.cpp
```diff
@@ -255,6 +256,27 @@ void Gameplay::GameplayModule::savePlaybook(const string& playbookFile,
     PyGILState_Release(state);
 }
 
+void Gameplay::GameplayModule::clearPlays() {
+    PyGILState_STATE state = PyGILState_Ensure();
+    getMainModule().attr("clear")();
+    PyGILState_Release(state);
+}
+
+bool Gameplay::GameplayModule::checkPlayStatus() {
```
joshhting: I feel like this function could be named something that's more specific to what it does, like `playStatusChanged`?

soccer/gameplay/GameplayModule.cpp
```diff
@@ -255,6 +256,27 @@ void Gameplay::GameplayModule::savePlaybook(const string& playbookFile,
...
+  PyGILState_STATE state = PyGILState_Ensure();
+  static int prevStatus = extract<int>(getMainModule().attr("checkPlayStatus")());
+  bool static change = false;
```
joshhting: Does this need to be static?

chachmu:
@jgkamat Does this look good to you?

jgkamat:
Yup, I'm good to go with this, we'll handle issues if they come up :P

PR #1065, opened 2017-11-19
it said i needed to commit these files in order to pull from rj master

gkarnati3:


jgkamat:
I think you might have had some stale files in your repo, when this happens, you usually want to trash the stale files and only commit the ones you want to contribute. Merging this dosen't help much (as it's mostly committing stale files), so I'm closing this. You probably want to do a hard reset to the master branch before starting any work.

Issue #1046, opened 2017-10-29
Remove Face for the Goalie Intercept State


JNeiger:
When trying to intercept the ball, the goalie has a "face" and "move to" command. We should remove the "face" because it causes some lag in terms of moving to block the ball. We should double check some of the edge cases, but it seemed to work just fine on the field.

Issue #947, opened 2017-03-16
Refactor/Move goalie.py to avoid confusion


gtrosawa:
# TODO: merge this back into the regular goalie?

jgkamat:
This is not feasable, and would break our current goalie strategy. `goalie.py` is deprecated, and it should be moved to a 'deprecated' folder to avoid confusion, if it isn't being used anywhere. I'll make this a 'refactor/remove goalie.py' issue instead.

jgkamat:
To solve this issue, we'll need to move goalie.py to a place where it can't be mistaken as our goalie code, or put some warnings somewhere. This has caused at least 2 people to do work they shouldn't have, so I would like this fixed before next year ;P

PR #1066, opened 2017-11-21
Goalie intercepting without concern for face

liquidmetal9015:
Issue #1046 

Note: I don't currently know how to run check style or the make pretty, If style (or function for that matter) is a problem then I can revise at the meeting tomorrow. 

Three changes:
1. Removed redundant goalie code:
        In intercept_execute there was code that was duplicating the functionality of the intercept sub behavior while the intercept sub behavior was also being called. This has been removed. 

2. Made goalie not always face the ball
        There was a function being called continuously in goalie that would make it face the ball, a condition has been added so that this function will not be called in the intercept state, but still in all other states.

3. Made facing the ball during the intercept sub-behavior optional
       Previously the intercept sub behavior made the active robot face the ball, a parameter has been added to the intercept constructor to make this optional for faster intercepts.

I'm happy to make any changes needed.

### DIFF
--- a/soccer/gameplay/skills/intercept.py
+++ b/soccer/gameplay/skills/intercept.py
@@ -6,7 +6,6 @@ import main
 
 # wraps up OurRobot.move() into a Skill so we can use it in the play system more easily
 class Intercept(single_robot_behavior.SingleRobotBehavior):
-
     def __init__(self, pos=None, faceBall=True):
         super().__init__(continuous=True)
 
--- a/soccer/gameplay/skills/intercept.py
+++ b/soccer/gameplay/skills/intercept.py
@@ -10,7 +10,7 @@ class Intercept(single_robot_behavior.SingleRobotBehavior):
     #controls if the robot should face the ball while intercepting
     faceBall = False
 
-    def __init__(self, pos=None, faceBall=False):
+    def __init__(self, pos=None, faceBall=True):
         super().__init__(continuous=True)
 
         self.faceBall = faceBall
--- a/soccer/gameplay/skills/intercept.py
+++ b/soccer/gameplay/skills/intercept.py
@@ -7,9 +7,6 @@ import main
 # wraps up OurRobot.move() into a Skill so we can use it in the play system more easily
 class Intercept(single_robot_behavior.SingleRobotBehavior):
 
-    #controls if the robot should face the ball while intercepting
-    faceBall = False
-
     def __init__(self, pos=None, faceBall=True):
         super().__init__(continuous=True)
 
--- a/soccer/gameplay/skills/intercept.py
+++ b/soccer/gameplay/skills/intercept.py
@@ -17,7 +17,8 @@ class Intercept(single_robot_behavior.SingleRobotBehavior):
 
         self._shape_constraint = None
 
-        self.ball_line = lambda: robocup.Segment(main.ball().pos, main.ball().pos + (main.ball().vel.normalized() * 8.))
+        self.ball_line = lambda: robocup.Segment(main.ball(
+        ).pos, main.ball().pos + (main.ball().vel.normalized() * 8.))
 
         self.add_transition(behavior.Behavior.State.start,
                             behavior.Behavior.State.running, lambda: True,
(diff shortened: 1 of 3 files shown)


soccer/gameplay/skills/intercept.py
```diff
@@ -6,9 +6,15 @@
 
 # wraps up OurRobot.move() into a Skill so we can use it in the play system more easily
 class Intercept(single_robot_behavior.SingleRobotBehavior):
-    def __init__(self, pos=None):
+
+    #controls if the robot should face the ball while intercepting
+    faceBall = False
```
JNeiger: This version of `faceBall` isn't being used anywhere. If you ever want to access this one, you generally have to do `Intercept.faceBall`. Most of the time for these class variables, we only ever use them as constants (See adaptive formation for example).

soccer/gameplay/tactics/positions/goalie.py
```diff
@@ -164,17 +170,15 @@ def on_exit_clear(self):
...
+        i = skills.intercept.Intercept(None, False)
         self.add_subbehavior(i, 'intercept', required=True)
 
     def execute_intercept(self):
-        ball_path = robocup.Segment(
-            main.ball().pos,
-            main.ball().pos + main.ball().vel.normalized() * 10.0)
-        dest = ball_path.nearest_point(self.robot.pos)
-        self.robot.move_to(dest)
+        pass
```
JNeiger: If we don't need `execute_intercept` anymore, it's probably better to just completely remove the function stub.

soccer/gameplay/tactics/positions/goalie.py
```diff
@@ -106,7 +110,9 @@ def robot_is_facing_our_goal(self, robot):
     # note that execute_running() gets called BEFORE any of the execute_SUBSTATE methods gets called
     def execute_running(self):
         if self.robot != None:
-            self.robot.face(main.ball().pos)
+            #face the ball in all states but intercept
+            if not self.isInter:
```
JNeiger: So I've never actually used this, but there might be a function `self.is_in_state(Goalie.intercept)` that returns whether you are in that state or not.

If that doesn't work, you can also just check if `self.has_subbehavior_with_name('intercept')` is true and it should do the same thing.

JNeiger:
Overall, it looks pretty good! Just a few things here or there. 

Checkstyle is super easy to fix, all you have to do is run `make pretty` before you commit your code and it should fix all the checkstyle stuff. Worst comes to worst, you can just click the `Details` beside `circle/style`, copy/paste that git command in terminal, and re-commit everything.

soccer/gameplay/tactics/positions/goalie.py
```diff
@@ -21,6 +21,10 @@ class Goalie(single_robot_composite_behavior.SingleRobotCompositeBehavior):
         robocup.Point(MaxX, constants.Robot.Radius + OFFSET))
```
jgkamat: Are you sure that this is the correct goalie? I think that this goalie isn't used unless in extreme situations (I think during unknown states), I think the goalie you want to modify is [submissive_goalie](https://github.com/RoboJackets/robocup-software/blob/master/soccer/gameplay/tactics/positions/submissive_goalie.py), which runs under the [defense tactic](https://github.com/RoboJackets/robocup-software/blob/master/soccer/gameplay/tactics/defense.py).

Also see https://github.com/RoboJackets/robocup-software/issues/947 :( 

The reason for this is because it allows more advanced coordination between the goalie and the defense (which is needed for effective defense).

liquidmetal9015: I'm sure that this is the goalie that runs if you just start soccer and press the green check, beyond that I don't know.

Edit: Well now that you mention it submissive goalie does appear to be the one that runs when the other plays are running. I think I can modify that one using the exact same changes, as I think the relevant sections are identical.

soccer/gameplay/skills/intercept.py
jgkamat: I think this is the only issue left. 

The only time `goalie.py` is used is during penalty kicks (which is basically untested), does this seem like a good idea for penalty kicks too?

liquidmetal9015: I'm not really sure what change to make here. I was under the impression that I was using that variable when I use self.faceBall in line 49. I'm not sure where I can put it that will still give me access to that information.

liquidmetal9015:
I think this may be ready to merge.

Issue #1051, opened 2017-11-04
Robots past the target in the Kick Evaluator should scale based on the CDF instead of the PDF


JNeiger:
[This line](https://github.com/RoboJackets/robocup-software/blob/master/soccer/KickEvaluator.cpp#L101) should be ~~`1 - ( 1 + erf(distPastTarget / (*robot_std_dev * sqrt(2))));`~~.
Additionally, [this line](https://github.com/RoboJackets/robocup-software/blob/master/soccer/KickEvaluator.cpp#L99) can be removed.

Edit: This is the right formula `1 - erf(distPastTarget / (*robot_std_dev * sqrt(2)))`

PR #1069, opened 2017-11-29
Fixed Robots past the target in the Kick Evaluator should scale based on the CDF instead of the PDF

gkarnati3:
Resolved issue #1051

### DIFF
--- a/soccer/KickEvaluator.cpp
+++ b/soccer/KickEvaluator.cpp
@@ -96,7 +96,7 @@ KickResults KickEvaluator::eval_pt_to_seg(Point origin, Segment target) {
         // If robot is past target, only use the chance at the target segment
         if (distPastTarget > 0 && fabs(get<1>(loc)) < M_PI / 2) {
             // Evaluate a normal distribution at dist away and scale
-            float stdev2 = pow(*robot_std_dev, 2);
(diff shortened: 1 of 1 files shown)


soccer/KickEvaluator.cpp
```diff
@@ -96,9 +96,8 @@ KickResults KickEvaluator::eval_pt_to_seg(Point origin, Segment target) {
         // If robot is past target, only use the chance at the target segment
         if (distPastTarget > 0 && fabs(get<1>(loc)) < M_PI / 2) {
             // Evaluate a normal distribution at dist away and scale
-            float stdev2 = pow(*robot_std_dev, 2);
             botVertScales.push_back(
-                fast_exp(-0.5 * pow(distPastTarget, 2) / stdev2));
+                1 - erf(distPastTarget / (*robot_std_dev * sqrt(2)));
```
joshhting: I believe you're missing a `(1 +` here

joshhting: If you try and run `make run sim` it doesn't compile because of the parantheses

JNeiger: So I think this is actually correct. In the issue, I messed up the formula. It was corrected in #1064 before we closed it

PR #1073, opened 2017-11-29
Flip joysticks

ldricci3:
Flipped translation and rotation joystick mapping.

Left stick now controls translation, right stick now controls rotation.

### DIFF
--- a/soccer/joystick/GamepadController.cpp
+++ b/soccer/joystick/GamepadController.cpp
@@ -166,38 +166,38 @@ void GamepadController::update() {
      */
     // Logitech F310 Controller
     _controls.rotation =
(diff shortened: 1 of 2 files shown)


JNeiger:
This is looking pretty good. I'm not sure about having all those extra cmake files included. Could you remove those as well as the `which_half` files?

The accepted answer [here](https://stackoverflow.com/questions/12481639/remove-files-from-git-commit) seems like a reasonable way to do it.

Also, you can run `make pretty` to clean up those style problems.

joshhting:
Also can you add the issue number to the description?

chachmu:
This looks good, I'm going to test it with an actual joystick at the next meeting and if everything checks out I'll go ahead and merge this.

PR #1074, opened 2017-11-29
Fixed Robots past the target in the Kick Evaluator should scale based on the CDF instead of the PDF

### LINKED ISSUE #1051: Robots past the target in the Kick Evaluator should scale based on the CDF instead of the PDF

gkarnati3:
fixes #1051

### DIFF
--- a/soccer/KickEvaluator.cpp
+++ b/soccer/KickEvaluator.cpp
@@ -97,7 +97,7 @@ KickResults KickEvaluator::eval_pt_to_seg(Point origin, Segment target) {
         if (distPastTarget > 0 && fabs(get<1>(loc)) < M_PI / 2) {
             // Evaluate a normal distribution at dist away and scalet
             botVertScales.push_back(
-                1 - (1 + erf(distPastTarget / (*robot_std_dev* sqrt(2)))));
+                1 - (1 + erf(distPastTarget / (*robot_std_dev * sqrt(2)))));
         } else {
             botVertScales.push_back(1);
         }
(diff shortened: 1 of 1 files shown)


chachmu:
You don't have to make a new pull request to make changes, you can just push to the branch that you used to make the pull request and it will update automatically.

soccer/KickEvaluator.cpp
```diff
@@ -95,10 +95,9 @@ KickResults KickEvaluator::eval_pt_to_seg(Point origin, Segment target) {
 
         // If robot is past target, only use the chance at the target segment
         if (distPastTarget > 0 && fabs(get<1>(loc)) < M_PI / 2) {
-            // Evaluate a normal distribution at dist away and scale
-            float stdev2 = pow(*robot_std_dev, 2);
+            // Evaluate a normal distribution at dist away and scalet
             botVertScales.push_back(
-                fast_exp(-0.5 * pow(distPastTarget, 2) / stdev2));
+                1 - (1 + erf(distPastTarget / (*robot_std_dev * sqrt(2)))));
```
JNeiger: I think this got reverted back to the old equation. 

Here's the right one. `1 - erf(distPastTarget / (*robot_std_dev * sqrt(2)))`

PR #1075, opened 2017-11-29
Add C++ Geometry 2d Rotate Point to the Python side

### LINKED ISSUE #965: Add C++ Geometry 2d Rotate Point to the Python side
C++ defaults to rotating a point around the origin if it is only given an angle. This needs to be added to robocup-py.cpp so it will be available in python too.

CollinAvidano:
Added a rotate_origin wrapper in robocup-py.cpp for the single argument definition of rotate from point.

Closes #965

### DIFF
--- a/soccer/gameplay/robocup-py.cpp
+++ b/soccer/gameplay/robocup-py.cpp
@@ -194,6 +194,10 @@ void Point_rotate(Geometry2d::Point* self, Geometry2d::Point* origin,
     self->rotate(*origin, angle);
 }
 
+void Point_rotate_origin(Geometry2d::Point* self, float angle) {
+    self->rotate(angle);
+}
+
 void CompositeShape_add_shape(Geometry2d::CompositeShape* self,
                               Geometry2d::Shape* shape) {
     if (shape == nullptr) throw NullArgumentException("shape");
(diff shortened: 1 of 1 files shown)


soccer/gameplay/robocup-py.cpp
```diff
@@ -646,6 +650,7 @@ BOOST_PYTHON_MODULE(robocup) {
         .def("__repr__", &Point_repr)
         .def("normalized", &Geometry2d::Point::normalized, Point_overloads())
         .def("rotate", &Point_rotate)
+        .def("rotate_origin", &Point_rotate_origin)
```
jgkamat: I think since `Point_rotate_origin` is a single line, this might be able to just be `&Geometry2d::Point::angle` maybe? I haven't tested it though.

chachmu: Unless I am misunderstanding your question I think that you have to make a wrapper for any of the functions that directly alter the point that it is passed through pointers since python doesn't seem to directly support that.

jgkamat: ah, that's right, I forgot about that, this seems good then to me!

PR #1079, opened 2017-12-03
Jay/restore plays

jgkamat:
Restores training plays we removed in 2017


jgkamat:
Merging as the failure in style is due to `__init__.py` (?!?)

PR #1080, opened 2017-12-05
Defense Improvement

### LINKED ISSUE #999: Rewrite Defense.py
While the logic in Defense.py is mostly good, the file is really unmaintainable, not because it has no documentation, but because it's very poorly organized (and is kind of a mess).

When we rewrite it, we should consider using adaptive formation to improve defense placement.

JNeiger:
Closes #999 

- [x] Get opponent robot scores
- [x] Get area scores
- [ ] Categorize the scores based on location to separate the strikers and wingers
- [ ] Do some sort of area / man2man based on opponent robot formation
- [ ] Tune all the constants
- [ ] Clean up and comment


JNeiger:
Closing until I get motivation to finish this

PR #1081, opened 2017-12-05
Improves Motion Control

JNeiger:
~~Adds...~~
- ~~Feed forward for the path velocity at the next step~~
- ~~Velocity PID to keep it close to the path~~
- ~~Position PID which outputs into the velocity PID~~
TODO:
- [x] ~~Implement basic implementation~~
- [x] ~~Tune it~~
- [x] ~~Test on robots and sim~~
- [x] ~~Clean up all the code~~

### DIFF
--- a/soccer/RobotConfig.cpp
+++ b/soccer/RobotConfig.cpp
@@ -27,6 +27,7 @@ RobotConfig::Chipper::Chipper(Configuration* config, QString prefix)
 
 RobotConfig::RobotConfig(Configuration* config, QString prefix)
     : translation(config, QString("%1/translation").arg(prefix)),
+      velTranslation(config, QString("%1/velTranslation").arg(prefix)),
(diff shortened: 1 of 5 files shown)


JNeiger:
At least in sim, this produces better results. Instead of using the current heading for this entire time step, I think it's better to "predict" the average heading for the entire time step and rotate the world velocities by that average amount. In sim, this removes almost all the problems when doing that rotating repeated line up. There is a small problem when it rotates a large amount and is trying to accelerate, but that can be fixed I think.

I need to test this on robots to see if it actually makes that much of a difference IRL. If it doesn't completely solve the problem, I think I can use the commanded "sideways" translation velocity and add that to the commanded rotation velocity to improve the straight line strafing of the robot.

chachmu:
Closing in favor of the new motion control rewrite

PR #1082, opened 2017-12-06
Fix recursive transition in penalty

jgkamat:
This is a very small change, please look at the last two commits for the actual diff, the first commit is reformat.


jgkamat:
This is actually conflicting a lot with #1066, so I'm going to make this as small as possible (and trash the other fixes)

jgkamat:
Tinychange, so merging.

PR #1087, opened 2018-01-07
Switch to new version of stylize

justbuchanan:
I made some updates to stylize and rewrote it in go (it was python before). The old version is still registered with pypi (so `pip install stylize` still works) and the code is still available at https://github.com/justbuchanan/stylize/tree/python. Everything should continue to work as-is.

Since robocup is the only project using stylize (that I know of), I figured I'd send an update in case you're interested in trying the new version. It's a bit faster than before and handles some edge cases better.

In addition to the setup script changes in this PR, you'll also have to set the `GOPATH` environment variable and uninstall the old python version (`pip uninstall stylize`).

Note: the new version of stylize needs some testing - let me know if you run into any issues.

### DIFF
--- a/.stylize.yml
+++ b/.stylize.yml
@@ -1,0 +1,14 @@
+formatters:
+    .py: yapf
+    .h: clang
+    .hpp: clang
+    .c: clang
+    .cc: clang
+    .cpp: clang
+yapf_style: .style.yapf
+clang_style: file
+exclude_dirs:
+    - build
+    - external
+    - old
+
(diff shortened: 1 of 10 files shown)


jgkamat:
I'd be happy to take point on handling any issues (to beta test this). The `GOPATH` issue is a bit of a bummer though, in the past [i've manually built a go environment](https://github.com/RoboJackets/beekeeper/blob/master/smoker/makefile) to get reproducable builds and let people not have to worry about a `GOPATH`. I'll add a `GOPATH` to the docker images, but I'd rather the scripts print a warning (rather than failing) when `GOPATH` is not available.

jgkamat:
I'm moving this into the rj namespace, if you have any updates, you can push to the stylize-go branch :smile: #1088

PR #1088, opened 2018-01-07
Update stylize to new framework

jgkamat:
Continuation of #1087


jgkamat:
Hmm, seems like the compiled binary isn't being put into place, I'll have to look into that later...

justbuchanan:
I see this line in the build logs: "[WARN] No GOPATH set, not installing stylize.", but I also see this happening earlier: "Step 12 : ENV GOPATH /home/developer/golang". Not sure why GOPATH isn't staying set.

justbuchanan:
I extended the config handling a bit, so exclude directories and formatter args can be specified there (.stylize.yml). `make pretty/checkstyle` will still work the same, but the actual stylize commands are much simpler and can be done without make. Try `stylize -g origin/master` (add `-i` for in-place formatting).

One other thought:
If you think the naming conflict between the new go program and the old python program will be a hassle for people, you can change from `go get github.com/justbuchanan/stylize` to `go get gopkg.in/justbuchanan/stylize.v1`, which will name the binary "stylize.v1". The gopkg.in site is a redirector that just forwards to versions of the github repo based on tagged releases.

jgkamat:
Hmm, that's a good idea, since the argument format is changing, this would avoid confusion, thanks for letting me know about that feature :D.

jgkamat:
Ah, the bug is because ENV vars are being dropped by sudo, I should have seen that sooner :P

I wonder what a good solution to this in general is, I don't like passing `-E` to sudo for everyone.

JNeiger:
@jgkamat What's the status on this?

jgkamat:
iirc, the reason this stalled was because the go version of stylize didn't do the magic "find what clang_format is *actually* called" the python version currently does. On ubuntu, the clang formatter isn't named `clang-format`, but rather `clang-format-<version>`. 

```
Loaded config from file .stylize.yml
Formatter clang not installed
makefile:135: recipe for target 'checkstyle' failed
make: *** [checkstyle] Error 1
```

So if this is to be merged, we'll either need to add support for the magic guessing again into stylize or add an alternative which makes a `clang-format` binary from the options. I'm not sure what the best option is currently.....

jgkamat:
> Merged #1088 into master.

FWIW, I'm not sure if this was ready or not, it was at least still failing CI
(I think because the binary wasn't installed properly). I'm not sure though...

chachmu:
As far as I was aware the only check on CI that this was failing was the style check which would understandably fail since it was the problem in the first place

PR #1093, opened 2018-01-24
Add a new theme

chachmu:
This adds a nyan theme. I think I might need to switch the rainbow background to one that isn't wavy since it looks weird when it gets rescaled.


chachmu:
@jgkamat opinions?

soccer/MainWindow.cpp
```diff
@@ -1218,6 +1218,10 @@ void MainWindow::on_actionDarculizedStyle_triggered() {
 void MainWindow::on_action1337h4x0rStyle_triggered() {
     StyleSheetManager::changeStyleSheet(this, "1337H4X0R");
 }
+void MainWindow::on_actionNyanStyle_triggered() {
+    StyleSheetManager::changeStyleSheet(this, "NYAN");
+    std::cout << "STYLE CHANGE" << std::endl;
```
jgkamat: Please remove this debug printout

soccer/ui/themes/nyan.stylesheet
```diff
@@ -0,0 +1,36 @@
+/*
+	Copyright 2013 Emanuel Claesson
```
jgkamat: We should probably not have this header :P

chachmu: for some reason its in all of the stylesheets :/

soccer/MainWindow.cpp
chachmu: whoops 👍

soccer/ui/themes/nyan.stylesheet
jgkamat: ah, it was part of the dark stylesheet since the existing sheets were derived from that, since this is original, I'll remove it.

chachmu:
I feel like the rainbow should be one with straight lines instead since it looks super weird when the bar is moving because the image just gets scaled

chachmu:
@jgkamat I have replaced the rainbow.svg with a 7 pixel version which both fixes the issue of the wavy trail looking bad and also takes up less space. Do you want to rebase out the old version of the image or should we just merge as is?

jgkamat:
since SVG files are actually text, I wouldn't worry about that, it's not worth messing with in my opinion.

PR #1099, opened 2018-02-18
Fix makefile for multi-agent soccer

jgkamat:
Also add warning about proper port and IP

Adding a warning is the only good way to do this, we really shouldn't be broadcasting over localhost, but I suppose that's fine since it solves our interference issues.


jgkamat:
Gonna just merge this since it dosen't affect anything exept for fixing a broken feature :)

PR #1100, opened 2018-02-18
Fix defenders trying to occupy the same space

### LINKED ISSUE #1098: Defenders try to occupy the same space
There is an issue we found at comp where the defenders sometimes try to defend exactly the same spot. This leads to them trying to push each other out of the way.

chachmu:
Closes #1098.


soccer/gameplay/tactics/defense.py
```diff
@@ -78,435 +70,297 @@ def debug(self, value):
...
+                    pass_line = robocup.Segment(main.ball().pos, threat[0])
                     main.system_state().draw_line(
-                        pass_line, constants.Colors.Red, "Defense")
-                    main.system_state().draw_text(
-                        "Pass: " + str(int(threat.ball_acquire_chance * 100.0))
-                        + "%", pass_line.center(), constants.Colors.White,
-                        "Defense")
+                        pass_line, constants.Colors.Red, "Defense-Pass Line")
+
+            #keep defenders from occupying the same spot
```
jgkamat: You should add a note where handlers is defined to look here as well if that's changed, because this will break if handlers is changed.

soccer/gameplay/tactics/defense.py
```diff
@@ -78,435 +70,297 @@ def debug(self, value):
...
+            #only matters if there are 2 defenders (and the goalie)
```
jgkamat: You should have a space after `#` in comments :P

soccer/gameplay/tactics/defense.py
```diff
@@ -78,435 +70,297 @@ def debug(self, value):
...
+            if len(handlers) == 3:
```
jgkamat: Is there any reason you can't go over every combination of 2 points here instead of just using the two defenders? Does that break the goalie? I think thats a bit cleaner and less likely to break.

chachmu: I could but since this is the only case that matters it seems like a waste of time ¯\_(ツ)_/¯

chachmu:
@jgkamat Does this look good now?

soccer/gameplay/tactics/defense.py
```diff
@@ -78,435 +70,300 @@ def debug(self, value):
...
+                handler2 = handlers[2]
+
+                #vector between the 2 points
+                overlap = handler2.move_target - handler1.move_target
+
+                #if the robots overlap
+                if overlap.mag() < (2 * constants.Robot.Radius) + .005:
+                    #move the robots away from each other
+                    overlap = overlap - (overlap.normalized() *
+                                         (constants.Robot.Radius + .1))
```
jgkamat: Weren't you going to make this 0.1 into Radius * 2? Or was that not large enough?

PR #1101, opened 2018-02-20
More Soccer Tests

Jason27chan:
Unit tests for functions in soccer/gameplay/evaluations 
-shooting
-passing
-opponent


soccer/gameplay/robocup-py.cpp
```diff
@@ -828,10 +835,12 @@ BOOST_PYTHON_MODULE(robocup) {
            bases<Robot>>("OpponentRobot", init<int>());
 
     class_<Ball, std::shared_ptr<Ball>>("Ball", init<>())
+        .def("set_pos_for_testing", &Ball_set_pos_for_testing)
         .def_readonly("pos", &Ball::pos)
         .def_readonly("vel", &Ball::vel)
         .def_readonly("valid", &Ball::valid);
 
+
```
JNeiger: Extra newline was accidentally committed

soccer/gameplay/tests/test_opponent.py
```diff
@@ -0,0 +1,112 @@
...
+	def __init__(self, *args, **kwargs):
+		super(TestOpponent, self).__init__(*args, **kwargs)
+		self.system_state = robocup.SystemState()
+
+	def setUp(self):
+		main.init(False)
+		main.set_system_state(self.system_state)
+
+		for robot in main.system_state().their_robots:
+
```
JNeiger: I would take out this newline too

soccer/gameplay/tests/test_opponent.py
```diff
@@ -0,0 +1,112 @@
...
+	def test_get_closest_opponent(self):
+		length = constants.Field.Length
+		width = constants.Field.Width
+
+		# Tests a point for get_closest_opponent
+		# 
+		# @param x: x coordinate of test point
+		# @param y: y coordinate of test point
+		#
+		def test_point(x, y, directional_weight=1, excluded_robots=[]):
```
JNeiger: Should this be `directional_weight=0` per #1094?

soccer/gameplay/tests/test_passing.py
```diff
@@ -0,0 +1,99 @@
...
+		# Test a point passing to itself. 
+		# Supposed to return 0. 
+		# This is an edge case that we discussed to be okay. 
+		# In the future, if we want a pass that starts and ends at the same point 
+		# to be considered successful, we need to modify the eval_pass function. 
+		self.assertEqual(self.eval_pass(0, 0, 0, 0), self.failure, "point pass to itself fail")
+
+		# Test passing outside the fields.
+		# Succeeds when it should fail?
+		self.assertEqual(self.eval_pass(0, 0, 0, 2 * self.length), self.success, "pass to the end of the field fail")
```
JNeiger: Yeahhhh, neither KickEval nor WinEval do any sort of input or bounds checking.

soccer/gameplay/tests/test_passing.py
```diff
@@ -0,0 +1,99 @@
...
+		
+		# Test a point passing from bottom left to top right of field
+		self.assertEqual(self.eval_pass(self.left_side, 0, self.right_side, self.length), self.success, "pass from bottom left to top right fail")
+
+		# Test horizontal pass across the center of the field
+		self.assertEqual(self.eval_pass(self.left_side, self.center_y, self.right_side, self.center_y), self.success, "horizontal pass across center of field")
+
+	def test_eval_pass_with_bots_and_exclusion(self):
+		their_bot1, their_bot2, their_bot3, their_bot4, their_bot5, their_bot6 = self.their_robots[0:6]		
+		our_bot1, our_bot2, our_bot3, our_bot4, our_bot5, our_bot6 = self.our_robots[0:6]
```
JNeiger: Do we need to specify all 6 for each team when only the first robot on each side is being used?

soccer/gameplay/tests/test_shooting.py
```diff
@@ -0,0 +1,99 @@
...
+
+	@unittest.skip("Skip Problematic Cases")
+	def test_eval_shot_problem_cases(self):
+		# This location is very outside the field but we consider it successful
+		# This case return almost 1 when it should fail
+		self.assertEqual(self.eval_shot(0, 2 * self.length), self.failure)
+
+		# Does not allow shot from midfield
+		# This case returns 0 
+		self.assertEqual(self.eval_shot(0, self.length / 2), self.success)
```
JNeiger: I think this is ok. We cannot score from our side of the field, so we just return a 0% shot chance when we try and evaluate it.

soccer/gameplay/tests/test_shooting.py
```diff
@@ -0,0 +1,99 @@
...
+
+		# Corner shots should not have 100% chance to hit the goal
+		self.assertGreater(self.eval_shot(self.width / 2, self.length), 0.99)
```
JNeiger: Just out of curiosity, what happens when you are just inside the field at the corner? This might be some sort of extreme edge case that isn't accounted for but should be.

soccer/gameplay/tests/test_shooting.py
```diff
@@ -0,0 +1,99 @@
...
+
+	def test_eval_shot_with_bots_and_exclusion(self):
+		their_bot1, their_bot2, their_bot3, their_bot4, their_bot5, their_bot6 = self.their_robots
+		our_bot1, our_bot2, our_bot3, our_bot4, our_bot5, our_bot6 = self.our_robots
```
JNeiger: Same comment as above about having all 6 variables written out for each team

JNeiger:
Looks pretty good! Just a few things here and there

Jason27chan:
@JNeiger 
should num_on_offense in soccer/gameplay/evaluation/opponent.py have an excluded robots parameter?

soccer/gameplay/tests/test_opponent.py
Jason27chan: Should I change this in opponent.py too?

JNeiger: Yeah. It make sense for the normal case to just get the closest opponents

JNeiger:
Sure,  I don't think it's being used right now so it's not a big deal either way

soccer/gameplay/tests/test_shooting.py
Jason27chan: It still returns 100% for a point just inside the field at the corner

chachmu:
@jgkamat @JNeiger Does this look good to merge to you guys?

JNeiger:
I think we fixed the field size mismatch between python and C++. If that's the case, we can move those one or two test cases that were failing back into the normal section and it should be good to go.

ashaw596:
This seems nice to have. You should merge this (eventually)!! =P

JNeiger:
@Jason27chan Can you merge master into this? CircleCI isn't updated in this branch.

Issue #1068, opened 2017-11-28
Improve the Midfielders in Adaptive Formation


JNeiger:
Right now we are using the `SimpleZoneMidfielder` which just keeps the midfielders behind the ball. 

It would be nice to create another midfielder that predicts the likely pass location and moves there before adaptive formation commands them to. Any other ideas for improvement are welcome too

PR #1102, opened 2018-02-21
Adaptive Midfielder Formation Update

tjones320:
Created a new midfielder set up to make the robots in midfield more useful. This sets one robot to go to the best passing position while the other to go to the next best passing position (the best passing position of the location of the current best passing position). Resolve issue #1068


soccer/gameplay/tactics/advance_zone_midfielder.py
```diff
@@ -0,0 +1,138 @@
...
+            self.optimal = 1
+            other = 0
+
+        # sets the second point
+        points[
+            other], value2 = evaluation.passing_positioning.eval_best_receive_point(
+                points[self.optimal],
+                main.our_robots(), AdvanceZoneMidfielder.FIELD_POS_WEIGHTS,
+                AdvanceZoneMidfielder.NELDER_MEAD_ARGS,
+                AdvanceZoneMidfielder.PASSING_WEIGHTS)
```
JNeiger: You should be able to simplify lines 93-114. Let points[0] always be the optimal position and points[1] be the second point.

```
points = [best_pos_from_ball, robocup.Point(0,0)]
points[1], _ = evaluation. ....
```

soccer/gameplay/tactics/advance_zone_midfielder.py
```diff
@@ -0,0 +1,138 @@
...
+
+        # moves the robots
+        for i in range(2):
+            if (self.moves[i] is None):
+                self.moves[i] = skills.move.Move(points[i])
+                self.add_subbehavior(
+                    self.moves[i], self.names[i], required=False, priority=1)
```
JNeiger: I'm not sure if this is needed, but if we do that simplification from above, we can set the optimal point to have a higher priority than the second point

soccer/gameplay/tactics/advance_zone_midfielder.py
```diff
@@ -0,0 +1,138 @@
...
+                self.moves[i].pos = robocup.Point(self.x_positions[i],
+                                                  target_y)
+
+    def execute_passSet(self):
+        #gets the best position to travel to for ball reception
+        best_Pos_From_Ball, value = evaluation.passing_positioning.eval_best_receive_point(
+            main.ball().pos,
+            main.our_robots(), AdvanceZoneMidfielder.FIELD_POS_WEIGHTS,
+            AdvanceZoneMidfielder.NELDER_MEAD_ARGS,
+            AdvanceZoneMidfielder.PASSING_WEIGHTS)
```
JNeiger: Since this is the same call that we are doing in `adaptive_formation` it might be more efficient to just pass it. There should be an example of this type of passing in the defense.py <-> submissive_defender.py interaction that you can look at

soccer/gameplay/tactics/advance_zone_midfielder.py
```diff
@@ -0,0 +1,138 @@
...
+import skills.move
+import evaluation.ball
+import evaluation.passing_positioning
+import evaluation.field
+import evaluation.passing
+import evaluation.shooting
+import functools
+
+
+#2 offenders are related to do a certain event base off of great code
```
JNeiger: ..Typo?

soccer/gameplay/tactics/advance_zone_midfielder.py
```diff
@@ -0,0 +1,138 @@
...
+class AdvanceZoneMidfielder(composite_behavior.CompositeBehavior):
+    # Weights for the general field positioning
+    FIELD_POS_WEIGHTS = (0.01, 3, 0.02)
+    # Weights for finding best pass
+    PASSING_WEIGHTS = (2, 2, 15, 1)
+    # Initial arguements for the nelder mead optimization in passing positioning
+    NELDER_MEAD_ARGS = (robocup.Point(0.5, 2), \
+                        robocup.Point(0.01, 0.01), 1, 2, \
+                        0.75, 0.5, 50, 1, 0.1)
```
JNeiger: We might want to pass these in from `adaptive_formation` too since these constants significantly change the behavior of  the passing positioning method.

soccer/gameplay/tactics/advance_zone_midfielder.py
```diff
@@ -0,0 +1,138 @@
...
+        self.add_state(AdvanceZoneMidfielder.State.passSet,
+                       behavior.Behavior.State.running)
+        #To be imblemented later when wanting to play a more conservative 
+        self.add_state(AdvanceZoneMidfielder.State.passBack,
+                       behavior.Behavior.State.running)
+        # old constants in Simple zone midfieler
+        self.y_hold_percent = .8
+        self.x_positions = [-constants.Field.Width / 3, \
+                            constants.Field.Width / 3]
+        self.x_positions_alternate = [-constants.Field.Width/2, \
```
JNeiger: This isn't being used anywhere

soccer/gameplay/tactics/advance_zone_midfielder.py
```diff
@@ -0,0 +1,138 @@
...
+                                      constants.Field.Width/2]
+        self.names = ['left_mid', 'right_mid']
+        # Which robot would be best to move to first move point
+        self.optimal = 0
```
JNeiger: This isn't going to be used either

JNeiger:
Looking good! I'll probably add some more comments later after I run it

soccer/gameplay/tactics/advance_zone_midfielder.py
JNeiger: It's in the execute and on_enter dribbling state. It's also stored in the class variable self.pass_target. I'll link the line numbers in a second

JNeiger: https://github.com/RoboJackets/robocup-software/blob/master/soccer/gameplay/plays/offense/adaptive_formation.py#L249
Here is where it gets calculated when just dribbling.
https://github.com/RoboJackets/robocup-software/blob/master/soccer/gameplay/plays/offense/adaptive_formation.py#L298
Here is where we calculate it when trying to clear the ball

soccer/gameplay/tactics/advance_zone_midfielder.py
tjones320: There is a problem with having the the optimal point as 0. After one of the robots transition into a passing play both robots then get set to the same optimal point. So it may be better to set the optimal point as 1, and the secondary point as 0.

JNeiger:
```
    PASSING_WEIGHTS = (2, 2, 15, 10)
    NELDER_MEAD_ARGS = (robocup.Point(0.75, 1), \
                        robocup.Point(0.01, 0.01), 1, 1.1, \
                        0.5, 0.9, 100, 1, 0.1)
```

These weights seem to preform somewhat well with the couple test cases I tried.

JNeiger:
Closing in favor of #1135

PR #1111, opened 2018-03-31
Update soccer to use newer common and remove debug communication code

jpfeltracco:
Software and firmware have diverged in their versions of common due to changes in constants by software, and a restructure and clean up in firmware.

This is making it hard to make changes to common that are needed in both software and firmware, specifically sending encoder values back to software (a motion control project) and adding the robot model to common.

This PR recitifes those differences. Note that firmware's version of common deleted both the debug communication header and the status header. Since debug communication is getting replaced by @petersonev by comp, I've deleted the debug communcation header and related code. I've also re-added the status.h header until we decide on what exactly will replace the rtp header (we may be moving back to protobuf).


jpfeltracco:
@jgkamat Any idea what's happening with checkstyle?

jgkamat:
Hmm, there seems to be new git integration in clang-format, making a binary called `git-clang-format`. However, stylize's guess on which clang format to use asks it to use the new file, which dosen't support the same options...

I'm pushing a new commit which will try to resolve this. I'm not sure why it's only happening on your branch though.

soccer/MainWindow.cpp
```diff
@@ -625,8 +594,8 @@ void MainWindow::updateViews() {
             }
             statusWidget->setRobotModel(robotModel);
 
-// uncomment this #define to test the display of a variety of different errors
-// #define DEMO_ROBOT_STATUS
+            // uncomment this #define to test the display of a variety of
+            // different errors #define DEMO_ROBOT_STATUS
```
jgkamat: This is probably unintended (shakes fist at clang)

soccer/Robot.hpp
```diff
@@ -4,30 +4,30 @@
...
+#include "planning/RotationCommand.hpp"
 
+#include <protobuf/Control.pb.h>
 #include <protobuf/RadioRx.pb.h>
 #include <protobuf/RadioTx.pb.h>
-#include <protobuf/Control.pb.h>
 #include <Utils.hpp>
 
+#include <stdint.h>
+#include <Eigen/Dense>
```
ashaw596: This stuff should probably be in the cpp if possible

soccer/Robot.hpp
```diff
@@ -4,30 +4,30 @@
...
 #include "planning/DynamicObstacle.hpp"
+#include "planning/RotationCommand.hpp"
 
+#include <protobuf/Control.pb.h>
 #include <protobuf/RadioRx.pb.h>
 #include <protobuf/RadioTx.pb.h>
-#include <protobuf/Control.pb.h>
 #include <Utils.hpp>
 
+#include <stdint.h>
```
ashaw596: IS this suppose to be here?

ashaw596:
Yeahhh, this stuff should defs be rewritten. THe packets were designed really hacky =P

soccer/Robot.hpp
jpfeltracco: There's a function returning Eigen stuff so it should be in the header, same with stdint.h, there's a function using uint8_t

jpfeltracco:
@petersonev has made good progress on a replacement, it'll just be a little longer before it's in. Is this branch okay to merge? The diff is a little nasty because of clang-format but the changes were pretty minimal.

jpfeltracco:
Cool, merging.

Issue #1112, opened 2018-04-01
Pause before transitioning to another play after capture


jgkamat:
We need to add a pause after we get the ball to migitate issues with the dribbler bouncing. Just waiting a bit will allow the ball to stabilize.

https://github.com/RoboJackets/robocup-software/blob/master/soccer/gameplay/tools/sleep.py will probably be helpful

PR #1113, opened 2018-04-04
Capture delay

Jason27chan:
Addresses #1112
Also fixes test_capture with relation to new goal zone

### DIFF
--- a/soccer/gameplay/skills/capture.py
+++ b/soccer/gameplay/skills/capture.py
@@ -68,7 +68,7 @@ class Capture(single_robot_behavior.SingleRobotBehavior):
             Capture.State.delay,
             Capture.State.fine_approach,
             lambda: not self.bot_near_ball(
-                constants.Robot.Radius + constants.Ball.Radius),
+                constants.Robot.Radius + constants.Ball.Radius * 2),
             'lost ball during delay')
 
         self.add_transition(
--- a/soccer/gameplay/skills/capture.py
+++ b/soccer/gameplay/skills/capture.py
@@ -16,7 +16,7 @@ class Capture(single_robot_behavior.SingleRobotBehavior):
     CourseApproachErrorThresh = 0.8
     CourseApproachDist = 0.4
     CourseApproachAvoidBall = 0.10
-    DelayTime = .25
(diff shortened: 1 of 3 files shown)


soccer/gameplay/skills/capture.py
```diff
@@ -49,9 +54,23 @@ def __init__(self, faceBall=True):
...
             Capture.State.fine_approach,
-            behavior.Behavior.State.completed, lambda: self.bot_near_ball(
+            Capture.State.delay, lambda: self.bot_near_ball(
                 constants.Robot.Radius + constants.Ball.Radius), 'has ball')
 
+        self.add_transition(
+            Capture.State.delay, 
+            behavior.Behavior.State.completed, 
+            lambda: time.time() - self.start_time > Capture.DelayTime and self.bot_near_ball(
+                constants.Robot.Radius + constants.Ball.Radius),
```
joshhting: Would using has_ball() make more sense here than a proximity check? Since has_ball() checks if the ball is actually detected by the mouth of the robot.

soccer/gameplay/skills/capture.py
```diff
@@ -49,9 +54,23 @@ def __init__(self, faceBall=True):
...
+            'delay before finish')
+
+        self.add_transition(
+            Capture.State.delay,
+            Capture.State.fine_approach,
+            lambda: not self.bot_near_ball(
+                constants.Robot.Radius + constants.Ball.Radius * 2),
```
joshhting: Same here

soccer/gameplay/skills/capture.py
jgkamat: the ball sensor is actually quite unreliable on the robots currently (the testing bot we have has it always on). We really need to build a more generic function which uses ball sense only if it's available, and disables using it if we think it's broken.

joshhting: At the very least, I think we should be checking if the robot is facing the ball here, not just if the ball is touching the robot

jgkamat:
LGTM, so I'm merging now. Let me know if anyone spots any issues.

PR #1116, opened 2018-04-11
Update rc-fshare, remove pid tuner temporarily.

jpfeltracco:
Firmware can't compile with the pid tuner. The pid tuner was also owned by the pid class, which prevents the pid tuner from just being moved to soccer. @chachmu will refactor to re-enable the pid tuner functionality, but for right now it has been ripped out.


jpfeltracco:
@chachmu Just as a sanity check, can I merge this?

chachmu:
LGTM 👍

PR #1124, opened 2018-04-22
Add better error reporting for role assignment

### LINKED ISSUE #1123: Better role assignment debugging
We got a role assignment error recently:
```
ERROR:root:Failed Assignment:
Robots:
	<Robot us[2], pos=Point(-1.02215, 1.46373)>
	<Robot us[6], pos=Point(0.474425, 2.44997)>
Roles:
	has_ball=False; pos=Point(-1.45546, 1.03075); chip_pref=0; prev_id=6; required=True; can_kick=True
	has_ball=False; pos=Point(-0.933876, 1.56105); chip_pref=0; prev_id=2; required=True; can_kick=False
ERROR:root:Unable to satisfy role assignment constraints.  Dropping and temp. blacklisting current play...
```

I'm not sure why it wasn't able to full-fill that constraint, we need better debugging output at least telling us constraints that weren't able to be fulfilled.

jgkamat:
Closes #1123 

I haven't tested this too well, although I think this should be pretty good.

### DIFF
--- a/soccer/gameplay/role_assignment.py
+++ b/soccer/gameplay/role_assignment.py
@@ -21,7 +21,7 @@ class RoleRequirements:
 
         # multiply this by the distance between two points to get the cost
         self.position_cost_multiplier = 1.0
-        
+
         # A lambda function property that allows customization of cost
         # Has exactly one parameter, which is a robot
         self.cost_func = lambda r: 0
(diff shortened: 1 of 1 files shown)


jgkamat:
I'll merge this in a couple days if no one gets a chance to review this, since I think it's good :cry: 


I want to use this to possibly squash a few role req bugs.

ashaw596:
Seems fine just looking at it. Might be a bit confusing since it might print stuff for every robot tried in each role, but its debugging. Slight possibility too many string operations slowing down runtime speed since python strings are immutable i guess.

jgkamat:
Yah, I agree it's a bit annoying that it prints out all the robot's messages, but we need all the robot's fail reasons to know why the overall play failed. I'm not sure of a better solution than that.

I don't think there's an easy way to get mutable strings in python, so I don't think we should worry about that unless we notice any perf issues (I haven't so far :P)

ashaw596:
lgtm

jgkamat:
Ok, merging now, let me know if anyone sees any issues or dosen't like the output format :)

PR #1127, opened 2018-05-13
Fix initialization of simradio on yellow for a frame

### LINKED ISSUE #1126: Can't bind to to the %%%% team's radio port
We have this problem when trying to run an instance of soccer -sim -b, but then it says:

```"what():  Can't bind to the yellow team's radio port."```

image

jgkamat:
Closes #1126, very simple.

@chachmu, can I get a quick review? :)

### DIFF
--- a/soccer/Processor.cpp
+++ b/soccer/Processor.cpp
@@ -14,7 +14,6 @@
 #include <Robot.hpp>
 #include <RobotConfig.hpp>
 #include <Utils.hpp>
-#include <rc-fshare/git_version.hpp>
 #include <joystick/GamepadController.hpp>
 #include <joystick/GamepadJoystick.hpp>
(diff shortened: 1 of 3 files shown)


jgkamat:
Merging now, let me know if anyone sees any issues. :)

jgkamat:
@maxaubel, can you ensure that your grSim broadcast IP is set to the proper multicast address:

https://github.com/RoboJackets/robocup-software/blob/master/makefile#L49-L51

We set our fork to broadcast on localhost, which prevents multicast from going over the network (so many people can debug at the same time), but prevents multicast from working with two clients.

Otherwise, can you screenshot your grSim settings?  Here are mine:

2018-05-15-235349_329x368_scrot

PR #1135, opened 2018-06-05
Advance Midfielder

tjones320:
This new tactic builds off of adaptive formation and moves the non-ball carriers to move to an ideal ball reception position and a follow up reception position.
Resolve issue #1068

For QA this method will move the other midfielder to another viable pass position out of the way from the current robot we want to pass to goal. Please allow this program to run and determine if this effectively makes shots on goal and if there are no clear cases where robots get stuck (ideally something repeatable). Also please look at the code and if it's readable and works to purpose.

### DIFF
--- a/soccer/gameplay/plays/offense/adaptive_formation.py
+++ b/soccer/gameplay/plays/offense/adaptive_formation.py
@@ -10,6 +10,7 @@ import evaluation.passing_positioning
 import tactics.coordinated_pass
 import tactics.defensive_forward
 import tactics.simple_zone_midfielder
+import tactics.advance_zone_midfielder
 import skills.move
 import skills.capture
 
@@ -94,54 +95,51 @@ class AdaptiveFormation(standard_play.StandardPlay):
                             AdaptiveFormation.State.collecting, lambda: True,
                             'immediately')
 
-        self.add_transition(
-            AdaptiveFormation.State.collecting,
-            AdaptiveFormation.State.dribbling,
-            lambda: self.subbehavior_with_name('defend').state == behavior.Behavior.State.completed,
-            'Ball Collected')
+        self.add_transition(AdaptiveFormation.State.collecting,
+                            AdaptiveFormation.State.dribbling, lambda: self.
+                            subbehavior_with_name('defend').state == behavior.
+                            Behavior.State.completed, 'Ball Collected')
 
         self.add_transition(
             AdaptiveFormation.State.dribbling, AdaptiveFormation.State.passing,
-            lambda: self.dribbler_has_ball() and self.should_pass_from_dribble() and not self.should_shoot_from_dribble(),
-            'Passing')
+            lambda: self.dribbler_has_ball() and self.should_pass_from_dribble(
+            ) and not self.should_shoot_from_dribble(), 'Passing')
 
         self.add_transition(
             AdaptiveFormation.State.dribbling,
-            AdaptiveFormation.State.shooting,
-            lambda: self.dribbler_has_ball() and self.should_shoot_from_dribble(),
-            'Shooting')
+            AdaptiveFormation.State.shooting, lambda: self.dribbler_has_ball(
+            ) and self.should_shoot_from_dribble(), 'Shooting')
 
         self.add_transition(
             AdaptiveFormation.State.dribbling,
-            AdaptiveFormation.State.clearing,
-            lambda: self.dribbler_has_ball() and self.should_clear_from_dribble() and not self.should_pass_from_dribble() and not self.should_shoot_from_dribble(),
-            'Clearing')
+            AdaptiveFormation.State.clearing, lambda: self.dribbler_has_ball(
+            ) and self.should_clear_from_dribble(
+            ) and not self.should_pass_from_dribble(
+            ) and not self.should_shoot_from_dribble(), 'Clearing')
 
         self.add_transition(
             AdaptiveFormation.State.passing, AdaptiveFormation.State.dribbling,
-            lambda: self.subbehavior_with_name('pass').state == behavior.Behavior.State.completed,
-            'Passed')
+            lambda: self.subbehavior_with_name(
+                'pass').state == behavior.Behavior.State.completed, 'Passed')
 
         # Reset to collecting when ball is lost at any stage
         self.add_transition(AdaptiveFormation.State.dribbling,
-                            AdaptiveFormation.State.collecting,
-                            lambda: not self.dribbler_has_ball(),
-                            'Dribble: Ball Lost')
+                            AdaptiveFormation.State.collecting, lambda:
+                            not self.dribbler_has_ball(), 'Dribble: Ball Lost')
         self.add_transition(
             AdaptiveFormation.State.passing,
-            AdaptiveFormation.State.collecting,
-            lambda: self.subbehavior_with_name('pass').state == behavior.Behavior.State.cancelled or self.subbehavior_with_name('pass').state == behavior.Behavior.State.failed,
-            'Passing: Ball Lost')
-        self.add_transition(
-            AdaptiveFormation.State.shooting,
-            AdaptiveFormation.State.collecting,
-            lambda: self.subbehavior_with_name('kick').is_done_running(),
-            'Shooting: Ball Lost / Shot')
-        self.add_transition(
-            AdaptiveFormation.State.clearing,
-            AdaptiveFormation.State.collecting,
-            lambda: self.subbehavior_with_name('clear').is_done_running(),
-            'Clearing: Ball Lost')
+            AdaptiveFormation.State.collecting, lambda: self.
+            subbehavior_with_name('pass').state == behavior.Behavior.State.
+            cancelled or self.subbehavior_with_name('pass').state == behavior.
+            Behavior.State.failed, 'Passing: Ball Lost')
+        self.add_transition(AdaptiveFormation.State.shooting,
+                            AdaptiveFormation.State.collecting, lambda: self.
+                            subbehavior_with_name('kick').is_done_running(),
+                            'Shooting: Ball Lost / Shot')
+        self.add_transition(AdaptiveFormation.State.clearing,
+                            AdaptiveFormation.State.collecting, lambda: self.
+                            subbehavior_with_name('clear').is_done_running(),
+                            'Clearing: Ball Lost')
 
     @classmethod
     def score(cls):
@@ -207,8 +205,8 @@ class AdaptiveFormation(standard_play.StandardPlay):
         return True
 
     def dribbler_has_ball(self):
-        return any(evaluation.ball.robot_has_ball(r)
-                   for r in main.our_robots())
+        return any(
+            evaluation.ball.robot_has_ball(r) for r in main.our_robots())
 
     def on_enter_collecting(self):
         self.remove_all_subbehaviors()
@@ -227,8 +225,8 @@ class AdaptiveFormation(standard_play.StandardPlay):
         # Dribbles toward the best receive point
 
         self.dribbler.pos, _ = evaluation.passing_positioning.eval_best_receive_point(
-            main.ball().pos, main.our_robots(),
-            AdaptiveFormation.FIELD_POS_WEIGHTS,
+            main.ball().pos,
+            main.our_robots(), AdaptiveFormation.FIELD_POS_WEIGHTS,
             AdaptiveFormation.NELDER_MEAD_ARGS,
             AdaptiveFormation.DRIBBLING_WEIGHTS)
 
@@ -237,20 +235,24 @@ class AdaptiveFormation(standard_play.StandardPlay):
         self.check_dribbling_timer = 0
 
         if (not self.has_subbehavior_with_name('midfielders')):
-            self.midfielders = tactics.simple_zone_midfielder.SimpleZoneMidfielder(
+            self.midfielders = tactics.advance_zone_midfielder.AdvanceZoneMidfielder(
             )
-            self.add_subbehavior(self.midfielders,
-                                 'midfielders',
-                                 required=False,
-                                 priority=10)
+            self.add_subbehavior(
+                self.midfielders, 'midfielders', required=False, priority=10)
+        self.midfielders.dribble_state = True
+        self.midfielders.kick = False
 
     def execute_dribbling(self):
         # Grab best pass
         self.pass_target, self.pass_score = evaluation.passing_positioning.eval_best_receive_point(
-            main.ball().pos, main.our_robots(),
-            AdaptiveFormation.FIELD_POS_WEIGHTS,
+            main.ball().pos,
+            main.our_robots(), AdaptiveFormation.FIELD_POS_WEIGHTS,
             AdaptiveFormation.NELDER_MEAD_ARGS,
             AdaptiveFormation.PASSING_WEIGHTS)
+
+        self.midfielders.dribble_state = True
+        self.midfielders.kick = False
+        self.midfielders.passing_point = self.pass_target
 
         # Grab shot chance
         self.shot_chance = evaluation.shooting.eval_shot(main.ball().pos)
@@ -260,8 +262,8 @@ class AdaptiveFormation(standard_play.StandardPlay):
         if (self.check_dribbling_timer > self.check_dribbling_timer_cutoff):
             self.check_dribbling_timer = 0
             self.dribbler.pos, _ = evaluation.passing_positioning.eval_best_receive_point(
-                main.ball().pos, main.our_robots(),
-                AdaptiveFormation.FIELD_POS_WEIGHTS,
+                main.ball().pos,
+                main.our_robots(), AdaptiveFormation.FIELD_POS_WEIGHTS,
                 AdaptiveFormation.NELDER_MEAD_ARGS,
                 AdaptiveFormation.DRIBBLING_WEIGHTS)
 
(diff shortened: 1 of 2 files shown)


guyfleeman:
Does this go with #1102 ?

JNeiger:
@guyfleeman  Yeah, I think so. This PR is better though since it doesn't have all those extra files included. I'll close the other one

soccer/gameplay/plays/offense/adaptive_formation.py
```diff
@@ -10,6 +10,7 @@
 import tactics.coordinated_pass
 import tactics.defensive_forward
 import tactics.simple_zone_midfielder
```
joshhting: Remove line

soccer/gameplay/tactics/advance_zone_midfielder.py
```diff
@@ -0,0 +1,174 @@
...
+        x_temp_hold = constants.Field.Width / 3
+
+        self.hold_point = [robocup.Point(x_temp_hold,y_temp_hold), robocup.Point(x_temp_hold,y_temp_hold)]
+
+        self.dribble_state = True
+
+        self.kick = False
+
+        for s in AdvanceZoneMidfielder.State:
+            self.add_state(s, behavior.Behavior.State.running)
```
joshhting: Isn't this already being done at the beginning of this method?

soccer/gameplay/tactics/advance_zone_midfielder.py
```diff
@@ -0,0 +1,174 @@
...
+
+        self.names = ['left','right']
+        #intially should be passSet method
+        self.add_transition(behavior.Behavior.State.start,
+                            AdvanceZoneMidfielder.State.passSet, lambda: True,
+                            "Immediately")
+        self.add_transition(AdvanceZoneMidfielder.State.passSet,
+                            AdvanceZoneMidfielder.State.hold, lambda: False,
```
JNeiger: I assume this is a holdover from testing and should actually be using some real lambda function here. Otherwise it will only be in the passSet state

soccer/gameplay/tactics/advance_zone_midfielder.py
```diff
@@ -0,0 +1,174 @@
...
+            self.other = 0
+        else :
+            points = [robocup.Point(0, 0), robocup.Point(0, 0)]
+            self.optimal = 1
+            # sets either the left or the right side of the ball to the best ball position
+            # and then notes which one is the other
+            # left side of the ball
+            points[self.optimal] = best_Pos_From_Ball
+            self.other = 0
+        """
```
JNeiger: This can be removed if you don't need the commented stuff

soccer/gameplay/tactics/advance_zone_midfielder.py
```diff
@@ -0,0 +1,174 @@
...
+        for i in range(2):
+            if (self.moves[i] is None):
+                self.moves[i] = skills.move.Move(points[i])
+                self.add_subbehavior(
+                    self.moves[i], self.names[i], required=False, priority= self.priorities[i])
+            else:
+                self.moves[i].pos = points[i]
+
+    def on_enter_hold(self):
+        print("kicking")
```
JNeiger: This can be removed

soccer/gameplay/tactics/advance_zone_midfielder.py
```diff
@@ -0,0 +1,174 @@
...
+            # and then notes which one is the other
+            # left side of the ball
+            points[self.optimal] = best_Pos_From_Ball
+            self.other = 0
+        """
+        points = [robocup.Point(0, 0), robocup.Point(0, 0)]
+        points[0] = best_Pos_From_Ball;
+
+        # sets the second point
+        pointer, value2 = evaluation.passing_positioning.eval_best_receive_point(
```
JNeiger: You should be able to just do `points[1], value2 = evaluation.passing_positioning.....`

soccer/gameplay/tactics/advance_zone_midfielder.py
```diff
@@ -0,0 +1,174 @@
...
+    def on_enter_hold(self):
+        print("kicking")
+
+    def execute_hold(self):
+
+        y_temp_hold = 0.8 * self.passing_point.y
+
+        x_temp_hold = constants.Field.Width / 3
+
+        self.hold_point = [robocup.Point(x_temp_hold,y_temp_hold), robocup.Point(x_temp_hold,y_temp_hold)]
```
JNeiger: These are the same definitions that are in the constructor. I would rather remove these and leave the ones in the constructor if they are just going to be static variables.

soccer/gameplay/tactics/advance_zone_midfielder.py
```diff
@@ -0,0 +1,174 @@
...
+        self._dribble_state = value
+
+    @property
+    def kick(self):
+        return self._kick
+
+    @kick.setter
+    def kick(self, value):
+        self._kick = value
+        print("Change Value to kick on")
```
JNeiger: This print statement can be removed

soccer/gameplay/tactics/advance_zone_midfielder.py
```diff
@@ -0,0 +1,174 @@
...
+                            "Immediately")
+        self.add_transition(AdvanceZoneMidfielder.State.passSet,
+                            AdvanceZoneMidfielder.State.hold, lambda: False,
+                            "When a kick play begins")
+        self.add_transition(AdvanceZoneMidfielder.State.hold,
+                            AdvanceZoneMidfielder.State.passSet, lambda: True,
+                            "When not in a kick play")
+
+    def on_enter_passSet(self):
+        print("Entering PassSet");
```
JNeiger: This can be removed

soccer/gameplay/tactics/advance_zone_midfielder.py
```diff
@@ -0,0 +1,174 @@
...
+            points = [robocup.Point(0, 0), robocup.Point(0, 0)]
+            self.optimal = 1
+            # sets either the left or the right side of the ball to the best ball position
+            # and then notes which one is the other
+            # left side of the ball
+            points[self.optimal] = best_Pos_From_Ball
+            self.other = 0
+        """
+        points = [robocup.Point(0, 0), robocup.Point(0, 0)]
+        points[0] = best_Pos_From_Ball;
```
JNeiger: You should be able to just initialize points with `self.passing_point` already in the first position. This will remove the need for `best_Pos_From_Ball`

soccer/gameplay/tactics/advance_zone_midfielder.py
```diff
@@ -0,0 +1,174 @@
...
+#2 midfielder rely on the future location of their teammate to pass quickly
+class AdvanceZoneMidfielder(composite_behavior.CompositeBehavior):
+    # Weights for the general field positioning
+    FIELD_POS_WEIGHTS = (0.01, 3, 0.02)#AdaptiveFormation.FIELD_POS_WEIGHTS
+    # Weights for finding best pass
+    PASSING_WEIGHTS = (2, 2, 15, 10)#AdaptiveFormation.PASSING_WEIGHTS
+    # Initial arguements for the nelder mead optimization in passing positioning
+    NELDER_MEAD_ARGS = (robocup.Point(0.75, 1), \
+                        robocup.Point(0.01, 0.01), 1, 1.1, \
+                        0.5, 0.9, 100, 1, 0.1)#AdaptiveFormation.NELDER_MEAD_ARGS
```
jgkamat: I don't care what the style thing says please put a space or newline after the `)` :P

soccer/gameplay/tactics/advance_zone_midfielder.py
```diff
@@ -0,0 +1,174 @@
...
+
+
+#2 midfielder rely on the future location of their teammate to pass quickly
+class AdvanceZoneMidfielder(composite_behavior.CompositeBehavior):
+    # Weights for the general field positioning
+    FIELD_POS_WEIGHTS = (0.01, 3, 0.02)#AdaptiveFormation.FIELD_POS_WEIGHTS
+    # Weights for finding best pass
+    PASSING_WEIGHTS = (2, 2, 15, 10)#AdaptiveFormation.PASSING_WEIGHTS
+    # Initial arguements for the nelder mead optimization in passing positioning
+    NELDER_MEAD_ARGS = (robocup.Point(0.75, 1), \
```
jgkamat: `\` is unneeded inside parens

soccer/gameplay/tactics/advance_zone_midfielder.py
```diff
@@ -0,0 +1,174 @@
...
+import evaluation.passing
+import evaluation.shooting
+import functools
+import plays.offense.adaptive_formation
+
+
+#2 midfielder rely on the future location of their teammate to pass quickly
+class AdvanceZoneMidfielder(composite_behavior.CompositeBehavior):
+    # Weights for the general field positioning
+    FIELD_POS_WEIGHTS = (0.01, 3, 0.02)#AdaptiveFormation.FIELD_POS_WEIGHTS
```
jgkamat: This is just the same as `AdaptiveFormation.FIELD_POS_WEIGHTS`, can't you copy it in here?

I'd rather have one source for constants and tweak them if needed (unless the number is completely different).

soccer/gameplay/tactics/advance_zone_midfielder.py
```diff
@@ -0,0 +1,174 @@
...
+
+        self.add_state(AdvanceZoneMidfielder.State.hold, 
+                        behavior.Behavior.State.running)
+
+        self.moves = [None, None]
+
+
+        self.priorities = [1, 2]
+
+        self.passing_point = main.ball().pos
```
jgkamat: sometimes, the ball position is unknown. To avoid crashing when it is, wrap this in `if main.ball().valid:` with a backup.

Maybe we can just wait, worst case, we can cancel the play.

soccer/gameplay/tactics/advance_zone_midfielder.py
```diff
@@ -0,0 +1,174 @@
...
+
+        y_temp_hold = 0.8 * self.passing_point.y
+
+        x_temp_hold = constants.Field.Width / 3
+
+        self.hold_point = [robocup.Point(x_temp_hold,y_temp_hold), robocup.Point(x_temp_hold,y_temp_hold)]
```
jgkamat: Why make a list with two identical values inside of it? I'm a bit confused here.

soccer/gameplay/tactics/advance_zone_midfielder.py
```diff
@@ -0,0 +1,174 @@
...
+        self.names = ['left','right']
+        #intially should be passSet method
+        self.add_transition(behavior.Behavior.State.start,
+                            AdvanceZoneMidfielder.State.passSet, lambda: True,
+                            "Immediately")
+        self.add_transition(AdvanceZoneMidfielder.State.passSet,
+                            AdvanceZoneMidfielder.State.hold, lambda: False,
+                            "When a kick play begins")
+        self.add_transition(AdvanceZoneMidfielder.State.hold,
+                            AdvanceZoneMidfielder.State.passSet, lambda: True,
```
jgkamat: This lambda should be real too, piggybacking from above.

soccer/gameplay/tactics/advance_zone_midfielder.py
```diff
@@ -0,0 +1,174 @@
...
+        # sets the second point
+        pointer, value2 = evaluation.passing_positioning.eval_best_receive_point(
+                best_Pos_From_Ball,
+                main.our_robots(), AdvanceZoneMidfielder.FIELD_POS_WEIGHTS,
+                AdvanceZoneMidfielder.NELDER_MEAD_ARGS,
+                AdvanceZoneMidfielder.PASSING_WEIGHTS)
+        
+        points[1] = pointer
+        # moves the robots
+        for i in range(2):
```
jgkamat: I really hate hard coding this. Maybe you can use `len(moves)` or define a num of robots contstants.

soccer/gameplay/tactics/advance_zone_midfielder.py
```diff
@@ -0,0 +1,174 @@
...
+                self.add_subbehavior(self.moves[i],
+                                     self.names[i],
+                                     required=False,
+                                     priority=1)
+            else:
+                self.moves[i].pos = self.hold_point[i]
+
+
+
+    @property
```
jgkamat: You don't actually need all this property stuff, if you want to simplify stuff, we can set kick directly. I think you should just be able to remove the property and setter for all these variables and everything should just work.

these are mainly useful only if we want to change how a variable is get/set without using functions.

soccer/gameplay/tactics/advance_zone_midfielder.py
tjones320: I switched it over to actually holding during a kick. I'm not sure this is a desired behavior. But this should prevent us going behind the goal during a kick and getting in the way of the kick.

soccer/gameplay/tactics/advance_zone_midfielder.py
tjones320: Removed

soccer/gameplay/tactics/advance_zone_midfielder.py
tjones320: removed the method.

soccer/gameplay/tactics/advance_zone_midfielder.py
tjones320: one was intended to be a negative my error

soccer/gameplay/tactics/advance_zone_midfielder.py
tjones320: I need it to receive a already calculated best point from adaptive formation. It greatly increases the efficiency of the class. I also need to know whether we are kicking to stop the passSet otherwise sometimes the best location to pass is directly in front of the clear kick to goal.

tjones320:
Please reexamine the new code

JNeiger:
Sorry about that, I got caught up in work. Overall, it looks good. In terms of behavior, there are 2 cases that may cause some problems. 

The first is that sometimes two robots are still going to the same spot. It may be beneficial to build in some sort of minimum distance between the current pass receiver and the next pass receiver. I only had it happen once so it may not be that reproducible.

This second happened quite a bit. When we are near the opposing goal, it would be normal to just take a shot after receiving the ball. Due to the way our scoring function works, the best pass position is directly between the receiving robot and the goal, blocking the shot. (See picture). (This may not even be that big of deal, someone else should chime in)

You could fix it a few different ways. One of the easiest would be implementing a different behavior when the shot chance or y value of the first pass receiver hits a certain cutoff.

image

tjones320:
Thanks JNeiger. That's what hold was intended to fix, I didn't see this issue when running it on the computers by the field but now that i'm running it on my laptop I see the issue. I'll try to modify the hold command to actually move out of the way during a kick.

soccer/gameplay/tactics/advance_zone_midfielder.py
tjones320: Resolved

soccer/gameplay/plays/offense/adaptive_formation.py
tjones320: Resolved

soccer/gameplay/tactics/advance_zone_midfielder.py
```diff
@@ -0,0 +1,182 @@
...
+                self.add_subbehavior(
+                    self.moves[i],
+                    self.names[i],
+                    required=False,
+                    priority=self.priorities[i])
+            else:
+                self.moves[i].pos = points[i]
+
+    def execute_hold(self):
+        # old methodology for simple zone midfielder (check comments on that method)
```
JNeiger: Where are the comments?

soccer/gameplay/tactics/advance_zone_midfielder.py
```diff
@@ -0,0 +1,182 @@
...
+        return self._kick
+
+    @kick.setter
+    def kick(self, value):
+        self._kick = value
+
+    def on_exit_passSet(self):
+        self.remove_all_subbehaviors()
+
+    def on_exit_hold(self):
```
JNeiger: This might be super trivial, but it looks better when reading code to have all the setter and getter's below almost all the other code. I also personally like having the functions that we use be directly below the functions they are being called. So if function A calls function B which calls function C. The order would be A, B ,C in the file.

JNeiger:
Make sure to unhide my comments from today. I did them all before the last commit that changed a few things.

soccer/gameplay/tactics/advance_zone_midfielder.py
```diff
@@ -0,0 +1,186 @@
...
+        self.remove_all_subbehaviors()
+
+    # this method determines if the 2nd midfielders kicking point is actually in the way of the potential shot
+    def in_shot_triangle(self, best_point, alt_point):
+        # get the two points of the enemies goal
+        goalSegment = constants.Field.TheirGoalSegment
+        goalCenter = goalSegment.center()
+        length = goalSegment.length()
+        right_post = robocup.Point(goalCenter.x + length / 2, goalCenter.y)
+        left_post = robocup.Point(goalCenter.x - length / 2, goalCenter.y)
```
JNeiger: You can do goalSegment.pt[0] and goalSegment.pt[1] here instead of the goalCenter + length/2

soccer/gameplay/tactics/advance_zone_midfielder.py
```diff
@@ -0,0 +1,186 @@
...
+            self.closest_post = right_post
+        # if interior angles near sum to 0 then robot is inside the the zone. 
+        return abs(shot_angle - left_post_alt_pos_angle - right_post_alt_pos_angle) < self.O_of_h
+
+    # finds closest point to leave zone and goes four robot radius outside the zone
+    def remove_obstruction(self, best_point, alt_point):
+        # use line projection
+        post_to_best_vector = best_point - self.closest_post
+        best_to_alt_vector = best_point - alt_point
+        newCoef = best_to_alt_vector.dot(post_to_best_vector) / post_to_best_vector.dot(post_to_best_vector)
```
JNeiger: What's newCoef represent?

soccer/gameplay/tactics/advance_zone_midfielder.py
tjones320: Moved it to the end

PR #1141, opened 2018-06-08
bumped y_start value to avoid weird path planning in repeated lineups

guyfleeman:
The bottom bot on lineups barely intersects with the new defense area, causing vertical replans that are pretty gross. I didn't see this fixed in master so I bumped the y_start so the robots stay in a line (at least in sim).

### DIFF
--- a/soccer/gameplay/plays/testing/repeated_line_up.py
+++ b/soccer/gameplay/plays/testing/repeated_line_up.py
@@ -38,7 +38,7 @@ class RepeatedLineUp(play.Play):
     def generate_line(self, x_multiplier):
         x = (constants.Field.Width / 2 - constants.Robot.Radius * 2
              ) * x_multiplier
-        y_start = 1.0
+        y_start = 1.2
         line = robocup.Segment(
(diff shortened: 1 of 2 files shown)


guyfleeman:
how do I pull the style patch file from circleci?

ashaw596:
@jgkamat circleci says?
[1 / 2] files need formatting
Writing patch to file: '/tmp/clean.patch'
makefile:155: recipe for target 'checkstyle' failed
make: *** [checkstyle] Error 1

jgkamat:
> how do I pull the style patch file from circleci?

You can find the patch files under the artifacts tab in circleci, for this
build it is:

https://5365-15662105-gh.circle-artifacts.com/0/tmp/clean.patch

I don't think the url is stable anymore, so we can't guess it during build
(before the artifact is uploaded).

I don't really care about style when things look OK already (esp for something
as trivial as this) so I'm fine with just leaving it :P

guyfleeman:
We just got all bots up, so will test soon.

PR #1143, opened 2018-06-10
Offense and Defense in Basic122 will both go for the ball

chachmu:
The offensive robots in Basic122 don't communicate at all with defense so sometimes the defense robots will try to clear the ball while the offensive robots are trying to pivot kick it resulting in them fighting over it.

### DIFF
--- a/soccer/gameplay/plays/offense/basic_122.py
+++ b/soccer/gameplay/plays/offense/basic_122.py
@@ -7,6 +7,7 @@ import robocup
 import evaluation
 import constants
 import math
+from enum import Enum
 
 
 class Basic122(standard_play.StandardPlay):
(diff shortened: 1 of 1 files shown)


chachmu:
@jgkamat Should I add a state to pivot kick that has the robot get out of the way? That seems really jank but it would probably at least work for during comp :/

jgkamat:
>          # make the kicker try again if it already kicked
> -        if not striker.is_in_state(behavior.Behavior.State.running):
> +        if self.subbehavior_with_name("defense").state == tactics.defense.Defense().State.clearing:
subbehavior_with_name could return None, we should check for that here.
> +            striker.restart()
> +        elif not striker.is_in_state(behavior.Behavior.State.running):
This could just be an or in the above statement, right?
>              striker.restart()

> @jgkamat Should I add a state to pivot kick that has the robot get out of
> the way? That seems really jank but it would probably at least work for
> during comp :/

No, that would be really bad, pivot kick shouldn't handle that at all :P

Instead of having the robot sitting around, we could assign it a marking
position (so we have 3 marking robots) or assign another robot to the defense.
Unfortunately, it looks like basic122 has 2 markers pretty hard-coded.

It might be easier to have the robot move up the field (with a move
subbheavior), or assign another robot to the defense temporarily.

I'm fine with having the robot sit around for a couple seconds as well, it
shouldn't hurt too much.

In my tests at least, the robot does get out of the way currently, because
pivot kick lays an obstacle so our robots try to avoid the shot.

chachmu:
I'm probably just going to add a new behavior state to Basic122 to get out of the way if the defense is clearing

jgkamat:
> I'm probably just going to add a new behavior state to Basic122 to get out
> of the way if the defense is clearing

Hmm, I really don't think we should be sitting around doing nothing during the
new clear state. Maybe we could just remove the striker or something? We don't
want to loose the blockers as well, right?

chachmu:
AFAIK the support robots just position themselves up near the opponents goal, I don't think they really block anything and the defenders should keep blocking even during the clear state.

soccer/gameplay/plays/offense/basic_122.py
```diff
@@ -169,5 +189,12 @@ def execute_running(self):
 
             # TODO: this would be a good place for a "keep-trying container behavior"
             # make the kicker try again if it already kicked
+        if self.subbehavior_with_name("defense").state == tactics.defense.Defense().State.clearing:
+            print("clearing")
```
JNeiger: x

Issue #1138, opened 2018-06-06
Jiggling on the defense line


Jason27chan:
Robots on defense are jiggling because they're not allowed in the goal zone. Should be easy to offset robots by a small distance

jgkamat:
I think that the fix for #1098 was not fully tested. I think that if we bump
the range a bit more this should go away, but the problem could be deeper.

Jason Chan writes:

Jason27chan:
we're finding a problem where robots are path-planning through each other.

joshhting:
Apologies for not looking at this sooner, I'll tackle it this afternoon

Jason27chan:
Is this still a problem @chachmu?

chachmu:
@Jason27chan Probably, I haven't seen it quite as much since we improved motion control and better calibrated vision but we haven't really run the robots enough to say it is gone for sure.

PR #1145, opened 2018-06-15
Offset defenders from goal zone to prevent jiggling

joshhting:
Address #1138. I'm using 0.1 as a fudge number for now, it looks ok on sim but should still be tested irl.

### DIFF
--- a/soccer/gameplay/tactics/positions/submissive_defender.py
+++ b/soccer/gameplay/tactics/positions/submissive_defender.py
@@ -95,6 +95,7 @@ class SubmissiveDefender(
             threat_point = self._block_line.get_pt(0)
 
             intersection_center = top_seg.line_intersection(self._block_line)
+            intersection_center += robocup.Point(0, 0.1)  # offset from goal zone
 
             if threat_point.x < 0:
                 intersections_left = left_seg.line_intersection(self._block_line)


jgkamat:
Josh Ting writes:


> @@ -95,6 +95,7 @@ class SubmissiveDefender(
>              threat_point = self._block_line.get_pt(0)
>
>              intersection_center = top_seg.line_intersection(self._block_line)
> +            intersection_center += robocup.Point(0, 0.1)  # offset from goal zone

intersection_center can be None if the lines don't intersect, we should add
an if here to prevent a crash.

Also, if this is an issue with running into the obstacle (I think it is),
maybe we should add an offset of a robot radius (plus a bit more) here:

>         top_seg = robocup.Segment(
>            robocup.Point(
>                -constants.Field.PenaltyLongDist / 2,
>                constants.Field.PenaltyShortDist),
>            robocup.Point(
>                constants.Field.PenaltyLongDist / 2,
>                constants.Field.PenaltyShortDist))


>
>              if threat_point.x < 0:
>                  intersections_left = left_seg.line_intersection(self._block_line)

I'm now wondering if we need to offset the sides as well (in which case we
need to offset {left,right}_seg in a similar way. This dosen't fix the
jiggling completely for me (when two robots are close to each other), but we
probably should have realized that we were planning too close to the goal zone
obstacle earlier on :).

Otherwise, LGTM, feel free to merge once you get it done!

jgkamat:
LGTM but I haven't tested it :P

PR #1149, opened 2018-06-16
Robot_Has_Ball improvements

JNeiger:
It uses a triangle now instead of whatever it was before


JNeiger:
Closing since it's already in comp2018

PR #1171, opened 2018-06-27
Add ability for soccer to read logs

jgkamat:
This is pretty rough, still a WIP, but I think it's most of the way there.

This allows us to use the normal soccer interface to read logs, and hopefully can replace the external log viewer. This means we get a lot of things like shortcuts and drawing debug symbols for free.


jgkamat:
For me, a command like `./run/soccer -vlog run/logs/20181028-114754.log` works. When you see that error, do you get a stack trace or something that could pin it to some line in our code (without that I don't think I can do much) :P

Jason27chan:
JK I forgot to make the change to the path that I commented on. It's running right now, but no robots are showing up, so I dont know if the log is running

PR #1172, opened 2018-06-27
Remove Model reset from play reloading

jgkamat:
According to [this doc page](http://doc.qt.io/qt-5/qabstractitemmodel.html#modelReset) the modelReset signal shouldn't be fired by us. Additionally it collapses the treeview and dosen't seem to do anything else (from my tests at least).

I also fixed a couple crashes while using flycheck (due to reloading classes that already exist twice), by adding the ability to easily blacklist files to the play reloading system.

### DIFF
--- a/soccer/gameplay/main.py
+++ b/soccer/gameplay/main.py
@@ -112,6 +112,11 @@ def init(log_errors=True):
                         # FIXME: this logic should go inside the play_registry
                         play_reg_node = _play_registry.node_for_module_path(
                             module_path[1:])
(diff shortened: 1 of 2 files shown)


chachmu:
@jgkamat Do you want me to review and merge this?

jgkamat:
> @jgkamat Do you want me to review and merge this?

Yah, that would be nice :D. (let me know if you see any issues)

It should make it easier to interact with plays while editing/hot-reloading
them.

PR #1174, opened 2018-06-30
New Path Planner to Intercept Ball

JNeiger:
Intercepts ball as quickly as possible. It tries to stop in the path of the ball if there is enough time, but if there isn't, it will just hit the ball while the robot is in motion. This is mostly geared towards the goalie

@ashaw596 I'm getting some segfaults when I try to test this. It's getting through the run method completely and dies sometime after I return the path. Any ideas where I should look to try and debug it?

Also, what happens when I pass a final velocity that is much higher than can realistically be reached in the distance between the start and end of the path? Is it actually smart enough to cut it down to something realistic? Here's specifically the code I'm talking about.

```
path = RRTPlanner::generatePath(startEndPoints, obstacles, motionConstraints,
                                        startInstant.vel, 
                                        mag*motionConstraints.maxSpeed * botToTargetNorm);
```

### DIFF
--- a/soccer/Robot.cpp
+++ b/soccer/Robot.cpp
@@ -213,7 +213,7 @@ void OurRobot::lineKick(Point target) {
 
     disableAvoidBall();
     _motionCommand =
-        std::make_unique<Planning::LineKickCommand>(std::move(target));
+        std::make_unique<Planning::LineKickCommand>(target);
 }
 
 void OurRobot::intercept(Point target) {
@@ -221,7 +221,7 @@ void OurRobot::intercept(Point target) {
 
     disableAvoidBall();
     _motionCommand =
-        std::make_unique<Planning::InterceptCommand>(std::move(target));
+        std::make_unique<Planning::InterceptCommand>(target);
 }
 
 void OurRobot::worldVelocity(Geometry2d::Point v) {
--- a/soccer/Robot.cpp
+++ b/soccer/Robot.cpp
@@ -216,6 +216,14 @@ void OurRobot::lineKick(Point target) {
         std::make_unique<Planning::LineKickCommand>(std::move(target));
 }
 
+void OurRobot::intercept(Point target) {
+    if (!visible) return;
+
+    disableAvoidBall();
+    _motionCommand =
+        std::make_unique<Planning::InterceptCommand>(std::move(target));
+}
+
 void OurRobot::worldVelocity(Geometry2d::Point v) {
     _motionCommand = std::make_unique<Planning::WorldVelTargetCommand>(v);
     setPath(nullptr);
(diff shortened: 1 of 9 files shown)


ashaw596:
Hmm. I'll have to take to alook at the at segfault. Have a stack trace from make debug? It accelerates to the max it can so a higher end vel gets cut down to whatever is feasible.

JNeiger:
```
Thread 19 "QThread" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7fffaaffd700 (LWP 26141)]
Planning::angleFunctionForCommandType (command=...) at ../soccer/planning/SingleRobotPathPlanner.cpp:92
92	    switch (command.getCommandType()) {
(gdb) backtrace
#0  Planning::angleFunctionForCommandType (command=...) at ../soccer/planning/SingleRobotPathPlanner.cpp:92
#1  0x00007ffff7a8d1fb in Processor::run (this=0x82e4e0) at ../soccer/Processor.cpp:529
#2  0x00007ffff6cf97be in ?? () from /usr/lib/x86_64-linux-gnu/libQt5Core.so.5
#3  0x00007ffff6a3e6ba in start_thread (arg=0x7fffaaffd700) at pthread_create.c:333
#4  0x00007ffff5b5541d in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:109

```

soccer/Robot.cpp
```diff
@@ -216,6 +216,14 @@ void OurRobot::lineKick(Point target) {
         std::make_unique<Planning::LineKickCommand>(std::move(target));
 }
 
+void OurRobot::intercept(Point target) {
+    if (!visible) return;
+
+    disableAvoidBall();
+    _motionCommand =
+        std::make_unique<Planning::InterceptCommand>(std::move(target));
```
ashaw596: Don't move point. It's faster to just copy it actually than to reference pass it. Also it has nothing thats movable about it.

ashaw596: If you could fix the one in the LineKickplanenr too. That would be nice.

ashaw596:
Hmmm. thats a weird one. Don't really see why that would occur. You didn't really change anything to do with that. Does it work if you put the face command back in there?

JNeiger:
Nope :( Same error

JNeiger:
I think we're corrupting the robot object. If I comment out that line `processor.cpp`, it just straight up fails on [this line](https://github.com/RoboJackets/robocup-software/blob/master/soccer/motion/MotionControl.cpp#L47). I was also printing out the address of the robot object and the frame that it dies on the address jumps from ~0x83202 to over ~0x1b47b30. If that's true, then this is going to be fun

ashaw596:
That is very strange. We really shouldn't be corrupting the Robot object.....

ashaw596:
Hmm. Yeah I checked it in the clion debugger. The robot pointer is a nullptr.

ashaw596:
Yeah the entire MotionControl is just zero initialized. So I assume the robot object was too. Hmm.

ashaw596:
Yeah no. All the robot data got corrupted or something. idk. All the shell_ids are messed up.

JNeiger:
... I was coping the `planRequest.systemState` variable instead of taking a reference to it

JNeiger:
The intercept times are really consistent now with the new vision stuff. I'll test it a few more times on real robots then merge it

Issue #1166, opened 2018-06-20
Preferred robot direction as a factor in role assignment for capture


JNeiger:
Role assignment picks the robot to capture based on the linear distance between the ball and the robots. If the ball is moving, it is better to pick a robot on the path to move forward into the ball. This way we don't have to chase down the ball and pivot around it

jgkamat:
There's a hacky way of doing this on comp right now, which sort of works.
Hoping we find a better way of doing this though (and that we remove the magic
number).

Joseph Neiger writes:

JNeiger:
I think the best approach at this moment is to create a custom `cost_func` that is linearly added to the distance cost. We can then tune the knobs using the `position_cost_multiplier` and our own cost function to get the desired behavior. If we need some sort of non-linear combination, we can just remove the `destination_shape` and do everything in the `cost_func` itself. 

If other subbehaviors decide they want this functionality, we can have the conversation later about whether it's worth adding this role_assignment itself.

@jgkamat Thoughts?

JNeiger:
Hopefully that description overall was enough to give a general idea of the goal. [Here ](https://robojackets.github.io/robocup-software/t20163.html) is a good description of role assignment and how to do different operations. [Here](https://github.com/RoboJackets/robocup-software/blob/master/soccer/gameplay/role_assignment.py#L28) is how the `cost_func` is defined, so if you set `cost_func`, it'll have to follow this form. [Here](https://github.com/RoboJackets/robocup-software/blob/master/soccer/gameplay/role_assignment.py#L299) is where it actually gets added into the total cost for each robot in each role. Finally, [here](https://github.com/RoboJackets/robocup-software/blob/master/soccer/gameplay/skills/mark.py#L94) is the role requirements in another play that gives the general form.

kylestach:
@JNeiger I'm closing this, as it was apparently fixed by capture rewrite. Feel free to reopen it if it should stay open.

Issue #756, opened 2016-06-30
Coordinated pass's receiver should always look for a robot close to the receive point


justbuchanan:
Relevant to #755.  The role requirements for the receiver's subbehaviors should be overridden to ensure that the assignment system looks for a robot close to the receive point, not the current ball position.

jcarn:
I'll look into this and #755

jgkamat:
@jcarn this issue *might* be fixed already (or will be fixed by #755). Someone needs to double check that though.

jgkamat:
This is *not* fixed yet, but #897 gets us most of the way there!

Right now, pass_receive overrides role requirements to prefer the robot closest to where the selected robot is going, which works, but prevents us from using the best robot to receive the ball. Ideally, we would delegate this out to capture, but it's role requirements don't take the ball velocity into account (so it assigns the kicker to get the ball, which fails due to the pass_receive's failure checking rules.

We need to either:

Change pass_receive's role requirements to prefer the robot which actually has the best chance of getting the ball
Fix capture's role assignment to consider ball velocity.
I would much prefer (2), but I'm not sure about how feasible that would be.

I would /really/ like to have this done by japan, and I think this is a pretty good issue for someone new who's interested in role requirements to look at!

jcarn:
I'm messing around with role requirements anyway for the stopped pass thing, so I'll keep working at it

jgkamat:
this is basically kind of a dupe of #375. Solving #375 will solve this issue too though.

kylestach:
@JNeiger, was this fixed by #1180?

JNeiger:
Yep, I think it was. Both due to the capture rewrite as well as the vision rewrite

PR #1180, opened 2018-07-10
Capture Rewrite

### LINKED ISSUE #1179: Rewrite capture using the pathplanner
Closing #1148 #1142 #1128 #433 in favor of this issue due to duplication or being OBE.

My current thoughts are to split up capture into two different path planners. The first part will be to settle the ball. It will be similar to the intercept state where we will get in front of the ball then try to dampen it with a quick motion backwards. I would like to add a feature here where the user can place a target point that specifies the direction the settling touch is in so the play is smoother overall when we start chaining stuff together.

The second part will move into collecting (state name wip) the ball. It should come up to the ball and start dribbling it. Most likely, there will be a velocity profile that starts out with a low delta speed when about to touch the ball that will accelerate once the ball is touched to satisfy the minimum time needed for the dribbler to place a significant spin on the ball.

On the python side, there will most likely be 2 new skills. One will be the settle and the other will just be capture. I don't think collect needs to be exposed.

JNeiger:
Closes #1179 , #1166 
May solve #756

Settle
- [x] Rename file/class to settle and remove anything relating to the collect phase
- [x] Get it to compile with basically whats already in there (See other PR for some of the boilerplate)
- [x] Added partial replan
- [x] Tune interception so we will get in front of the ball correctly
- [x] Implement the damping movement. This could be either the final velocity on the initial path or a separate path / state
- [x] Limit motion if dampen or capture point is outside the field
- [x] Add target point to settle to first touch ball in a certain direction. Limit to `maxAngleoffBallForDampen`
- [x] Add target intercept point as an argument for cases like one touch, where we just want to intercept the ball, but not move towards it
- [x] Try different search techniques to get intercept point
- [x] Create skill in python

Collect
- [x] Find a better name
- [x] Understand the requirements of the dribbler such that we can actually control the ball
- [x] Create a velocity acceleration profile to match those requirements

Capture
- [x] Ctrl-A, Delete
- [x] Add transitions between settle and collect in python

Both
- [x] Move all tuning coefficients to the soccer config
- [x] Minimize the number of tuning coefficients to make the search space smaller
- [ ] Test a lot. Any bugs will crash soccer since it's in C++ now

General
 ~~- [ ] Implement path like trapezoidal but with a custom velocity curve~~
- [x] Look at all the skills and tactics to see if we can standardize the implementations

@CollinAvidano Look at #1174 to get an example of how the pathplanner stuff works. Renaming it and getting it to compile should be a pretty good start while you figure it all out. Feel free to ping me if you have any questions

Edit:
Things left to figure out before comp
 - [x] Figure out why the path uses the ball as the target face point on the control phase of collect
 - [x] Try to improve the velocity non-continuity with the averager or path rejection
 - [ ] Understand why the math of the velocity and distance to stop is so bad on the control phase of collect
 - [x] See if there is a way to improve the collect after receiving a pass and the ball velocity / direction is super sketch


JNeiger:
@ashaw596 Was there any reason subpaths weren't added to TrapezoidalPaths? I just added them and wanted to make sure I wasn't missing some something.

JNeiger:
@ashaw596 What's the difference between `RRTPlanner::generatePath` and using the `rrtPlanner.run(PlanRequest(...))`?

soccer/planning/CollectPathPlanner.cpp
```diff
@@ -0,0 +1,295 @@
...
+        unique_ptr<Path> path =
+            RRTPlanner::generatePath(startEndPoints, obstacles, motionConstraints, startInstant.vel, target.vel);
+
+        path->setDebugText("approaching");
+
+        // Append to the partial path
+        if (path) {
+            if (partialPath) {
+                path = make_unique<CompositePath>(move(partialPath),
+                                                  move(path));
```
kylestach: Is this expected to run every frame? If so this is going to turn the path into basically a linked-list, right?

soccer/planning/CollectPathPlanner.cpp
```diff
@@ -0,0 +1,295 @@
...
+    // If we are close enough to the target point near the ball
+    // and almost the same speed we want, start slowing down
+    // TODO: Check for ball sense?
+    if (dist < *_distCutoffToControl && speedDiff < *_velCutoffToControl && currentState == Approach) {
+        currentState = Control;
+    }
+
+    // Check if we need to go back into approach
+    if ((ball.pos - startInstant.pos).mag() > 2*Robot_Radius && currentState == Control) {
+        cout << "Reseting" << endl;
```
kylestach: *resetting.

Also, can we not have `cout`s in the middle of random code? (Or could it be a little bit more specific as to what's happening?)

soccer/planning/SettlePathPlanner.cpp
```diff
@@ -0,0 +1,419 @@
...
+        // TimeNow     EndPartialPrevPath  FinalTargetPoint
+        //                     |-----------------|
+        //          Amount of the path we can change this iteration
+        if (timeIntoPreviousPath < prevPath->getDuration() - partialReplanLeadTime &&
+            timeIntoPreviousPath > 0ms) {
+            partialPath =
+                prevPath->subPath(0ms, timeIntoPreviousPath + partialReplanLeadTime);
+            partialPathTime = partialPath->getDuration() - timeIntoPreviousPath;
+            startInstant = partialPath->end().motion;
+        }
```
kylestach: Is this copy-pasted from the earlier one? If so it might be good to refactor some of this (Also, if so, why is `partialReplanLeadTime` multiplied by 2 in the other one and not this one?)

soccer/planning/SettlePathPlanner.cpp
```diff
@@ -0,0 +1,419 @@
...
+            currentState == Intercept) {
+            
+            startInstant = pathSoFar->end().motion;
+            currentState = Dampen;
+            std::cout << "Transitioned to dampen" << std::endl;
+        }
+    }
+
+    switch (currentState) {
+    case Intercept: {
```
kylestach: Similar comment about switch-cases here. These seem a bit complex to be nested in an already large switch statement, maybe refactor to helper functions? (something something cyclomatic complexity?)

soccer/motion/TrapezoidalMotion.cpp
```diff
@@ -194,6 +195,7 @@ bool TrapezoidalMotion(double pathLength, double maxSpeed, double maxAcc,
         /// past the end of the path
         posOut = pathLength;
         speedOut = finalSpeed;
+        //std::cout << "Past End of Path" << std::endl;
```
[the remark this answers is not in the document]
JNeiger: @ all the cout's and extra useless comments: I'll get them fixed before I merge stuff in, things aren't working yet so Imma get that stuff fixed then do a style run afterwards

soccer/planning/CollectPathPlanner.cpp
```diff
@@ -0,0 +1,295 @@
...
+            // Approach the ball from behind
+            // TODO: May need to rethink average here since it'll be based on the first ball vel only
+            approachDirection = averageBallVel.norm();
+        }
+    }
+
+    switch (currentState) {
+    // Moves from the current location to the ball
+    // Controls the robot up until just before the point of contact
+    case Approach: {
```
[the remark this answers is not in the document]
JNeiger: Yeah, I gotta think about it a little more, but something should definitely be done. I based the code setup on the line kick path planner.

It's possible there will be one or two more states, but it'll likely be the same behavior overall, just split up more

soccer/planning/SettlePathPlanner.cpp
JNeiger: The should both be the same, rip copy paste.

soccer/planning/CollectPathPlanner.cpp
kylestach: @JNeiger I can't remember, did we talk about this in person?

soccer/config/realShopField.xml
```diff
@@ -57,11 +57,18 @@
    <p value="0"/>
    <i value="0"/>
    <d value="0"/>
+   <i_windup value="0"/>
```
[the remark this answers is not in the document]
JNeiger: The configs will be relevant since I have added and tuned capture parameters, but anything outside of that will be reverted

soccer/planning/SettlePathPlanner.cpp
```diff
@@ -0,0 +1,507 @@
...
+                                                   const RJ::Time curTime,
+                                                   const MotionInstant& startInstant,
+                                                   unique_ptr<Path> partialPath,
+                                                   const RJ::Seconds partialPathTime,
+                                                   const ShapeSet& obstacles) {
+    const Ball& ball = planRequest.systemState.ball;
+
+    // Search settings when trying to find the correct intersection location
+    // between the fast moving ball and our collecting robot
+    // TODO: Either switch to a time or change config to dist
```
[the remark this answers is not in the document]
JNeiger: Specifically the for loop below, it's in meters from the ball, while the config values for that are in seconds.

soccer/planning/SettlePathPlanner.cpp
```diff
@@ -0,0 +1,507 @@
...
+    const RJ::Seconds searchIncTime = RJ::Seconds(*_searchIncTime);
+
+    // How much earlier should we be at the point compaired to the ball
+    const RJ::Seconds bufferInterceptTime = RJ::Seconds(*_interceptBufferTime);
+
+    bool foundInterceptPath = false;
+
+    // Try find best point to intercept using old method
+    // where we check ever X distance along the ball velocity vector
+    for (float dist = 0; dist < 5; dist += 0.05) {
```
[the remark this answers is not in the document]
JNeiger: This was that seconds comment above

soccer/planning/SettlePathPlanner.cpp
```diff
@@ -0,0 +1,507 @@
...
+        float botDistToBallMovementLine = ballMovementLine.distTo(pathSoFar->end().motion.pos);
+
+        // Intercept -> Dampen
+        //  Almost interescting the ball path and
+        //  Almost at end of the target path and
+        //  Actually in front of the ball
+        //
+        // TODO: Check ball sense?
+        if (timeIntoPreviousPath < prevPath->getDuration() - 1.5*partialReplanLeadTime &&
+            (pathSoFar->end().motion.pos - prevPath->end().motion.pos).mag() < 3*Robot_Radius &&
```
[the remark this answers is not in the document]
JNeiger: Yeah, I need to do that. These are kinda the same thing so I need to decide if I want to combine them

soccer/planning/SettlePathPlanner.cpp
```diff
@@ -0,0 +1,538 @@
...
+
+        ctr = 0;
+        if (!Field_Dimensions::Current_Dimensions.FieldRect().containsPoint(interceptTarget)) {
+            interceptTarget += 10*averageBallVel;
+        }
+
+        while (!Field_Dimensions::Current_Dimensions.FieldRect().containsPoint(interceptTarget) && ctr < 100) {
+            interceptTarget += 0.1*averageBallVel;
+            ctr++;
+        }
```
JNeiger: If someone has a clean way to do this, lmk. I'm having trouble finding a good search term in the couple minutes that I looked. Basically I want to limit the target point from moving outside the field. It the target point is outside, I want to select the intersection point of the ball velocity line with the field rectangle.

I can do the line rectangle intersection and pick the intersect point if it's closer, but I feel like there's a more direct way of doing this that we can throw into the geometry2d class.

kylestach: Huh? Why isn't a line-rect intersection what you want here? I guess technically you want to intersect with a ray, but I'm confused as to why that's still less "clean" than this.

JNeiger: It's not that it's less clean, but that I feel like there is a more direct calculation that can be made to do this. If I do the intersection, it'll be the rect line intersection, a dot product to get the one in front of the moving ball, then a comparison to get the closest.

This class of problems has a lot of work out there due to the whole game design and graphics industry that I figured there might be a more unique and direct approach since I can see this use case in a variety of locations

JNeiger: If we get a more direct approach, then we could just throw this into the rect class itself

kylestach: I don't think there is one, but a line segment-rectangle intersection is pretty fast anyway (our code for that allocates a vector for some reason, but I'm pretty sure it's still going to be faster than what you have here for most inputs and should only return two points: the robot's origin and the intersection point).

ashaw596: I think we have a Line and and a Line segment class. You can immitate a velocity ray by making a long line segment and doing that intersection with the rectangle if it's implemented right.

soccer/planning/SettlePathPlanner.cpp
```diff
@@ -0,0 +1,507 @@
...
+    }
+
+    // Check if we should transition from intercept to dampen
+    // Start instant may be changed in that case since we want to start changing the path
+    // as soon as possible
+    processStateTransition(ball, prevPath.get(), timeIntoPreviousPath, startInstant);
+
+    // Run state code
+    // TODO: Do I need to do a move or is it implicit?
```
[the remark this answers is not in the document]
ashaw596: This actually can't do Return value optimization since there's multiple return paths. But function calls return temporary values anyways. THe std:move just casts values to temporaries so it's basically a no-op to have it here.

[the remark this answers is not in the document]
ashaw596: By temporaries I mean rvalue.

ashaw596:
> @ashaw596 What's the difference between `RRTPlanner::generatePath` and using the `rrtPlanner.run(PlanRequest(...))`?

Uhh. sort of late. sorry. I think rn, rrtPlanner.run partially wraps the generatePath call, to follow the Planner interface. But it also has replanning logic. generatePath just generates a VelocityProfiled RRT path.

JNeiger:
I'm starting to get to the point that I can clean up this branch. The code should be basically done, but I have a little tuning left. (Imma leave those print statements in until that is done).

Just to update on the overall content of this PR. 

Capture:
Rewritten and broken into two new skills, Settle and Collect.

Some big new features on the python side is the improved role assignment cost function and averaging on the ball sense.

Settle:
Moves in front of the ball to force the ball to hit the dribbler and when it is in the path, just before the ball hits the dribbler, we move backwards to dampen the ball. (The dampen isn't used currently since the dribbler dampens really well anyways).

2 big states, intercept and then dampen.
Intercept finds the first point that we can reach before the ball does and moves there. There's a little jank here to try and solve the issue with non-continuous velocity profiles.
Dampen stays in the ball path and moves backwards and stops. The ball should be reaching the robot while it stll has some backwards velocity

Collect:
Approaches the ball at any speed, some distance before hitting, it slows down to some slow speed (relative to the ball), then moves to touch the ball, once it touches, it slows down to a stop.

3 big states, course, fine, and control.
Course just moves at the ball until the point where it should slow down to a constant speed.
Fine stays at the constant speed until it reaches the ball.
Control moves through the ball and stops.

Coordinated Pass:
Since passing is one of the easiest ways to test this, I improved the passing as well. I mostly just changed some constants so we aim a little better.

One Touch Pass:
I made settle with the idea that it should be used to one touch things in since the only difference is the angle receive. This is the iffy spot since I never actually fully read the code to understand it. I read just enough to make it work.

A lot of the config motion control values were changed since we actually need to be precise in our movements.

Imma start just adding people to review this. Anything that shouldn't be changed, anything that doesn't make sense etc just throw a comment in. You don't have to review everything, just the parts you're comfortable with.

soccer/planning/SettlePathPlanner.cpp
```diff
@@ -0,0 +1,528 @@
...
+        std::move(path),
+        angleFunctionForCommandType(FacePointCommand(planRequest.systemState.ball.pos)));
+}
+
+void SettlePathPlanner::calcDeltaPosForDir(const Ball& ball, const MotionInstant& startInstant, 
+                                           Geometry2d::Point& deltaRobotPos, Geometry2d::Point& facePos) {
+    // If we have a valid bounce target
+    //
+    // TODO: Simplify the math
+    if (targetBounceDirection) {
```
JNeiger: Check my math here. The goal is to offset the position of the robot so that when the robot is facing some other point besides the ball, the ball will hit the mouth of the robot. `deltaRobotPos` is the offset from the ball movement line for the robot settling the ball. `facePos` is the point that we end up trying to face.

soccer/gameplay/skills/capture.py
```diff
@@ -1,270 +1,302 @@
...
+        if (evaluation.ball.robot_has_ball(self.robot)): #self.robot.has_ball()):
+        #if (self.robot is not None and self.robot.vel.mag() < Collect.STOP_SPEED):
+            self.probably_held_cnt = min(self.probably_held_cnt + 1, Capture.PROBABLY_HELD_HISTORY_LENGTH)
+        else:
+            self.probably_held_cnt = max(self.probably_held_cnt - 1, 0)
 
-    def on_enter_delay(self):
-        self.start_time = time.time()
+    @staticmethod
+    def role_assignment_cost(robot):
```
JNeiger: Check this math out too. My goal was to halfway estimate which robot will reach the ball first and choose that one

soccer/SystemState.cpp
```diff
@@ -139,7 +139,7 @@ RJ::Time Ball::estimateTimeTo(const Geometry2d::Point& point,
 
 double Ball::estimateSecondsTo(const Geometry2d::Point &point) const {
     const auto time = estimateTimeTo(point);
-    return RJ::Seconds(RJ::now() - time).count();
+    return RJ::Seconds(time - RJ::now()).count();
```
JNeiger: I think this was a bug. `estimateTimeTo` returns the current time + the time for the ball to move that distance. To get the time between that estimate and now, you have to subtract now otherwise the time is always negative

soccer/gameplay/skills/capture.py
```diff
@@ -1,270 +1,302 @@
...
-    InterceptVelocityThresh = 0.15
 
-    # Multiplied by the speed of the ball to find a "dampened" point to move to during an intercept
-    DampenMult = 0.0
+class Capture(single_robot_composite_behavior.SingleRobotCompositeBehavior):
 
-    # The distance to transition from coarse approach to fine
-    # TODO: The correct way to do this would be using our official max acceleration and current speed
-    CoarseToFineApproachDistance = 0.3
+    DRIBBLE_SPEED = 50
```
JNeiger: I want these to be that `constants.StandardPower`, but at the current moment, if you bump up the dribbler, the robot ends up sucking up the ball and lifting it's own wheels off the ground. At the current moment, I'm leaving them separate so we can bump it up for the initial settle, the lower it for control portions

soccer/gameplay/skills/capture.py
```diff
@@ -1,270 +1,302 @@
...
+
+        self.prev_ball = main.ball()
+        self.probably_held_cnt = Capture.PROBABLY_HELD_START
+        self.frames_in_captured = 0
+
+        
+        # By default, move into the settle state since we almost never start with ball
+        self.add_transition(behavior.Behavior.State.start,
+                            Capture.State.settle,
+                            lambda: self.robot is not None and not evaluation.ball.robot_has_ball(self.robot), #self.robot.has_ball(),
```
JNeiger: Until the robots get break beam sensors to detect the ball, we have to use this jank way. There's two ways we can deal with this, have both and just assume that if the break beam is valid and the ball is in the right spot, then we are good.

The only challenge is that when the ball gets covered by the robot so the cameras can't see it, it won't trigger since the robot will move the ball slightly

soccer/gameplay/skills/capture.py
```diff
@@ -1,270 +1,241 @@
...
+        # Don't have ball so try to intercept the moving ball
+        settle = 2
 
-    class State(Enum):
-        intercept = 0
-        coarse_approach = 1
-        fine_approach = 2
-        delay = 3
+        # Don't have ball so try to do the fine approach to control the ball
+        collect = 1
```
tjones320: Put these in order

soccer/gameplay/skills/capture.py
```diff
@@ -1,270 +1,241 @@
...
+        self.add_transition(Capture.State.collect,
+                            behavior.Behavior.State.completed,
+                            lambda: self.subbehavior_with_name('collector').is_done_running(),
+                            'captured')
+
+        # Go back to settle if we loose the ball when completed
+        self.add_transition(behavior.Behavior.State.completed,
+                            Capture.State.settle,
+                            lambda: self.probably_held_cnt < Capture.PROBABLY_NOT_HELD_CUTOFF,
+                            'captured')
```
tjones320: Is this the name you want here? Kinda picky

soccer/gameplay/skills/capture.py
```diff
@@ -1,270 +1,241 @@
...
-    def on_exit_coarse_approach(self):
-        self.lastApproachTarget is None
+        # Use straight line distance to closest point along line
+        # Then movement down the line
 
-    def execute_fine_approach(self):
-        self.robot.disable_avoid_ball()
-        self.robot.set_dribble_speed(self.dribbler_power)
+        # Represents ball movement line
+        ball_move_line = robocup.Line(main.ball().pos, main.ball().pos + main.ball().vel * 10)
```
tjones320: What's the justification for 10 seconds?

soccer/gameplay/skills/capture.py
```diff
@@ -1,270 +1,241 @@
...
+        ball_to_robot = robot.pos - main.ball().pos
+        ball_to_intercept = ball_line_intercept - main.ball().pos
+        robot_to_intercept = ball_line_intercept - robot.pos
 
-    def on_enter_delay(self):
-        self.start_time = time.time()
+        # If we are actually in front of the ball angle wise
+        # OR close enough that it could be a bounce off the robot due to noisy vision
+        in_front = ball_move_line.delta().angle_between(ball_to_robot) < math.pi/2 or \
+                   ball_to_robot.mag() < 1.5*(constants.Robot.Radius + constants.Ball.Radius)
```
tjones320: Are you ok with the fact that the ball could be moving away from the robot here?

soccer/gameplay/skills/capture.py
```diff
@@ -1,270 +1,241 @@
...
+        max_acc = robocup.MotionConstraints.MaxRobotAccel.value
+
+        # How long it will take the robot/ball to reach that intercept point
+        ball_secs = main.ball().estimate_seconds_to(ball_line_intercept)
+        robot_secs = robocup.get_trapezoidal_time(robot_to_intercept.mag(), robot_to_intercept.mag(), max_speed, max_acc, 0, 0)
+
+        time_to_hit = robot_secs
+
+        # If we aren't in front of the ball, we need to get time to catch up
+        if (not in_front):
```
tjones320: Also, can local variables be seen across the same state? This is nice to know if they can.

soccer/gameplay/skills/capture.py
```diff
@@ -1,270 +1,241 @@
...
+            # if we are slower than the ball, don't even try and just exclude it automatically
+            if (delta_speed < 0):
+                time_to_hit += 100
+            # If we are fast, figure out how much more time we need
+            else:
+                delta_time = robocup.get_trapezoidal_time(ball_to_intercept.mag(), ball_to_intercept.mag(), delta_speed, max_acc, 0, 0)
+                time_to_hit += delta_time
+        
+
+        main.system_state().draw_text(str(time_to_hit), robot.pos + robocup.Point(0.1,0.1), (255,255,255), "test")
```
tjones320: Possibly give this a better name if you're going to keep it

PR #1181, opened 2018-07-10
Add Indirect and Direct buttons

### LINKED ISSUE #1178: Add buttons for restarts to the soccer ui
It is becoming more important that we test our restart plays so we should add buttons to trigger the different restarts to the soccer ui to make it easier.

chachmu:
Closes #1178.


chachmu:
Having to hit stop before the restart is part of the rules. Normal refbox is run the same way.

PR #1183, opened 2018-07-11
Kickoff improvements

joshhting:
@Jason27chan do you want to keep these commits that you made on my laptop during comp?


Jason27chan:
naw most of it doesnt use the new find_shooting_gap function so we can abandon this

PR #1185, opened 2018-07-14
Add Movement to the Test_Capture play

### LINKED ISSUE #1159: Add ball movement to the test capture play
It would be nice to move the ball a little bit when we run test capture to make sure we are actually capturing the ball. Something like the rotation around the ball that we see in pivot kick etc

CollinAvidano:
Closes #1159 Added a rotating state to test capture that runs aim subbehavior to pivot around the ball until facing towards the goal after intial capture to check that it actually has it.

### DIFF
--- a/soccer/gameplay/plays/testing/test_capture.py
+++ b/soccer/gameplay/plays/testing/test_capture.py
@@ -12,16 +12,17 @@ class TestCapture(play.Play):
     class State(enum.Enum):
         setup = 1
         capturing = 2
+        rotating = 3
 
     def __init__(self):
         super().__init__(continuous=True)
 
         self.shell_id = None
+        self.dribbler_power = 128
+        self.face_target = robocup.Point(0,0)
 
-        self.add_state(TestCapture.State.setup,
-                       behavior.Behavior.State.running)
-        self.add_state(TestCapture.State.capturing,
-                       behavior.Behavior.State.running)
+        for state in TestCapture.State:
+            self.add_state(state, behavior.Behavior.State.running)
 
         self.add_transition(behavior.Behavior.State.start,
                             TestCapture.State.setup, lambda: True,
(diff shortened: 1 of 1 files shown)


jcarn:
Looks solid, works well in sim. I would like to see it rotate for at least a few full rotations to get a full picture of how well we are hanging on to the ball.

soccer/gameplay/plays/testing/test_capture.py
```diff
@@ -33,8 +34,13 @@ def __init__(self):
...
 
         self.add_transition(
-            TestCapture.State.capturing, TestCapture.State.setup,
+            TestCapture.State.capturing, TestCapture.State.rotating,
             lambda: self.subbehavior_with_name('capture').state == behavior.Behavior.State.completed,
+            'rotating to check capture')
+
+        self.add_transition(
+            TestCapture.State.rotating, TestCapture.State.setup,
+            lambda: self.subbehavior_with_name('aim').state == skills.aim.Aim.State.aimed,
```
jgkamat: If aim goes into the failed state, this play will be stuck in this state and won't do anything.

soccer/gameplay/plays/testing/test_capture.py
```diff
@@ -33,8 +34,13 @@ def __init__(self):
             'robot away from ball')
 
         self.add_transition(
-            TestCapture.State.capturing, TestCapture.State.setup,
+            TestCapture.State.capturing, TestCapture.State.rotating,
```
jgkamat: Could rotation be gated behind a  boolean at the top of the file? I think this might get annoying on the field.

You could do this by just going into/out of rotation in a single frame, to keep things simple.

chachmu: I'm not sure if this would really be annoying during comp since it's almost impossible to tell if capture is actually working without the aim added.

PR #1186, opened 2018-07-15
Disable Dribbler When Out of Bounds

### LINKED ISSUE #1163: Stop dribbler when outside the field
We're having problems where the dribblers are on when we drive into walls which is kinda bad. This can also be applied to kicking

CollinAvidano:
Closes #1163  Added a quick check in Robot.cpp that checks if the robot is outside of configurable field bounds and disables the dribblers. May need to move config out of pathplanning tab.

### DIFF
--- a/soccer/Robot.cpp
+++ b/soccer/Robot.cpp
@@ -47,6 +47,7 @@ REGISTER_CONFIGURABLE(OurRobot)
 ConfigDouble* OurRobot::_selfAvoidRadius;
 ConfigDouble* OurRobot::_oppAvoidRadius;
 ConfigDouble* OurRobot::_oppGoalieAvoidRadius;
+ConfigDouble* OurRobot::_dribbleOutOfBoundsMult;
 
 void OurRobot::createConfiguration(Configuration* cfg) {
     _selfAvoidRadius =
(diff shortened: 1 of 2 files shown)


soccer/Robot.cpp
```diff
@@ -265,10 +268,15 @@ float OurRobot::kickTimer() const {
 }
 
 void OurRobot::dribble(uint8_t speed) {
-    uint8_t scaled = *config->dribbler.multiplier * speed;
-    control->set_dvelocity(scaled);
+    Geometry2d::Rect modifiedField = Field_Dimensions::Current_Dimensions.FieldRect() * (*_dribbleOutOfBoundsMult);
+    if (modifiedField.containsPoint(pos)) {
```
JNeiger: I'm worried this doesn't actually extend the "playable field" behind our goal line since the y coordinates are 0. Did we check that?

JNeiger:
Do we want a static distance from the field edges or some sort of percentage? I can see the percentage causing some problems when we use a larger field, or even a smaller field since it could disable our dribbler when we can still interact with the ball inside the field

soccer/Robot.cpp
```diff
@@ -265,10 +268,16 @@ float OurRobot::kickTimer() const {
 }
 
 void OurRobot::dribble(uint8_t speed) {
-    uint8_t scaled = *config->dribbler.multiplier * speed;
-    control->set_dvelocity(scaled);
+    Geometry2d::Rect modifiedField = Field_Dimensions::Current_Dimensions.FieldRect();
+    modifiedField.expand(*new Point(*_dribbleOutOfBoundsOffset,*_dribbleOutOfBoundsOffset));
```
JNeiger: For future reference, you are are trying to pass a object by value to a function, just do `modifiedField.expand(Point(...))`. Here you are creating an object on the heap then instantly throwing away the resulting address of that object on the heap which causing a memory leak. Every `new` needs a corresponding delete (in most cases)

CollinAvidano: Thanks Joe. I should of known about the memory leak. Though in regards to the second comment I asked Woodward if that would work.

JNeiger:
I think this is good. We just need to test it and get the right offset value. Also test what happens when the robot is running without a shell, I'm curious to see where the default position is when vision can't see it.

PR #1192, opened 2018-07-17
Remove Random statics

ashaw596:
Static doesn't really mean anything in this context.

### DIFF
--- a/common/Utils.hpp
+++ b/common/Utils.hpp
@@ -14,7 +14,7 @@
 #include <QtWidgets>
 #include <memory>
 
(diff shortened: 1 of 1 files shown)


jgkamat:
I (and CI) are getting compiler errors with this, could you take a look at
that?

https://circleci.com/gh/RoboJackets/robocup-software/6677

ashaw596 writes:

ashaw596:
Hmm. It seems like a weird name collision. These should probably be in the RJ namespace...

ashaw596:
I think if we do #1193 we can remove static. I think the problem is global name collisions.

PR #1203, opened 2018-09-09
Latency Testing Play

CollinAvidano:
The robots switch between every play in the behavior sequence. The play itself seems to be restarting.


JNeiger:
Is this good for a review or are things still being messed with?

CollinAvidano:
This is being removed in favor of Chris' tests.

PR #1212, opened 2018-10-10
Add fixes for downgraded git version in dockerhub

jgkamat:
At some point, the docker hub downgraded their version of git to bootstrap the dockerfiles (or they're using faulty scripts somewhere), which causes the docker hub baseimage builds to fail. This fixes that by turning off submodules in the docker hub build (they should still get populated on circle since they should have an up-to-date git version).

[here's the basimage build page for the curious](https://hub.docker.com/r/robojackets/robocup-software/builds/)

This incorporates the fixes I found in https://github.com/RoboJackets/robocup-firmware/pull/66

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -34,4 +34,4 @@ ENV LANG en_US.UTF-8
 COPY . ~/robocup-software
 WORKDIR ~/robocup-software
 
-RUN sudo ./util/ubuntu-setup --yes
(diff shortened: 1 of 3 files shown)


JNeiger:
LGTM! I'm not super familiar with this, but it looks ok to me

jgkamat:
I'm just going to merge this to save my email inbox, let me know if it causes any issues and I'll fix it :)

PR #1214, opened 2018-10-25
Fixed Memory Leak in Nelder Mead Python Function Pointer

JNeiger:
I still need to test this but and clean up the code a little, but I'm 90% sure this is needed. Otherwise we were leaking python objects every time we called this.

Double check me and make sure I actually need this and it's done correctly. In my mind, we have to take control of the `PyObject` when it gets passed to us since we are using it, not the parent method. If we take control, we then would need to release control and let the garbage collector eat the leftovers.

I'm also not sure about setting the pointer to `nullptr` after `decref`'ing it. It makes it an easy test so I can tell people about this weird idiosyncrasy, but it's kinda jank.

### DIFF
--- a/soccer/gameplay/robocup-py.cpp
+++ b/soccer/gameplay/robocup-py.cpp
@@ -26,6 +26,7 @@ using namespace boost::python;
 #include "motion/TrapezoidalMotion.hpp"
 #include "optimization/NelderMead2D.hpp"
 #include "optimization/NelderMead2DConfig.hpp"
+#include "optimization/PythonFunctionWrapper.hpp"
 #include "planning/MotionConstraints.hpp"
 
 #include <boost/python/exception_translator.hpp>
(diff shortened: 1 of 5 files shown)


soccer/gameplay/robocup-py.cpp
```diff
@@ -598,35 +598,6 @@ float Point_get_y(const Geometry2d::Point* self) { return self->y(); }
 void Point_set_x(Geometry2d::Point* self, float x) { self->x() = x; }
```
jgkamat: I think you need to include `PythonFunctionWrapper.hpp` here now

soccer/optimization/PythonFunctionWrapper.hpp
```diff
@@ -0,0 +1,50 @@
+#include <functional>
+#include <Python.h>
+#include <Geometry2d/Point.hpp>
+
+
+float cpp_function_cb(Geometry2d::Point p, PyObject* pyfunc) {
+    if (pyfunc == nullptr) {
+        std::cerr << "Pyfunction is null. Does the PythonFunctionWrapper have the same lifetime as the NelderMead object?"
```
jgkamat: Missing endl and semicolon ;P

soccer/gameplay/robocup-py.cpp
```diff
@@ -958,9 +929,8 @@ BOOST_PYTHON_MODULE(robocup) {
         .def("eval_pt_to_our_goal", &KickEval_eval_pt_to_our_goal)
         .def("eval_pt_to_seg", &KickEval_eval_pt_to_seg);
 
-    class_<std::function<float(Geometry2d::Point)>,
-           std::function<float(Geometry2d::Point)>*>("stdfunction", no_init)
-        .def("__init__", make_constructor(&stdfunction_constructor));
+    class_<PythonFunctionWrapper>("PythonFunctionWrapper")
+        .def("get_function", &PythonFunctionWrapper::getFunction);
```
jgkamat: This line is now complaining about missing arguments to the `PythonFunctionWrapper` constructor. I think you need to define a similar constructor like `NelderMead2DConfig`.

tjones320:
Do you we need include more comments/documentation about this process and what's going on?

JNeiger:
@tjones320 @CollinAvidano Is this good?

JNeiger:
Imma merge this in a little bit unless there are more requests

Issue #1177, opened 2018-07-02
Record strategic game information


joshhting:
Developing plays that make more strategic, reactive, and holistic decisions requires extracting information on the game state and the opposing team's capabilities, and making this information readily available for the play system to access. To start, we are interested in gathering the following information:
- Who last touched the ball
- Opponent's chipper capability and whether a chip is in progress
  - Tiger's Autoref
- Opponent's demonstrated max speed

JNeiger:
Are we planning to do some sort of simplified chip detection or should I just build that into the ball filter stuff? All of tigers chip detection is in the ball filter module itself

joshhting:
Do that in the ball filter code, that makes the most sense to me

JNeiger:
Other things to include is...

- PDF of probable robot locations
- Whether the other team can actually kick
- A kick event object from the ball filter

PR #1215, opened 2018-10-31
Indicate which team last touched ball

joshhting:
Resolves part of #1177


kylestach:
Not really sure what to do with this - it doesn't look like it's in a working state yet, and there hasn't been any activity here (and it's definitely not urgent) so I'm closing it for now.

PR #1218, opened 2018-11-07
Moved defensive forward and simple zone midfielder zone to tactics/po…

mdmarshmallow:
…sitions closes #1218

### DIFF
--- a/soccer/gameplay/tactics/positions/defensive_forward.py
+++ b/soccer/gameplay/tactics/positions/defensive_forward.py
@@ -1,0 +1,129 @@
+import main
+import robocup
+import behavior
+import constants
+import enum
+import math
+
+import composite_behavior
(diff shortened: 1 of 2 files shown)


JNeiger:
I forgot to add this to the issue, but the file path changes must be reflected to wherever this is being used (AKA adaptive formation). Just go in there and change everything to be `tactics.positions.simple_zone_midfielder` and `tactics.positions.defensive_forward` instead of the non-position path

JNeiger:
Actually, on second thought, it fits better if this stuff stays in the tactics folder. The stuff in the `positions` folder right now are the single robot type behaviors that make up the defense. It may be better to move the positions into the skill folder depending on whether we want skills to represent "simple" behaviors or single robot behaviors

PR #1219, opened 2018-11-07
Deleted a blank line

bopas2:


JNeiger:
I think in this specific case, it's ok to have the extra white space since it's more so a collection of functions in a file instead of a class.

joshhting:
Idk if we use PEP 8 or not, but PEP 8 style actually uses double blank lines between top-level functions

PR #1227, opened 2018-11-16
Vision rewrite

### LINKED ISSUE #1202: Rewrite vision filter using multi-hypothesis localization
The overall process will be somewhat following [this approach](https://www.researchgate.net/publication/326464230_Relative_Pose_Estimation_using_Range-only_Measurements_with_Large_Initial_Uncertainty). 

The idea is that on the first camera frame, we will create a "kalman filter" object intialized at each ball position possiblity as stated in the camera packet (Referenced as "camera ball" from here out).
Each frame from there we will try to attach a measurement to the existing "kalman filter" objects using some cutoff like chi-squared test or something even simpler like checking a static distance. If it doesn't match a previous filter, we will just create a new one at that point. 

Each time a point matches an existing filter, a counter referred to as health is increased up to a max value. Every time-step we also decreases the health for all existing filters. In the linked paper above, this health is referred to as the chi count.

The filters should also taken into account bounces and such off of other robots.

A kick detection system will also need to be developed. This will need to use some factors both relating to changes in the ball state as well as surrounding robots. Ball state stuff would be things like changes in speed, changes in direction etc. The robot stuff will mostly be distance based and within the mouth.

Additionally, it is very important to figure out where the kick is going. In this case we will use some sort of least squares or nonlinear model on the raw vision packets attributed to this filter to figure out where the ball is going quickly.

The same approach for the filters will be applied to the robots themselves. The only difference is that it will be a 6 state (xy position/velocity and angle/angular velocity) filter instead of 4 state (xy position/velocity). The same health and such will be used.

This will basically replace everything in the `soccer/modeling` folder.

JNeiger:
Closes #1202 

I haven't tried compiling this yet so ¯\\\_(ツ)\_/¯. I'll add some documentation eventually. This is also going to be super fun to review. Should be done in a week or two

- [x] Finish camera robot filter
- [x] Fill out world ball/robot
- [x] Ball reflection math
- [x] Finish up world object
- [x] Fast/Slow kick detection
- [x] Kick Events
~~Kick/Chip trajectory prediction~~
- [x] Finish adding tests

Summary

We get camera measurements from SSLVision. They all get packed into my own CameraRobot/Ball objects to get rid of all the extra data and not deal with protobuf stuff. There are two different algorithms to figure out which Kalman Filters those measurements should go to. AKF which averages all the measurements each frame for each camera and assigns the measurements to the one kalman filter in the camera for that specific object. MHKF which averages all the measurements which are a specific distance from each of the kalman filters in the camera for that specific object. If one of the measurements isn't close enough to an existing kalman filter, a new one is created. These Kalman Filters are called KalmanRobot/Ball. Finally, we need to take the best kalman filter object from each camera and merge them together to find the best across the entire world. They are averaged together based on their variance/standard deviation. These best estimates of the objects are called WorldRobot/Ball's. I predict ball's bouncing off of robots also to help improve our filters.

Edit: 11/23 First round of compile fixes done. Only a few unimplemented functions left that are causing errors
Edit 12/3 Everything is implemented. Adding tests for all the functions before I begin trying out with real vision data


JNeiger:
Anything in the `Ball`, `Camera`, `Filter` and `Robot` folder inside the new `vision` folder is probably good to start reviewing. Nothing actually runs yet, but there's quite a bit of style etc that needs to be reviewed

soccer/vision/ball/BallBounce.cpp
```diff
@@ -0,0 +1,267 @@
...
+        Geometry2d::Point mouthHalfUnitVec = Geometry2d::Point(0, 1).rotate(robot.getTheta());
+        // TODO: Get mouth dist from center
+        Geometry2d::Point mouthCenterPos = Geometry2d::Point(.1, 0).rotate(robot.getTheta()) + robot.getPos();
+        Geometry2d::Line mouthLine = Geometry2d::Line(mouthCenterPos + mouthHalfUnitVec,
+                                                      mouthCenterPos - mouthHalfUnitVec);
+
+        Geometry2d::Point mouthIntersect;
+        bool intersects = intersectLine.intersects(mouthLine, &mouthIntersect);
+
+        // The mouth is a coord across the circle.
```
chachmu: This is pretty pedantic but a line that cuts across a circle is a "chord"

soccer/vision/ball/BallBounce.hpp
```diff
@@ -0,0 +1,57 @@
...
+
+class BallBounce {
+public:
+    /**
+     * Calculates whether the given kalman ball will bounce against another robot and
+     * the resulting velocity vector
+     *
+     * @param ball Kalman ball we are trying to test
+     * @param yellowRobots Best estimation of the yellow robots states
+     * @param blueRobots Best esetimation of the yellow robots states
```
chachmu:
```suggestion
     * @param blueRobots Best estimation of the yellow robots states
```

soccer/vision/camera/Camera.hpp
```diff
@@ -0,0 +1,167 @@
...
+    static void createConfiguration(Configuration* cfg);
+
+private:
+    /**
+     * MHKF refers to the Multi-Hypothesis Kalman Filter
+     *   MHKF averages "near" the current predicted position before using that as the measurement
+     *   Any measurements not "near" a filter are used as the initial values to create a new filter
+     *
+     * AKF refers to the Average Kalman Filter
+     *   AKF averages all mesurements and then uses that as the measurement to the filter
```
chachmu:
```suggestion
     *   AKF averages all measurements and then uses that as the measurement to the filter
```

soccer/vision/camera/Camera.hpp
```diff
@@ -0,0 +1,167 @@
...
+     * Removes any invalid kalman balls that may be too old etc
+     *
+     * Done every iteration to keep things clean
+     */
+    void removeInvalidBalls();
+    void removeInvalidRobots();
+
+    /**
+     * Predicts all robots in one of the yellow/blue lists
+     * Simplifies some copy past
```
chachmu:
```suggestion
     * Simplifies some copy paste
```

soccer/vision/camera/Camera.hpp
```diff
@@ -0,0 +1,167 @@
...
+    /**
+     * Removes any invalid kalman balls that may be too old etc
+     *
+     * Done every iteration to keep things clean
+     */
+    void removeInvalidBalls();
+    void removeInvalidRobots();
+
+    /**
+     * Predicts all robots in one of the yellow/blue lists
```
chachmu:
```suggestion
     * Predicts all robots in the given list
```

soccer/vision/camera/World.cpp
```diff
@@ -0,0 +1,152 @@
...
+void World::calcBallBounce() {
+    for (Camera& camera : cameras) {
+        if (camera.getIsValid()) {
+            camera.processBallBounce(robotsYellow, robotsBlue);
+        }
+    }
+}
+
+void World::updateWorldObjects() {
+    // Fill robotsYellow/Blue with what robobts we want and remove the rest
```
chachmu:
```suggestion
    // Fill robotsYellow/Blue with what robots we want and remove the rest
```

soccer/vision/filter/KalmanFilter.cpp
```diff
@@ -0,0 +1,32 @@
+#include "KalmanFilter.hpp"
```
chachmu: I have been informed that there is a wikipedia page with most of this information. It would be nice if you at least added the link

soccer/vision/kick/detector/FastKickDetector.hpp
```diff
@@ -0,0 +1,27 @@
...
+
+#include <Geometry2d/Point.hpp>
+#include <Utils.hpp>
+
+#include "vision/ball/WorldBall.hpp"
+#include "vision/robot/WorldRobot.hpp"
+#include "vision/kick/KickEvent.hpp"
+#include "vision/kick/VisionState.hpp"
+
+class FastKickDetector {
```
chachmu: It would be good to get a quick explanation of what this does. Does it detect kicks quickly or does it detect kicks that give the ball a high velocity. Same for the SlowKickDetector

JNeiger:
What did you have in mind for the kalman filter "merger"? I would like to keep one super general interface so that if we decide to use this elsewhere, we can choose different A, B, C, D matrices. I might be able to merge the 2D and 3D together since they have somewhat of the same form, but I'm slightly worried how that would affect the clarity of which matrices are being used. Also the whole getPos, getTheta stuff will have to be generalized into a single getPosState() where the user will have to pull out the position and heading. If you think that would look clearer, I don't mind switching over to it

chachmu:
I think I was mostly looking at kalmanBall and kalmanRobot and had seen a lot of overlap for some of the update code but if they need to be separate since the robots have a theta then that's fine

JNeiger:
I still want to go back and clean things up a little bit, but it works!

Here's the ball speed after a kick, little bit of oscillation, but it reaches max speed about instantly
image

JNeiger:
@jgkamat So the coverage is failing but the c++ tests is passing? Are the not running the same thing?

JNeiger:
This is all good to review. Any changes from here on out will be just comments/documentation

soccer/vision/ball/BallBounce.hpp
```diff
@@ -0,0 +1,70 @@
...
+#include <vector>
+
+#include <Geometry2d/Point.hpp>
+#include <Configuration.hpp>
+
+#include "KalmanBall.hpp"
+#include "vision/robot/WorldRobot.hpp"
+
+class BallBounce {
+public:
```
ashaw596: =P you can probs go ahead and delete the constructor or destructor (or just turn this into a namespace if you want)

soccer/vision/ball/BallBounce.hpp
```diff
@@ -0,0 +1,70 @@
...
+private:
+    /**
+     * Returns whether the ball is most likely intersecting the robots
+     *
+     * Note: Ignores the extra mouth calculations
+     *
+     * @param ball The ball we want to check for intersection with
+     * @param robot The robot we what to check for intersection with
+     */
+    static bool BallInRobot(KalmanBall& ball, WorldRobot& robot);
```
ashaw596: Can most of the variables and stuff be const ref in this file?

soccer/vision/ball/KalmanBall.cpp
```diff
@@ -0,0 +1,96 @@
...
+
+    previousMeasurements.push_back(initMeasurement);
+}
+
+void KalmanBall::predict(RJ::Time currentTime) {
+    lastPredictTime = currentTime;
+
+    // Decrement but make sure you don't go too low
+    health = std::max(health - *VisionFilterConfig::filter_health_dec,
+                      (int)*VisionFilterConfig::filter_health_min);
```
ashaw596: Change to static cast if you want.

soccer/vision/ball/KalmanBall.cpp
```diff
@@ -0,0 +1,96 @@
...
+void KalmanBall::predictAndUpdate(RJ::Time currentTime, CameraBall updateBall) {
+    lastPredictTime = currentTime;
+    lastUpdateTime = currentTime;
+
+    // Increment but make sure you don't go too high
+    health = std::min(health + *VisionFilterConfig::filter_health_inc,
+                      (int)*VisionFilterConfig::filter_health_max);
+
+    // Keep last X camera observations in list for kick detection and filtering
+    previousMeasurements.push_back(updateBall);
```
ashaw596: Would a boost::circular_buffer work?

soccer/Robot.hpp
```diff
@@ -120,7 +109,6 @@ class Robot : public RobotPose {
 private:
     unsigned int _shell;
```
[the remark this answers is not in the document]
JNeiger: I assume so? It's old code and doesn't blow up so I don't see why not

soccer/vision/VisionFilter.cpp
```diff
@@ -0,0 +1,133 @@
...
+    threadEndLock.lock();
+    threadEnd = true;
+    threadEndLock.unlock();
+
+    // Wait for it to die
+    worker.join();
+}
+
+void VisionFilter::addFrames(std::vector<CameraFrame>& frames) {
+    frameLock.lock();
```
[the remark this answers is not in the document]
JNeiger: Double check my implementation for the multiple lock case. It feels wrong

soccer/vision/VisionFilter.hpp
```diff
@@ -0,0 +1,66 @@
...
+private:
+    void workerThread();
+
+    std::thread worker;
+
+    std::mutex worldLock;
+    World world;
+
+    std::mutex threadEndLock;
+    bool threadEnd;
```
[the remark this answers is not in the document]
JNeiger: Is `memory_order_seq_cst` the correct memory order to use?

soccer/vision/ball/BallBounce.hpp
JNeiger: I think I'm going to keep it a class just to hid some of the helper functions

soccer/vision/ball/KalmanBall.cpp
JNeiger: ooo. I didn't realize that was a thing

JNeiger:
Thanks for the comments Albert!

soccer/vision/VisionFilter.cpp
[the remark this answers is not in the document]
ashaw596: Lockguard is just syntatic sugar. You can use {} to set the same scope as the manual lock and unlock. Just makes sure you can't forget. {std::lockguard guard(frameLock); do things... }.
 You're double locks look fine. You could use scoped_lock if you want to be safer if you try to double lock somewhere else.

soccer/vision/VisionFilter.hpp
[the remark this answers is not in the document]
ashaw596: Um. I think that's safe for pretty much most uses. But maybe slow in general. For this case. Bool is atomic reguardless in most any system so it probs doesn't really matter.

soccer/vision/ball/BallBounce.hpp
ashaw596: Well if you just want to hide it namespaces (anonymous if you really want to hide it from any other file) are great.

soccer/vision/VisionFilter.cpp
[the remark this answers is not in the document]
JNeiger: Unfortunately, scoped_lock is C++17 and I think we're still on C++11 (or 14)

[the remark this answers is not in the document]
ashaw596: Oh right. Forgot we never ended up updating it. You guys should update! 20 is getting finalized =P My 2 year our branch is still around. https://github.com/RoboJackets/robocup-software/tree/albert/c++17 We're on 14 btw.

soccer/vision/ball/CameraBall.cpp
```diff
@@ -0,0 +1,34 @@
...
+
+RJ::Time CameraBall::getTimeCaptured() const {
+    return timeCaptured;
+}
+
+Geometry2d::Point CameraBall::getPos() const {
+    return pos;
+}
+
+CameraBall CameraBall::CombineBalls(const std::list<CameraBall>& balls) {
```
kylestach: Why does this take a `std::list` instead of a front/back iterator?

soccer/vision/camera/Camera.cpp
```diff
@@ -0,0 +1,389 @@
...
+    }
+}
+
+void Camera::updateRobotsAKF(RJ::Time calcTime,
+                             const std::list<CameraRobot>& singleRobotList,
+                             const WorldRobot& previousWorldRobot,
+                             std::list<KalmanRobot>& singleKalmanRobotList) {
+    
+    std::list<CameraRobot> singleRobotList2;
+    std::copy(singleRobotList.begin(), singleRobotList.end(), std::back_inserter(singleRobotList2));
```
kylestach: Unnecessary copies. Iterator pls?

soccer/vision/camera/Camera.cpp
```diff
@@ -0,0 +1,389 @@
...
+        CameraRobot avgRobot = CameraRobot::CombineRobots(singleRobotList2);
+        singleKalmanRobotList.emplace_back(cameraID, calcTime, avgRobot, previousWorldRobot);
+
+        return;
+    }
+
+    // Average everything and add as measuremnet
+    CameraRobot avgRobot = CameraRobot::CombineRobots(singleRobotList2);
+    // Kinda cheating, but we are only keeping a single element in the list
+    singleKalmanRobotList.front().predictAndUpdate(calcTime, avgRobot);
```
kylestach: Should you remove the rest of the list if you're going to ignore the other estimates in the future? (Or set them to unhealthy and let the autoremoval code take care of it?)

soccer/vision/robot/CameraRobot.cpp
```diff
@@ -0,0 +1,49 @@
...
+
+double CameraRobot::getTheta() const {
+    return theta;
+}
+
+int CameraRobot::getRobotID() const {
+    return robotID;
+}
+
+CameraRobot CameraRobot::CombineRobots(const std::list<CameraRobot>& robots) {
```
kylestach: If we're combining robot measurements, are we going to do anything about the covariance estimate of the measurement step? Multiple measurements resolving as the same robot should be less certain, _probably_. Also _probably_ a pain to implement (you'd need to keep track of actual covariance instead of just variance everywhere), so it's _probably_ not worth bothering. Just a thought though.

soccer/vision/filter/KalmanFilter2D.cpp
```diff
@@ -0,0 +1,116 @@
...
+    // The noise can be propogated through the model resulting in a process noise
+    // of the form
+    //
+    //  [1/3 T^3     1/2 T^2] * sigma^2
+    //  [1/2 T^2           T]
+    // Where sigma is the standard deviation of the process noise
+    // the change in velocity over one time step should be around sqrt(T * sigma^2)
+    // Note: T is the sample period
+    // Taken from Tiger's AutoRef. Most likely found through integration of error through the
+    // state matrices
```
kylestach: More information here: https://en.wikipedia.org/wiki/Discretization#Discretization_of_process_noise. This is basically just an expansion of the matrix exponential.

soccer/vision/camera/Camera.cpp
```diff
@@ -0,0 +1,389 @@
...
+            CameraBall avgBall = CameraBall::CombineBalls(measurementBalls);
+            kalmanBall.predictAndUpdate(calcTime, avgBall);
+
+        // There aren't any measurements so just predict
+        } else {
+            kalmanBall.predict(calcTime);
+        }
+    }
+
+    // Any balls not used, create a kalman ball at that position
```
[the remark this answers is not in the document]
JNeiger: Yep, if there are two measurements of the same physical object, really close together, they would both be used to initialize the filter instead of doing a init then update.

soccer/vision/camera/Camera.cpp
```diff
@@ -0,0 +1,389 @@
...
+            kalmanBallList.emplace_back(cameraID, calcTime, cameraBall, previousWorldBall);
+        }
+    }
+}
+
+void Camera::updateBallsAKF(RJ::Time calcTime,
+                            const std::vector<CameraBall> ballList,
+                            const WorldBall& previousWorldBall) {
+    std::list<CameraBall> ballList2;
+    std::copy(ballList.begin(), ballList.end(), std::back_inserter(ballList2));
```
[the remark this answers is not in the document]
JNeiger: Whoops, all of these copies were from a previous iteration that I forgot to remove

soccer/vision/camera/Camera.cpp
JNeiger: There should only be a single item in the list. We only add one if there's nothing in the list, then we only refer to the first one while it's still valid. 

In the case that it swaps from MHEKF to the AKF method, all of the other filters should die within a half a second, and this one should move to the correct state somewhere within a second or two. I think it's reasonable to assume that if someone is swapping the config values, it's ok if the estimates are bad for a short time.

soccer/vision/camera/World.cpp
```diff
@@ -0,0 +1,221 @@
...
+    : cameras(*VisionFilterConfig::max_num_cameras),
+      robotsYellow(Num_Shells, WorldRobot()),
+      robotsBlue(Num_Shells, WorldRobot()) {}
+
+void World::updateWithCameraFrame(RJ::Time calcTime, const std::vector<CameraFrame>& newFrames) {
+    calcBallBounce();
+
+    std::vector<bool> cameraUpdated(*VisionFilterConfig::max_num_cameras, false);
+
+    // TODO: Take only the newest frame if 2 come in for the same camera
```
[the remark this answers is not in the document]
JNeiger: It should be about 60 hz. The filters are running at 100 hz.

I'm not sure I see how ignoring frames will lead to anomalies though. We would just be skipping the update phase so it'll never "fix" the estimation, but it shouldn't cause any major problems.

Additionally, if this happens where two frames do arrive within a small period in time due to network/SSL Vision lag on the older one, is it even a good idea to run the update step on both measurements at the current time since one will be at least 1/60th of a second out of date? There are filters that take that into account and do some cool stuff to account for that, but is it really worth it since the models themselves are not close to the physical processes they represent?

In the current state without removing old frames, it's possible that we may do 2 predicts and updates in a single frame which would cause some weird problems. I'll do a quick test tonight to see how often this actually happens.

(This isn't implemented right now, but I should probably be rejecting any out of order frames anyways)

soccer/vision/ball/CameraBall.cpp
JNeiger: I'll be honest, it's more probably personal preference. I started in C98 so this feels better to me. I also can't say that I've seen people pass iterators around instead of the list itself in anything except the standard libraries and things that are operations directly on lists (sorting, search etc).

soccer/vision/robot/CameraRobot.cpp
JNeiger: Probably not this time. It's there if anyone wants to go back and play with it, but I don't really feel like trying to do the math and figure it out

soccer/vision/ball/CameraBall.cpp
kylestach: The only reason I ask is that you do a lot of copy operations later to turn vectors into lists so you can pass them to this function, which isn't great. Alternatively I suppose you could pass in a reference to a collection of arbitrary type via templates, but iterators are more idiomatic.

JNeiger: I'll probably just swap them over to vectors to match. This is part of the whole copy fiasco that I forgot about

kylestach: It looks like they're lists for the sake of arbitrary removal from the container, right? (although now that I think about it, vectors are still probably better for any sane estimate count - I suppose if big-O is coming into play here something is _seriously_ wrong).

soccer/vision/robot/CameraRobot.cpp
kylestach: Yeah, makes sense. :+1:

JNeiger:
I'm going to merge this soon unless there's any other complaints.

soccer/vision/ball/CameraBall.cpp
ashaw596: Yeah. You can only take in iterators if you have a templated library/functions. Vector is faster in most every use case of reasonable sizes tho!!!

ashaw596: Yeah. You can only take in iterators if you have a templated library/functions. Vector is faster in most every use case of reasonable sizes tho!!!

ashaw596: Yeah. You can only take in iterators if you have a templated library/functions. Vector is faster in most every use case of reasonable sizes tho!!!

ashaw596: Yeah. You can only take in iterators if you have a templated library/functions. Vector is faster in most every use case of reasonable sizes tho!!!

ashaw596: Yeah. You can only take in iterators if you have a templated library/functions. Vector is faster in most every use case of reasonable sizes tho!!!

ashaw596: Yeah. You can only take in iterators if you have a templated library/functions. Vector is faster in most every use case of reasonable sizes tho!!!

ashaw596: Yeah. You can only take in iterators if you have a templated library/functions. Vector is faster in most every use case of reasonable sizes tho!!!

soccer/vision/VisionFilter.cpp
```diff
@@ -0,0 +1,131 @@
...
+        state->ball.vel = wb.getVel();
+        state->ball.time = wb.getTime();
+    } else {
+        state->ball.valid = false;
+    }
+}
+
+void VisionFilter::fillRobotState(SystemState* state, bool usBlue) {
+    std::lock_guard<std::mutex> lock(worldLock);
+    std::vector<WorldRobot> yellowTeam = world.getRobotsYellow();
```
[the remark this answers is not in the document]
JNeiger: So if I switch this to a `const &` it won't copy the vector being returned?

soccer/vision/camera/World.cpp
```diff
@@ -0,0 +1,221 @@
...
+    }
+
+    // If we haven't had a kick in a while, reset out kick estimate
+    if (bestKickEstimate.getIsValid() && 
+        RJ::Seconds(bestKickEstimate.getKickTime() - calcTime) > RJ::Seconds(*slow_kick_timeout + *fast_kick_timeout)) {
+        bestKickEstimate = KickEvent();
+    }
+}
+
+WorldBall World::getWorldBall() const {
```
ashaw596: Do all of these need to return copies?

soccer/vision/ball/BallBounce.cpp
```diff
@@ -0,0 +1,279 @@
...
+    robot_mouth_lin_dampen = new ConfigDouble(cfg, "VisionFilter/Bounce/robot_mouth_lin_dampen", .3);
+    robot_body_angle_dampen = new ConfigDouble(cfg, "VisionFilter/Bounce/robot_body_angle_dampen", 0);
+    robot_mouth_angle_dampen = new ConfigDouble(cfg, "VisionFilter/Bounce/robot_mouth_angle_dampen", 0);
+}
+
+bool BallBounce::CalcBallBounce(const KalmanBall& ball,
+                                const std::vector<WorldRobot>& yellowRobots,
+                                const std::vector<WorldRobot>& blueRobots,
+                                Geometry2d::Point& outNewVel) {
+    std::vector<WorldRobot> robots = yellowRobots;
```
ashaw596: WorldRobot seems to be a fairly large type and you're making alot of copies. I would try to see if you can get rid of most of the unnecessary copies with refferences.
Do you really need copies of all the Robots here?
Couldn't this be a vector of const pointers? or a vector<std::reference_wrapper>

These copies seem to be happening alot rn.

soccer/vision/camera/CameraFrame.hpp
```diff
@@ -0,0 +1,38 @@
...
+
+#include <Utils.hpp>
+
+#include "vision/ball/CameraBall.hpp"
+#include "vision/robot/CameraRobot.hpp"
+
+/**
+ * Simple non-protobuf object representing a camera frame
+ */
+class CameraFrame {
```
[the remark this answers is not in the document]
JNeiger: Sweet, I was worried those were too much of a C-ism to use. How should I handle the `vector.emplace_back` used in a few places? Should I add the constructor for the struct or just swap to a `push_back`?

soccer/vision/filter/KalmanFilter.hpp
```diff
@@ -0,0 +1,88 @@
...
+    void predict();
+
+    /**
+     * Predicts with update
+     * z_k must be set with the observation
+     */
+    void predictWithUpdate();
+
+protected:
+    Eigen::VectorXd x_k1_k1;
```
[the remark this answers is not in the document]
JNeiger: I can't think of a way off the top of my head. The robot kalman filters have 6 states while the ball only has 4. I can copy and paste the implementation into the two child classes but that feels counter intuitive.

If there's something like a "virtual member", where I can declare it in a child class but still have the main implementation in the parent, that would work too.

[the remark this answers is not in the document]
ashaw596: Could you template it?

soccer/vision/kick/estimator/ChipKickEstimator.hpp
```diff
@@ -0,0 +1,6 @@
+#pragma once
+
+/**
+ * Takes a kick event and tries to match a parabolic model to the trajectory
+ */
+class ChipKickEstimator {};
```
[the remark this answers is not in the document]
JNeiger: The original intent was to estimate the velocity/trajectory of a kicked ball, but due to time constraints and the fact that this was more of a stretch goal, it wasn't completed. I would like for it to be completed at a later time though.

There's two ways I could go with this. I could either just delete it and let whomever wants to do this figure out their own implementation or I could build a good skeleton for them to use that would meld well into the current structure of the filter as a whole. As an aside, I've been tossing around whether this should be used as a new member type project also

soccer/Processor.cpp
```diff
@@ -196,77 +196,54 @@ bool Processor::joystickValid() const {
...
-                                            frame->frame_number());
-        }
+        frames.emplace_back(time, frame->camera_id(), ballObservations, yellowObservations, blueObservations);
     }
 
-    _ballTracker->run(ballObservations, &_state);
+    _vision->addFrames(frames);
+
+    // Fill the list of our robots/balls based on _blue
+    _vision->fillBallState(&_state);
```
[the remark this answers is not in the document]
JNeiger: It looks like there's some stuff in systemState related to the drawing interface. We would have to go through everything else and make sure it won't cause problems

soccer/vision/VisionFilter.cpp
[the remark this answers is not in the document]
JNeiger: I would make the getter `std::vector<WorldRobot>& getRobotsYellow() const` and then this line would be `const std::vector<WorldRobot>& yellowTeam = world.getRobotsYellow()`.

soccer/vision/filter/KalmanFilter2D.cpp
```diff
@@ -0,0 +1,117 @@
...
+
+#include <cmath>
+
+#include "vision/util/VisionFilterConfig.hpp"
+
+REGISTER_CONFIGURABLE(KalmanFilter2D)
+
+ConfigDouble* KalmanFilter2D::ball_init_covariance;
+ConfigDouble* KalmanFilter2D::ball_process_noise;
+ConfigDouble* KalmanFilter2D::ball_observation_noise;
```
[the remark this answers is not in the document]
JNeiger: I think it's probably ok in this case just because of how specialized the A,B,C matrices already are.

soccer/vision/kick/detector/FastKickDetector.cpp
```diff
@@ -0,0 +1,125 @@
...
+    // Velocity at the start and end measurements
+    Geometry2d::Point vStart = dpStart / *VisionFilterConfig::vision_loop_dt;
+    Geometry2d::Point vEnd = dpEnd / *VisionFilterConfig::vision_loop_dt;
+
+    // Change in velocity between start and end measurements
+    Geometry2d::Point dv = vEnd - vStart;
+
+    // Acceleration between the start and final velocity
+    // This is weird when the history length is > 3, but it allows you not to have to retune it
+    Geometry2d::Point accel = dv / *VisionFilterConfig::vision_loop_dt;
```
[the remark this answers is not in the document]
JNeiger: Yep, you're right. I forgot to add that when I generalized this from the N=3 case to just N=K

JNeiger:
Im going to merge this Sunday after testing it one last time.

PR #1228, opened 2018-11-18
Added information about Advance Zone Midfielder in Documentation

tjones320:
Added the description about recently added tactic Advance Zone Midfielder into the documentation PlaysTacticSkillsSummary.md in docs.


JNeiger:
I think this should go in the positions table instead of the tactics one to better fit

JNeiger:
I should probably move defense down to that section too. My goal was to try and split up the things like coordinated pass from the higher level ones like submissive defender etc

tjones320:
Alright I'll fix it should be simple

PR #1231, opened 2019-01-13
Mark skills

Alan-Penkar:
Comments added on the mark skill (gameplay/skills/mark.py).


chachmu:
I think you accidentally included a bunch of cmake stuff

soccer/gameplay/plays/restarts/our_free_kick.py
```diff
@@ -16,7 +16,7 @@ class OurFreeKick(standard_play.StandardPlay):
     BumpKickPower = 0.01
     FullKickPower = 1
     MaxShootingAngle = 80
-    # Untested as of now
+    # Untested as of now - I didn't test this either!!!
```
jgkamat: This comment made my day :)

Alan-Penkar:
well sh**, clearly I'm struggling with git. I tried pulling master from rj and creating a new branch- should I do a hard reset and then reintegrate my changes?

soccer/gameplay/plays/restarts/our_free_kick.py
Alan-Penkar: on the one hand - good :), on the other- I didn't mean to include this commit... so there's that

JNeiger:
If it's only a few files, it's probably easier to just do the reset. There probably is a proper way to do this through git itself, but I'm not sure what that would be.

PR #1232, opened 2019-01-15
Remove spaces in project name to fix CMake build

kylestach:
The build was broken on Ubuntu 18.10+, because some tool was generating a directory named after the project name which would then break Make.

To fix this, replace the spaces in the project name with underscores to ensure it'll be a valid token in the Makefile (`gt_robojackets_robocup`).


kylestach:
Never mind, disregard this. Issue was on my end in how I was building it. It works when using the Ninja backend, which is the default, but breaks using the `make` backend.

kylestach:
@jgkamat the make backend seems to work on my system with this change added, so I suppose there's some value in merging this. Re-opened.

PR #1233, opened 2019-01-21
Update to friction coefficient

JNeiger:
This is an update to the friction coefficient since I feel like the difference in velocity estimates between the two vision filters creates non-trivial difference.

I created a new branch `friction_estimation` that uses the new vision filter to record the ball velocity and position as a function of time. It should record into the `test.csv` file in the `robocup-software` root directory. I then held the ball outside vision and rolled it across the field. I killed soccer once it reached the other side and bounced off the wall. Inside the `csv`, I clipped the data to remove the windup on the velocity estimate at the beginning and remove the wall bounce. If you plot the data (speed vs time), you can easily do a best fit line to get the slope.

I ran this 7 times across different directions of the field. I think the outlier across the goal at fast speed was due to how fast it was compared to the distance it had to travel. The velocity basically didn't decrease over the distance and the noise started messing with the best fit.

| Acceleration measured | Description of run |
| ----------------------------- | ----------------------- |
| -0.217 | Directly down the side of the field at medium speed |
| -0.249 | Directly down the side of the field at medium speed |
| -0.208 | Across the field diagonally at medium speed |
| -0.180 | From our corner to their goal at medium speed |
| -0.312 | Across our goal at fast speed |
| -0.189 | Across our goal at slow speed |
| -0.195 | Across our goal at slow speed |

This is the first run's truncated graph with only the windup removed and the bounce at the end removed.
image

The second run's graph is as follows.
image

Averaging everything but that weird outlier results in a acceleration of -.206

This is blocked based on the vision filter merge. I'll add a doc page on how to do this in the future just to make life easier.

### DIFF
--- a/doc/FrictionEstimation.md
+++ b/doc/FrictionEstimation.md
@@ -1,0 +1,34 @@
+# Friction Estimation
+
+Due to the heavy reliance on predicting ball location, a good estimation of friction is needed. Unfortuantly, getting this estimation is sometimes difficult. A seperate branch has been created to directly write a list of positions/velocities as a function of time into a csv file.
+
+## Procedure
+
+1. Checkout the "friction_estimation" branch from robojackets/robocup-software
+2. Remove the ball from the view of the cameras
+3. Start soccer
+4. Wait until the field has updated with the geometry packet (Usually a second or two)
+5. Slowly roll the ball onto the field. It should be fast enough to reach the other side in a somewhat straight line, but slow enough to take a second or two at least.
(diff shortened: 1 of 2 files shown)


JNeiger:
I synced up the sim friction coefficient so it matched the estimate here and it looks about right so I'm going to merge this after the vision filter changes go through.

JNeiger:
Merging since this was highly dependent on the vision filter being used and #1227 just got merged

PR #1236, opened 2019-01-23
Mark skills

Alan-Penkar:
Updated mark and created goalside mark.

### DIFF
--- a/common/Field_Dimensions.hpp
+++ b/common/Field_Dimensions.hpp
@@ -55,6 +55,12 @@ struct Field_Dimensions {
     Geometry2d::Rect OurHalf() const { return _OurHalf; }
     Geometry2d::Rect TheirHalf() const { return _TheirHalf; }
     Geometry2d::Rect FieldRect() const { return _FieldRect; }
+    Geometry2d::Rect OurGoalZoneShapePadded(float padding){
+      Geometry2d::Rect tmp = Geometry2d::Rect(_OurGoalZoneShape);
+      tmp.pad(padding);
+      return tmp;
+    };
+
 
     std::vector<Geometry2d::Line> FieldBorders() const { return _FieldBorders; }
 
--- a/common/Field_Dimensions.hpp
+++ b/common/Field_Dimensions.hpp
@@ -44,23 +44,41 @@ struct Field_Dimensions {
     float FloorWidth() const { return _FloorWidth; }
 
     Geometry2d::Point CenterPoint() const { return _CenterPoint; }
+    
     Geometry2d::Rect OurGoalZoneShape() const {
         return _OurGoalZoneShape;
     }
     Geometry2d::Rect TheirGoalZoneShape() const {
         return _TheirGoalZoneShape;
     }
+
+    /*
+    * Provides a rect that is a padded version of their goalbox
+    * used mostly for movement at the play level
+    * exposed to python via constants.Field
+    */
+    Geometry2d::Rect TheirGoalZoneShapePadded(float padding){
+      Geometry2d::Rect tmp = Geometry2d::Rect(_TheirGoalZoneShape);
+      tmp.pad(padding);
+      return tmp;
+    };
+
     Geometry2d::Segment OurGoalSegment() const { return _OurGoalSegment; }
     Geometry2d::Segment TheirGoalSegment() const { return _TheirGoalSegment; }
     Geometry2d::Rect OurHalf() const { return _OurHalf; }
     Geometry2d::Rect TheirHalf() const { return _TheirHalf; }
     Geometry2d::Rect FieldRect() const { return _FieldRect; }
+    
+    /*
+    * Provides a rect that is a padded version of our goalbox
+    * used mostly for movement at the play level
+    * exposed to python via constants.Field
+    */
     Geometry2d::Rect OurGoalZoneShapePadded(float padding){
       Geometry2d::Rect tmp = Geometry2d::Rect(_OurGoalZoneShape);
       tmp.pad(padding);
       return tmp;
     };
-
 
     std::vector<Geometry2d::Line> FieldBorders() const { return _FieldBorders; }
 
--- a/common/Field_Dimensions.hpp
+++ b/common/Field_Dimensions.hpp
@@ -44,10 +44,10 @@ struct Field_Dimensions {
     float FloorWidth() const { return _FloorWidth; }
 
     Geometry2d::Point CenterPoint() const { return _CenterPoint; }
-    Geometry2d::CompositeShape OurGoalZoneShape() const {
+    Geometry2d::Rect OurGoalZoneShape() const {
         return _OurGoalZoneShape;
     }
-    Geometry2d::CompositeShape TheirGoalZoneShape() const {
+    Geometry2d::Rect TheirGoalZoneShape() const {
         return _TheirGoalZoneShape;
     }
     Geometry2d::Segment OurGoalSegment() const { return _OurGoalSegment; }
@@ -121,15 +121,13 @@ struct Field_Dimensions {
     void updateGeometry() {
         _CenterPoint = Geometry2d::Point(0.0, _Length / 2.0);
 
-        _OurGoalZoneShape = Geometry2d::CompositeShape();
-        _OurGoalZoneShape.add(std::make_shared<Geometry2d::Rect>(
-          Geometry2d::Point(-_PenaltyLongDist / 2, 0), 
-          Geometry2d::Point(_PenaltyLongDist / 2, _PenaltyShortDist)));
+        _OurGoalZoneShape = Geometry2d::Rect(
+          Geometry2d::Point(_PenaltyLongDist / 2, _PenaltyShortDist),
+          Geometry2d::Point(-_PenaltyLongDist / 2, 0));
 
-        _TheirGoalZoneShape = Geometry2d::CompositeShape();
-        _TheirGoalZoneShape.add(std::make_shared<Geometry2d::Rect>(
-            Geometry2d::Point(-_PenaltyLongDist / 2, _Length),
-            Geometry2d::Point(_PenaltyLongDist / 2, _Length - _PenaltyShortDist)));
+        _TheirGoalZoneShape = Geometry2d::Rect(
+          Geometry2d::Point(-_PenaltyLongDist / 2, _Length),
+          Geometry2d::Point(_PenaltyLongDist / 2, _Length - _PenaltyShortDist));
 
         _TheirGoalSegment =
             Geometry2d::Segment(Geometry2d::Point(_GoalWidth / 2.0, _Length),
@@ -176,8 +174,8 @@ private:
     float _FloorWidth;
 
     Geometry2d::Point _CenterPoint;
-    Geometry2d::CompositeShape _OurGoalZoneShape;
-    Geometry2d::CompositeShape _TheirGoalZoneShape;
+    Geometry2d::Rect _OurGoalZoneShape;
+    Geometry2d::Rect _TheirGoalZoneShape;
     Geometry2d::Segment _OurGoalSegment;
     Geometry2d::Segment _TheirGoalSegment;
     Geometry2d::Rect _OurHalf;
--- a/common/Geometry2d/Rect.cpp
+++ b/common/Geometry2d/Rect.cpp
@@ -7,6 +7,13 @@ using namespace std;
 
 namespace Geometry2d {
 
+//constants used for the rect-segment intersection
+const int INSIDE = 0; // 0000
+const int LEFT = 1;   // 0001
+const int RIGHT = 2;  // 0010
+const int BOTTOM = 4; // 0100
+const int TOP = 8;    // 1000
+
 Shape* Rect::clone() const { return new Rect(*this); }
 
 bool Rect::intersects(const Rect& other) const {
@@ -19,11 +26,6 @@ bool Rect::intersects(const Rect& other) const {
 }
 
 int Rect::CohenSutherlandCode(const Point& other) const{
-    const int INSIDE = 0; // 0000
-    const int LEFT = 1;   // 0001
-    const int RIGHT = 2;  // 0010
-    const int BOTTOM = 4; // 0100
-    const int TOP = 8;    // 1000
     int code;
     double x,y;
     x = other.x();
@@ -43,13 +45,6 @@ int Rect::CohenSutherlandCode(const Point& other) const{
 }
 
 bool Rect::intersects_(const Segment& other) const{
-    //These constants are needed in here and in the CohenSutherlandCode method - is there
-    //a place where we should make these definitions globally?
-    const int INSIDE = 0; // 0000
-    const int LEFT = 1;   // 0001
-    const int RIGHT = 2;  // 0010
-    const int BOTTOM = 4; // 0100
-    const int TOP = 8;    // 1000
 
     //Code aggressively borrowed from wikipedia entry Cohen-Sutherland Line Clipping
     Point p0 = other.pt[0];
--- a/common/Geometry2d/Rect.cpp
+++ b/common/Geometry2d/Rect.cpp
@@ -18,6 +18,111 @@ bool Rect::intersects(const Rect& other) const {
     }
 }
 
+int Rect::CohenSutherlandCode(const Point& other) const{
+    const int INSIDE = 0; // 0000
+    const int LEFT = 1;   // 0001
+    const int RIGHT = 2;  // 0010
+    const int BOTTOM = 4; // 0100
+    const int TOP = 8;    // 1000
+    int code;
+    double x,y;
+    x = other.x();
+    y = other.y();
+
+    code = INSIDE;          // initialised as being inside of [[clip window]]
+    if (x < minx())           // to the left of clip window
+        code |= LEFT;
+    else if (x > maxx())      // to the right of clip window
+        code |= RIGHT;
+    if (y < miny())           // below the clip window
+        code |= BOTTOM;
+    else if (y > maxy())      // above the clip window
+        code |= TOP;
+
+    return code;
+}
+
+bool Rect::intersects_(const Segment& other) const{
+    //These constants are needed in here and in the CohenSutherlandCode method - is there
+    //a place where we should make these definitions globally?
+    const int INSIDE = 0; // 0000
+    const int LEFT = 1;   // 0001
+    const int RIGHT = 2;  // 0010
+    const int BOTTOM = 4; // 0100
+    const int TOP = 8;    // 1000
+
+    //Code aggressively borrowed from wikipedia entry Cohen-Sutherland Line Clipping
+    Point p0 = other.pt[0];
+    double x0 = p0.x();
+    double y0 = p0.y();
+    Point p1 = other.pt[1];
+    double x1 = p1.x();
+    double y1 = p1.y();
+
+    int outcode0 = CohenSutherlandCode(p0);
+    int outcode1 = CohenSutherlandCode(p1);
+    bool accept = false;
+
+    while (true) {
+        if (!(outcode0 | outcode1)) {
+            // bitwise OR is 0: both points inside window; trivially accept and exit loop
+            accept = true;
+            break;
+        } else if (outcode0 & outcode1) {
+            // bitwise AND is not 0: both points share an outside zone (LEFT, RIGHT, TOP,
+            // or BOTTOM), so both must be outside window; exit loop (accept is false)
+            break;
+        } else {
+            // failed both tests, so calculate the line segment to clip
+            // from an outside point to an intersection with clip edge
+            double x, y;
+
+            // At least one endpoint is outside the clip rectangle; pick it.
+            int outcodeOut = outcode0 ? outcode0 : outcode1;
+
+            // Now find the intersection point;
+            // use formulas:
+            //   slope = (y1 - y0) / (x1 - x0)
+            //   x = x0 + (1 / slope) * (ym - y0), where ym is ymin or ymax
+            //   y = y0 + slope * (xm - x0), where xm is xmin or xmax
+            // No need to worry about divide-by-zero because, in each case, the
+            // outcode bit being tested guarantees the denominator is non-zero
+            if (outcodeOut & TOP) {           // point is above the clip window
+                x = x0 + (x1 - x0) * (maxy() - y0) / (y1 - y0);
+                y = maxy();
+            } else if (outcodeOut & BOTTOM) { // point is below the clip window
+                x = x0 + (x1 - x0) * (miny() - y0) / (y1 - y0);
+                y = miny();
+            } else if (outcodeOut & RIGHT) {  // point is to the right of clip window
+                y = y0 + (y1 - y0) * (maxx() - x0) / (x1 - x0);
+                x = maxx();
+            } else if (outcodeOut & LEFT) {   // point is to the left of clip window
+                y = y0 + (y1 - y0) * (minx() - x0) / (x1 - x0);
+                x = minx();
+            }
+
+            // Now we move outside point to intersection point to clip
+            // and get ready for next pass.
+            if (outcodeOut == outcode0) {
+                x0 = x;
+                y0 = y;
+                Point tmp = Point(x0, y0);
+                outcode0 = CohenSutherlandCode(tmp);
+            } else {
+                x1 = x;
+                y1 = y;
+                Point tmp = Point(x1, y1);
+                outcode1 = CohenSutherlandCode(tmp);
+            }
+        }
+    }
+
+    return accept;
+
+}
+
+
+
 bool Rect::containsRect(const Rect& other) const {
     // return other.pt[0].inRect(*this) && other.pt[1].inRect(*this);
     return this->containsPoint(other.pt[0]) && this->containsPoint(other.pt[1]);
--- a/common/Geometry2d/Rect.cpp
+++ b/common/Geometry2d/Rect.cpp
@@ -133,13 +133,6 @@ std::vector<Point> Rect::corners(){
     return tmp;
 }
 
-std::vector<Point> Rect::pointList(){
-    std::vector<Point> tmp;
-    tmp.push_back(pt[0]);
-    tmp.push_back(pt[1]);
-    return tmp;
-}
-
 bool Rect::containsRect(const Rect& other) const {
     // return other.pt[0].inRect(*this) && other.pt[1].inRect(*this);
     return this->containsPoint(other.pt[0]) && this->containsPoint(other.pt[1]);
--- a/common/Geometry2d/Rect.cpp
+++ b/common/Geometry2d/Rect.cpp
@@ -44,7 +44,7 @@ int Rect::CohenSutherlandOutCode(const Point& other) const{
     return code;
 }
 
-std::tuple<bool, std::vector<Point> > Rect::intersects_(const Segment& other) const{
+std::tuple<bool, std::vector<Point> > Rect::intersects(const Segment& other) const{
 
     //Code aggressively borrowed from wikipedia entry Cohen-Sutherland Line Clipping
     Point p0 = other.pt[0];
--- a/common/Geometry2d/Rect.cpp
+++ b/common/Geometry2d/Rect.cpp
@@ -106,6 +106,7 @@ bool Rect::intersects_(const Segment& other, Point* intr1, Point* intr2) const{
                 y0 = y;
                 Point pt = Point(x0, y0);
                 outcode0 = CohenSutherlandCode(pt);
+                //Save point iff it is inside the Rect
                 if (outcode0==INSIDE){
                     **nextPoint = pt;
                     nextPoint = &intr2;
@@ -116,6 +117,7 @@ bool Rect::intersects_(const Segment& other, Point* intr1, Point* intr2) const{
                 Point pt = Point(x0, y0);
                 **nextPoint = Point(x0, y0);
                 outcode1 = CohenSutherlandCode(pt);
+                //Save point iff it is inside the Rect
                 if (outcode1==INSIDE){
                     **nextPoint = pt;
                     nextPoint = &intr2;
@@ -123,9 +125,7 @@ bool Rect::intersects_(const Segment& other, Point* intr1, Point* intr2) const{
             }
         }
     }
-    std::cout << "Returning" << std::endl;
     return accept;
-
 }
 
 std::vector<Point> Rect::corners(){
--- a/common/Geometry2d/Rect.cpp
+++ b/common/Geometry2d/Rect.cpp
@@ -8,11 +8,11 @@ using namespace std;
 namespace Geometry2d {
 
 //constants used for the rect-segment intersection
-const int INSIDE = 0; // 0000
-const int LEFT = 1;   // 0001
-const int RIGHT = 2;  // 0010
-const int BOTTOM = 4; // 0100
-const int TOP = 8;    // 1000
+const int INSIDE = 0x00; // 0000
+const int LEFT = 0x01;   // 0001
+const int RIGHT = 0x02;  // 0010
+const int BOTTOM = 0x04; // 0100
+const int TOP = 0x08;    // 1000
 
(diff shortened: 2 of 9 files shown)


soccer/gameplay/skills/goalside_mark.py
```diff
@@ -0,0 +1,194 @@
...
+
+        self.add_transition(behavior.Behavior.State.start,
+                            behavior.Behavior.State.running, lambda: True,
+                            "immediately")
+        self.kick_eval = robocup.KickEvaluator(main.system_state())
+
+    def execute_running(self):
+        #pylint: disable=no-member
+        #Skill does nothing if mark point isn't given AND the ball or robot to mark can't be found
+        self.kick_eval = robocup.KickEvaluator(main.system_state())
```
JNeiger: I don't think you need to create a new kick eval object every frame. You can just use the same one and clear the excluded robots whenever

soccer/gameplay/skills/goalside_mark.py
```diff
@@ -0,0 +1,194 @@
...
+                          constants.Field.PenaltyShortDist + offset))
+        right_seg = robocup.Segment(
+            robocup.Point(constants.Field.PenaltyLongDist / 2 + offset, 0),
+            robocup.Point(constants.Field.PenaltyLongDist / 2 + offset,
+                          constants.Field.PenaltyShortDist + offset))
+        top_seg = robocup.Segment(
+            robocup.Point(-constants.Field.PenaltyLongDist / 2,
+                          constants.Field.PenaltyShortDist + offset),
+            robocup.Point(constants.Field.PenaltyLongDist / 2,
+                          constants.Field.PenaltyShortDist + offset))
```
JNeiger: I see why this is here, but I wish there was a better way to do this (Intersection of a segment/line with the goal box with some offset). If you are comfortable making some other changes to the C++ interface to allow for this that would be awesome. If not, I can add it sometime.

Basically, the [field](https://github.com/RoboJackets/robocup-software/blob/master/soccer/gameplay/constants.py#L34) in python refers back to it's C++ [counterpart](https://github.com/RoboJackets/robocup-software/blob/f0cf06638ee427563ca7162bbbd65809c30f59f2/common/Field_Dimensions.hpp#L47). Here, there is something called our goal zone shape which is a composite shape. In short a [composite shape](https://github.com/RoboJackets/robocup-software/blob/master/common/Geometry2d/CompositeShape.hpp) is a list of shapes which are or'ed together. In old times (like a year or two ago), the defensive area was two arc's with a rectangle in the middle so this made sense. When we redid it for rectangles, we left the composite shape wrapper for some reason. 

Anyways, now, the composite shape just contains a single [rectangle](https://github.com/RoboJackets/robocup-software/blob/f0cf06638ee427563ca7162bbbd65809c30f59f2/common/Field_Dimensions.hpp#L127). The [rectangle](https://github.com/RoboJackets/robocup-software/blob/master/common/Geometry2d/Rect.hpp) class doesn't allow for segment intersection, but that's a good thing to add in my mind. If we overload the `bool intersects` function and take in a segment, very close to how it's done in the [segment `intersect` function](https://github.com/RoboJackets/robocup-software/blob/master/common/Geometry2d/Segment.hpp#L51), I think it'll work out pretty well. In terms of actual implementation, you should be able to follow the same way they do the [`nearPoint` function](https://github.com/RoboJackets/robocup-software/blob/master/common/Geometry2d/Rect.cpp#L105) where they create the 4 segments and just call the segment functions. It's up to you, but I think returning both intersections would be the best bet and let the user figure out which to use (and maybe return the number of intersect points). Finally, to access in the python side, you'll have to add the function call to [this location in robocup-py](https://github.com/RoboJackets/robocup-software/blob/master/common/Geometry2d/Rect.cpp#L105), but you'll have to follow the setup in the [segment intersect function](https://github.com/RoboJackets/robocup-software/blob/master/soccer/gameplay/robocup-py.cpp#L683) (and [here](https://github.com/RoboJackets/robocup-software/blob/master/soccer/gameplay/robocup-py.cpp#L242)) to fully deal with the weird python <-> C++ conversion issues. A test or two should be added into [this file](https://github.com/RoboJackets/robocup-software/blob/master/common/Geometry2d/RectTest.cpp) also.

In the future, composite shape should probably be removed, and just a rectangle should be used instead, but that may be out of scope for this.

soccer/gameplay/skills/goalside_mark.py
```diff
@@ -0,0 +1,194 @@
...
+
+        #Sets the position to mark as the given mark position 
+        #or robot position if no mark position is given
+        pos = self.robot.pos
+        mark_pos = self.mark_point if self.mark_point is not None else self.mark_robot.pos
+
+        #Find best shot point from threat
+        self.kick_eval.add_excluded_robot(self.robot) #FIX: Should we really exclude all of our robots but the goalie?
+        shot_pt, shot_score = self.kick_eval.eval_pt_to_our_goal(mark_pos)
+        self.kick_eval.excluded_robots.clear()
```
JNeiger: I think these are just copy paste from the caller function?

soccer/gameplay/skills/goalside_mark.py
```diff
@@ -0,0 +1,194 @@
...
+
+        
+        #Finds the line from the mark position to the shot point and creates a line between them
+        #removing the overlap with the ball on one side and robot on the other
+        #This assumes even with mark position parameter that there is a robot there to avoid
+        mark_line_dir = (mark_pos - shot_pt).normalized()
+        mark_line, shot_pt = self.get_mark_line()
+        print(mark_line)
```
JNeiger: In the final version, prints should probably be removed so we don't spam the console too much

soccer/gameplay/skills/goalside_mark.py
```diff
@@ -0,0 +1,194 @@
...
+        #Sets target point to place on line defined by ratio- 
+        #   - 0 being close to ball, 1 close to mark pt
+        self._target_point = None
+        if mark_line_dist > self.mark_line_thresh:
+            self._target_point = mark_line.nearest_point(pos)
+        else:
+            if self.mark_point is None:
+                adjusted_mark_pos = mark_pos - mark_line_dir * 2 * constants.Robot.Radius
+            else:
+                adjusted_mark_pos = mark_pos - mark_line_dir * constants.Ball.Radius
```
JNeiger: These two are also calculated already in the `get_mark_line()` function. It feels weird to re-calculate them

soccer/gameplay/skills/goalside_mark.py
JNeiger: Oh, and to add the offsets, you can use the [`expand`](https://github.com/RoboJackets/robocup-software/blob/master/common/Geometry2d/Rect.hpp#L67) function of rectangles with the [mixx/maxx/miny/maxy](https://github.com/RoboJackets/robocup-software/blob/master/common/Geometry2d/Rect.hpp#L73) to create the X meter offsets outside the given rectangle.

soccer/gameplay/skills/goalside_mark.py
Alan-Penkar: Good call- changed this trying to debug the kickevaluator and never changed it back

soccer/gameplay/skills/goalside_mark.py
Alan-Penkar: I'll give this a go - but expect to do a pretty heavy review on my C++ skills.

soccer/gameplay/skills/goalside_mark.py
JNeiger: It's not bad by itself, but if a big chunk of code is being copy pasted and the same calculations are being done, it's a good idea to find a way to merge them. That way if someone wants to make a tweak later on, they don't have to try and edit two places. I'll have to look at the code a little more to see if that's the case here, but it felt a little deja vu looking at the two functions

common/Geometry2d/Rect.cpp
```diff
@@ -18,6 +18,111 @@ bool Rect::intersects(const Rect& other) const {
...
+        code |= BOTTOM;
+    else if (y > maxy())      // above the clip window
+        code |= TOP;
+
+    return code;
+}
+
+bool Rect::intersects_(const Segment& other) const{
+    //These constants are needed in here and in the CohenSutherlandCode method - is there
+    //a place where we should make these definitions globally?
```
JNeiger: If you don't need them outside these two functions, putting them at the top of this file should be fine

common/Geometry2d/Rect.cpp
```diff
@@ -18,6 +18,111 @@ bool Rect::intersects(const Rect& other) const {
...
+    const int INSIDE = 0; // 0000
+    const int LEFT = 1;   // 0001
+    const int RIGHT = 2;  // 0010
+    const int BOTTOM = 4; // 0100
+    const int TOP = 8;    // 1000
+
+    //Code aggressively borrowed from wikipedia entry Cohen-Sutherland Line Clipping
```
JNeiger: I would just link it here

common/Geometry2d/Rect.cpp
```diff
@@ -18,6 +18,111 @@ bool Rect::intersects(const Rect& other) const {
...
+        code |= RIGHT;
+    if (y < miny())           // below the clip window
+        code |= BOTTOM;
+    else if (y > maxy())      // above the clip window
+        code |= TOP;
+
+    return code;
+}
+
+bool Rect::intersects_(const Segment& other) const{
```
JNeiger: You shouldn't need the underscore as C++ will allow method overloads depending on the arguments

common/Geometry2d/Rect.cpp
```diff
@@ -18,6 +18,111 @@ bool Rect::intersects(const Rect& other) const {
     }
 }
 
+int Rect::CohenSutherlandCode(const Point& other) const{
```
JNeiger: I'm on the fence about the naming for this function. At first glance it felt like just saying it's the Cohen Sutherland C++ code, but after looking at the algorithm, I see it's referring to the out code. Making it `CohenSutherlandOutCode`, `ComputeOutCode` etc may make it clearer.

common/Geometry2d/Rect.cpp
```diff
@@ -18,6 +18,111 @@ bool Rect::intersects(const Rect& other) const {
...
+    const int INSIDE = 0; // 0000
+    const int LEFT = 1;   // 0001
+    const int RIGHT = 2;  // 0010
+    const int BOTTOM = 4; // 0100
+    const int TOP = 8;    // 1000
```
JNeiger: You can use `0x01` to make this a hex literal or `1<<2` if you want to specify it's a bitwise field. Either of those makes it much clearer that you are using them in their binary form

common/Geometry2d/Rect.cpp
```diff
@@ -18,6 +18,111 @@ bool Rect::intersects(const Rect& other) const {
...
+    double y0 = p0.y();
+    Point p1 = other.pt[1];
+    double x1 = p1.x();
+    double y1 = p1.y();
+
+    int outcode0 = CohenSutherlandCode(p0);
+    int outcode1 = CohenSutherlandCode(p1);
+    bool accept = false;
+
+    while (true) {
```
JNeiger: There is no way this can become an infinite loop right? Even with a 0 by 0 rectangle and other possible edge cases?

common/Geometry2d/Rect.hpp
```diff
@@ -76,6 +76,9 @@ class Rect : public Shape {
     bool nearSegment(const Segment& seg, float threshold) const;
 
     bool intersects(const Rect& other) const;
+    bool intersects_(const Segment& other) const;
+
+    int CohenSutherlandCode(const Point& other) const;
```
JNeiger: This function can be private since it shouldn't be used anywhere else

common/Geometry2d/Rect.cpp
```diff
@@ -18,6 +18,111 @@ bool Rect::intersects(const Rect& other) const {
...
+    const int BOTTOM = 4; // 0100
+    const int TOP = 8;    // 1000
+
+    //Code aggressively borrowed from wikipedia entry Cohen-Sutherland Line Clipping
+    Point p0 = other.pt[0];
+    double x0 = p0.x();
+    double y0 = p0.y();
+    Point p1 = other.pt[1];
+    double x1 = p1.x();
+    double y1 = p1.y();
```
JNeiger: If you don't want to deal with writing `other.pt[1]` every single time, you can also do the `xmin/xmax/ymin/ymax` stuff too.

soccer/gameplay/skills/goalside_mark.py
```diff
@@ -69,18 +62,14 @@ def execute_running(self):
         #   - 0 being close to ball, 1 close to mark pt
         self._target_point = None
         if mark_line_dist > self.mark_line_thresh:
-            self._target_point = mark_line.nearest_point(pos)
+            self._target_point = mark_line.nearest_point(self.robot.pos)
```
JNeiger: self.robot can be none at times if we don't have enough robots to fill roles

soccer/gameplay/skills/goalside_mark.py
```diff
@@ -114,6 +103,11 @@ def mark_point(self) -> robocup.Point:
     @mark_point.setter
     def mark_point(self, value: robocup.Point):
         self._mark_point = value
+    
+    # Sets the position to mark as the given mark position 
+    # or robot position if no mark position is given
+    def _reset_mark_pos(self):
+        self.mark_pos = self.mark_point if self.mark_point is not None else self._mark_robot.pos
```
JNeiger: Is it ever possible that `mark_point` will be None and `_mark_robot` will be None?

JNeiger:
Looking good so far!

For the tests, you can edit[ this file](https://github.com/RoboJackets/robocup-software/blob/master/common/Geometry2d/RectTest.cpp). The functions are in the form `TEST([Class], [test_name])`. There's been two ways I've been doing tests. The first is one test case for each specific function I want to test and the other is a test case for each edge case.

For Geometry2d I think we've been doing a test case for each function. Inside that we just test the function against a few "good" results to make sure it performs correct (Both successes and failures), then against every single edge case that it could hit.

In terms of the `EXPECT` stuff, there's TRUE, FALSE, NEAR, LT, GT, EQ are the big ones I use. This is GTest so there should be quite a few examples elsewhere.

You should be able to then run `make tests` to actually run the test suite. If you scroll up, you should see your function under the Rectangle class (IIRC).

common/Geometry2d/Rect.cpp
```diff
@@ -18,6 +25,125 @@ bool Rect::intersects(const Rect& other) const {
...
+    }
+    return accept;
+}
+
+std::vector<Point> Rect::corners(){
+    std::vector<Point> tmp;
+    tmp.push_back(Point(minx(),miny()));
+    tmp.push_back(Point(minx(),maxy()));
+    tmp.push_back(Point(maxx(),maxy()));
+    tmp.push_back(Point(maxx(),miny()));
```
JNeiger: If you do `tmp.emplace_back(minx(), miny())` it's more efficient since it doesn't have to create the Point object, then copy it into memory. It just creates it in memory

common/Geometry2d/Rect.hpp
```diff
@@ -63,19 +63,61 @@ class Rect : public Shape {
...
     bool nearPoint(Point pt, float threshold) const override;
     bool nearSegment(const Segment& seg, float threshold) const;
 
     bool intersects(const Rect& other) const;
+    
+    /*
+    * Calculates intersection point(s) between the rectangle and a line segment
+    * Uses Cohen Sutherland line clipping algorithm
+    */
+    bool intersects_(const Segment& other, Point* intr1, Point* intr2) const;
```
JNeiger: The `intr1` and `intr2` should default to `nullptr`

common/Geometry2d/Rect.cpp
```diff
@@ -18,6 +25,125 @@ bool Rect::intersects(const Rect& other) const {
...
+    double y0 = p0.y();
+    Point p1 = other.pt[1];
+    double x1 = p1.x();
+    double y1 = p1.y();
+
+    int outcode0 = CohenSutherlandCode(p0);
+    int outcode1 = CohenSutherlandCode(p1);
+    
+    //intr1, intr2 = nullptr;
+    Point **nextPoint = &intr1;
```
JNeiger: `intr1` can be a nullptr if it's not passed into the function. This would break the program. You should have two function variables if you really want to keep their state then just set intr1 and intr2 to their correct values if they are non-null at the end

JNeiger: This wouldn't need to be a double pointer then. You could just make it a list and append it

common/Geometry2d/Rect.cpp
```diff
@@ -18,6 +25,125 @@ bool Rect::intersects(const Rect& other) const {
...
+    tmp.push_back(Point(maxx(),miny()));
+    return tmp;
+}
+
+std::vector<Point> Rect::pointList(){
+    std::vector<Point> tmp;
+    tmp.push_back(pt[0]);
+    tmp.push_back(pt[1]);
+    return tmp;
+}
```
JNeiger: I'm not sure if this is a good idea to add. It's already there through the `.pt[0]` and `.pt[1]` values. If it's for debugging and you were trying to print it out or something, it's better to just overload the `<<` operator and do it the correct way

JNeiger: That can be a different pr. You don't have to add that

common/Geometry2d/Rect.cpp
```diff
@@ -35,17 +161,33 @@ bool Rect::hit(const Segment& seg) const {
...
-    pt[1].y() = max(pt[1].y(), p.y());
+    float _minx = minx();
+    float _miny = miny();
+    float _maxx = maxx();
+    float _maxy = maxy();
+
+    pt[0].x() = min(_minx, (float)p.x());
+    pt[0].y() = min(_miny, (float)p.y());
+    pt[1].x() = max(_maxx, (float)p.x());
+    pt[1].y() = max(_maxy, (float)p.y());
```
JNeiger: For this and the pad function, you should make it clear in the hpp that this changes which corners pt1 and pt2 are. This makes it bottom left and upper right always.

common/Field_Dimensions.hpp
```diff
@@ -44,17 +44,41 @@ struct Field_Dimensions {
...
+    /*
+    * Provides a rect that is a padded version of their goalbox
+    * used mostly for movement at the play level
+    * exposed to python via constants.Field
+    */
+    Geometry2d::Rect TheirGoalZoneShapePadded(float padding){
+      Geometry2d::Rect tmp = Geometry2d::Rect(_TheirGoalZoneShape);
+      tmp.pad(padding);
+      return tmp;
+    };
```
JNeiger: I'm on the fence about these two. This struct is built like it's just supposed to be the raw field information from the packet and make it very simple to pull that data. Additionally, this only saves one or two lines in python where it's used. 

The feature is nice, but is that enough reason to add it to a place it might not fit in

common/Geometry2d/Rect.hpp
JNeiger: This can just be named `intersects` and function overloading will determine which one they reach

soccer/gameplay/robocup-py.cpp
```diff
@@ -251,6 +251,29 @@ boost::python::object Segment_segment_intersection(Geometry2d::Segment* self,
     }
 }
 
+boost::python::object Our_goalzone_padded(float padding){
+    Geometry2d::Rect tmp = Geometry2d::Rect(Field_Dimensions::Current_Dimensions.OurGoalZoneShape());
+    tmp.expand(Geometry2d::Point(-Field_Dimensions::Current_Dimensions.PenaltyLongDist()/2 - padding, Field_Dimensions::Current_Dimensions.PenaltyShortDist() + padding));
+    tmp.expand(Geometry2d::Point(Field_Dimensions::Current_Dimensions.PenaltyLongDist()/2 + padding, Field_Dimensions::Current_Dimensions.PenaltyShortDist() + padding));
+    return boost::python::object(tmp);
+}
```
JNeiger: I think this is obe since you made a `pad` function

soccer/gameplay/robocup-py.cpp
```diff
@@ -715,8 +741,12 @@ BOOST_PYTHON_MODULE(robocup) {
         .def("min_y", &Geometry2d::Rect::miny)
         .def("max_x", &Geometry2d::Rect::maxx)
         .def("max_y", &Geometry2d::Rect::maxy)
+        .def("corners", &Geometry2d::Rect::corners)
+        .def("points", &Geometry2d::Rect::pointList)
+        .def("pad", &Geometry2d::Rect::pad)
         .def("near_point", &Geometry2d::Rect::nearPoint)
         .def("intersects_rect", &Geometry2d::Rect::intersects)
+        .def("intersects_segment", &Rect_segment_intersection)
```
JNeiger: Just to make the naming convention the same between pts, lines, and segments, this should flip. IE: `segment_intersection` instead of `intersects_segment`. Feel free to change up  `intersects_rect` as well since you need to change the function it points to. It'll need a wrapper similar to how you do `segment` intersection

soccer/gameplay/robocup-py.cpp
```diff
@@ -681,6 +704,9 @@ BOOST_PYTHON_MODULE(robocup) {
         .def("nearly_equals", &Geometry2d::Point::nearlyEquals)
         .staticmethod("direction");
 
+    class_<std::vector<Geometry2d::Point>>("vector_Point")
+        .def(vector_indexing_suite<std::vector<Geometry2d::Point>>());
```
JNeiger: If this is for the `corners` function, it might be reasonable to just create a wrapper and return and `boost::python::list`. It'll be less "weird" objects to deal with in python.

common/Geometry2d/Rect.hpp
```diff
@@ -63,19 +63,61 @@ class Rect : public Shape {
...
     float maxy() const { return std::max(pt[0].y(), pt[1].y()); }
 
+    /*
+    * The corners() function lists the 4 corners of the rectangle
+    * in a predictable order regardless of the 2 corners defined on
+    * construction.
+    * BottomLeft, TopLeft, TopRight, BottomRight
+    * exposed to python as corners()
+    */
+    std::vector<Point> corners();
```
JNeiger: Is this even being used anywhere?

soccer/gameplay/skills/goalside_mark.py
```diff
@@ -0,0 +1,163 @@
...
+
+        self.add_transition(behavior.Behavior.State.start,
+                            behavior.Behavior.State.running, lambda: True,
+                            "immediately")
+        self.kick_eval = robocup.KickEvaluator(main.system_state())
+        self.mark_pos = None
+        self.adjusted_mark_pos = None
+
+    def execute_running(self):
+        #pylint: disable=no-member
```
JNeiger: Which variable is giving the pylint problem?

soccer/gameplay/skills/goalside_mark.py
```diff
@@ -0,0 +1,163 @@
...
+            self._target_point = self.adjusted_mark_pos - (
+                self.mark_pos -
+                shot_pt).normalized() * self.ratio * mark_line.length()
+
+        #Drawing for simulator
+        main.system_state().draw_circle(self.mark_pos, constants.Robot.Radius * 1.2,
+                                        (0, 127, 255), "Mark")
+
+        #Move robot into position and face the ball
+        self.robot.move_to(self._target_point)
```
JNeiger: This should be a move subbehavior since its there. `SingleRobotCompositeBehavior` allows for a single subbehavior

common/Geometry2d/Rect.cpp
Alan-Penkar: it was getting f***ed up in boost and I didn't really want to figure out how to fix it....

common/Geometry2d/Rect.cpp
Alan-Penkar: That seems doable

Alan-Penkar: refactored to CohenSutherlandOutCode

common/Geometry2d/Rect.cpp
Alan-Penkar: using hex literals

common/Geometry2d/Rect.cpp
Alan-Penkar: Fixed... GTest cases included in latest commit

common/Geometry2d/Rect.hpp
Alan-Penkar: changed

common/Geometry2d/Rect.cpp
Alan-Penkar: meh.

soccer/gameplay/skills/goalside_mark.py
Alan-Penkar: added self.robot is not None to the conditions required to execute_running

soccer/gameplay/skills/goalside_mark.py
Alan-Penkar: I mean... mark_robot could be a string.  But specifically stated that one of these is required in the docstring

common/Geometry2d/Rect.hpp
Alan-Penkar: no longer relevant

common/Geometry2d/Rect.cpp
Alan-Penkar: huh- how did that work exactly?  it's a Point vector, emplace is allowing me to add non-points?
(done)

common/Geometry2d/Rect.cpp
Alan-Penkar: done

common/Geometry2d/Rect.cpp
Alan-Penkar: removed

common/Geometry2d/Rect.cpp
Alan-Penkar: done

common/Geometry2d/Rect.hpp
Alan-Penkar: frankly... I just didn't want to deal with it in boost- how big a deal is this?

soccer/gameplay/robocup-py.cpp
Alan-Penkar: you're right - removed

soccer/gameplay/robocup-py.cpp
Alan-Penkar: I'm not sure why it needs a wrapper?  I just changed the name and it can call the same C++ function.

common/Geometry2d/Rect.hpp
Alan-Penkar: yes, recreationally

common/Geometry2d/Rect.cpp
JNeiger: You basically pass in the direct variables that usually go to the constructor directly into the `emplace_back`. It then forwards those to the real constructor

https://stackoverflow.com/questions/23717151/why-emplace-back-is-faster-than-push-back/23717223#23717223

common/Geometry2d/Rect.hpp
JNeiger: I think it's a pretty big deal, but not earth shattering. It would mess up the function overloading in C++ land. I made a patch for it since I was super familiar with how to do it. I don't think boost allows overloading for python functions so you have to build a second function, that calls the overloaded function.

Here's a link to the [patch](https://gist.github.com/JNeiger/96f2fa64c92cd0736d757dd4bd240122). Basically download the zip. Unzip it into the `robocup-software` directory. Run `git apply fix_underscore.patch` to apply the patch. You'll have to commit this afterwards then.

Feel free to ignore this and try it yourself if you would like, but it sounds like you've fully experienced all the boost::python you wanted to :P

soccer/gameplay/robocup-py.cpp
JNeiger: Long story short, the callbacks boost does to convert a python function call to a C++ function call doesn't really deal with types super well. I believe it uses the type definition of the target function to figure out what to convert the python objects to. If there are two definitions with the same name, it doesn't know which one to pick and dies. If you use the wrapper functions, it picks the correct one

soccer/gameplay/robocup-py.cpp
Alan-Penkar: done

soccer/gameplay/skills/goalside_mark.py
Alan-Penkar: This is not a composite behavior - I should change that?

common/Field_Dimensions.hpp
Alan-Penkar: so- I agree this doesn't totally fit in.  The reason I wanted it wrapped up here was because it directly changes the object you call it on in C++, so trying to do this in python leaves us open to someone accidentally changing the goalbox that gets called in all of python...  Tried it in the bridge file, but because we keep info in an instance of the class, that didn't work right.

common/Geometry2d/Rect.hpp
Alan-Penkar: ha- if you can't tell from the degradation of my commit messages, I think I'm done with C++ now and want to get back to the data driven stuff we've discussed.... trying so hard to close this out.  But I'll refactor this.

Alan-Penkar: done

soccer/gameplay/robocup-py.cpp
Alan-Penkar: got it- it's required because of the overloading only.  It wouldn't have been with the underscore

soccer/gameplay/skills/goalside_mark.py
Alan-Penkar: this is the one thing I'm not sure how to deal with- don't know how pylint works or what its looking for

common/Field_Dimensions.hpp
JNeiger: I didn't even think about that. This is probably ok then

soccer/gameplay/skills/goalside_mark.py
JNeiger: Eh, I was looking at some of the other skills some more and it's like 50-50 whether they do this. It's up to you

common/Geometry2d/RectTest.cpp
```diff
@@ -18,16 +20,63 @@ TEST(Rect, hit) {
...
+    EXPECT_EQ(intersectExample.CohenSutherlandOutCode(Point(1.5,2.5)), 0x0A);
+}*/
+
+TEST(Rect, degenerageCohenCodes){
+    EXPECT_EQ(nullExample.CohenSutherlandOutCode(Point(0,0)), 0x00);
+    EXPECT_EQ(nullExample.CohenSutherlandOutCode(Point(1,0)), 0x02);
+    EXPECT_EQ(nullExample.CohenSutherlandOutCode(Point(-1,0)), 0x01);
+    EXPECT_EQ(nullExample.CohenSutherlandOutCode(Point(0,1)), 0x08);
+    EXPECT_EQ(nullExample.CohenSutherlandOutCode(Point(0,-.1)), 0x04);
+}
```
JNeiger: This function needs to be commented out too since `CohenSutherlandOutCode` is private

soccer/gameplay/skills/goalside_mark.py
JNeiger: So I removed this and ran `make pylint`. There's a bunch of crap that comes up, but nothing similar to the `self.mark_robot` has no member `pos` that came up earlier. Try taking it out and seeing if circle ci blows up.

soccer/gameplay/robocup-py.cpp
JNeiger: This part can be removed now I think

common/Geometry2d/RectTest.cpp
Alan-Penkar: Done

soccer/gameplay/skills/goalside_mark.py
Alan-Penkar: when I run make pylint - the only thing that comes up is related to a network util file that I've definitely never touched

PR #1237, opened 2019-02-03
basic indirect kick

Jason27chan:
One goal of this PR was to separate our indirect restart and our direct restart into two different files. 

This specific restart is meant to be a basic indirect kickoff that we can use if we need a restart and are unsure of how our more complex restarts will work

This depends on the chipper being functional, may want to reconsider that


soccer/gameplay/plays/restarts/basic_indirect.py
```diff
@@ -0,0 +1,53 @@
+import standard_play
+import behavior
+import skills.move
+import skills.pivot_kick
+import constants
+import robocup
+import math
+import main
+import tactics.coordinated_pass
+import evaluation.passing_positioning
```
JNeiger: skills.move, skills.pivot_kick, and evaluation.passing_positioning aren't used

JNeiger: I don't think you need enum either since there aren't any states

soccer/gameplay/plays/restarts/basic_indirect.py
```diff
@@ -0,0 +1,53 @@
...
+
+    def __init__(self, indirect=None):
+        super().__init__(continuous=True)
+
+        self.add_transition(behavior.Behavior.State.start,
+                            behavior.Behavior.State.running, lambda: True,
+                            'immediately')
+
+        # FIXME: this could also be a PivotKick
+        kicker = skills.line_kick.LineKick()
```
JNeiger: I think for any restarts, we've been doing line kick so that we don't have to worry about double touch stuff and it's a little more consistent I think

soccer/gameplay/plays/restarts/basic_indirect.py
```diff
@@ -0,0 +1,53 @@
...
+        self.add_transition(
+            behavior.Behavior.State.running, behavior.Behavior.State.completed,
+            lambda: self.subbehavior_with_name('kicker').is_done_running() and self.subbehavior_with_name('kicker').state != tactics.coordinated_pass.CoordinatedPass.State.timeout,
+            'kicker completes')
+
+    @classmethod
+    def score(cls):
+        gs = main.game_state()
+        return 0 if behavior.Behavior.State.running or (
+            gs.is_ready_state() and gs.is_our_free_kick()) else float("inf")
```
JNeiger: I haven't looked, but is `is_our_free_kick()` only for our indirect kicks, or is it both direct and indirect?

soccer/gameplay/plays/restarts/basic_indirect.py
```diff
@@ -0,0 +1,53 @@
...
+
+        self.add_transition(
+            behavior.Behavior.State.running, behavior.Behavior.State.completed,
+            lambda: self.subbehavior_with_name('kicker').is_done_running() and self.subbehavior_with_name('kicker').state != tactics.coordinated_pass.CoordinatedPass.State.timeout,
+            'kicker completes')
+
+    @classmethod
+    def score(cls):
+        gs = main.game_state()
+        return 0 if behavior.Behavior.State.running or (
```
JNeiger: I'm not sure this `behavior.Behavior.State.running` is doing anything? I think this always is true assuming the actual int this represents is 0

soccer/gameplay/plays/restarts/basic_indirect.py
```diff
@@ -0,0 +1,53 @@
...
+            None,
+            (kicker, lambda x: True),
+            receiver_required=False,
+            kicker_required=False,
+            prekick_timeout=9)
+        self.add_subbehavior(pass_behavior, 'kicker')
+
+        self.add_transition(
+            behavior.Behavior.State.running, behavior.Behavior.State.completed,
+            lambda: self.subbehavior_with_name('kicker').is_done_running() and self.subbehavior_with_name('kicker').state != tactics.coordinated_pass.CoordinatedPass.State.timeout,
```
JNeiger: Why are we waiting for if it is done running and not timed out? Is this to always keep trying if we fail the kick?

soccer/gameplay/plays/restarts/basic_indirect.py
```diff
@@ -0,0 +1,53 @@
...
+        receive_pt = robocup.Point(0, 3 * constants.Field.Length / 4)
+        
+        pass_behavior = tactics.coordinated_pass.CoordinatedPass(
+            receive_pt,
+            None,
+            (kicker, lambda x: True),
+            receiver_required=False,
+            kicker_required=False,
+            prekick_timeout=9)
+        self.add_subbehavior(pass_behavior, 'kicker')
```
JNeiger: This name should probably be changed to represent it's a pass not a pivot/line kick

soccer/gameplay/plays/restarts/basic_indirect.py
Jason27chan: Yes it is for both

soccer/gameplay/plays/restarts/basic_indirect.py
Jason27chan: honestly I am not sure. This was in the previous our_free_kick file

soccer/gameplay/plays/restarts/basic_indirect.py
Jason27chan: I am not sure. If the play has timed out, then it has failed, so it shouldn't just stop right? we should transition into our play

Jason27chan: Tried removing the timeout check. 
Leaving the time out check in Pros: If the pass fails, it allows the play to transition into the main play. 
leaving the time out check in Cons: If the pass fails, it will still try to pass, which is good if the pass point is temporarily blocked.

JNeiger: I think I'm ok with leaving it in then. That seems reasonable

PR #1238, opened 2019-02-03
Updated sim constants

JNeiger:


JNeiger:
If you want to update your friction coeff in sim, you have to delete the `~/.grsim.xml` file. Be warned that passing and capture goes downhill really quick

Issue #1234, opened 2019-01-23
Wall Tactic


Alan-Penkar:
Based on what was discussed today with @JNeiger, @Jason27chan, and @bopas2 for adaptive defense - this is the spec I came up with for the wall tactic.  Please feel free to comment or add anything I missed.

Wall Tactic - 
    This tactic should create a wall between the mark_point (usually will be the ball) 
and the defend_point (usually a spot on the goal line)  The robots will line up equally spaced based on the
defender_spacing parameter on a line or curve (depending on wall_curvature) so that they are on the middle of the block line.

Possible enhancements (ignore for now):
-Determine 'optimal' number of defenders for wall
-Alter spacing based on the direction the mark_robot is facing to make more effective use of fewer defenders
 --This would need to consider camera lag and other things to make sure we don't leave open windows that the opponent could easily pass through

Just 1 state for execute wall
'''
@param mark_point: default to ball location
@param mark_robot: if we don't use mark_point (it's set to None), then use mark_robot.pos as the mark point.
@param defend point: default to shot_pt from KickEval (or center of goal?)
@param wall_distance: distance from mark point to center of wall (runs along the block line).  default to the minimum opp distance for restart?
@param wall_curvature: Curvature (1/radius) of the circle that players line up on.  Default to 0, which should be a straight line.  Need to set a minimum depending on wall_distance, number of robots on the wall, and robot_radius, still need to determine details
@param defender_spacing: default to ball radius / 2 for now.
@param num_defenders: default to 3 based on what we discussed today (easily could change)
'''

jgkamat:
This sounds very useful. It can also be used in many other places like in
line_up and the defense against direct (and indirect) restarts.

The main problem that will need to be solved is finding a distance between the
robots that they can protect against shots but far enough so they don't
interfere with each other's path planning. (or maybe, the path planner can be
altered to make this cleaner)

kylestach:
I believe this was closed by #1240. @bopas2 please re-open this issue if that's not correct.

PR #1240, opened 2019-02-10
Wall Tactic

bopas2:
Wall tactic for the new defense strategy.

Functionality is completed, I'm unsure as to whether or not the transitions and states are correct. (Do we want a state for when the tactic is completed?) 

#1234


soccer/gameplay/tactics/wall.py
```diff
@@ -0,0 +1,65 @@
...
+                 curvature = 0,                             # curvature (in radians) of the wall 0 - pi/2
+                 mark_point = None,                         # what point we are defending against (default is ball)
+                 defender_point = robocup.Point(0, 0),      # what point we are defending (default is goal)
+                 defender_spacing = 3,                      # number of robot radii between the centers of the defenders in the wall
+                 dist_from_mark = 1):                       # distance from the mark point we want to build the wall
+        super().__init__(continuous=True)
+
+        self.number_of_defenders = num_defenders
+        self.curvature = curvature
+        self.mark_point = main.ball().pos if mark_point == None else mark_point
```
JNeiger: Ball can be None if its outside vision

soccer/gameplay/tactics/wall.py
```diff
@@ -0,0 +1,65 @@
...
+                       behavior.Behavior.State.running)
+
+        self.add_transition(behavior.Behavior.State.start,
+                            Wall.State.defense_wall, lambda: True,
+                            "immediately")
+
+    def on_enter_defense_wall(self):
+        self.remove_all_subbehaviors()
+        midpoint = self.arc_midpoint()
+        angle = (math.atan2(midpoint.x - self.defense_point.x, midpoint.y - self.defense_point.y))
```
JNeiger: You can subtract the Points then do a .angle() call.

eg `(b - a).angle()`

soccer/gameplay/tactics/wall.py
```diff
@@ -0,0 +1,65 @@
...
+                skills.move.Move(pt),
+                name="robot" + str(i),
+                required=False,
+                priority=self.number_of_defenders - i)
+
+    # Finds the point on the wall (arc) at which the defender robot should move to
+    def point_on_arc(self, midpoint, angle, defender_number):
+        x_pt = midpoint.x - defender_number * (constants.Robot.Radius * self.defender_spacing) * math.cos(angle + (defender_number * self.curvature))
+        y_pt = midpoint.y + defender_number * (constants.Robot.Radius * self.defender_spacing) * math.sin(angle + (defender_number * self.curvature))
+        return robocup.Point(x_pt,y_pt)
```
JNeiger: You can do some vector math here to make life easy. You can get the perpendicular vector by `.rotate(pi/2)` and `.norm()` to normalize, then rotate another `defender_number * curvature`, scale by `defender_number * robotRadius * defender_space` and add that to the original midpoint

soccer/gameplay/tactics/wall.py
```diff
@@ -0,0 +1,65 @@
...
+
+    # Finds the point on the line between the defense and mark point that is the distance specified from the mark point
+    def arc_midpoint(self):
+        dist_from_defense_and_mark = (self.mark_point - self.defense_point).mag()
+        x_pt = self.mark_point.x + self.dist_from_mark / dist_from_defense_and_mark * (self.defense_point.x - self.mark_point.x)
+        y_pt = self.mark_point.y + self.dist_from_mark / dist_from_defense_and_mark * (self.defense_point.y - self.mark_point.y)
+        return robocup.Point(x_pt, y_pt)
```
JNeiger: This can be done by `mark_point + (b - a).norm() * dist_from_mark`

soccer/gameplay/tactics/wall.py
```diff
@@ -0,0 +1,65 @@
...
+    def on_enter_defense_wall(self):
+        self.remove_all_subbehaviors()
+        midpoint = self.arc_midpoint()
+        angle = (math.atan2(midpoint.x - self.defense_point.x, midpoint.y - self.defense_point.y))
+        for i in range(self.number_of_defenders):
+            pt = None
+            if self.number_of_defenders % 2 != 0:
+                pt = self.point_on_arc(midpoint, angle, i - int(self.number_of_defenders / 2))
+            else:
+                pt = self.point_on_arc(midpoint, angle, i - self.number_of_defenders / 2 + .5)
```
JNeiger: Is this supposed to be a float still? If you want to round up, you can use the `ceil` function in the math library to round up to the nearest int

soccer/gameplay/tactics/wall.py
```diff
@@ -0,0 +1,65 @@
...
+class Wall(composite_behavior.CompositeBehavior):
+    class State(enum.Enum):
+        defense_wall = 1
+
+    def __init__(self,
+                 num_defenders = 3,                         # number of defenders we're making the wall with (default 3)
+                 curvature = 0,                             # curvature (in radians) of the wall 0 - pi/2
+                 mark_point = None,                         # what point we are defending against (default is ball)
+                 defender_point = robocup.Point(0, 0),      # what point we are defending (default is goal)
+                 defender_spacing = 3,                      # number of robot radii between the centers of the defenders in the wall
```
JNeiger: This feels super high in terms of spacing. We'll have to test it a little more, but it should probably be lower

soccer/gameplay/tactics/wall.py
```diff
@@ -0,0 +1,65 @@
...
+            pt = None
+            if self.number_of_defenders % 2 != 0:
+                pt = self.point_on_arc(midpoint, angle, i - int(self.number_of_defenders / 2))
+            else:
+                pt = self.point_on_arc(midpoint, angle, i - self.number_of_defenders / 2 + .5)
+            self.add_subbehavior(
+                skills.move.Move(pt),
+                name="robot" + str(i),
+                required=False,
+                priority=self.number_of_defenders - i)
```
JNeiger: This feels off in terms of priority. Look at defense.py to get the range of numbers for defense. You can also pass priority in as a argument if you want

soccer/gameplay/tactics/wall.py
bopas2: Just looking at the simulator, the spacing looked fine, but I agree some onfield testing would probably be good to verify that and would be useful in general (for path planning as well).

soccer/gameplay/tactics/wall.py
bopas2: The float value allows for a symmetrical, 'balanced,' wall when we have an even number of defenders. When we have an odd number of defenders, the direct shot is blocked by a single defender. When we have an even number of defenders, we have two defenders block the direct shot together, and using the float value, the wall is symmetric against the shot (if that makes any sense). Perhaps this should be re-thought because it may allow for small gap in the wall for a direct shot. 
image (odd number of defenders)
image (even number of defenders)

soccer/gameplay/tactics/wall.py
bopas2: How should we handle that?

soccer/gameplay/tactics/wall.py
bopas2: Interestingly enough when I change my code to what you have written, the result breaks the wall's functionality. Is keep what I have ok, or should I use what you recommend for readability?

JNeiger: Oh whoops, didn't realize you swapped the x and y in the atan2. Do a .rotate(pi/2).angle(). If that doesn't work I'll actually have to pull out pen and paper

bopas2: I'll look into it

soccer/gameplay/tactics/wall.py
JNeiger: Yeah, I think it's an easy change based on testing. Let's just keep the current behavior since it's the "more complicated" version.

If I'm reading the code right, you actually don't need this if statement either. The `i - self.number_of_defenders / 2 + .5` works in both cases, even or odd. I came up the below which is the equivalent if you distribute the `-1` out of the parentheses.

`defender_number = (0:num_robots - 1) - (num_robots - 1) / 2`

JNeiger:
Imma put this here since it affects a few different places.

I'm working through all your math to do this and I think you have some spots that don't quite make sense from a geometric point of view. I'm going to walk through what I think you are doing and then the "corrected" version. Feel free to let me know if anything is wrong about my assumptions or math, it's 1:30am at the time of writing lol.

# Current Version
## Arc Midpoint
This looks find as far as I can tell. There's nothing inherently wrong with it, though it can be simplified using vector math.

## Angle Calculations
This is the whole atan2 stuff.
I don't think this makes sense from a math POV. `atan2` takes in `(Y, X)` in that order. You flipped the values here so it's really a cotangent. Here is a graph of them both. Blue is `atan2(x,y)` and red is `atan2(y,x)`. The red is correct.

image

It looks like the goal was to get the angle of the vector going from the defense point to the midpoint.

## Point on Arc
This takes the angle of the vector from defense point to the midpoint, rotates it by 90 + curvature offset degrees, then scales it by X amount to find the position of each specific robot on the arc.

This accidentally corrects the earlier mistake by messing up the negative signs. If you have an angle and you want the perpendicular angle, you build the vector like `[-sin(theta), cos(theta)]`, you did it like `[-cos(theta), sin(theta)]` which corrects the cotangent. 

Blue is vector from the goal to the midpoint. Red is the rotated vector way of doing it. Yellow is the vector represented by the angle when you do `atan2(x,y)`. Purple is the accidental correction vector when you do the `[-cos(theta), sin(theta)]`. Green is the correct way to calculate the perpendicular angle to the yellow vector.
image


# Fixed Version
(In the same structure you have)
## Arc Midpoint
`return mark_point + dist_from_mark*(defense_point - mark_point).norm()`

The norm function basically divides the vector by it's length to give it a unit length of 1.

## Angle Calculations

`angle = (midpoint - defense_point).angle()`

Angle does all the atan2 stuff so you don't have to thing about it.

## Point on Arc

This takes a vector pointing along the X axis, rotates it by the angle we found earlier + the curvature offset + pi/2 to get it to be perpendicular. Another way to do this is to just do `robocup.Point(0,1)` which is by default perpendicular to the x axis, then rotate the coordinate system so the x axis is along the angle we found earlier. This would also remove the `pi/2` constant in the `rotate` argument. 

`return midpoint + defender_number * constants.RobotRadius * self.defender_spacing * robocup.Point(1,0).rotate(angle + defender_number * self.curvature + pi/2)`

# Better Fixed Version
You do a lot of the same math and recalculate that `defense_point - mark_point` vector a few times. You can make just a single function for all of this since they are all almost 1 liners.

Remove everything but `point_on_arc` and that for loop where you figure out the offsets for the robots with that if statement. Just take `defender_number` as an arg. 

```
// Direction from the ball/robot to the goal (for example)
risk_dir = (self.defense_point - self.mark_point).norm()

// Location of the wall center
midpoint = self.mark_point + self.dist_from_mark * risk_dir

// Fake curvature through linearization
// Account for perpendicularity
curved_angle = defender_number * self.curvature + pi/2
defender_offset = defender_number * self.defender_spacing * constants.Robot.Radius

return midpoint + defender_offset * risk_dir.rotate(curved_angle)
```

soccer/gameplay/tactics/wall.py
JNeiger: Sorry, that's wrong on my end. Ball will always have a value, but it may be an invalid (`main.ball().valid = false`) ball. There are two common cases for an invalid ball, the first is if someone grabs the ball and literally leaves vision. The position will be the last known location. The second common case is when a robot blocks the view of the camera from seeing the ball. The ball will still be in play and actionable. 

Just due to the second case, it's probably a better idea to keep it like this and just always use the latest ball position

JNeiger: On another note, this only builds a wall at the ball location at the ball location when you create the subbehavior. It doesn't update as the ball moves. Is that the wanted behavior? @Alan-Penkar

soccer/gameplay/tactics/wall.py
```diff
@@ -0,0 +1,96 @@
...
+
+    @property
+    def mark_point(self):
+        return self._mark_point
+
+    @defense_point.setter
+    def defense_point(self, point):
+        self.remove_all_subbehaviors()
+        self._defense_point = point
+        self.on_enter_defense_wall()
```
JNeiger: Calling this feels wrong. I think our general practice has been to create a list of move behaviors in the on_enter or constructor and then have a seperate function that loops through all moves and assigns the updated move to point target. See [here](https://github.com/RoboJackets/robocup-software/blob/master/soccer/gameplay/tactics/line_up.py#L61)

soccer/gameplay/tactics/wall.py
```diff
@@ -0,0 +1,96 @@
...
+        self.add_state(Wall.State.defense_wall, 
+                       behavior.Behavior.State.running)
+
+        self.add_transition(behavior.Behavior.State.start,
+                            Wall.State.defense_wall, lambda: True,
+                            "immediately")
+
+        self.add_transition(
+            Wall.State.defense_wall,
+            behavior.Behavior.State.completed, lambda: self.completed(), "robot kicked")
```
JNeiger: Make sure to also have a transition from completed to defense_wall again if one of the robots get's knocked out of the way or if we end up changing the mark point after all the robots have hit their target locations

bopas2: does it make sense to even have a completed state?

bopas2:
> Imma put this here since it affects a few different places.
> 
> I'm working through all your math to do this and I think you have some spots that don't quite make sense from a geometric point of view. I'm going to walk through what I think you are doing and then the "corrected" version. Feel free to let me know if anything is wrong about my assumptions or math, it's 1:30am at the time of writing lol. You'll get the hang of vectors and stuff once you start using them more. They make life so much easier. Feel free to make [ascii art](https://github.com/RoboJackets/robocup-software/blob/master/soccer/vision/ball/BallBounce.cpp#L65) in comments if it helps describe the math :P
> 
> # Current Version
> ## Arc Midpoint
> This looks find as far as I can tell. There's nothing inherently wrong with it, though it can be simplified using vector math.
> 
> ## Angle Calculations
> This is the whole atan2 stuff.
> I don't think this makes sense from a math POV. `atan2` takes in `(Y, X)` in that order. You flipped the values here so it's really a cotangent. Here is a graph of them both. Blue is `atan2(x,y)` and red is `atan2(y,x)`. The red is correct.
> 
> image
> 
> It looks like the goal was to get the angle of the vector going from the defense point to the midpoint.
> 
> ## Point on Arc
> This takes the angle of the vector from defense point to the midpoint, rotates it by 90 + curvature offset degrees, then scales it by X amount to find the position of each specific robot on the arc.
> 
> This accidentally corrects the earlier mistake by messing up the negative signs. If you have an angle and you want the perpendicular angle, you build the vector like `[-sin(theta), cos(theta)]`, you did it like `[-cos(theta), sin(theta)]` which corrects the cotangent. This can be found by working it out through the math, but I always mix up which is which and trig sucks. If you look at the [2d rotation matrix](https://en.wikipedia.org/wiki/Rotation_matrix), due to the structure, the first column is the x axis along the vector at angle theta and the second column is the y axis, perpendicular to the vector at angle theta.
> 
> Blue is vector from the goal to the midpoint. Red is the rotated vector way of doing it. Yellow is the vector represented by the angle when you do `atan2(x,y)`. Purple is the accidental correction vector when you do the `[-cos(theta), sin(theta)]`. Green is the correct way to calculate the perpendicular angle to the yellow vector.
> image
> 
> # Fixed Version
> (In the same structure you have)
> 
> ## Arc Midpoint
> `return mark_point + dist_from_mark*(defense_point - mark_point).norm()`
> 
> The norm function basically divides the vector by it's length to give it a unit length of 1.
> 
> ## Angle Calculations
> `angle = (midpoint - defense_point).angle()`
> 
> Angle does all the atan2 stuff so you don't have to thing about it.
> 
> ## Point on Arc
> This takes a vector pointing along the X axis, rotates it by the angle we found earlier + the curvature offset + pi/2 to get it to be perpendicular. Another way to do this is to just do `robocup.Point(0,1)` which is by default perpendicular to the x axis, then rotate the coordinate system so the x axis is along the angle we found earlier. This would also remove the `pi/2` constant in the `rotate` argument.
> 
> `return midpoint + defender_number * constants.RobotRadius * self.defender_spacing * robocup.Point(1,0).rotate(angle + defender_number * self.curvature + pi/2)`
> 
> # Better Fixed Version
> You do a lot of the same math and recalculate that `defense_point - mark_point` vector a few times. You can make just a single function for all of this since they are all almost 1 liners.
> 
> Remove everything but `point_on_arc` and that for loop where you figure out the offsets for the robots with that if statement. Just take `defender_number` as an arg.
> 
> ```
> // Direction from the ball/robot to the goal (for example)
> risk_dir = (self.defense_point - self.mark_point).norm()
> 
> // Location of the wall center
> midpoint = self.mark_point + self.dist_from_mark * risk_dir
> 
> // Fake curvature through linearization
> // Account for perpendicularity
> curved_angle = defender_number * self.curvature + pi/2
> defender_offset = defender_number * self.defender_spacing * constants.Robot.Radius
> 
> return midpoint + defender_offset * risk_dir.rotate(curved_angle)
> ```

bopas2:
Accidentally closed the PR...oops

soccer/gameplay/tactics/wall.py
JNeiger: The only case it would be useful is if the parent behavior wants to know when all the robots are in their specified locations. Other than that, it makes no difference at all. If you would use that information in the defense, go for it and add it in, otherwise it doesn't really make a difference

soccer/gameplay/tactics/wall.py
bopas2: In my mind, the defense tactic above the wall tactic will be updating the mark_point every-time it iterates, which will then force the wall to adapt. Does that seem correct or no?

soccer/gameplay/tactics/wall.py
```diff
@@ -0,0 +1,91 @@
...
+                 curvature =  0,                            # 'curvature' (in radians) of the wall 
+                 mark_point = None,                         # what point we are defending against (default is ball)
+                 defender_point = robocup.Point(0, 0),      # what point we are defending (default is goal)
+                 defender_spacing = 3.5,                    # number of robot radii between the centers of the defenders in the wall
+                 dist_from_mark = 1,                        # distance from the mark point we want to build the wall
+                 defender_priorities = [20, 19, 18, 17, 16]): # default defense priorities                       
+        super().__init__(continuous=True)
+
+        self.number_of_defenders = num_defenders
+        self.curvature = -1 * curvature
```
JNeiger: I'm not sure I understand why this is inverted

soccer/gameplay/tactics/wall.py
```diff
@@ -0,0 +1,91 @@
...
+                name="robot" + str(i),
+                required=False,
+                priority=priority)
+
+    # Finds the point on the arc the defender should move to
+    def calculate_destination(self, robot_number):
+        defender_number = robot_number - self.number_of_defenders / 2 + .5
+        direct = (self.mark_point - self.defense_point).normalized()
+        arc_angle = defender_number * self.curvature + math.pi/2
+        direct.rotate(robocup.Point(0,0), arc_angle)
```
JNeiger: You can do `direct.rotate_origin(arc_angle)` and it does the same thing.

JNeiger:
I haven't run it yet, but does this update the wall position if you start the play then move the ball?

Other than those one or two things, it all looks really good!

JNeiger:
Is that kick thing something we should block this PR on or should that be another addition at different point in time? It's your call based on how integral this is to adaptive defense as a whole. I'll work on getting it in either way though

I was thinking about this a little more and it's also your call on whether we want that blocking state in this subbehavior one one higher. I will add that since the above subbehavior is the one continuously setting the ball position as the block point, it (in my mind) should be the one to hold the wall at a specific point. You guys have a better idea of how this is coming together so it's up to you

bopas2:
I agree with Joe, I think we can just set the location we're blocking to an enemy robot who currently has possession of the ball and I imagine that solves the problem of the defenders breaking formation.

JNeiger:
From what we have talked about, this sounds like you guys are ready to merge?

bopas2:
Yup! @JNeiger

PR #1242, opened 2019-02-17
Four Corner Pass Play 

### LINKED ISSUE #1204: Four Corners Passing Drill Play
Four Corners is this passing drill where there are 3 players on one team are playing keep away from 1 player on the other team. The three players can only move and pass on the edges of a square on the field. You cannot pass diagonally across the square. 

https://www.youtube.com/watch?v=BVZ90KGQV1o

```
A --------- B
|       E   |
|           |
|  F        |
|           |
C --------- D
```

If the player in corner A has the ball, the two other "offensive" players are usually on corners B and C. 

Let's say they pass to C, the player at corner B will move over to corner D to stay open. There will now be a player in corner A, C, and D.

The general decision on which way to pass is based on the relative angle of the defender. So if corner A has the ball and the defender is at point E, it would be best to pass to corner C. The opposite is true if the defender is at point F, it would be best to pass to corner B.

In terms of implementation, you will have to write two different plays. One on the offensive side and one on the defensive side. The offensive side will control the 3 players in the corners and the defensive side will deal with the defender. If there is no defensive robots, just always pass to the left or the right. The picture below describes about the size of the square in black. (Bonus points if you draw the square on the field too) (Extra bonus points. If the ball really leaves the square, pause the play until one of the offensive robots brings it back)

image

This should be a good step into evaluating and improving our inter-robot passing. In a perfect world, the defender should never get the ball. It will also give us a really good demo

tjones320:
Resolves #1204 . A good play for identifying issues with receiving and kicking passes. Also a good for easy demonstrations.

### DIFF
--- a/soccer/gameplay/plays/testing/four_courners_pass.py
+++ b/soccer/gameplay/plays/testing/four_courners_pass.py
@@ -1,0 +1,146 @@
+import robocup
+import play
+import behavior
+import skills.move
+import skills.capture
+import tactics.coordinated_pass
+import constants
+import main
+import enum
+
+
+## A demo play written during a teaching session to demonstrate play-writing
+# Three robots form a triangle on the field and pass the ball A->B->C->A and so on.
+class FourCornerPass(play.Play):
+	class State(enum.Enum):
+		## 2 robots get on the corners of a triangle,
+		# while a third fetches the ball
+		setup = 1
+
+		## The robots continually pass to each other
+		passing = 2
+
+		## Once the enemy robot captures the ball
+		#reset = 3
+
+	def __init__(self):
+		super().__init__(continuous=True)
+
+		# register states - they're both substates of "running"
+		self.add_state(FourCornerPass.State.setup,
+					   behavior.Behavior.State.running)
+		self.add_state(FourCornerPass.State.passing,
+					   behavior.Behavior.State.running)
+
+		self.add_transition(behavior.Behavior.State.start,
+							FourCornerPass.State.setup, lambda: True,
+							'immediately')
+		self.add_transition(
+			FourCornerPass.State.setup, FourCornerPass.State.passing, lambda: self.is_setup_done() , 'all subbehaviors completed')
+
+		self.add_transition(
+			FourCornerPass.State.passing, FourCornerPass.State.setup, lambda: self.subbehavior_with_name('passer').state == behavior.Behavior.
+			State.completed or self.subbehavior_with_name('passer').state == behavior.Behavior.
+			State.failed, 'all subbehaviors completed')
+
+		self.square_points = [
+			robocup.Point(1.5, 6),
+			robocup.Point(-1.5, 6),
+			robocup.Point(-1.5, 3),
+			robocup.Point(1.5, 3)]
+
+		self.direction = 1
+		self.chaser_pos = robocup.Point(0,0)
+		
+	def execute_running(self):
+		if (self.has_subbehavior_with_name('move1')) :
+			self.subbehavior_with_name('move1').get_robot().set_max_speed(2.0)
+
+		if (self.has_subbehavior_with_name('move2')) :
+			self.subbehavior_with_name('move2').get_robot().set_max_speed(2.0)
+
+		if (self.has_subbehavior_with_name('capture')):
+			self.subbehavior_with_name('capture').get_robot().set_max_speed(2.0)
+
+		if (self.has_subbehavior_with_name('chaser')):
+			self.subbehavior_with_name('chaser').get_robot().set_max_speed(1.0)
+
+
+		self.paint_the_field()
+		#self.chaser_pos = self.chaser.robot_location
+
+	def on_enter_setup(self):
+		closestPt = min(self.square_points, key=lambda pt: pt.dist_to(main.ball().pos))
+		self.otherPts = list(self.square_points)
+		self.otherPts.remove(closestPt)
+
+		farthestPt = max(self.otherPts,
+						key=lambda pt: pt.dist_to(main.ball().pos))
+
+		self.otherPts.remove(farthestPt)
+
+		self.add_subbehavior(skills.move.Move(self.otherPts[0]), 'move1')
+		self.add_subbehavior(skills.move.Move(self.otherPts[1]), 'move2')
+		self.add_subbehavior(skills.capture.Capture(), 'capture')
+
+	def on_exit_setup(self):
+		#self.remove_subbehavior('move1')
+		#self.remove_subbehavior('move2')
+		self.remove_subbehavior('capture')
+
+	def on_enter_passing(self):
+		# pick pass from and to points
+		kickFrom = min(self.square_points,
+					   key=lambda pt: pt.dist_to(main.ball().pos))
+		kickFromIdx = self.square_points.index(kickFrom)
+		self.safer_pass()
+		kickToIdx = (kickFromIdx + self.direction) % len(self.square_points)
+		kickToPt = self.square_points[kickToIdx]
+		if (kickToPt is self.otherPts[0]) :
+			self.remove_subbehavior('move1')
+			self.text = 'move2'
+		else :
+			self.remove_subbehavior('move2')
+			self.text = 'move1'
+		self.passing_machine = tactics.coordinated_pass.CoordinatedPass(kickToPt)
+
+		self.add_subbehavior(self.passing_machine, 'passer')
+
+		self.chaser = skills.move.Move(self.cut_of_pos(main.ball().pos))
+		self.add_subbehavior(self.chaser, 'Chasing', required = True, priority = 5)
+
+	#def execute_passing(self):
+			
+
+	def on_exit_passing(self):
+		self.remove_all_subbehaviors()
+		#self.remove_subbehavior(self.text)
+		#self.remove_subbehavior('passing')
+
+	def paint_the_field(self):
+		for i in range(len(self.square_points)) :
+			main.system_state().draw_line(robocup.Line(self.square_points[i], self.square_points[(i + 1) % 4]), (135, 0, 255), "Square")
+
+	def safer_pass(self):
+		if (True):
+			self.direction = 1
+		else :
+			self.direction = -1
+
+	def is_setup_done(self):
+		return (self.subbehavior_with_name('capture').state == behavior.Behavior.
+			State.completed)
+
+	def cut_of_pos(self, point):
+		x = point.x
+		if (point.x > 1.25):
+			x = 1.25
+		elif (point.x < -1.25):
+			x = -1.25
+
+		y = point.y
+		if (point.y > 5.75):
+			y = 5.75
+		elif(point.y < 3.25):
+			y = 3.25
+		return robocup.Point(x,y)
--- a/soccer/gameplay/plays/testing/four_courners_pass.py
+++ b/soccer/gameplay/plays/testing/four_courners_pass.py
@@ -9,19 +9,19 @@ import main
 import enum
 
 
-## A demo play written during a teaching session to demonstrate play-writing
-# Three robots form a triangle on the field and pass the ball A->B->C->A and so on.
+## A testing play to demonstrate our ability to pass and recieve balls
+# One robot will pursue the ball while three other robots will pass the ball amongst themselves
+# they can only pass the ball at the corners and recieve at the corners and they cannot move
+# through the center of the square
 class FourCornerPass(play.Play):
 	class State(enum.Enum):
-		## 2 robots get on the corners of a triangle,
-		# while a third fetches the ball
+		## One robot goes captures the ball while the other two gets on corners
+		# pursuing robot does not move on the first instance
 		setup = 1
 
-		## The robots continually pass to each other
+		## The robots pass to each other and begin setting up the next pass
+		# while the chasing robot goes for the ball.
 		passing = 2
-
-		## Once the enemy robot captures the ball
-		#reset = 3
 
 	def __init__(self):
 		super().__init__(continuous=True)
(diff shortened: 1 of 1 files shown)


soccer/gameplay/plays/testing/four_courners_pass.py
```diff
@@ -0,0 +1,229 @@
...
+			self.remove_subbehavior('move2')
+			self.text = 'move1'
+
+		# while pass is preparing get ready for the next pass
+		premove = skills.move.Move(self.square_points[(kickToIdx + self.direction) % len(self.square_points)])
+		# get ready to pass the ball
+		self.passing_machine = tactics.coordinated_pass.CoordinatedPass(kickToPt)
+
+		#if (self.has_subbehavior_with_name('Chasing')) :
+		#	self.remove_subbehavior('Chasing')
```
JNeiger: Extra comments that can be removed

soccer/gameplay/plays/testing/four_courners_pass.py
```diff
@@ -0,0 +1,229 @@
...
+			self.direction = -1
+
+	# checks if capture is completed and setup is finish
+	def is_setup_done(self):
+		return (self.subbehavior_with_name('capture').state == behavior.Behavior.
+			State.completed)
+
+	# if a point is outside of a smaller box that is the square points - 3 robot radius'
+	# return the closest point inside that box.
+	def cut_of_pos(self, point):
```
JNeiger: of -> off?

soccer/gameplay/plays/testing/four_courners_pass.py
```diff
@@ -0,0 +1,229 @@
...
+			x = self.max_x - 3 * self.bot
+		elif (point.x < (self.min_x + 3 *self.bot)):
+			x = self.min_x + 3 *self.bot
+
+		y = point.y
+		if (point.y > (self.max_y - 3 * self.bot)):
+			y = self.max_y - 3 * self.bot
+		elif(point.y < (self.min_y + 3 * self.bot)):
+			y = self.min_y + 3 * self.bot
+		return robocup.Point(x,y)
```
JNeiger: You can make a rectangle using opposite corners, then rect.nearPoint(point to test, threshold around the rect)

soccer/gameplay/plays/testing/four_courners_pass.py
```diff
@@ -0,0 +1,229 @@
...
+		self.remove_subbehavior('capture')
+
+	def on_enter_passing(self):
+		# get the point passing from
+		kickFrom = min(self.square_points,
+					   key=lambda pt: pt.dist_to(main.ball().pos))
+		# get the index of where you're passing from
+		kickFromIdx = self.square_points.index(kickFrom)
+		# decide which direction you're passing to
+		self.safer_pass()
```
JNeiger: This feels weird to have safer pass set a global variable when it's not being used anywhere else. I think it would be better to return which one

soccer/gameplay/plays/testing/four_courners_pass.py
```diff
@@ -0,0 +1,229 @@
...
+							'immediately')
+		#transition when ball is captured
+		self.add_transition(
+			FourCornerPass.State.setup, FourCornerPass.State.passing, lambda: self.is_setup_done() , 'all subbehaviors completed')
+
+		#transition when ball is passed or failed to pass
+		self.add_transition(
+			FourCornerPass.State.passing, FourCornerPass.State.setup, lambda: self.subbehavior_with_name('passer').state == behavior.Behavior.
+			State.completed or self.subbehavior_with_name('passer').state == behavior.Behavior.
+			State.failed, 'all subbehaviors completed')
```
JNeiger: You should fix up the formatting on this so the lines aren't as long, you should split the lambda to it's own line and break on the or

JNeiger: Same for the above transition

soccer/gameplay/plays/testing/four_courners_pass.py
```diff
@@ -0,0 +1,229 @@
...
+		variable_square = 4.5
+
+		#NEVER CHANGE THIS
+		length = constants.Field.Length
+		width = constants.Field.Width
+
+		#if the square size is smaller than the width shrink it down otherwise
+		#max value is the width of the field.
+		if (variable_square < width) :
+			width = variable_square
```
JNeiger: Why is this needed?

soccer/gameplay/plays/testing/four_courners_pass.py
```diff
@@ -0,0 +1,229 @@
...
+
+		# checks if there is a chasing robot and set their speed to defense speed
+		if (self.has_subbehavior_with_name('Chasing')):
+			robo = self.subbehavior_with_name('Chasing').robot
+			if (robo != None) :
+				robo.set_max_speed(self.defense_speed)
+
+		# keep the iso point stuck in the corner
+		if (self.has_subbehavior_with_name('Iso')) :
+			self.iso.pos = robocup.Point(-constants.Field.Width/2, 0)
```
JNeiger: iso?

soccer/gameplay/plays/testing/four_courners_pass.py
```diff
@@ -0,0 +1,229 @@
...
+		# get that point
+		kickToPt = self.square_points[kickToIdx]
+
+		# remove the subbehavior that's not recieving the pass
+		if (kickToPt is self.otherPts[0]) :
+			self.remove_subbehavior('move1')
+			self.text = 'move2'
+		else :
+			self.remove_subbehavior('move2')
+			self.text = 'move1'
```
JNeiger: You can name these both the same so you don't have to save the text

soccer/gameplay/plays/testing/four_courners_pass.py
```diff
@@ -0,0 +1,229 @@
...
+
+		#points that the other robots can move to
+		self.otherPts = list(self.square_points)
+		self.otherPts.remove(closestPt)
+
+		# remove the point that's diagonal of the closest point
+		farthestPt = max(self.otherPts,
+						key=lambda pt: pt.dist_to(main.ball().pos))
+
+		self.otherPts.remove(farthestPt)
```
JNeiger: I like the way you did this to find the two move/pass points

soccer/gameplay/plays/testing/four_courners_pass.py
```diff
@@ -0,0 +1,229 @@
...
+		# if we have too many robots isolate one of the robots so they don't help in the play
+		if (not self.has_subbehavior_with_name('Iso') and len(main.our_robots()) == 6) :
+			self.iso = skills.move.Move(robocup.Point(-constants.Field.Width/2, 0))
+			self.add_subbehavior(self.iso,'Iso', required = True, priority = 1)
+			
+	# remove subbehaviors 
+	def on_exit_passing(self):
+		self.remove_subbehavior(self.text)
+		self.remove_subbehavior('passer')
+		self.remove_subbehavior('Chasing')
```
JNeiger: The change in caps :P

soccer/gameplay/plays/testing/four_courners_pass.py
```diff
@@ -0,0 +1,229 @@
...
+
+		# the four courners
+		self.square_points = [
+			robocup.Point(self.max_x, self.max_y),
+			robocup.Point(self.min_x, self.max_y),
+			robocup.Point(self.min_x, self.min_y),
+			robocup.Point(self.max_x, self.min_y)]
+
+		# speed of the hunting robots
+		self.normal_speed = 2.0
```
JNeiger: You should be able to read the specific config value from the config

soccer/gameplay/plays/testing/four_courners_pass.py
```diff
@@ -0,0 +1,229 @@
...
+			self.remove_subbehavior('move1')
+			self.text = 'move2'
+		else :
+			self.remove_subbehavior('move2')
+			self.text = 'move1'
+
+		# while pass is preparing get ready for the next pass
+		premove = skills.move.Move(self.square_points[(kickToIdx + self.direction) % len(self.square_points)])
+		# get ready to pass the ball
+		self.passing_machine = tactics.coordinated_pass.CoordinatedPass(kickToPt)
```
JNeiger: You should flip the order so they match the add subbehavior below

soccer/gameplay/plays/testing/four_courners_pass.py
```diff
@@ -0,0 +1,229 @@
...
+
+		#if (self.has_subbehavior_with_name('Chasing')) :
+		#	self.remove_subbehavior('Chasing')
+
+		self.add_subbehavior(self.passing_machine, 'passer', required = True, priority = 10)
+		self.add_subbehavior(premove, 'premove', required = True)
```
JNeiger: Should probably set the priority of this too

soccer/gameplay/plays/testing/four_courners_pass.py
```diff
@@ -0,0 +1,229 @@
...
+
+	#TODO decide which the better direction to pass is
+	def safer_pass(self):
+		if (True):
+			self.direction = 1
+		else :
+			self.direction = -1
+
+	# checks if capture is completed and setup is finish
+	def is_setup_done(self):
```
JNeiger: This can just be in the lambda if you want, it's pretty short

soccer/gameplay/plays/testing/four_courners_pass.py
JNeiger: If we wait for #1236 to get merged, you can do this in two lines using the `pad` function in line with the `intersects` function

soccer/gameplay/plays/testing/four_courners_pass.py
```diff
@@ -0,0 +1,229 @@
...
+		# speed of the hunting robots
+		self.normal_speed = 2.0
+		# speed of the defending robots can decrease value to make it easier for offense
+		self.defense_speed = variable_square/3.0
+
+		# picks the direction to pass to. TODO make actual smart pass selection
+		self.direction = 1
+
+	# constanly running changes and updates
+	def execute_running(self):
```
JNeiger: I don't really like the way this looks. It feels super repetitive without much changing. I believe there is a `all_subbehaviors` function that we can use. If not, try some of the others here
https://github.com/RoboJackets/robocup-software/blob/master/soccer/gameplay/composite_behavior.py#L65

soccer/gameplay/plays/testing/four_courners_pass.py
```diff
@@ -0,0 +1,229 @@
...
+# through the center of the square
+class FourCornerPass(play.Play):
+	class State(enum.Enum):
+		## One robot goes captures the ball while the other two gets on corners
+		# pursuing robot does not move on the first instance
+		setup = 1
+
+		## The robots pass to each other and begin setting up the next pass
+		# while the chasing robot goes for the ball.
+		passing = 2
```
JNeiger: In terms of states, the current setup uses a move and capture to get the robots into the first two points and get the ball. Then the robot picks someone to pass to, passes there, and commands the second offensive robot to move to the next corner.

Pass has a built in capture and move for the target robots. I think we could do this where the state is the pass and movement to the next corner. The start might be a little jank looking, but we could have a true setup state where we pass the ball into the first corner instead of starting where ever the ball started. This way there is no chance that the robots change between the capture and pass. This play can transition to itself whenever the pass completes. If you want finer control of the defender during this point, you can have a second state between the passes to mess with him.

This will also allow only a single move subbehavior for the offense and a pass. This will change the behavior so we may not have the ball under control for the pass before we start moving the second robot. If that's the case, having an intermediate state where we transition after the start the kick part of the coordinated pass may be a good idea.

soccer/gameplay/plays/testing/four_courners_pass.py
```diff
@@ -0,0 +1,229 @@
...
+	def on_exit_passing(self):
+		self.remove_subbehavior(self.text)
+		self.remove_subbehavior('passer')
+		self.remove_subbehavior('Chasing')
+		self.remove_subbehavior('premove')
+
+	# takes in the square points and form lines and create a square on the field
+	def paint_the_field(self):
+		for i in range(len(self.square_points)) :
+			main.system_state().draw_line(robocup.Line(self.square_points[i], self.square_points[(i + 1) % 4]), (135, 0, 255), "Square")
```
JNeiger: There is a `draw_polygon` where you can pass in a rectangle if you want it filled in. It's a personal preference

soccer/gameplay/plays/testing/four_courners_pass.py
```diff
@@ -0,0 +1,229 @@
...
+		passing = 2
+
+	def __init__(self):
+		super().__init__(continuous=True)
+
+		# register states - they're both substates of "running"
+		self.add_state(FourCornerPass.State.setup,
+					   behavior.Behavior.State.running)
+		self.add_state(FourCornerPass.State.passing,
+					   behavior.Behavior.State.running)
```
JNeiger: This can be a for loop

soccer/gameplay/plays/testing/four_courners_pass.py
tjones320: Iso is isolating a robot in the corner so he can do nothing.

soccer/gameplay/plays/testing/four_courners_pass.py
tjones320: I meant this is more offense's speed want it to be editable too

soccer/gameplay/plays/testing/four_courners_pass.py
tjones320: Will do it originally checked three different things and I never switched it back. Probably will

soccer/gameplay/plays/testing/four_courners_pass.py
Jason27chan: I think a cool idea would be a modification of coordinated pass that implements the 'premove' idea that Tanner implemented here, where the passer aiming and the receiver getting into position happen concurrently

soccer/gameplay/plays/testing/four_courners_pass.py
```diff
@@ -0,0 +1,201 @@
...
+
+		self.add_transition(behavior.Behavior.State.start,
+							FourCornerPass.State.setup, lambda: True,
+							'immediately')
+		#transition when ball is captured
+		self.add_transition(
+			FourCornerPass.State.setup, FourCornerPass.State.passing, 
+			lambda: self.subbehavior_with_name('capture').state == behavior.Behavior.
+			State.completed , 'all subbehaviors completed')
```
JNeiger: The text on the transitions usually say what happened. Like "ball captured" etc. Super minor nitpick though

soccer/gameplay/plays/testing/four_courners_pass.py
```diff
@@ -0,0 +1,201 @@
...
+
+		#transition when ball is passed or failed to pass
+		self.add_transition(
+			FourCornerPass.State.passing, FourCornerPass.State.setup, 
+			lambda: self.subbehavior_with_name('passer').state == behavior.Behavior.State.completed
+			or self.subbehavior_with_name('passer').state == behavior.Behavior.
```
JNeiger: This might look cleaner if you did

```
lambda: self.subbehavior_with_name('passer').state == behavior.Behavior.State.completed or
	    self.subbehavior_with_name('passer').state == behavior.Behavior.State.failed
````

File: soccer/gameplay/plays/testing/four_courners_pass.py (hunk 11 above)
Comment (JNeiger) [reply soccer/gameplay/plays/testing/four_courners_pass.py]: Looking over this again, I still think we should pull the max speed from the config for the offense robots and then do a percent of that for the defense. I would hate for us to tune down the speed to make things cleaner then forget to change this config value to be lower when we demo

File: soccer/gameplay/plays/testing/four_courners_pass.py [hunk 21]
```diff
@@ -0,0 +1,201 @@
...
+			robocup.Point(self.min_x, self.max_y),
+			robocup.Point(self.min_x, self.min_y),
+			robocup.Point(self.max_x, self.min_y)]
+
+		# speed of the hunting robots
+		self.normal_speed = 2.0
+		# speed of the defending robots can decrease value to make it easier for offense
+		self.defense_speed = variable_square/3.0
+
+		# picks the direction to pass to. TODO make actual smart pass selection
```
JNeiger: I think you can remove this todo and either place it in the correct function or just leave it off

soccer/gameplay/plays/testing/four_courners_pass.py
```diff
@@ -0,0 +1,201 @@
...
+
+		# add a robot to chase after the ball.
+		if (not self.has_subbehavior_with_name('chasing')):
+			self.chaser = skills.move.Move(self.cut_of_pos(main.ball().pos))
+			self.add_subbehavior(self.chaser, 'chasing', required = True, priority = 2)
+
+		# if we have too many robots isolate one of the robots so they don't help in the play
+		if (not self.has_subbehavior_with_name('iso') and len(main.our_robots()) == 6) :
+			self.iso = skills.move.Move(robocup.Point(-constants.Field.Width/2, 0))
+			self.add_subbehavior(self.iso,'iso', required = True, priority = 1)
```
JNeiger: A better way of doing this is to take the number of robots currently on our team and then creating a move subbehavior for each one not required. This could even happen in the constructor of the play

soccer/gameplay/plays/testing/four_courners_pass.py
```diff
@@ -0,0 +1,201 @@
...
+			
+	# remove subbehaviors 
+	def on_exit_passing(self):
+		#self.text)
```
JNeiger: x

soccer/gameplay/plays/testing/four_courners_pass.py
JNeiger: This can probably be removed since the subbehavior will always try to move to a specific point. You shouldn't need to keep setting the target point

soccer/gameplay/plays/testing/four_courners_pass.py
```diff
@@ -0,0 +1,201 @@
...
+		# chasing robot positon should always follow the ball within a smaller inside box of the
+		# four corners.
+		if (self.has_subbehavior_with_name('chasing')) :
+			self.chaser.pos = self.cut_of_pos(main.ball().pos)
+
+		#draw the four courner field
+		# takes in the square points and form lines and create a square on the field
+		for i in range(len(self.square_points)) :
+			main.system_state().draw_line(robocup.Line(self.square_points[i], 
+			self.square_points[(i + 1) % 4]), (135, 0, 255), "Square")
```
JNeiger: I would add a tab here so it's clear it's one line of code

soccer/gameplay/plays/testing/four_courners_pass.py
```diff
@@ -0,0 +1,201 @@
...
+		# remove the point that's diagonal of the closest point
+		farthestPt = max(self.otherPts,
+						key=lambda pt: pt.dist_to(main.ball().pos))
+
+		self.otherPts.remove(farthestPt)
+
+		# decide which direction you're passing to
+		self.direction = self.safer_pass()
+		# move the other two robots to the other point locations
+		self.add_subbehavior(skills.move.Move(self.square_points[(closestPtIdx + self.direction) % 4]), 'move1', required = False)
```
JNeiger: There's only 1 move so you can probably just name it 'move'?

soccer/gameplay/plays/testing/four_courners_pass.py
JNeiger: You could probably even do the lineup tactic and just make the line somewhere out of the way

soccer/gameplay/plays/testing/four_courners_pass.py
JNeiger: Oh wait, this could be done even easier. This is the same thing you are doing here, but the computation is in other places. (I haven't tested this, but it should work)

```
# Projects the point given onto the edge of the smaller rectangle
# Each coordinate is found independently so we don't have to project to each
# segment of the rectangle and choose the closest
segment_x = robocup.Segment(robocup.Point(self.min_x + 3*self.bot, 0),
                            robocup.Point(self.max_x - 3*self.bot, 0))
segment_y = robocup.Segment(robocup.Point(0, self.min_y + 3*self.bot),
                            robocup.Point(0, self.max_y - 3*self.bot))

x_pos = segment_x.nearest_point(point)
y_pos = segment_y.nearest_point(point)

return robocup.Point(x_pos.x(), y_pos.y())
```

soccer/gameplay/plays/testing/four_courners_pass.py
```diff
@@ -0,0 +1,201 @@
...
+		#NEVER CHANGE THIS
+		length = constants.Field.Length
+		width = constants.Field.Width
+
+		#if the square size is smaller than the width shrink it down otherwise
+		#max value is the width of the field.
+		variable_square = min(variable_square, width, length)
+
+		#radius of robot
+		self.bot = constants.Robot.Radius
```
JNeiger: You can make this `self.bot_buffer = constants.Robot.Radius * 3` since AFAICT it's only being used in that case

JNeiger:
Looking good! Just some style and simplifications left, functionality looked great

soccer/gameplay/plays/testing/four_courners_pass.py
tjones320: Even if it's completed?

soccer/gameplay/plays/testing/four_courners_pass.py
tjones320: This is not possible at the current state of Python config values

soccer/gameplay/plays/testing/four_courners_pass.py
tjones320: But it's only like two lines. I can if it's really needed

soccer/gameplay/plays/testing/four_courners_pass.py
tjones320: Really I thought the operation always went on the second line?

soccer/gameplay/plays/testing/four_courners_pass.py
tjones320: Do TODO's come up in our code base?

soccer/gameplay/plays/testing/four_courners_pass.py
tjones320: That's probably a smarter idea how would I give them labels?

soccer/gameplay/plays/testing/four_courners_pass.py
tjones320: That's good advice. Thanks

soccer/gameplay/plays/testing/four_courners_pass.py
JNeiger: Move should always go from completed -> running if the robot gets bumped too far away from the target point

soccer/gameplay/plays/testing/four_courners_pass.py
JNeiger: Nah, it's just for future reference

soccer/gameplay/plays/testing/four_courners_pass.py
JNeiger: It's personal preference, but I was more-so commenting on the white space. If you make the lines in the lambda line up, it's really obvious it's part of the lambda

soccer/gameplay/plays/testing/four_courners_pass.py
JNeiger: They stay and if you search for them, you can find them all, but there's no build stuff that warns or anything

soccer/gameplay/plays/testing/four_courners_pass.py
JNeiger: Labels? Are you talking about the string part of the arg?

If so see here https://github.com/RoboJackets/robocup-software/blob/master/soccer/gameplay/tactics/line_up.py#L60

soccer/gameplay/plays/testing/four_courners_pass.py
tjones320: It's removed

soccer/gameplay/plays/testing/four_courners_pass.py
tjones320: fixed

soccer/gameplay/plays/testing/four_courners_pass.py
tjones320: There were originally two I'll change it now

soccer/gameplay/plays/testing/four_courners_pass.py
```diff
@@ -0,0 +1,215 @@
+import robocup
+import play
+import behavior
+import skills.move
+import skills.capture
+import tactics.coordinated_pass
+import constants
+import main
+import enum
+import copy
```
JNeiger: You can take this out since we aren't using it for the config stuff

soccer/gameplay/plays/testing/four_courners_pass.py
```diff
@@ -0,0 +1,215 @@
...
+			robocup.Point(self.min_x, self.max_y),
+			robocup.Point(self.min_x, self.min_y),
+			robocup.Point(self.max_x, self.min_y)]
+
+		# speed of the hunting robots
+		#TODO Create python pull from Config values. Currently this breaks the world.
+		#tmp = robocup.Configuration.FromRegisteredConfigurables().nameLookup("MotionConstraints/Max Velocity").value
+		self.normal_speed = 1.0 #tmp 
+		# speed of the defending robots can decrease value to make it easier for offense
+		self.defense_speed = self.normal_speed * 2 / 3 #self.normal_speed/2.0
```
JNeiger: I would take out the `#self.normal_speed/2.0`

soccer/gameplay/plays/testing/four_courners_pass.py
```diff
@@ -0,0 +1,215 @@
...
+		for i in range(len(self.square_points)) :
+			main.system_state().draw_line(robocup.Line(self.square_points[i], 
+				self.square_points[(i + 1) % 4]), (135, 0, 255), "Square")
+
+		# speed of the hunting robots
+		#TODO Create python pull from Config values. Currently this breaks the world.
+		#tmp = copy.deepcopy(robocup.Configuration.FromRegisteredConfigurables().nameLookup("MotionConstraints/Max Velocity").value)
+		self.normal_speed = 1.0 # tmp
+		#speed of the defending robots can decrease value to make it easier for offense
+		self.defense_speed = 2 * self.normal_speed/ 3.0 #self.normal_speed/2.0#self.variable_square/(min(self.width,self.length) * 2) * self.normal_speed
```
JNeiger: You can take out the `self.variable_square...`

soccer/gameplay/plays/testing/four_courners_pass.py
```diff
@@ -0,0 +1,215 @@
...
+
+
+
+	def on_enter_start(self):
+		# if we have too many robots isolate one of the robots so they don't help in the play
+		goalie = main.root_play().goalie_id#.system_state().game_state.get_goalie_id()
```
JNeiger: You can take out the `.system_state().game_state...`

soccer/gameplay/plays/testing/four_courners_pass.py
```diff
@@ -0,0 +1,215 @@
...
+		print(goalie)
```
JNeiger: Can take out the print

soccer/gameplay/plays/testing/four_courners_pass.py
```diff
@@ -0,0 +1,215 @@
...
+
+		#draw the four courner field
+		# takes in the square points and form lines and create a square on the field
+		for i in range(len(self.square_points)) :
+			main.system_state().draw_line(robocup.Line(self.square_points[i], 
+				self.square_points[(i + 1) % 4]), (135, 0, 255), "Square")
+
+		# speed of the hunting robots
+		#TODO Create python pull from Config values. Currently this breaks the world.
+		#tmp = copy.deepcopy(robocup.Configuration.FromRegisteredConfigurables().nameLookup("MotionConstraints/Max Velocity").value)
```
kylestach: I'd rather not commit commented code. Can we get rid of this line entirely?

soccer/gameplay/plays/testing/four_courners_pass.py
kylestach: Seconding Joe's comment

kylestach:
@tjones320 can we either revive this or kill it off? I'd rather not just have it sit here forever. I think it's pretty close though, so if we could get it through soon that would be great :slightly_smiling_face:

tjones320:
@kylestach Yeah should be able to finish it in the first week of actual work. I think it got side lined when things started heating up for competition.

PR #1244, opened 2019-03-01
Fix race condition in radio code

kylestach:
There's a race condition in the old USBRadio code when it writes to the reverse packet buffer: https://github.com/RoboJackets/robocup-software/blob/master/soccer/radio/USBRadio.cpp#L323. Because this function will be called asynchronously by libusb, and the rpackets can be accessed at any time through https://github.com/RoboJackets/robocup-software/blob/master/soccer/radio/Radio.hpp#L30 (and are accessed through the main thread), there can be writes to/from that vector concurrently, which is not good (Undefined Behavior :tm:) and could lead to any of the following:
- Dropped packets
- Corrupted packets
- Corruption in other random possibly unrelated sections of code
- Crashes (almost impossible, but theoretically a possibility under certain conditions)

### DIFF
--- a/soccer/Processor.cpp
+++ b/soccer/Processor.cpp
@@ -382,7 +382,8 @@ void Processor::run() {
         // Read radio reverse packets
         _radio->receive();
 
-        for (Packet::RadioRx& rx : _radio->reversePackets()) {
+        while (_radio->hasReversePackets()) {
+            Packet::RadioRx rx = _radio->popReversePacket();
             _state.logFrame->add_radio_rx()->CopyFrom(rx);
 
             curStatus.lastRadioRxTime =
@@ -398,7 +399,7 @@ void Processor::run() {
                 _state.self[board]->radioRxUpdated();
             }
         }
-        _radio->clear();
+
         for (Joystick* joystick : _joysticks) {
             joystick->update();
         }
--- a/soccer/radio/Radio.hpp
+++ b/soccer/radio/Radio.hpp
@@ -41,8 +41,9 @@ public:
     void clear() {
         std::lock_guard<std::mutex> lock(_reverse_packets_mutex);
 
-        // Clear the reverse packets queue
-        _reversePackets = {};
+        // Clear the reverse packets queue by copy-swap
+        std::queue<Packet::RadioRx> cleared;
+        std::swap(cleared, _reversePackets);
     }
 
 protected:
--- a/soccer/radio/Radio.hpp
+++ b/soccer/radio/Radio.hpp
@@ -23,15 +23,23 @@ public:
 
     int channel() const { return _channel; }
 
-    const std::vector<Packet::RadioRx>& reversePackets() const {
-        return _reversePackets;
+    bool hasReversePackets() {
+        std::lock_guard<std::mutex> lock(_reverse_packets_mutex);
+        return _reversePackets.size();
     }
 
-    std::vector<Packet::RadioRx>& reversePackets() { return _reversePackets; }
+    const Packet::RadioRx popReversePacket() {
+        std::lock_guard<std::mutex> lock(_reverse_packets_mutex);
+        Packet::RadioRx packet = _reversePackets.back();
+        _reversePackets.pop_back();
+        return packet;
+    }
 
     void clear() { _reversePackets.clear(); }
 
 protected:
+    std::mutex _reverse_packets_mutex;
+
     std::vector<Packet::RadioRx> _reversePackets;
     int _channel;
 };
--- a/soccer/radio/Radio.hpp
+++ b/soccer/radio/Radio.hpp
@@ -3,6 +3,9 @@
 #include <protobuf/RadioRx.pb.h>
 #include <protobuf/RadioTx.pb.h>
 
+#include <mutex>
+#include <queue>
+
 /**
  * @brief Sends and receives information to/from our robots.
  *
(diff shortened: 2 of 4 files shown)


soccer/Processor.cpp
```diff
@@ -382,7 +382,8 @@ void Processor::run() {
         // Read radio reverse packets
         _radio->receive();
 
-        for (Packet::RadioRx& rx : _radio->reversePackets()) {
+        while (_radio->hasReversePackets()) {
```
jgkamat: (I haven't read too much into this code, but), I think this was iterating in order (low->high index) before, and now this is iterating in reverse (high->low index). I'm worried this will change the value of `curStatus.lastRadioRxTime`.

If we decide that makes a difference, maybe we can use a queue instead of a vector for storing these packets.

soccer/radio/SimRadio.cpp
```diff
@@ -120,7 +120,8 @@ void SimRadio::receive() {
                                          : kicker_status_ready);
         rx.set_kicker_voltage(200);
 
-        _reversePackets.push_back(rx);
+        std::lock_guard<std::mutex> lock(_reverse_packets_mutex);
```
ashaw596: Wrap just those 2 statements in prackets to scope the mutex lockguard.

soccer/radio/USBRadio.cpp
```diff
@@ -320,7 +320,8 @@ void USBRadio::handleRxData(uint8_t* buf) {
         packet.set_fpga_status(FpgaStatus(msg->fpgaStatus));
     }
 
-    _reversePackets.push_back(packet);
+    std::lock_guard<std::mutex> lock(_reverse_packets_mutex);
```
ashaw596: Wrap just those 2 statements in prackets to scope the mutex lockguard.

soccer/radio/Radio.hpp
```diff
@@ -23,15 +26,31 @@ class Radio {
...
+        _reversePackets.pop();
+        return packet;
+    }
 
-    void clear() { _reversePackets.clear(); }
+    void clear() {
+        std::lock_guard<std::mutex> lock(_reverse_packets_mutex);
+
+        // Clear the reverse packets queue by copy-swap
+        std::queue<Packet::RadioRx> cleared;
```
ashaw596: Any reason why you want to use a swap instead of just clear?

jgkamat: std::queue doesn't have clear. You could use deque but I don't think it matters

ashaw596: Hey Jay! Is there a reason we're using a queue? It just wraps a queue. Anyways. You can make this simpler and save a copy by saying _reverse_packets = std::queue<Packet::RadioRx>{}; Or probably just _reverse_packets={};

soccer/radio/Radio.hpp
```diff
@@ -23,15 +26,31 @@ class Radio {
...
 
-    const std::vector<Packet::RadioRx>& reversePackets() const {
-        return _reversePackets;
+    bool hasReversePackets() {
+        std::lock_guard<std::mutex> lock(_reverse_packets_mutex);
+        return _reversePackets.size();
     }
 
-    std::vector<Packet::RadioRx>& reversePackets() { return _reversePackets; }
+    inline const Packet::RadioRx popReversePacket() {
```
JNeiger: Matt mentioned this on my use of inline, I figure I'd copy it here.

"I'd encourage you to actually measure if this inline keyword is necessary. My guess is it's not, but as with all performance things, it doesn't matter unless you measure it. (If you can't be bothered to check, then you don't need the optimization)"

As an aside, while I was googling around for inline stuff, I found this faq from the ioc. The last 3 q&a's are interesting in there's of where to put the inline decorator. I don't think it matters in this case, but it's an interesting thought none the less.
https://isocpp.org/wiki/faq/inline-functions#inline-member-fns

soccer/radio/SimRadio.cpp
kylestach: It'll only run on these two statements. Lock guards go into effect at construct time and go out of effect at scope close, so this is already doing that I think.

soccer/radio/Radio.hpp
kylestach: `_reverse_packets = {}` breaks the CI, and this copy-swap is a well-known C++ idiom that in pretty sure is going to end up being optimized away (maybe I'm overly optimistic about the optimizer's quality, though). Probably better to replace it with a dequeue anyway.

soccer/radio/Radio.hpp
kylestach: The inline wasn't for performance. I'm aware that it's basically ignored by modern compilers. This is a (non-template/constexpr) method definition in a header, which does in fact need inline. If there were two things that had separately included Radio.hpp before this commit, it would have been a link-time error.

kylestach: Actually, my bad here. Forgot that C++ automatically inserts an inline for class members defined in the class declaration, so this is really redundant. I'll take it out.

soccer/radio/SimRadio.cpp
ashaw596: Ah. Forgot c++ doesn't reorder mutex statements. You're right.

kylestach:
I'll merge this once it's been tested on hardware.

JNeiger:
@kylestach Was this tested?

kylestach:
Tested today.

PR #1248, opened 2019-03-11
Update RRT library to fix reverse-checking

### LINKED ISSUE #1208: Processing Hung freezes sim

kylestach:
This should fix #1208.


JNeiger:
I just had it happen again while testing this branch. I don't have time to debug at the moment, but I think it's still a problem.

kylestach:
@JNeiger so to be clear: you haven't actually reproduced the issue with this fix, right?

JNeiger:
Yep, as far as I could tell, this fixed it

PR #1250, opened 2019-03-31
Add network radio

kylestach:
This is completely untested (we don't even have the robot side of anything set up yet!) and uses a protocol that is very much subject to change. I'm just PR'ing it early.


soccer/radio/NetworkRadio.cpp
```diff
@@ -0,0 +1,187 @@
...
+            control.yvelocity() * rtp::ControlMessage::VELOCITY_SCALE_FACTOR);
+        msg->message.controlMessage.bodyW = static_cast<int16_t>(
+            control.avelocity() * rtp::ControlMessage::VELOCITY_SCALE_FACTOR);
+        msg->message.controlMessage.dribbler =
+            clamp(static_cast<uint16_t>(control.dvelocity()) * 2, 0, 255);
+
+        msg->message.controlMessage.shootMode = control.shootmode();
+        msg->message.controlMessage.kickStrength = control.kcstrength();
+        msg->message.controlMessage.triggerMode = control.triggermode();
+        msg->message.controlMessage.song = control.song();
```
oswinso: I feel like this could be separated into a separate function like `buildMessage` or something to make this function cleaner.

soccer/radio/NetworkRadio.cpp
```diff
@@ -0,0 +1,187 @@
...
+}
+
+void NetworkRadio::send(Packet::RadioTx& packet) {
+    // Get a list of all the IP addresses this packet needs to be sent to
+    for (int robot_idx = 0; robot_idx < packet.robots_size(); robot_idx++) {
+        const Packet::Control& control = packet.robots(robot_idx).control();
+        uint32_t robot_id = packet.robots(robot_idx).uid();
+
+        // Build the control packet for this robot.
+        std::array<uint8_t, rtp::HeaderSize + sizeof(rtp::RobotTxMessage)> forward_packet;
```
JNeiger: Since we will have to change everything in rtp.hpp, we should just use the `ForwardSize` defined there instead of `HeaderSize + sizeof(...)`.

This does remind me that we have to change all of that when we change radios

soccer/radio/NetworkRadio.cpp
```diff
@@ -0,0 +1,187 @@
...
+
+        rtp::Header* header = reinterpret_cast<rtp::Header*>(&forward_packet[0]);
+        header->port = rtp::PortType::CONTROL;
+        header->address = rtp::BROADCAST_ADDRESS;
+        header->type = rtp::MessageType::CONTROL;
+
+        size_t offset = sizeof(rtp::Header);
+        rtp::RobotTxMessage* msg = reinterpret_cast<rtp::RobotTxMessage*>(
+                &forward_packet[offset]);
```
JNeiger: This can be `HeaderSize`

soccer/radio/NetworkRadio.cpp
```diff
@@ -0,0 +1,187 @@
...
+        msg->uid = robot_id;
+
+        msg->message.controlMessage.bodyX = static_cast<int16_t>(
+            control.xvelocity() * rtp::ControlMessage::VELOCITY_SCALE_FACTOR);
+        msg->message.controlMessage.bodyY = static_cast<int16_t>(
+            control.yvelocity() * rtp::ControlMessage::VELOCITY_SCALE_FACTOR);
+        msg->message.controlMessage.bodyW = static_cast<int16_t>(
+            control.avelocity() * rtp::ControlMessage::VELOCITY_SCALE_FACTOR);
+        msg->message.controlMessage.dribbler =
+            clamp(static_cast<uint16_t>(control.dvelocity()) * 2, 0, 255);
```
JNeiger: *2?

soccer/radio/NetworkRadio.cpp
```diff
@@ -0,0 +1,187 @@
...
+
+        // Fetch the IP address
+        auto range = _robot_ip_map.left.equal_range(robot_id);
+
+        // Loop over all addresses for this robot; we want to send multiple copies.
+        for (auto it = range.first; it != range.second; it++) {
+            // Send to the given IP address
+            ip::address robot_ip = it->second;
+            udp::endpoint _robot_endpoint(robot_ip, _robot_port);
+            _socket.async_send_to(boost::asio::buffer(forward_packet), _robot_endpoint,
```
JNeiger: I'm worried about the `forward_packet` object not remaining valid until the handler is called. I don't think we have a guarantee that it is

soccer/radio/NetworkRadio.cpp
kylestach: Uh, why do we have to change any of it? It's just a struct packing, none of it looks platform specific.

soccer/radio/NetworkRadio.cpp
kylestach: Good question. Short answer is that this is copy pasted from USBRadio, which is another issue, but I didn't want to refactor the entire radio stack because I thought this was more urgent than it actually was.

soccer/radio/NetworkRadio.cpp
```diff
@@ -0,0 +1,187 @@
...
+        // auto to_remove = std::remove_if(mid_begin, mid_end, matching_ip);
+        // _robot_ip_map.left.erase(to_remove, _robot_ip_map.left.end());
+        registerRobot(msg->uid, _robot_endpoint.address());
+    }
+
+    packet.set_hardware_version(Packet::RJ2015);
+    packet.set_battery(msg->battVoltage *
+            rtp::RobotStatusMessage::BATTERY_SCALE_FACTOR);
+
+    if (Packet::BallSenseStatus_IsValid(msg->ballSenseStatus)) {
```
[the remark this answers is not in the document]
kylestach: All of the proto/packed struct conversation crap was copied verbatim. It'd probably be worth breaking it out into its own file.

soccer/radio/NetworkRadio.cpp
JNeiger: My b, I mean to imply that we would have to glance through and change a few things here and there. Most of it is good, but we are changing what information we are sending to the robots. Specifically, we are sending information on a per robot basis instead of all the robots' information to all the robots. I think it'll only be [this line](https://github.com/RoboJackets/robocup-fshare/blob/3dba0305a600eccadfdf80c4a22995e384860654/include/rc-fshare/rtp.hpp#L116), but I haven't gone through and really checked to be sure.

soccer/radio/NetworkRadio.cpp
kylestach: Good catch. Not sure how I missed that, this is entirely wrong.

soccer/radio/NetworkRadio.cpp
```diff
@@ -0,0 +1,187 @@
...
+    // Restart receiving
+    startReceive();
+}
+
+void NetworkRadio::switchTeam(bool) {}
+
+void NetworkRadio::registerRobot(int robot, ip::address ip) {
+    // Insert the robot into the bimap, going both directions
+    _robot_ip_map.right.insert(RobotIpMap::right_map::value_type(ip, robot));
+    _robot_ip_map.left.insert(RobotIpMap::left_map::value_type(robot, ip));
```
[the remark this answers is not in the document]
kylestach: You're right. I actually had my multimap completely backwards, which is why stuff is weird everywhere.

soccer/radio/NetworkRadio.cpp
kylestach: Gotcha. Yeah, that's why I wasn't using `ForwardSize` anywhere.

soccer/Processor.cpp
```diff
@@ -97,7 +98,7 @@ Processor::Processor(bool sim, bool defendPlus, VisionChannel visionChannel,
 
     // Create radio socket
     _radio = _simulation ? static_cast<Radio*>(new SimRadio(_state, _blueTeam))
-                         : static_cast<Radio*>(new USBRadio());
+                         : static_cast<Radio*>(new NetworkRadio(1337, 1337));
```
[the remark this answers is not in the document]
kylestach: This is definitely something we need to discuss. The difference would be server port (not robot port), but I think that would be the most logical option.

kylestach:
I've tested receive with a shim client and it works, but it doesn't send data unless it has vision so I'll have to wait to test send functionality.

Still waiting on actual radio.

soccer/Processor.cpp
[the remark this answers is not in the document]
kylestach: After thinking about it a little more, I think defining two different server ports, one for yellow and one for blue, should work correctly.

soccer/radio/NetworkRadio.cpp
```diff
@@ -0,0 +1,126 @@
...
+#include "Geometry2d/Util.hpp"
+#include "status.h"
+#include "PacketConvert.hpp"
+
+using namespace boost::asio;
+using ip::udp;
+
+NetworkRadio::NetworkRadio(int server_port)
+    : _socket(_context, udp::endpoint(udp::v4(), server_port)),
+      _send_buffers(6) {
```
JNeiger: The 6 can be `Robots_Per_Team` https://github.com/RoboJackets/robocup-software/blob/7d4e49eb543f6d07b512f61065e2c415f48da4ba/common/Constants.hpp#L16

soccer/radio/USBRadio.cpp
```diff
@@ -187,53 +189,11 @@ void USBRadio::send(Packet::RadioTx& packet) {
...
-        } else {
-            // empty slot
-            msg->uid = rtp::INVALID_ROBOT_UID;
-        }
-    }
+    fill_header(reinterpret_cast<rtp::Header*>(forward_packet));
+    rtp::RobotTxMessage* body = reinterpret_cast<rtp::RobotTxMessage*>(
+            forward_packet + sizeof(rtp::Header));
+
+    convert_tx_proto_to_rtp(packet, body, 6);
```
JNeiger: This 6 can be changed too

JNeiger:
We're about to start integrating with the radio. Since it's been a few months, I don't remember where this is at. It should be mostly working right? @kylestach

kylestach:
Yes. This is tested against a local client and works. However we never really coordinated on the interop. This assumes that the robot knows the soccer computer's ip and spams it with rpackets whether or not it's getting forward packets. Also each forward packet only contains data for a single robot. Don't know how well that matches up with what you have for the robot side, I can check it over again and take a look.

JNeiger:
Perfect, that's exactly how I was setting it up

JNeiger:
These are just notes for other people/myself for comp. Eventually we should fix the behavior, but lets do it when we have more than a few days to do stuff.

Once a robot is registered into the endpoint table, it is never released so we continue to send it data even if it's not sending data back to soccer. Additionally, if you have multiple robots with the same id, (aka testing different radios on a single platform), there is significant packet loss or difficulties connecting unless soccer is restarted (and possibly opened) before power is applied to the robot

kylestach:
Closed by #1263

PR #1255, opened 2019-04-29
Trajectory execution

kylestach:
This is blocked on #1250 (new radio) and https://github.com/RoboJackets/robocup-fshare/pull/3 (new rtp packets) and also robot-side motion control, and currently it's nowhere near ready. I've tested in sim and it gives very strange results - my first thought is that maybe paths' velocities don't quite match up with the derivatives of their positions, and the old computer-side PID was compensating, but I haven't really spent a lot of time looking/thinking about it yet.

Edit: some of the "weird behavior" was me being an idiot and forgetting how reference frames work. It still acts really weird with respect to capture/collect/settle for some reason, though (i.e. it just quits once it gets near a ball, which is bizarre considering this code shouldn't touch gameplay).


kylestach:
(Also, I'll probably close and re-open this PR after #1250 gets merged in to fix the diffs so it's easier to read.)

common/protobuf/Control.proto
```diff
@@ -23,12 +23,35 @@ message Control
...
+
+    required float goal_velocity_x = 7          [default = 0];
+    required float goal_velocity_y = 8          [default = 0];
+    required float goal_velocity_theta = 9      [default = 0];
+
+    required float goal_acceleration_x = 10     [default = 0];
+    required float goal_acceleration_y = 11     [default = 0];
+    required float goal_acceleration_theta = 12 [default = 0];
+
+    required MotionMode motion_mode = 18        [default = POSITION_CONTROL];
```
JNeiger: Is this supposed to be in order?

kylestach: It doesn't technically matter, I think I was just being lazy. I'll fix it.

kylestach:
I think the issue is actually something to do with angle planning. It looks like maybe we don't always create an angle plan over our paths, and only create x-y plans (we probably should always plan angle). @JNeiger can you confirm that this is how path planning works right now?

JNeiger:
#770 has most of the discussion on this. Long story short its a PID on angle error between current and final target

soccer/motion/TrajectoryExecutor.cpp
```diff
@@ -0,0 +1,114 @@
...
+        robot->control->set_goal_velocity_y(-instant_now.motion.vel.x());
+
+        // Default angular values, in case there's no angular command
+        robot->control->set_goal_pose_theta(robot->angle);
+        robot->control->set_goal_velocity_theta(robot->angleVel);
+
+        // Default acceleration values, in case there's no next command
+        robot->control->set_goal_acceleration_x(0);
+        robot->control->set_goal_acceleration_y(0);
+        robot->control->set_goal_acceleration_theta(0);
```
oswinso: idk I feel like it'd be more readable to define a helper `set_goal_acceleration(x, y, theta)` and then instead of repeating ..._x(0), ...._y(0), ..._theta(0) you could just do `set_goal_acceleration(0, 0, 0)`

kylestach:
Closing this for now. It's pending a lot of changes to motion planning.

Issue #954, opened 2017-03-29
Make Soccer Halt play and stop logging when paused


chachmu:
When looking back into the log history using soccer it continues to run plays and log in the background. When the pause button is activated Soccer should Halt play and stop logging to prevent keep the logs from continuing on when they run out of space while someone is trying to look at something. It will probably also be necessary to make it so that restarting play using the normal or force start buttons also restarts logging and pushes the user to the current frame.

jgkamat:
Also, I'm debating if this should be an option or not. Probably we should lean on the safe side and say yes unless there's a specific reason not to.

chachmu:
I feel like we are getting to the point where we need a config file of some sort so people can have their own defaults. Especially since we have themes now.

jgkamat:
We can keep going the environement variable route, which would be easier, but I would love to see a real user-based preference system in place. =D

PR #1257, opened 2019-05-10
Stop writing to the log on pause

CollinAvidano:
Addresses #954 . This is probably not the way to do this but is a quick way to do it. Literally its like 5 lines Open to all suggestions.


soccer/MainWindow.cpp
```diff
@@ -362,6 +362,7 @@ void MainWindow::updateViews() {
         _procFPS->setText(
             QString("Proc: %1 fps").arg(_processor->framerate(), 0, 'f', 1));
 
+        // NOTE: why are we using magic numbers instead of defined constants?
```
kylestach: Probably should be a TODO instead of a NOTE, unless you just want to fix it now.

soccer/Processor.cpp
```diff
@@ -607,9 +607,10 @@ void Processor::run() {
         // Send motion commands to the robots
         sendRadioData();
 
-        // Write to the log unless we are viewing logs
-        if (_readLogFile.empty()) {
-            _logger.addFrame(_state.logFrame);
+        // Write to the log unless we are viewing logs or main window is paused
+        // NOTE: a config option could be added here for pausing
```
kylestach: What would this config option do? I'm not sure what you mean?

soccer/MainWindow.cpp
```diff
@@ -1302,6 +1305,7 @@ void MainWindow::on_logPlaybackPrevFrame_clicked() {
     _doubleFrameNumber -= 1;
 }
 
+// NOTE:Pause should possibly be locked
```
kylestach: I don't think pause needs to be locked. `bool` is an atomic type, so there's really no reason that would be necessary.

PR #1258, opened 2019-05-14
New Play System Setup

JNeiger:
@ All the play people

We've had this challenge for a while where the main offensive plays have to control all the robots at once. This makes it really annoying to do small play like crosses or switches etc where we don't really care about where all the robots are. This is an approach to rectify that situation.

The idea is that in IRL soccer, anybody who's not in the general vicinity of the ball stays in some sort of formation. It's only when the ball gets near those players, that the player begins to do something. I took this and expanded it to our system. 

In the final form, you would just inherit from some parent class and build your play. The play itself would be some very "small" action like crossing the ball. Somewhere inside the play, you would select which positions from the formation you want to use. The rest will stay in their position while the ones you activated are under your control. The score functions for plays would describe what situations it should be used in.

I have a prototype of this system below. It's all in one file so I didn't have to deal with all the inheritance stuff, but it would get split out if actually implemented. It's not 100% perfect, but it's good enough to see what I'm talking about.
The implemented plays right now are:
 - Passing
 - Overtake
 - Switch
 - Cross
 - Shoot

Some interesting results and features that come from this:
 - We can actually define the [traingles used in soccer](http://outsideoftheboot.com/2016/11/23/geometry-in-football-understanding-the-triangle/). TL;DR: The pass lines between robots in a formation
 - Passes are usually quicker since the robots are already in a good position most of the time
 - We cover a larger section of the field in general so their passes are much easier to intercept
 - Things like through passes and that are much easier to setup for

Thoughts? Other features to add? Is this something that I should finish up?

### DIFF
--- a/soccer/gameplay/composite_behavior.py
+++ b/soccer/gameplay/composite_behavior.py
@@ -50,6 +50,13 @@ class CompositeBehavior(behavior.Behavior):
     def subbehavior_with_name(self, name: str):
         return self._subbehavior_info[name]['behavior']
 
+    def subbehavior_info_with_behavior(self, bhvr: behavior.Behavior):
+        info = None
+        for name in self._subbehavior_info:
+            if (self._subbehavior_info[name]['behavior'] is bhvr):
+                info = self._subbehavior_info[name]
+        return info
+
     def subbehaviors_by_name(self) -> Dict[str, Dict]:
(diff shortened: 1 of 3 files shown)


JNeiger:
Gonna close this and get a something real going eventually

PR #1259, opened 2019-05-30
Update for Ubuntu 19.04

JNeiger:
Had to update to 19.04 due to firmware things. These are all the things I needed to change to get it working. I also need to change the rrt submodule due to some LZ4 linking errors

I think the avr stuff is due to some compiler things for mtrain so that could probably get removed before merging. Just fyi for anyone else having problems, I also had to run `sudo dpkg --force-all -i /var/cache/apt/archives/gcc-avr_1%3a5.4.0+Atmel3.6.1-2_amd64.deb` due to gcc-avr getting messed up with the different pinnings.

### DIFF
--- a/soccer/CMakeLists.txt
+++ b/soccer/CMakeLists.txt
@@ -114,9 +114,9 @@ set(version ${PYTHONLIBS_VERSION_STRING})
 if(APPLE)
     set(boost_py_version "")
 else()
-    STRING( REGEX REPLACE "[^0-9]" "" boost_py_version ${version} )
+    STRING( REGEX REPLACE "[^0-9]" "" boost_py_version "${version}" )
     # Get first two numbers of python string
-    STRING( REGEX MATCH "^[0-9][0-9]" boost_py_version ${boost_py_version} )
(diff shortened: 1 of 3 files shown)


JNeiger:
Yep, this doesn't fix 19.04 alone, but IIRC, cmake does die due to the lack of quotes which are fixed in this PR. The RRT will also need to be fixed as well since that also kills the compile for 19.04.

I think I would want to pull the 19.04 changes out of comp if possible so we can triage them separately and build a more robust solution to this update problem

guyfleeman:
This is only tangentially related, but we wanna look at making our own compiler install so it doesn't clobber the global path and cause errors like these

kylestach:
@guyfleeman are you referring to vendoring in the build toolchain as another build step, or just suggesting that we should create a ppa or similar for robojackets toolchains and have people install off of that?

guyfleeman:
I was actually reading about schroot and debbootstrap. It'd let us install the compilers in a chrooted area without having to deal with any of the other crap I think

PR #1261, opened 2019-06-29
Chip restarts

tjones320:
Activated the chipper in restarts

### DIFF
--- a/soccer/gameplay/plays/restarts/our_shootout_chip.py
+++ b/soccer/gameplay/plays/restarts/our_shootout_chip.py
@@ -25,6 +25,7 @@ class OurShootoutChip(play.Play):
         chipping = 1
         shooting = 2
         capture = 3
+        starting = 4
 
     def __init__(self):
         super().__init__(continuous=True)
@@ -33,8 +34,13 @@ class OurShootoutChip(play.Play):
             self.add_state(s, behavior.Behavior.State.running)
 
         self.add_transition(behavior.Behavior.State.start,
-                            OurShootoutChip.State.dribbling, lambda: True,
+                            OurShootoutChip.State.starting, lambda: True,
                             'immediately')
+
+        self.add_transition(OurShootoutChip.State.starting,
+                            OurShootoutChip.State.dribbling,
+                            lambda: main.game_state().is_ready_state(),
+                            'begin')
 
         self.add_transition(OurShootoutChip.State.dribbling,
                             OurShootoutChip.State.chipping, lambda: self.is_goalie_close(),
@@ -47,7 +53,8 @@ class OurShootoutChip(play.Play):
 
         self.add_transition(OurShootoutChip.State.chipping, 
                             OurShootoutChip.State.capture,
-                            lambda: self.subbehavior_with_name('chipping').state == behavior.Behavior.State.completed,
+                            lambda: self.has_subbehavior_with_name('chipping') and 
+                            self.subbehavior_with_name('chipping').state == behavior.Behavior.State.completed,
                             'recapture')
 
         self.add_transition(OurShootoutChip.State.capture,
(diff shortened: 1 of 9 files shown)


soccer/gameplay/plays/restarts/our_shootout_chip.py
```diff
@@ -0,0 +1,182 @@
...
+        dribble = skills.dribble.Dribble(dribble_point)
+        if (not self.has_subbehavior_with_name('dribble')):
+            self.add_subbehavior(dribble, 'dribble', required = False, priority = 5)
+
+    def on_exit_dribbling(self):
+        self.remove_all_subbehaviors()
+
+    def on_enter_chipping(self):
+            # setup chipper
+            kicker = skills.line_kick.LineKick()
```
JNeiger: Line kick may not have the best performance with moving balls. That would be a thing to check. Most of the time we try to use pivot kick since we make sure we actually control it first

soccer/gameplay/plays/restarts/our_shootout_chip.py
```diff
@@ -0,0 +1,182 @@
...
+            kicker.use_chipper = True
+            kicker.min_chip_range = our_free_kick.OurFreeKick.MinChipRange
+            kicker.max_chip_range = our_free_kick.OurFreeKick.MaxChipRange
+            #calculate target
+            gap = evaluation.shooting.find_gap(max_shooting_angle= our_free_kick.OurFreeKick.MaxShootingAngle)
```
JNeiger: Find gap is more for free kicks, the technically correct way to do this is win eval. Find gap does use win eval, but there's some extra stuff that we don't want here

soccer/gameplay/plays/restarts/our_shootout_chip.py
```diff
@@ -0,0 +1,182 @@
...
+            kicker.target = gap
+            # chip
+            if (not self.has_subbehavior_with_name('chipper')) :
+                self.add_subbehavior(kicker, 'chipper', required=False, priority=5)
+
+    def on_exit_chipping(self):
+        self.remove_all_subbehaviors()
+
+    def execute_shooting(self):
+        kicker = skills.line_kick.LineKick()
```
JNeiger: Same thing with the line kick here

soccer/gameplay/plays/restarts/our_shootout_chip.py
```diff
@@ -0,0 +1,182 @@
...
+        # kick
+        if (not self.has_subbehavior_with_name('kicker')) :
+                self.add_subbehavior(kicker, 'kicker', required=False, priority=5)
+
+    def on_exit_shooting(self):
+        self.remove_all_subbehaviors()
+
+    #capture the ball
+    def execute_capture (self):
+        capture = tactics.capture.Capture()
```
JNeiger: Since pivot kick and a few others already have this internally, you can possibly just replace this with that and save a quarter second on the capture eval

soccer/gameplay/plays/restarts/our_shootout_kick.py
```diff
@@ -0,0 +1,72 @@
...
+        # move the robot one robot width behind the ball away from the opponent's goal
+        start_point = behind_ball * 2 * constants.Robot.Radius + main.ball().pos
+        self.add_subbehavior(skills.move.Move(start_point), 'starting', required = False,
+                            priority = 5)
+
+    def on_exit_starting(self):
+        self.remove_all_subbehaviors()
+
+    def execute_shooting(self):
+        kicker = skills.line_kick.LineKick()
```
JNeiger: Line kick again

soccer/gameplay/plays/restarts/their_shoot_out.py
```diff
@@ -0,0 +1,149 @@
...
+import evaluation
+
+class TheirShootOut(play.Play):
+	class State(enum.Enum):
+		#where the goalie should start before the game begins
+		starting = 0
+		#prevent the shooting robot from chipping of the robot and shooting a straight shot
+		block = 1
+		#go recieve the ball
+		capture = 2
```
JNeiger: This should be the intercept skill rather than capture. It allows a nonzero final velocity to block the ball

soccer/gameplay/plays/restarts/their_shoot_out.py
```diff
@@ -0,0 +1,149 @@
...
+
+	@classmethod
+	def handles_goalie(cls):
+		return True
+
+	# test all robots in their fleet to see if one has the ball
+	def has_ball(self):
+		closeIndicator = False
+		for r in main.their_robots():
+			closeIndicator = closeIndicator or robocup.Segment(r.pos, main.ball().pos).length() < self.ball_lost_distance
```
JNeiger: The standard thing is points and .mag here but this should works

tjones320: Switched to .mag

soccer/gameplay/plays/restarts/their_shoot_out.py
tjones320: about to test this also this isn't my most recent change

PR #1265, opened 2019-07-02
Playbook 2019

tjones320:
WE ARE NOT USING BASIC 122!


JNeiger:
We might switch back to basic 122, but this at least makes it so I don't have to click stuff when starting soccer :P

PR #1267, opened 2019-07-03
Added functionality to make unfaced move commands faced directly towa…

liquidmetal9015:
We are considering making non faced moves face the move target point. This does that. A better version might be making the robots face tangent of their path.

### DIFF
--- a/soccer/gameplay/skills/move.py
+++ b/soccer/gameplay/skills/move.py
@@ -1,6 +1,6 @@
 import single_robot_behavior
 import behavior
-
+import robocup
 
 ## Behavior that moves a robot to a specified location
 # wraps up OurRobot.move() into a Skill so we can use it in the play system more easily
@@ -42,16 +42,11 @@ class Move(single_robot_behavior.SingleRobotBehavior):
     def threshold(self, value):
         self._threshold = value
 
-    def on_enter_running(self):
-        print(self.robot.is_facing())
-
-
-
     def execute_running(self):
         if self.pos != None:
             self.robot.move_to(self.pos)
             if(not self.robot.is_facing()):
-                self.robot.face(self.pos)
+                self.robot.face(robocup.Point(self.robot.pos.x + self.robot.vel.x * 5, self.robot.pos.y + self.robot.vel.y * 5))
            
     def role_requirements(self):
         reqs = super().role_requirements()
--- a/soccer/gameplay/skills/move.py
+++ b/soccer/gameplay/skills/move.py
@@ -1,6 +1,7 @@
 import single_robot_behavior
 import behavior
 import robocup
+import math
 
 ## Behavior that moves a robot to a specified location
 # wraps up OurRobot.move() into a Skill so we can use it in the play system more easily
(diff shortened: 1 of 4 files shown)


ashaw596:
Not sure if this all works the same it use to. But if it did, default angle command as part of the path was something I had built in as an option. Current default is no angle, but you can probably set it here https://github.com/RoboJackets/robocup-software/blob/c10033605427f84bcc33e73ed77d7489c4c52532/soccer/planning/RRTPlanner.cpp#L727
If you change it to 
entries.emplace_back({MotionInstant(point, vel * currentSpeed), totalTime, AngleInstant(vel.angle)});
This adds the angle command to the Interpolate Path waypoints. Entry is defined here. https://github.com/RoboJackets/robocup-software/blob/c10033605427f84bcc33e73ed77d7489c4c52532/soccer/planning/InterpolatedPath.hpp#L27

Well I apparently didn't implement it that well. So probs also have to change this line to use it 
https://github.com/RoboJackets/robocup-software/blob/c10033605427f84bcc33e73ed77d7489c4c52532/soccer/planning/InterpolatedPath.cpp#L250
```if (waypoints[i - 1].angle) {

return RobotInstant(MotionInstant(waypoints[i - 1].pos() * (1 - constant) +
                                          waypoints[i].pos() * (constant),
                                      waypoints[i - 1].vel() * (1 - constant) +
                                          waypoints[i].vel() * (constant)), 
                        waypoints[i - 1].angle().get() * (1 - constant)
);
}
else {
return RobotInstant(MotionInstant(waypoints[i - 1].pos() * (1 - constant) +
                                          waypoints[i].pos() * (constant),
                                      waypoints[i - 1].vel() * (1 - constant) +
                                          waypoints[i].vel() * (constant)));
}
```
Should probs interpolate, but its hard with radians.

soccer/gameplay/skills/move.py
```diff
@@ -45,8 +46,16 @@ def threshold(self, value):
...
+                velPoint = robocup.Point(self.robot.vel.x, self.robot.vel.y)
+                robotPoint = robocup.Point(math.cos(self.robot.angle) * 3, math.sin(self.robot.angle) * 3)
+                if(math.degrees(robotPoint.angle_between(velPoint)) < 90):
+                    self.robot.face(robocup.Point(self.robot.pos.x + self.robot.vel.x * 5, self.robot.pos.y + self.robot.vel.y * 5))
+                else:
+                    self.robot.face(robocup.Point(self.robot.pos.x + self.robot.vel.x * -5, self.robot.pos.y + self.robot.vel.y * -5))
+
 
-    def role_requirements(self):
+def role_requirements(self):
```
JNeiger: This tab is messed up

soccer/gameplay/skills/move.py
```diff
@@ -45,8 +46,16 @@ def threshold(self, value):
     def execute_running(self):
         if self.pos != None:
             self.robot.move_to(self.pos)
+            if(not self.robot.is_facing()):
+                velPoint = robocup.Point(self.robot.vel.x, self.robot.vel.y)
```
JNeiger: Vel is already a point

soccer/gameplay/skills/move.py
```diff
@@ -45,8 +46,16 @@ def threshold(self, value):
...
+                robotPoint = robocup.Point(math.cos(self.robot.angle) * 3, math.sin(self.robot.angle) * 3)
```
JNeiger: do `robocup.Point.direction(self.robot.angle) * 3`

soccer/gameplay/skills/move.py
```diff
@@ -45,8 +46,16 @@ def threshold(self, value):
...
+                if(math.degrees(robotPoint.angle_between(velPoint)) < 90):
+                    self.robot.face(robocup.Point(self.robot.pos.x + self.robot.vel.x * 5, self.robot.pos.y + self.robot.vel.y * 5))
```
JNeiger: do `self.robot.pos + 5*self.robot.vel`

JNeiger:
This will be one of the things we rip out after comp and do the correct way in the path planner. If i'm reading this correctly, this interpolates the target angle across the entire path based on the angle of the velocity vector at the specific point in time?

PR #1281, opened 2019-07-04
Slow Stop

tjones320:
ensure we go slow during stopped play currently set to 1 m per s


tjones320:
After comp we need to change this to check the config file, however since we cannot read the config files without major glitches (See four courners) we will manually have the self.speed set inside the file

PR #1285, opened 2019-07-05
Increase aim vel

JNeiger:
>45 degrees should be faster now

Fixed bug with fine timeouts


JNeiger:
Reverted the change, but using this to fix the fine timeout bug fix

PR #1292, opened 2019-07-06
Switched aiming from goal center to use window evaluator

liquidmetal9015:
Tested and appears to work, window returned none at some point so I've made the old behavior the default in that case.


JNeiger:
This can be one lined. I haven't tested it yet, but just doing `self.use_windowing = true` will automatically calculate the correct point. You may have to add things to the `shot_obstacle_ignoring_robots` list.

See https://github.com/RoboJackets/robocup-software/blob/42a712d529fef9698f9b7d07a97daa008d0902c8/soccer/gameplay/skills/_kick.py#L99

PR #1293, opened 2019-07-14
Upgrade to C++17

kylestach:
Now that GCC 7 is standard on Ubuntu 18.04, we can use newer versions of
C++.

This change does not actually take advantage of any new C++17 features.
New PRs will be coming shortly to take advantage of:
 - std::optional
 - New library functions
 - Probably some other features that I haven't run into yet (structured
   bindings are really nice, but I can't think of a specific place in
   our code that really needs them at the moment).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -103,11 +103,10 @@ jobs:
       - <<: *load_compile_cache
       - run: make test-python
 
-  # FIXME: pylint is broken on 18.04
   pylint:
     <<: *dir
     docker:
-      - image: robojackets/robocup-software:16
+      - image: robojackets/robocup-software
     steps:
       - <<: *load_src_cache
       - checkout
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -103,10 +103,11 @@ jobs:
       - <<: *load_compile_cache
       - run: make test-python
 
+  # FIXME: pylint is broken on 18.04
   pylint:
     <<: *dir
     docker:
-      - image: robojackets/robocup-software
+      - image: robojackets/robocup-software:16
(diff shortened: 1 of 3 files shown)


kylestach:
@JNeiger do we still care about the compile-16 CircleCI setup? I don't think anyone on the team still uses 16.04, is it a platform we still want to support?

kylestach:
Talked to Joe out-of-band, he confirmed that the 16.04 CI build is no longer important. I just removed it from the config.

.circleci/config.yml
```diff
@@ -195,17 +169,14 @@ workflows:
...
       - test-soccer:
           requires:
             - compile
       - test-python:
           requires:
             - compile
-      - pylint:
-          requires:
-            - compile-16
+      - pylint
```
JNeiger: Should this require compile? I honestly never do this stuff so I'm not sure

kylestach: It previously did back on 16.04. But I'm pretty sure there's no reason for it, it doesn't rely on compiled binaries or anything. I think it may have just been ordering of tasks, so we don't run pylint until the end? But I'm not sure about that at all.

kylestach: It definitely shouldn't require the 18.04 compile if pylint is still broken on that release for us though.

jgkamat: This was a hack. The `compile` target used a newer verison of ubuntu, which shipped a version of pylint which had bugs which caused a crash on our codebase. This was a hack to pin the pylint to the ubuntu 16 version. Assuming pylint dosen't crash (or maybe if you just want to get rid of it), it should be safe to remove :)

PR #1295, opened 2019-07-16
Use old checkstyle make target in CI

kylestach:
The new checkstyle make target is broken (I don't think stylize.v1
actually exists) so it's been breaking all of our CI builds for a while.
Until we put in the time to figure out what's wrong with it, this uses
the old target (make checkstyle-old) on CI builds.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -104,10 +104,11 @@ jobs:
       - <<: *load_compile_cache
       - run: make test-python
 
+  # FIXME: pylint is broken on 18.04
   pylint:
     <<: *dir
     docker:
-      - image: robojackets/robocup-software
+      - image: robojackets/robocup-software:16
     steps:
       - <<: *load_src_cache
       - checkout
@@ -134,7 +135,7 @@ jobs:
       - <<: *install_deps
       - <<: *load_workspace
       - <<: *load_compile_cache
-      - run: git fetch origin && PATH=$PATH:$GOPATH/bin STYLIZE_DIFFBASE=origin/master make checkstyle-old
+      - run: git fetch origin && STYLIZE_DIFFBASE=origin/master make checkstyle-old
       - store_artifacts:
           path: /tmp/clean.patch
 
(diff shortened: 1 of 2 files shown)


JNeiger:
We have to make sure this matches the `make pretty` command locally. That was one of the big problems in the past

kylestach:
@JNeiger this PR does not work at all, there are several layers of broken. It looks like there are some issues with the GOPATH on circleCI...I'll investigate but this may take a while :cry:

kylestach:
@justbuchanan out of curiosity, what was the rationale behind moving stylize to go from python?

justbuchanan:
@kylestach the main reason for go was to avoid the slow startup time of the python interpreter. Since `stylize` often only checks a few files, the time for python to start running was one of the most time-consuming parts. Go runs very quickly because it is compiled.

Beyond that, it was also just an exercise for me to learn go :).

justbuchanan:
Btw, setting an environment variable in a circleci `run` step will not persist the variable outside of that step because each step runs in its own subshell. You can set an environment variable on the container that applies to all run steps with: https://circleci.com/docs/2.0/env-vars/#setting-an-environment-variable-in-a-container.

There are two ways to `go get` the `stylize` program. Initially we were using the gopkg.in redirector site to avoid naming conflicts with the old python-based stylize. `go get github.com/justbuchanan/stylize.v1` downloads and builds stylize, naming the binary `stylize.v1`. The v1 just tells gopkg to scan through git tags on the repo and choose the latest `v1.x.x` version.

The other way is just to fetch from github with `go get github.com/justbuchanan/stylize`.

kylestach:
Okay, that fixed it so that stylize.v1 is at least running. I was mostly using this PR as a testing ground (hence the spam - sorry all, but local CircleCI doesn't support workflows so I couldn't test locally :cry:). I'll squash all of those commits and make a new PR off of master so that it doesn't register as having style diffs and this should be good to go.

PR #1296, opened 2019-07-17
Fix checkstyle in CI

kylestach:
This fixes the GOPATH in CI so that checkstyle will actually run.

This is a squash PR of #1295 rebased onto master.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,65 +1,82 @@
-defaults: &dir
+defaults_env: &setup_env
+  run:
+    name: Set $PATH
+    command: |
+      echo 'export GOPATH="$HOME/go"' >> $BASH_ENV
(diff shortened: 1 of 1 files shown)


kylestach:
@JNeiger you mentioned in the previous PR that we at some points have had issues with the CI formatter doing things slightly differently than local, right? My disposition would be to enable it anyway (right now it's failing no matter what we do) and revisit it later if it becomes an issue. It should be very easy to spot, and worst comes to worst we can disable it later on down the line.

My assumption (guess?) is that the conflicts likely came from different clang-format versions, which shouldn't be an issue if all CI is on 18.04 (which it is).

JNeiger:
Sounds good to me. I just wanted to make sure you were aware that it was an issue in the past so we could possibly kill two birds with one stone

kylestach:
@JNeiger in that case, is it alright if this gets merged?

Issue #175, opened 2015-01-08
Replace Geometry2d::Point with Eigen::Vector2f?


justbuchanan:
Right now we have our own class for representing 2d points, but there is one provided in Eigen, the matrix library we use for many things.  The benefit of switching is that if we natively use Eigen's `Vector2f` class, it makes it much easier to do other matrix operations on them.  This will come in handy as we start doing more advanced path-planning and motion control.

Here's the API reference for the `Matrix` class, which `Vector2f` is an instance of: http://eigen.tuxfamily.org/dox-devel/classEigen_1_1Matrix.html.

@barulicm, do you see any reasons not to switch?  I know there are a few convenience methods we added to the `Point` class that are nice, but we could just make those functions that take `Vector2f`s as parameters.

barulicm:
I'm not sure I completely agree with this idea. While I love Eigen with all my heart, I don't know that replacing Geometry2d::Point is worth the eigen compatibility. There are a few issues that should be seriously considered in this decision.

**Library Dependencies**
Turning Point into Vector2f breaks the independence of the Geometry2d library. This will permanently tie the two libraries together and turn Geometry2d into a mere extension of the Eigen framework. Now, as Geometry2d is a library developed specifically for this project, this may or may not be a concern. It definitely feels weird to replace such a fundamental part of a library with a type from another library entirely.

**Stylistic Effects**
"A few convenience methods" translates to "every public method on Point" based on my understanding of the Matrix and Point interface differences. Eigen is built as a generic linear algebra library with types such as Vector2f that can do anything but have built-in support for almost nothing. Our Point interface has been tuned for the actions we take on a Point in our current code base. Geometric concepts such as rotation, distance, and magnitude are all easily implemented on Vector2f, but not explicitly built in. We would end up with a header file containing nearly every Point method as an external function taking a *self*-style parameter. Thousands of calls on Point objects would have to change to match this new interface, all for the sake of a relatively few places that want to multiply them by matrices.

**Relative Difficulty of Alternatives**
[...]

PR #1297, opened 2019-07-17
Add Point<->Eigen conversions and Geometry2d tests

kylestach:
I added a bit of documentation to Arc and wrote some quick tests for it - it looked like it was the biggest nontrivial class in geometry still lacking tests. More importantly, I added conversions between Eigen::Vector3d and Geometry2d::Point, which I'll be using a lot once we get down to business on refactoring motion planning (I'm also planning on adding Pose and possibly Twist to the Geometry2d library, which will end up as the interfaces that the rest of motion planning will use - but that's outside of the scope of this PR).


kylestach:
I've now added tests and conversions for TransformMatrix as well, and fixed it to not break with the checkstyle CI fixes.

kylestach:
Fixes #175.

PR #1298, opened 2019-07-18
Add Pose and Twist classes.

kylestach:
These will be very useful for doing motion planning in more familiar terms

### DIFF
--- a/common/Geometry2d/Pose.hpp
+++ b/common/Geometry2d/Pose.hpp
@@ -219,7 +219,7 @@ public:
      * constant relative to the pose.
      *
      * In mathematical terms, this is the exponential mapping that takes the Lie
-     * algebra so(2) (twists) to the Lie group SO(2) (poses).
+     * algebra se(2) (twists) to the Lie group SE(2) (poses).
      */
     Pose deltaRelative(double t) const {
         // twist = (x', y', h')
--- a/common/Geometry2d/Pose.hpp
+++ b/common/Geometry2d/Pose.hpp
@@ -24,6 +24,11 @@ public:
         : _pose(position.x(), position.y(), heading) {}
 
     /**
+     * Component-wise constructor
+     */
+    Pose(double x, double y, double h) : _pose(x, y, h) {}
+
+    /**
      * Implicit conversion from Eigen::Vector3d
      */
     Pose(const Eigen::Vector3d& other) : _pose(other) {}
@@ -84,8 +89,10 @@ public:
     /**
      * Accessors
      */
-    Point position() const { return Point(_pose.template block<2, 1>(0, 0)); }
-    double heading() const { return _pose(2); }
+    Point const position() const {
+        return Point(_pose.template block<2, 1>(0, 0));
+    }
+    double const heading() const { return _pose(2); }
 
     /**
      * Operators
@@ -137,6 +144,11 @@ public:
         : _twist(linear.x(), linear.y(), angular) {}
 
     /**
+     * Component-wise constructor
+     */
+    Twist(double dx, double dy, double dh) : _twist(dx, dy, dh) {}
+
+    /**
      * Implicit conversion from Eigen::Vector3d
      */
     Twist(const Eigen::Vector3d& other) : _twist(other) {}
@@ -159,8 +171,10 @@ public:
     /**
      * Accessors
      */
-    Eigen::Vector2d linear() const { return _twist.template block<2, 1>(0, 0); }
-    double angular() const { return _twist(2); }
+    Point const linear() const {
+        return Point(_twist.template block<2, 1>(0, 0));
+    }
+    double const angular() const { return _twist(2); }
 
     /**
      * Find the resulting pose (delta) of an object starting at the origin and
@@ -221,7 +235,7 @@ public:
                                vx * cosine_frac + vy * sine_frac, vh * t};
     }
 
-    double curvature() const { return angular() / linear().norm(); }
+    double curvature() const { return angular() / linear().mag(); }
 
     /**
      * Operators
--- a/common/Geometry2d/Pose.hpp
+++ b/common/Geometry2d/Pose.hpp
@@ -173,9 +173,7 @@ public:
      *
      * Called deltaFixed because it operates fixed to the origin frame.
      */
-    Pose applyFixed(double t) const {
-        return Pose(t * _twist);
-    }
+    Pose applyFixed(double t) const { return Pose(t * _twist); }
 
     /**
      * Find the resulting pose (delta) of an object starting at the origin and
@@ -219,16 +217,11 @@ public:
             cosine_frac = (1 - std::cos(vh * t)) / vh;
         }
 
-        return Eigen::Vector3d {
-            vx * sine_frac - vy * cosine_frac,
-            vx * cosine_frac + vy * sine_frac,
-            vh * t
-        };
+        return Eigen::Vector3d{vx * sine_frac - vy * cosine_frac,
+                               vx * cosine_frac + vy * sine_frac, vh * t};
     }
 
-    double curvature() const {
-        return angular() / linear().norm();
-    }
+    double curvature() const { return angular() / linear().norm(); }
 
     /**
      * Operators
--- a/common/Geometry2d/Pose.hpp
+++ b/common/Geometry2d/Pose.hpp
@@ -15,23 +15,24 @@ public:
     /**
      * Default constructor - zero-initialize
      */
-    Pose() { _pose = Eigen::Vector3d::Zero(); }
+    Pose() : _position{}, _heading{0} {}
 
     /**
      * Point-heading constructor
      */
     Pose(Point position, double heading)
-        : _pose(position.x(), position.y(), heading) {}
+        : _position(position), _heading(heading) {}
 
     /**
      * Component-wise constructor
      */
-    Pose(double x, double y, double h) : _pose(x, y, h) {}
+    Pose(double x, double y, double h) : _position(x, y), _heading(h) {}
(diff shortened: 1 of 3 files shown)


kylestach:
Note: this is not quite ready, I still need to add two functions to get a Pose from a Twist and a double representing time (or similar differential). The first one is simple multiplication (i.e. you have a twist (1, 0, 1) for one second so you have a delta of (1, 0, 1). The second one could technically be called "exponentiation", but that's a terrible name for it: basically it is the continuous equivalent, where you combine several small multiplications where each one originates at the reference point (hence "exponentiation") - what this actually means is that you rotate your reference frame while translating at a fixed vector relative to that reference frame. So, something like moving at (1, 0, 1) for pi/2 seconds results in a pose delta of (1, 1, pi/2). 

All, please let me know if you have thoughts on the names.

common/Geometry2d/Pose.hpp
```diff
@@ -0,0 +1,193 @@
...
+     * origin of the coordinate system.
+     */
+    TransformMatrix transform() const {
+        return TransformMatrix(position(), heading());
+    }
+
+    /**
+     * Accessors
+     */
+    Point position() const { return Point(_pose.template block<2, 1>(0, 0)); }
```
ashaw596: People might expect these accessors to be references when they're using Pose btw.
```Pose pose{};
pose.position() += Point(1,1);
```
This effectively no-op code  unfortunately might or might not give a compiler warning.

kylestach: Hmm...I honestly can't immediately think of a good way to resolve this. Do you have a suggestion?

ashaw596: Hmmm. You could create a Point subclass or something that is created from the 3f (with references). Or you could back Pose with Point and float instead. Idk the class idea seems similar (though different) to the idea behind MotionInstants https://github.com/RoboJackets/robocup-software/blob/master/soccer/planning/MotionInstant.hpp and RobotPose.

kylestach: Backing it with point+double gives the same issue though when you try to expose a Eigen::Vector3d and the first one is really just abusing inheritance for no good reason.  I could probably get the right semantics by creating a `PoseLinearRef` class with defined `operator+=(Point)` and implicit conversion to Point (no ref). I'll try that and see if it gets too ugly.

common/Geometry2d/Pose.hpp
```diff
@@ -0,0 +1,258 @@
...
+    double heading() const { return _pose(2); }
```
oswinso: same as comment above: make a `double&` version so that you can do `pose.heading() += M_PI` or something?

common/Geometry2d/Pose.hpp
```diff
@@ -0,0 +1,258 @@
...
+    /**
+     * Implicit conversion to Eigen::Vector3d
+     */
+    operator Eigen::Vector3d() const { return _twist; }
+
+    /**
+     * Accessors
+     */
+    Eigen::Vector2d linear() const { return _twist.template block<2, 1>(0, 0); }
+    double angular() const { return _twist(2); }
```
oswinso: Same concern as above

common/Geometry2d/Pose.hpp
oswinso: or alternatively provide a setter?

kylestach: Read above comment chain. I don't want it to work for one accessor and not the other, and implementation for Point accessor is nontrivial.

common/Geometry2d/Pose.hpp
oswinso: Alternatively do this like ROS and express everything in terms of transforms, make `Pose` an alias for a transform, and then mutate the Pose using a transform with something like 
```c++
Pose pose{};
pose *= Transform(1, 1, 0);
```

kylestach: That's not the issue, Oswin. We have Point all over our codebase, so regardless of the "canonical" way to manipulate these at some point someone will try doing it the wrong way.

On that note, I wrote out a `PositionRef` and it's just kind of a pain to use. Unless I copy the entire Point API into the class (which I'd rather not do) it's going to not work with auto (expected, and that's honestly fine) _and_ it will be missing some of the methods that you would want to do on `Point`.

oswinso: Could you enforce that mutating these don't do anything by changing this to like a `const Point` so that the no-op code gives a compiler warning?

kylestach: Yeah that works (error not warning though, obviously). I'll do that.

ashaw596:
You could change the name of accessors we care less to use accessors.

Its really not that bad to subclass Point I think. It doesn't violate LSP.
You basically only have 2 useless private variables in the subclass but
that's fine. Just make sure the point class uses .x() accessors for
everything (and inline those) than override those with refs. Ehhh mistakes
could happen with pointer invalidation. I would say unless you really like
eigen3d I would just back this with point. Then add a getter for eigen3d.
We really use it less.

kylestach:
@ashaw596 I prefer oswin's suggestion. No chance of misuse and better communicates intent.

ashaw596:
Seems good to me too!

oswinso:
I'm still confused about how you would say rotate a pose by like pi degrees or something though.

kylestach:
@oswin `Pose(...).withOrigin(Pose({0, 0}, M_PI / 2));` should do it.

common/Geometry2d/Pose.hpp
```diff
@@ -0,0 +1,272 @@
...
+     *
+     * y.withOrigin(x):
+     *       |  x
+     *       | / \
+     *       |/   y
+     * ------|------
+     *       |
+     */
+    Pose withOrigin(Pose other) const {
+        double sh = std::sin(other.heading()), ch = std::cos(other.heading());
```
ashaw596: Aren't sh and ch not used?

ashaw596:
It seems a bit confusing. Unless there is something in particular you are doing with Eigen3f I would really not recommend backing with that. It seems even the internal rotate was written with the accessors instead of a transformation multiply so I don't really think people are going to use it. I think to actually rotate it is Pose(...).withOrigin(Pose({0, 0}, -M_PI / 2)); ?? Though not sure.

common/Geometry2d/Pose.hpp
```diff
@@ -0,0 +1,272 @@
...
+    /**
+     * Find the resulting pose (delta) of an object starting at the origin and
+     * continuing with constant (world-space) velocity for the specified time
+     * (in seconds).
+     *
+     * Throughout the movement, linear velocity relative to the origin is
+     * constant (but velocity in the pose's reference frame is changing in
+     * direction as the pose rotates)
+     *
+     * Called deltaFixed because it operates fixed to the origin frame.
```
JNeiger: I think this should be 'applyFixed' not 'deltaFixed'.

JNeiger: On a different note, applyX feels as if it should change the object it's acting on. I can't think of a better name off the top of my head so it's not a big deal, but I figured I'd throw it out there.

kylestach: Yeah that's why I didn't name it apply*. That has an expectation that it mutates.

common/Geometry2d/Pose.hpp
kylestach: Uh, oops...I originally implemented it using Eigen and then realized we already had rotation baked into geometry so I deleted it. My bad, I'll fix.

common/Geometry2d/Pose.hpp
JNeiger: Unless I'm looking at an old version of the code the one on line 190 is applyFixed and the one on line 207 is applyRelative

kylestach: Oops again. Didn't mean to resolve this, I fat fingered. These should have been named delta* instead of apply*.

kylestach:
> It seems a bit confusing. Unless there is something in particular you are doing with Eigen3f I would really not recommend backing with that. It seems even the internal rotate was written with the accessors instead of a transformation multiply so I don't really think people are going to use it. I think to actually rotate it is Pose(...).withOrigin(Pose({0, 0}, -M_PI / 2)); ?? Though not sure.

I think it depends on your use case - I tend to do a lot of matrix math in my code and I like to think of things as vectors, personally, but I suppose a lot of the users of this down the line (the goal is for this to be used anywhere a robot's pose is actually needed, including in the world state estimator, motion planning, etc) will probably have the other use case in mind that care more about the geometric properties. With that in mind, I suppose it makes sense to change it over (and I don't personally care much, as I'll just be using the conversions to Eigen when necessary anyway).

common/Geometry2d/Pose.hpp
kylestach: There, fixed.

common/Geometry2d/Pose.hpp
```diff
@@ -0,0 +1,313 @@
...
+     *
+     * Throughout the movement, linear velocity relative to the pose's frame of
+     * reference is constant (but linear velocity relative to the origin might
+     * change as the pose rotates)
+     *
+     * Called deltaRelative because it operates with velocities that remain
+     * constant relative to the pose.
+     *
+     * In mathematical terms, this is the exponential mapping that takes the Lie
+     * algebra so(2) (twists) to the Lie group SO(2) (poses).
```
oswinso: > Lie group SO(2) (poses)
Should this be SE(2) ?

kylestach: Yes.

PR #1299, opened 2019-07-20
Move UI to separate directory

kylestach:
This pushes the QT-level UI code out of the root directory of soccer into its own directory, and the .ui files (non-code) that were already in `ui` into `ui/qt`.

This seemed like a useful thing to do given that the soccer root directory was previously cluttered with UI files - my thoughts are that the only files in the root directory should be directly relevant to all other modules. Obviously this isn't quite the case - there are still some "misc" things in there - but this goes a long way towards cleaning it up. I'm open to discussion on whether or not this is necessary, though.


kylestach:
As a side note: this is going to fail style check, because it moves some old files that were never formatted properly, which counts as "touching them" for stylize's purposes. Now that style CI is back up and running, we should probably have a discussion on when we need to listen to it when it says to reformat old files in which maybe one or two lines are touched. There is `clang-format-diff` to only do formatting (or checkstyle) on _changed_ lines, but stylize doesn't currently use it.

PR #1302, opened 2019-07-23
Per-Line Checkstyle/Formatter

### LINKED ISSUE #1300: Per-Line Checkstyle/Pretty
It would be very helpful to use a tool like `clang-format-diff` to only run the formatter on changed _lines_ instead of on all changed _files_. Currently the style checker is useless for large PRs like `comp2019` that touch existing files that may or may not already be properly formatted. Unfortunately, I don't think there's an easy way to integrate this into `stylize` because the tool works by piping the output of git diff into `clang-format-diff` - @justbuchanan, do you have any thoughts?

I've got this set up halfway on my fork (not using stylize, and currently the diff doesn't fail CI but it should), and a sample diff for comp2019 is available at https://circleci.com/gh/kylestach/robocup-software/36.

kylestach:
Fixes #1300, although it doesn't use stylize.

An example of the result of failing diff of this PR (for comp2019) is at https://circleci.com/gh/kylestach/robocup-software/52.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -153,7 +153,7 @@ jobs:
       - <<: *install_deps
       - <<: *load_workspace
       - <<: *load_compile_cache
-      - run: git fetch origin && STYLIZE_DIFFBASE=origin/master make checkstyle
+      - run: git fetch origin && STYLIZE_DIFFBASE=origin/master make checkstyle-lines
       - store_artifacts:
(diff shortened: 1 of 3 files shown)


kylestach:
I didn't write the .py stuff...I copy-pasted it from someone else who in turn copy-pasted it from a script officially distributed with clang-format.

Issue #1301, opened 2019-07-23
(Tracking Issue) Refactor SystemState


kylestach:
We have as a stated goal to refactor Processor, but before we take a look at our logic we need to work on our data representation. Currently, all data is passed around in SystemState without any clear "ownership" over individual fields. For example, gameplay code, motion planning, and radio all read and write to fields in the `Robot` class.

Address this by splitting SystemState up into several smaller data structures that are more "atomic" (i.e. written as a unit, so WorldState written by vision filter, RobotStatus written by radio, RobotCommands written by gameplay, etc. with a Logger and DebugDrawer that are accessible everywhere), with each module containing a pointer to a Context that contains them all (instead of a SystemState pointer).

This doesn't necessarily fix all of our issues (we still have shared mutable state) but it clarifies who is responsible for which data and make each of the modules a little bit more independent, and prepares us well for a long-term move to ROS if that's the direction we want to go.

Current plan of attack (in chronological order):

- [x] Add a `Context` struct containing only `SystemState`, propagate a `Context` through the tree. **Done: #1303**
- [x] Move `GameState` from `SystemState` into `Context`. **Done: #1304**
- [x] Break out debug-drawing capability from `SystemState` into `DebugDrawer`. **Done: #1305**
- [x] ~Make Python-side Robot class to maintain existing gameplay API by wrapping reads/writes from structs (start by just forwarding calls to the old `OurRobot` class)~
- [x] Break out `WorldState` (ball and robot positions and velocities) **Review: #1333**
- [x] Break out `RobotPath`
- [x] Break out `RobotStatus` for robot status reports from radio
- [x] Break out `RobotIntent` for the goals set by the gameplay module
- [x] Break out `RobotCommand` for the final commands ready to be sent to the radio, with a function to generate a radio packet and a proto
- [x] Move configuration parameters into context instead of making it totally global
- [ ] **(Ongoing)** make individual parts of the code take in one struct and return another instead of directly mutating values in the `SystemState`: for example, a path planner should take in a `WorldState` and `RobotIntent` and produce a `RobotPath`. The following is an in-progress list of places we need to do this; if you think of something that needs to be added to this list, comment or DM me.
  - [ ] Planner: `(WorldState, RobotIntent) -> RobotPath`
  - [x] Radio: `RobotCommand -> (network) -> RobotStatus`
  - [x] Motion control/trajectory follower: `(RobotCommand, RobotPath, WorldState) -> RobotIntent`
[...]

PR #1303, opened 2019-07-25
Refactor systemstate context

kylestach:
First item on the checklist for #1301 (tracking issue).

This blocks all other items on that checklist.

This PR is almost 100% renaming things, with one small exception: the formatter ran into some python files that I changed that had tabs instead of spaces (yuck) and decided that it wanted to fix that file-wide instead of just on my diffs. I didn't argue with it.

### DIFF
--- a/soccer/CMakeLists.txt
+++ b/soccer/CMakeLists.txt
@@ -79,7 +79,7 @@ set(ROBOCUP_LIB_SRC
     "vision/util/VisionFilterConfig.cpp"
     "vision/VisionFilter.cpp"
     "WindowEvaluator.cpp"
-)
+        Context.hpp)
 
 
 if(APPLE)
--- a/soccer/CMakeLists.txt
+++ b/soccer/CMakeLists.txt
@@ -78,8 +78,7 @@ set(ROBOCUP_LIB_SRC
     "vision/robot/WorldRobot.cpp"
     "vision/util/VisionFilterConfig.cpp"
     "vision/VisionFilter.cpp"
-    "WindowEvaluator.cpp"
-        Context.hpp)
+    "WindowEvaluator.cpp")
 
 
 if(APPLE)
--- a/soccer/Context.hpp
+++ b/soccer/Context.hpp
@@ -5,5 +5,12 @@
 struct Context {
     Context() : state(this) {}
 
+    // Delete copy, copy-assign, move, and move-assign because
+    // many places are expected to hold Context* pointers.
+    Context(const Context&) = delete;
+    Context& operator=(const Context&) = delete;
+    Context(Context&&) = delete;
+    Context& operator=(Context&&) = delete;
+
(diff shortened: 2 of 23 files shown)


soccer/CMakeLists.txt
```diff
@@ -79,7 +79,7 @@ set(ROBOCUP_LIB_SRC
     "vision/util/VisionFilterConfig.cpp"
     "vision/VisionFilter.cpp"
     "WindowEvaluator.cpp"
-)
+        Context.hpp)
```
JNeiger: Tabbing feels off here

kylestach: Uh yeah...actually that shouldn't have even been added.

kylestach: I think CLion did it automatically. I'll get rid of it.

soccer/Context.hpp
```diff
@@ -0,0 +1,9 @@
+#pragma once
+
+#include "SystemState.hpp"
+
+struct Context {
+    Context() : state(this) {}
```
oswinso: Delete the copy constructor, copy assign and move constructor to prevent accidental copies?

soccer/NewRefereeModule.cpp
```diff
@@ -106,11 +106,11 @@ static const int KickVerifyTime_ms = 250;
 // the ref halts/stops, make this false
 static const bool CancelBallPlaceOnHalt = true;
 
-NewRefereeModule::NewRefereeModule(SystemState& state)
+NewRefereeModule::NewRefereeModule(Context* context)
```
oswinso: Kind of curious about the decision to move from references to pointers.

oswinso: Also maybe make this a const pointer, and then add a using declaration or something so instead of typing `Context* const context` all the time you can do something like `ContextPtr context` or something?

soccer/Context.hpp
kylestach: Yeah, these should all be deleted. Good call.

soccer/NewRefereeModule.cpp
kylestach: There's really no point in making the parameter a const pointer - it doesn't really add any information to the interface IMO.

kylestach: I have a vendetta against references. 

In all seriousness, I am not a fan of references that get mutated or stored outside of the scope of the function as parameters to that function. I prefer pointers in this case because it makes the semantics more explicit at the callsite.

oswinso: Maybe not here but would it be worth at places where you're storing the pointer, ie. `Context* _context`?

kylestach: Hmm, that's true. I can go through and do that.

PR #1304, opened 2019-07-26
Move GameState out of SystemState

kylestach:
Knocks out a line item on #1301.


soccer/gameplay/robocup-py.cpp
```diff
@@ -891,7 +891,6 @@ BOOST_PYTHON_MODULE(robocup) {
         .def_readonly("our_robots", &SystemState::self)
         .def_readonly("their_robots", &SystemState::opp)
         .def_readonly("ball", &SystemState::ball)
-        .def_readonly("game_state", &SystemState::gameState)
```
JNeiger: Is this being used anywhere in python?

kylestach: No. The `main` module has a global gamestate, that's what is used.

PR #1305, opened 2019-07-27
Refactor SystemState: Separate Debug-Drawing

kylestach:
Split out `DebugDrawer`.

### DIFF
--- a/soccer/SystemState.cpp
+++ b/soccer/SystemState.cpp
@@ -7,8 +7,8 @@
 #include <Robot.hpp>
 #include <RobotConfig.hpp>
 #include <SystemState.hpp>
-#include "planning/Path.hpp"
 #include "DebugDrawer.hpp"
+#include "planning/Path.hpp"
 
 using namespace Packet;
 using namespace std;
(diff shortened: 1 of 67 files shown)


soccer/SystemState.cpp
```diff
@@ -179,122 +179,6 @@ SystemState::~SystemState() {
...
-
-void SystemState::drawSegment(const Geometry2d::Segment& line, const QColor& qc,
-                              const QString& layer) {
-    DebugPath* dbg = logFrame->add_debug_paths();
-    dbg->set_layer(findDebugLayer(layer));
-    *dbg->add_points() = line.pt[0];
-    *dbg->add_points() = line.pt[1];
-    dbg->set_color(color(qc));
-}
-
```
oswinso: Oh god all these methods were in `SystemState`? thats so gross

soccer/WindowEvaluator.cpp
```diff
@@ -27,8 +28,11 @@ void WindowEvaluator::createConfiguration(Configuration* cfg) {
         new ConfigDouble(cfg, "WindowEvaluator/distScoreCoeff", 0.3);
 }
 
-WindowEvaluator::WindowEvaluator(SystemState* systemState)
-    : system(systemState) {}
+WindowEvaluator::WindowEvaluator(Context* context) : context(context) {
+    if (context == nullptr) {
+        std::cout << "NULL POINTER FOUND!" << std::endl;
```
JNeiger: I feel like this could be more descriptive. If this appears multiple places, it might be annoying to debug since you would have to narrow down exactly which one it is

kylestach: Oh god. I let my lazy debug prints in the code. That's embarrassing, it should just be deleted. I'm honestly surprised the text is even somewhat related to the actual error 😂

Issue #1314, opened 2019-08-25
Situation Analysis


kylestach:
## Description
Situation analysis is the first stage of our effort to move away from mono-plays towards plays with more narrow scope. It should identify the current "situation" (i.e. attack goal, defend goal, clear, defend opponent's clear, pileup, etc.) to allow plays to only be selected for certain situations.

## Requirements

 - Analyze the game state and robot positions to produce a single "situation" enumeration variable, accessible from python code.
 - Display the current situation in soccer.
 - Calculate and expose other "situation" variables for consumption by python plays.

## Design

[Design document](https://docs.google.com/document/d/14Izlly57bEsJ9LZM9k_msLDCzfsqQl8IkGVFywJVs5I/edit)

PR #1315, opened 2019-08-25
Situation Analysis Module Minimum Viable Product

liquidmetal9015:
## Description
Pull request for getting the situation analysis module to the point where we can start developing plays around it. See issue for short term goals.

## Associated Issue
Issue #1314 

## Design Documents
[TODO](www.robojackets.org)

### Test Case 1
1. Configure field to a given situation from the list and description 
2. Make sure the situation displays correctly 

Expected result: ???

##  Sub-issues 
- [ ] Fix pileup state transitions and sensitivity 
- [ ] Create Design Document
- [ ] Fix goalie clear state check
- [ ] Figure out basic information publishing scheme so plays can use info for scoring.


soccer/gameplay/situational_play_selection.py
```diff
@@ -0,0 +1,744 @@
...
+import math
+from enum import Enum
+import constants
+
+
+class SituationalPlaySelector:
+
+
+    #Enum for representing the current game situation, each of which acts as a catagory of play to be run
+    class situation(Enum):
```
kylestach: Are nested classes supposed to be `lowerCamel`? Not entirely familiar with the style guide we use.

soccer/gameplay/situational_play_selection.py
```diff
@@ -0,0 +1,744 @@
...
+            none = 0
```
kylestach: Similar, are these supposed to be capitalized i.e. `NONE`, `KICKOFF`, ... (see: http://google.github.io/styleguide/pyguide.html#3164-guidelines-derived-from-guidos-recommendations)

soccer/gameplay/situational_play_selection.py
```diff
@@ -0,0 +1,744 @@
...
+
+    #Enum for representing the possession of the ball
+    class ballPos(Enum):
+        ourBall = 1
+        freeBall = 2
+        theirBall = 3
+
+    def __init__(self):
+        print("Don't make an instance of this class you bafoon!") 
+        exit() #This is a joke I'll need to remove at some point
```
kylestach: Delete before merging. Would be great if you could add a docstring to the class that includes this information (along with general usage).

kylestach: Actually, thinking more: does it make sense for this to be class methods instead of just having a singleton SituationalPlaySelector object with regular instance methods? It's stateful and everything, so I don't see why it would make sense to do them as static methods.

soccer/gameplay/situational_play_selection.py
```diff
@@ -0,0 +1,744 @@
...
+    @classmethod
+    def ball_ingress(cls, ballPos, ballVel, robot):
+        robotx = robot.pos.x
+        roboty = robot.pos.y
+        
+        ballSpeed = math.sqrt(ballVel.x**2 + ballVel.y**2)
+        if(ballSpeed < 0.15):
+            return (None, None, None, None)
+        
+        robotToBall = (robotx - ballPos.x, roboty - ballPos.y)
```
kylestach:
```suggestion
        robotToBall = robot.pos - ballPos
```

soccer/gameplay/situational_play_selection.py
```diff
@@ -0,0 +1,744 @@
...
+        
+        ballDist = math.sqrt(robotToBall[0]**2 + robotToBall[1]**2)
```
kylestach:
```suggestion
        ballDist = robotToBall.norm()
```

soccer/gameplay/situational_play_selection.py
```diff
@@ -0,0 +1,744 @@
...
+        angle = math.degrees(math.atan2(ballVel.y, ballVel.x) - math.atan2(robotToBall[1], robotToBall[0]))
```
kylestach: I would really rather avoid using degrees everywhere...it's much cleaner to use standard units in all places so you just never have to think about your units.

kylestach: Also, this needs to wrap around properly. So, i.e. 359 degrees should wrap to -1.

soccer/gameplay/situational_play_selection.py
```diff
@@ -0,0 +1,744 @@
...
+        if(abs(angle) > 100):
+            return (None, None, None, None)
+       
+        robotToBallDOTBallVel = robotToBall[0] * ballVel.x + robotToBall[1] * ballVel.y
+        scalar = robotToBallDOTBallVel / (ballSpeed**2)
+        robotOntoVelocity = (scalar * ballVel.x, scalar * ballVel.y) #The robots position relative to the ball projected onto the balls velocity
+        projectedToRobot = (robotOntoVelocity[0] - robotToBall[0], robotOntoVelocity[1] - robotToBall[1]) #The vector from the projected vector to the robots position
+        
+        distanceFromPath = math.sqrt(projectedToRobot[0]**2 +  projectedToRobot[1]**2) 
+        interceptDistance = math.sqrt(robotOntoVelocity[0]**2 + robotOntoVelocity[1]**2)
```
kylestach: This is a lot of mathemagic. It would be great to put some comments on the derivation so the next person who has to look at it isn't lost.

soccer/gameplay/situational_play_selection.py
```diff
@@ -0,0 +1,744 @@
...
+        #print(cls.currentSituation.name)
+        #print(abs(time.time() - startTime))
+
+
+    #It would be interesting to evaluate characteristics about our enemy
+    #Like some kind of manuverability/speed characteristic
+    #Would have to have one for each team
+   
+    '''
+    def ballVelFactor(ballVel):
```
kylestach: Are these expected to continue to exist?

soccer/gameplay/situational_play_selection.py
```diff
@@ -0,0 +1,744 @@
...
+    def get_obstruct_dist(cls):
+        return 999999
+    '''
+
+
+
+
+    #These two functions are really a mess, its kind of bad
+    @classmethod
+    def in_ball_path(cls):
```
kylestach: All of these functions are expected to be used all over soccer and probably should have some nice documentation.

soccer/gameplay/situational_play_selection.py
```diff
@@ -0,0 +1,744 @@
...
+        projectedToRobot = (robotOntoVelocity[0] - robotToBall[0], robotOntoVelocity[1] - robotToBall[1]) #The vector from the projected vector to the robots position
+        
+        distanceFromPath = math.sqrt(projectedToRobot[0]**2 +  projectedToRobot[1]**2) 
+        interceptDistance = math.sqrt(robotOntoVelocity[0]**2 + robotOntoVelocity[1]**2)
+
+        return (distanceFromPath, interceptDistance, ballSpeed, angle)
+
+
+    @classmethod
+    def closestReciever(cls):
```
kylestach: Does this make sense only when preparing to shoot or while the ball is moving? (If it's supposed to cover the latter case it should probably take velocity into account)

soccer/gameplay/situational_play_selection.py
```diff
@@ -0,0 +1,744 @@
...
+
+        return (closestRobot, closestRobotDistance)
+
+        
+
+    #A function that determines if the ball is in the mouth of a given robot
+    @staticmethod
+    def possesses_the_ball(ballPos, robot, distThresh=0.14, angleThresh=35):
+        distance = math.sqrt((ballPos.x - robot.pos.x)**2 + (ballPos.y - robot.pos.y)**2)
+        angle = math.degrees(math.atan2(ballPos.y - robot.pos.y, ballPos.x - robot.pos.x) - robot.angle)
```
kylestach: You've got this pattern in a few places, I think it can be replaced by this:
https://github.com/RoboJackets/robocup-software/blob/master/common/Geometry2d/Point.hpp#L348

soccer/gameplay/situational_play_selection.py
```diff
@@ -0,0 +1,744 @@
...
+            if(toPrint != None):
+                print(toPrint)'''
+
+    #Second mess function
+    @classmethod
+    def ball_ingress(cls, ballPos, ballVel, robot):
+        robotx = robot.pos.x
+        roboty = robot.pos.y
+        
+        ballSpeed = math.sqrt(ballVel.x**2 + ballVel.y**2)
```
kylestach:
```suggestion
        ballSpeed = ballVel.norm()
```

soccer/gameplay/situational_play_selection.py
```diff
@@ -0,0 +1,744 @@
...
+    def possesses_the_ball(ballPos, robot, distThresh=0.14, angleThresh=35):
+        distance = math.sqrt((ballPos.x - robot.pos.x)**2 + (ballPos.y - robot.pos.y)**2)
+        angle = math.degrees(math.atan2(ballPos.y - robot.pos.y, ballPos.x - robot.pos.x) - robot.angle)
+        if(distance < distThresh and abs(angle) < angleThresh):
+            return True
+        else:
+            return False
+
+
+    hasBall = dict()
```
kylestach: I think I mentioned earlier, but I really don't think it makes sense for this to be a set of class methods with static variables instead of just a singleton instance of a class.

soccer/gameplay/situational_play_selection.py
```diff
@@ -0,0 +1,744 @@
...
+    posChangeTime = dict()
+    posDuration = dict()
+    recvProb = dict()
+    ballDist = dict()
+
+    lastSituation = None
+    situationChangeTime = None
+    playPreemptTime = 0.20 #The time after a situation changes before preempting the current play
```
kylestach: What does "preempt" mean in this context? I thought this class just publishes information and allows the plays to use it?

soccer/gameplay/situational_play_selection.py
```diff
@@ -0,0 +1,744 @@
...
+    #Update determining if we want to preempt the current play or not
+    @classmethod
+    def updatePreempt(cls):
+        if(cls.lastSituation != cls.currentSituaion and cls.situationChangeTime == None):
+            cls.situationChangeTime = time.time()
+
+        if(cls.currentPreempt):
+            cls.currentPreempt = False
+        elif(cls.situationChangeTime != None):
+            if(abs(time.time() - cls.situationChangeTime) > cls.playPreemptTime):
```
kylestach: I wonder if it makes sense to have some hysteresis on the published situation itself rather than on play preemption.

soccer/gameplay/situational_play_selection.py
```diff
@@ -0,0 +1,744 @@
...
+        pileupBufferTime = 0.8 #the number of seconds of pile up conditions before a pile up is declared
+
+        botsNearBall = cls.nearBallCount()
+        botsWithBall = cls.withBallCount()
+        totalNearBall = sum(botsNearBall)
+        totalWithBall = sum(botsWithBall)
+
+        currentPileup = False
+        
+        if(totalNearBall >= 3 and botsNearBall[0] > 0 and botsNearBall[1] > 0):
```
kylestach: Based on testing today, this may need to be adjusted to be more sensitive.

soccer/gameplay/situational_play_selection.py
```diff
@@ -0,0 +1,744 @@
...
+        elif(ballPos.y > midfield + (midfieldFactor / 2) * fieldLen):
+            return cls.fieldLoc.attackSide
+        else:
+            return cls.fieldLoc.midfield
+
+
+    @classmethod
+    def ballInGoalZone(cls, buff = constants.Robot.Radius + constants.Ball.Radius):
+        ballPos = cls.systemState.ball.pos
+        if(ballPos.x - buff > constants.Field.OurGoalZoneShape.min_x()):
```
kylestach:
```suggestion
        return (ballPos.x - buff > constants.Field.OurGoalZoneShape.min_x()) and (ballPos.x + buff < constants.Field.OurGoalZoneShape.max_x()) and (ballPos.y + buff < constants.Field.OurGoalZoneShape.max_y()) and (ballPos.y > constants.Field.OurGoalZoneShape.min_y())
```

soccer/gameplay/situational_play_selection.py
```diff
@@ -0,0 +1,744 @@
...
+        return goalieHasBall and cls.ballInGoalZone()
+
+        #print(constants.Field.OurGoalZoneShape)
+
+        #return False
+
+    #This function will detect if the ball is about to go out of bounds, or is headed towards the goal
+    @classmethod
+    def ballTrajectoryUpdate(cls, ballPos, ballVel, factor=0.5):
+        #Find function that determines if a point is in bounds
```
kylestach, line 657: https://github.com/RoboJackets/robocup-software/blob/master/common/Geometry2d/Rect.hpp#L60

soccer/gameplay/situational_play_selection.py
```diff
@@ -0,0 +1,744 @@
...
+        pass
+
+    @classmethod
+    def clearSituation(cls):
+        cls.currentSituation = cls.situation.none
+
+    @classmethod
+    def situationUpdate(cls):
+        #I've made all branches make an assignment for ease of debugging,
+        #none assignments have been marked
```
kylestach, line 671: Are `None` assignments supposed to eventually return something?

soccer/gameplay/situational_play_selection.py
```diff
@@ -0,0 +1,744 @@
...
+        elif(cls.pileupTime != None and not currentPileup):
+            cls.pileupTime = None
+
+        return cls.pileupTime != None and abs(time.time() - cls.pileupTime) >= pileupBufferTime 
+
+    @classmethod
+    def ballPossessionUpdate(cls):
+
+        minimumPassSpeed = 2.2 #The minumum speed for the ball to be traveling to look for recieving robots
+        ballRatioFactor = 6.0 #The ratio of robot closeness for automatic possession
```
kylestach: IMO these should be constants defined close to the top of the file.

soccer/gameplay/situational_play_selection.py
JNeiger: This should be `.mag()`

soccer/gameplay/situational_play_selection.py
JNeiger: Are you trying to get the angle between two vectors? fwiw `v1.angleBetween(v2)` exists

soccer/gameplay/situational_play_selection.py
```diff
@@ -0,0 +1,744 @@
...
+        
+        ballDist = math.sqrt(robotToBall[0]**2 + robotToBall[1]**2)
+        angle = math.degrees(math.atan2(ballVel.y, ballVel.x) - math.atan2(robotToBall[1], robotToBall[0]))
+
+        if(abs(angle) > 100):
+            return (None, None, None, None)
+       
+        robotToBallDOTBallVel = robotToBall[0] * ballVel.x + robotToBall[1] * ballVel.y
+        scalar = robotToBallDOTBallVel / (ballSpeed**2)
+        robotOntoVelocity = (scalar * ballVel.x, scalar * ballVel.y) #The robots position relative to the ball projected onto the balls velocity
```
JNeiger: `(robotToBall*ballVel.norm())*ballVel.norm()` produces the same results

soccer/gameplay/situational_play_selection.py
liquidmetal9015: I think my options are: 
- Leave it as a "static class" with class methods (just hope no one makes an instance, or throw an exception if they do)
- Make it an unenforced singleton which will have to live somewhere and be accessed from there
- Make it a module, so no class, just functions, classes and variables in a file
- Use some odd method of "enforcing" a singleton that lives in this file

I think it would be good if it lives in and is accessed from this file. Leaving it as a "static class" is easiest at this point and just throwing an exception in the constructor. idk

liquidmetal9015: Currently the only file that references this file is main.py, and I really don't feel super good about having it live there.

soccer/gameplay/situational_play_selection.py
kylestach: It's `v1.angle_between(v2)` in Python it looks like: https://github.com/RoboJackets/robocup-software/blob/staging/soccer/gameplay/robocup-py.cpp#L706

soccer/gameplay/situational_play_selection.py
kylestach: I think it's `.normalized()`.

So:

```py
ballDirection = ballVel.normalized()
robotOntoVelocity = ballDirection * ballDirection.dot(robotToBall)
```

soccer/gameplay/situational_play_selection.py
```diff
@@ -0,0 +1,620 @@
...
+    def isMidfield(self):
+        return self.ballLocation == self.fieldLoc.MIDFIELD
+
+    def isPileup(self):
+        return self.currentPileup
+
+    #Returns if we are in the specified situation without regard to capitalization
+    #If the check variable is true, it will check if the situation exists and throw an 
+    #exception is it does now.
+    def isSituation(self, situation, check=False):
```
kylestach, line 304: Do we actually use this anywhere? Seems like regular enum comparisons are more useful.

soccer/gameplay/situational_play_selection.py
```diff
@@ -0,0 +1,620 @@
+import main
+import time
+import robocup
+import math
+from enum import Enum
+import constants
+
+
+class SituationalPlaySelector:
```
Jason27chan, line 22: Can you add a comment here to describe the class

soccer/gameplay/situational_play_selection.py
```diff
@@ -0,0 +1,620 @@
...
+
+    pileupTime = time.time()  #The first time at which a pileup was detected
+
+    isSetup = False
+    gameState = None
+    systemState = None
+    robotList = list()
+    activeRobots = list()
+
+    def setupStates(self):
```
Jason27chan, line 112: comments

soccer/gameplay/situational_play_selection.py
```diff
@@ -0,0 +1,620 @@
...
+        self.systemState = main.system_state()
+        self.context = main.context()
+        for g in self.systemState.our_robots:
+            self.robotList.append(g)
+        for g in self.systemState.their_robots:
+            self.robotList.append(g)
+
+        self.updateRobotList()
+
+    def updateRobotList(self):
```
Jason27chan, line 127: Comments

soccer/gameplay/situational_play_selection.py
```diff
@@ -0,0 +1,620 @@
...
+        self.activeRobots.clear()
+        for g in self.robotList:
+            if (g.visible):
+                self.activeRobots.append(g)
+
+    #Calls all needed update functions
+    #Needs to be called in main loop to make the module work
+
+    def updateAnalysis(self):
```
Jason27chan, line 139: Move comments to right above function

soccer/gameplay/situational_play_selection.py
```diff
@@ -0,0 +1,620 @@
...
+        return robotsInBallPath
+
+    #Returns a tuple containing the robots distance from the balls path, the distance the ball has to travel to get to that intercept point, the balls speed, and the angle between the balls velocity and the robot's position
+    def ball_ingress(self, ballPos, ballVel, robot):
+
+        robotx = robot.pos.x
+        roboty = robot.pos.y
+
+        ballSpeed = ballVel.mag()
+        if (ballSpeed < 0.15):
```
Jason27chan: magic number?

soccer/gameplay/situational_play_selection.py
```diff
@@ -0,0 +1,620 @@
...
+            return (None, None, None, None)
+
+        robotToBall = robot.pos - ballPos
+        ballDist = robotToBall.mag()
+        angle = ballVel.angle_between(robotToBall)
+
+        if (abs(angle) > 100):
```
Jason27chan, line 191: magic number

soccer/gameplay/situational_play_selection.py
```diff
@@ -0,0 +1,620 @@
...
+            robotOntoVelocity.y - robotToBall.
+            y)  #The vector from the projected vector to the robots position
+
+        distanceFromPath = projectedToRobot.mag()
+        interceptDistance = robotOntoVelocity.mag()
+
+        return (distanceFromPath, interceptDistance, ballSpeed, angle)
+
+    #Returns the closest robot in the balls path as a tuple of the robot and its distance 
+    def closestReciever(self):
```
Jason27chan, line 213: Move function to evaluation?

soccer/gameplay/situational_play_selection.py
```diff
@@ -0,0 +1,620 @@
...
+        if (self.currentPreempt):
+            self.currentPreempt = False
+        elif (self.situationChangeTime != None):
+            if (abs(time.time() - self.situationChangeTime) >
+                    self.playPreemptTime):
+                self.situationChangeTime = None
+                self.currentPreempt = True
+                self.LastSituation = self.currentSituaion
+
+    '''def addPreempt(play) possibly a function to add transition out of every state to the completed state with preemptPlay as the lambda
```
Jason27chan, line 341: Take out comments and make it an issue

soccer/gameplay/situational_play_selection.py
```diff
@@ -0,0 +1,620 @@
...
+    def preemptPlay(self):
+        return self.currentPreempt
+
+    def ballToRobotDist(self, robot):
+        return math.sqrt((robot.pos.x - self.systemState.ball.pos.x)**2 + (
+            robot.pos.y - self.systemState.ball.pos.y)**2)
+
+    #Returns a tuple of the closest robot to the ball and the distance that robot is away from the ball
+
+    def closestRobot(self):
```
Jason27chan, line 356: didn't you make this function already?

soccer/gameplay/situational_play_selection.py
```diff
@@ -0,0 +1,620 @@
...
+
+    def ballClosenessRatio(self):
+        distances = self.bothTeamsClosest()
+        try:
+            return distances[0] / distances[1]
+        except:
+            return 1.0
+
+    #Returns the robot that last had the ball and how long it was since they had the ball
+
```
Jason27chan: Why do you keep leaving these awkward lines between the comment and the function lmao

soccer/gameplay/situational_play_selection.py
```diff
@@ -0,0 +1,620 @@
...
+    #Returns true if we had the ball last
+
+    def weHadBallLast(self):
+        if (self.hadBallLast()[0].is_ours()):
+            return True
+        else:
+            return False
+
+    def robotsWithTheBall(self):
```
Jason27chan, line 434: rip all the uncommented functions

soccer/gameplay/situational_play_selection.py
```diff
@@ -0,0 +1,620 @@
...
+    def updatePileup(self):
+
+        pileupBufferTimeIn = 0.3  #the number of seconds of pile up conditions before a pile up is declared
+        pileupBufferTimeOut = 0.3  #the number of seconds of not pile up conditions before a pile up is un declared
+
+        botsNearBall = self.nearBallCount(distance=0.35)
+        botsWithBall = self.withBallCount()
+        totalNearBall = sum(botsNearBall)
+        totalWithBall = sum(botsWithBall)
+        #print("Pileup info------------------")
```
Jason27chan: delete prints

soccer/gameplay/situational_play_selection.py
```diff
@@ -0,0 +1,620 @@
...
+        botsNearBall = self.nearBallCount(distance=0.35)
+        botsWithBall = self.withBallCount()
+        totalNearBall = sum(botsNearBall)
+        totalWithBall = sum(botsWithBall)
+        #print("Pileup info------------------") 
+        #print(botsNearBall)
+        #print(botsWithBall)
+        pileUpCalc = False
+
+        if (totalNearBall >= 3 and botsNearBall[0] > 0 and
```
Jason27chan, line 501: magic number

soccer/gameplay/situational_play_selection.py
```diff
@@ -0,0 +1,620 @@
...
+                botsNearBall[1] > 0):
+            pileUpCalc = True
+
+        if (totalWithBall >= 2 and botsWithBall[0] > 0 and
```
Jason27chan, line 505: magic number

soccer/gameplay/situational_play_selection.py
```diff
@@ -0,0 +1,620 @@
...
+            if (botsWithBall[0].is_ours()):
+                self.currentPossession = self.BallPos.OURBALL
+            else:
+                self.currentPossession = self.BallPos.THEIRBALL
+            return None
+
+        lastInfo = self.hadBallLast()
+        lastDurationThreshold = 0.5
+        lastDurationLengthThreshold = 0.5
+        #print(str(lastInfo[1]) + " " + str(lastInfo[2]))
```
Jason27chan: print statement

soccer/gameplay/situational_play_selection.py
```diff
@@ -0,0 +1,620 @@
...
+            return None
+        if (ballDistRatio < (1.0 / ballRatioFactor)):
+            self.currentPossession = self.BallPos.OURBALL
+            return None
+
+        #This should probably eventually be updated to take the intercept location into accoutn rather 
+        #than the current location for determining what situation we are in. Or not, I could see an arguement for both.
+        if (self.systemState.ball.vel.mag() > minimumPassSpeed):
+            recvr = self.closestReciever()
+            #if(recvr[0] != None):
```
Jason27chan: commented code

soccer/gameplay/main.py
```diff
@@ -201,6 +204,7 @@ def run():
     try:
         if root_play() is not None:
             root_play().spin()
+            situationAnalysis.updateAnalysis()
```
JNeiger, line 206: It's a 1 frame difference in lag, but is there any reason not to have this run before the root play?

soccer/gameplay/situational_play_selection.py
```diff
@@ -0,0 +1,742 @@
...
+
+        self.updatePileup()
+        self.locationUpdate()
+        self.ballPossessionUpdate()
+        self.situationUpdate()
+
+        #print(abs(startTime - time.time()))
+
+        #Print the current situation in the corner of the soccer gui
+        self.context.debug_drawer.draw_text(self.currentSituation.name,
```
JNeiger, line 155: It'd be nice at some point to throw this into the tree or something. That way we can see the state each frame since this is going to be very very important to have tuned correctly. There's a way to do it with normal plays, but I'm not sure about this specific case.

soccer/gameplay/situational_play_selection.py
```diff
@@ -0,0 +1,742 @@
...
+    isSetup = False
+
+    ##Holds the game state object
+    gameState = None
+
+    ##Holds the system state object
+    systemState = None
+
+    ##Holds a list of all the robots
+    robotList: List[robocup.Robot] = list()
```
JNeiger: I think you might have to do `robotList = list() # type: List[robocup.Robot]`

See https://www.python.org/dev/peps/pep-0484/#type-comments

soccer/gameplay/situational_play_selection.py
```diff
@@ -0,0 +1,742 @@
+import main
+import time
+import robocup
+import math
+from enum import Enum
+import constants
+from typing import List, Dict
```
JNeiger, line 7: "Don't use from xxxx import yyyy where yyyy is not a module, use import xxxx. This has to do with how python handles modules and the classes loaded from those modules. If you do it the wrong way, when a play file changes on disk and we reload it as a module, the yyyy that you imported from the module earlier won't get reloaded."

See https://github.com/RoboJackets/robocup-software/blob/63d982c288342243528836695e1bb5cec56a9745/soccer/gameplay2/readme.md.

kylestach: Good to know.

Jason27chan:
:angry: We're merging and testing the code without parameter documentation. We'll let it pass this time, but do it next time

Issue #1323, opened 2019-09-01
Angle planning part 1: path interface refactor


kylestach:
## Description
InterpolatedPath should interpolate angles, and we should populate these in the planners that we use. This is necessary for long-term angle planning (similar to #770, but this is actually actionable).

## Requirements

 - `InterpolatedPath` supports angle planning
 - `RRTPlanner` creates an `InterpolatedPath` with angles
 - `AngleFunctionPath` is removed
 - `RobotInstant` uses a `Pose` and `Twist` (angle is no longer optional)

## Design

[Link to design document](https://docs.google.com/document/d/1mveM6Q6uNT5T8Q3a5pQVPb19Xi-JDAWUPn6O1HKcsyc/edit#)

Issue #1318, opened 2019-08-27
Motion Benchmark Minimum Viable Product


liquidmetal9015:
## Description

A process to objectively measure and benchmark robot performance. This is already largly designed, see [design document](https://docs.google.com/document/d/1uA66JDNMuNfRuFQLymauNZkwIWz0kD3KOiUOSZp76vE/edit?usp=sharing) for details.

## Requirements

- Have a play that will run a robot through a sequence of motions while recording data about its performance.
- In the end, process that data into human readable scores and more refined data that can be meaningfully compared between runs, even across time.
- Ideally create a framework that can run different benchmarks consisting of different tests of which new ones can be added later on.

## Design

The design document contains the design details

[Link to design document](https://docs.google.com/document/d/1uA66JDNMuNfRuFQLymauNZkwIWz0kD3KOiUOSZp76vE/edit?usp=sharing)

PR #1324, opened 2019-09-01
Move angle planning to the planning phase

kylestach:
## Description
Currently, we effectively do angle planning at the "control" phase, because the angles from `AngleFunctionPath` are not evaluated until path execution time.

Instead, we want to allow `InterpolatedPath` to directly plan for angles, and the RRT planner to calculate angles.

Use cubic interpolation on paths (Hermite splines between position/velocity waypoints) for a properly smooth velocity profile.

This also allows us to interpolate over angles at the `InterpolatedPath` level. Currently this is not being used (we wrap the `InterpolatedPath` in an `AngleFunctionPath` in order to plan angles).

## Associated Issue
Issue #1323

## Design Documents
[Link](https://docs.google.com/document/d/1mveM6Q6uNT5T8Q3a5pQVPb19Xi-JDAWUPn6O1HKcsyc/edit#)

### Automated testing
(Run on CI)
Expected result: CI pass

### Soccer test
1. Run soccer
2. Select the `comp` playbook
3. There should be no measurable difference between this and the original.

### Motion benchmark
There should be no measurable difference on the motion benchmark before and after this PR. @liquidmetal9015's benchmark is a prerequisite for testing this.

### DIFF
--- a/soccer/planning/InterpolatedPath.cpp
+++ b/soccer/planning/InterpolatedPath.cpp
@@ -12,13 +12,13 @@ using namespace Geometry2d;
 
 namespace Planning {
 
-InterpolatedPath::InterpolatedPath(Point p0) {
-    waypoints.emplace_back(MotionInstant(p0, Point()), RJ::Seconds::zero());
+InterpolatedPath::InterpolatedPath(RobotInstant p0) {
+    waypoints.emplace_back(p0, RJ::Seconds::zero());
 }
 
-InterpolatedPath::InterpolatedPath(Point p0, Point p1) {
-    waypoints.emplace_back(MotionInstant(p0, Point()), 0ms);
-    waypoints.emplace_back(MotionInstant(p1, Point()), 0ms);
+InterpolatedPath::InterpolatedPath(RobotInstant p0, RobotInstant p1) {
+    waypoints.emplace_back(p0, 0ms);
+    waypoints.emplace_back(p1, 0ms);
 }
 
 float InterpolatedPath::length(unsigned int start) const {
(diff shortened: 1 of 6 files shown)


kylestach:
Validation blocked on #1318.

soccer/planning/InterpolatedPath.cpp
```diff
@@ -190,69 +195,76 @@ void InterpolatedPath::draw(DebugDrawer* const debug_drawer,
...
-        return RobotInstant(waypoints[i].instant);
-    }
-    float constant = (t - waypoints[i - 1].time) / deltaT;
 
-    return RobotInstant(MotionInstant(waypoints[i - 1].pos() * (1 - constant) +
-                                          waypoints[i].pos() * (constant),
-                                      waypoints[i - 1].vel() * (1 - constant) +
-                                          waypoints[i].vel() * (constant)));
+    assert(prev_it != waypoints.end());
+    assert(next_it != waypoints.end());
```
JNeiger, line 229: Due to the if statements above, this should never happen, but if for some reason we get to this while loop when `t > waypoints.back().time`, it will possibly crash on `next_it->time` inside the if statement on line 221 instead of this assert. If `next_it == waypoints.back()`, it enters the loop, increments `next_it` to `waypoints.end()` then tries to dereference a past-the-end iterator resulting in undefined behavior.

https://repl.it/repls/BigheartedAthleticConversion

soccer/planning/InterpolatedPath.cpp
```diff
@@ -190,69 +195,76 @@ void InterpolatedPath::draw(DebugDrawer* const debug_drawer,
...
+             Eigen::Vector3d(tangent_0) * (s * s * s - 2 * s * s + s) +
+             Eigen::Vector3d(pose_1) * (-2 * s * s * s + 3 * s * s) +
+             Eigen::Vector3d(tangent_1) * (s * s * s - s * s));
+
+    Twist interpolated_twist =
+        Twist(Eigen::Vector3d(pose_0) * (6 * s * s - 6 * s) +
+              Eigen::Vector3d(tangent_0) * (3 * s * s - 4 * s + 1) +
+              Eigen::Vector3d(pose_1) * (-6 * s * s + 6 * s) +
+              Eigen::Vector3d(tangent_1) * (3 * s * s - 2 * s)) /
+        RJ::numSeconds(dt);
```
JNeiger, line 267: If there is a source on these equations, even if it's a wikipedia link, that's always good to throw into a comment.

JNeiger (approved):
Everything else looks good

soccer/planning/InterpolatedPath.cpp
kylestach: Good catch. I think this can be fixed by moving the check above the iterator increments.

soccer/planning/InterpolatedPath.cpp
kylestach: ack

Zolkin1 (approved):
All looks good to me.

Issue #1312, opened 2019-08-22
Delete "old" directory


kylestach:
There exist some old prototypes for ball filter as well as some bezier curve code in the old/ directory. All of it is unused and as far as I can tell it is not expected to be used in the future. We should delete it, as at this point it just clutters the repo.

PR #1327, opened 2019-09-03
Delete old/ directory

kylestach:
## Description
Deletes unused files in old/*.

These files were either old code or prototypes that we didn't end up using. They probably shouldn't have been checked into the main repo anyway.

## Associated Issue
#1312 

## Design Documents
N/A

## Steps to test
Code was unused, testing unnecessary. As long as this doesn't break CI we're good to go!


CollinAvidano (approved):
Please burn it all...
Finish what I started

JNeiger (approved):
It looks like it was replaced in 2011, after being written in 2009. Imma go on a limb and say it dead.

Blast from the past though looking at the play system at that time.
https://github.com/RoboJackets/robocup-software/tree/0756a9fee19a94e13c888f099e2e7b8cc75f3a41/soccer/gameplay

Issue #1320, opened 2019-08-27
Clear play


HussainGynai:
## Description

A play that will activate on defense that will try to chip the ball up field while also moving robots up field

## Requirements

 - Chip the ball up field
 - Move robots up field

## Design

https://docs.google.com/document/d/1l8Yg3pnwk4c_xT5Ben8W-NOcU-Ope5kT-iGmuAXvBcY/edit# (link)

kylestach:
Scoping this for the next testing milestone based on your time estimate.

PR #1331, opened 2019-09-08
Clear play

HussainGynai:
## Description
A play that will kick the ball up field while also moving robots up field to try and retrieve the ball.

## Associated Issue
Issue #1320 

## Design Documents
[Link](https://docs.google.com/document/d/1l8Yg3pnwk4c_xT5Ben8W-NOcU-Ope5kT-iGmuAXvBcY/edit#)

### Sim Test
1. Open the sim
2. Select  Clear from the playbook
3. Run the play multiple times using different initial positions for both friendly and opposing robots, preferably starting the ball on the friendly side.

Expected result: Ball should be chipped up the field while other robots also move up to get the ball.


soccer/gameplay/plays/offense/clear.py
```diff
@@ -0,0 +1,95 @@
...
+
+        self.midfield_point_2 = robocup.Point(-0.40*constants.Field.Width,
+                                             0.60*constants.Field.Length)
+
+        self.points = [self.offense_point_1, self.offense_point_2, self.midfield_point_1, self.midfield_point_2]
+
+        self.add_transition(behavior.Behavior.State.start,
+                            Clear.State.get_ball,
+                            lambda: True,
+                            'Immidietley')
```
kylestach: Immediately

soccer/gameplay/plays/offense/clear.py
```diff
@@ -0,0 +1,95 @@
...
+            count ++
+            self.add_subbehavior(skills.move.Move(i),
+                                'move to point ' + count,
+                                required = True)
+
+    def on_enter_clear_ball(self):
+        #Chip ball to either offense points, while still moving robots up
+        self.remove_all_subbhevaiors()
+
+        num = random.randint(0,1)
```
kylestach, line 79: I would do `point = random.choice(offense_points)` and break the two types of points into two separate arrays.

soccer/gameplay/plays/offense/distraction.py
```diff
@@ -0,0 +1,256 @@
...
+import constants
+import enum
+import standard_play
+import evaluation
+import tactics.coordinated_pass
+import skills.move
+import skills.capture
+
+
+class Distraction(standard_play.StandardPlay):
```
kylestach: This was already merged. I'll see if I can fix the diff.

soccer/gameplay/plays/offense/clear.py
```diff
@@ -0,0 +1,95 @@
...
+        self.add_subbehavior(tactics.coordinated_pass.CoordinatePass(self.point[num], use_chipper = True),
```
kylestach:
```suggestion
        self.add_subbehavior(tactics.coordinated_pass.CoordinatedPass(self.point[num], use_chipper = True),
```

soccer/gameplay/plays/offense/clear.py
```diff
@@ -0,0 +1,95 @@
...
+                                                                     'clear',
+                                                                     required = True)
+        count = 0
+        for k in points
+            count ++
+            self.add_subbehavior(skills.move.Move(k),
+                                'move to point ' + count + ' 2',
+                                required = True)
```
kylestach: Should we skip adding the move subbehavior for the receiving point of CoordinatedPass? (Actually just asking, I'm not entirely sure how plays work)

kylestach (commented):
Mostly looks good! Here's some general feedback. I'll try to test later today.

kylestach:
Switching the base branch to master and back looks like it fixed the extra files.

soccer/gameplay/plays/offense/clear.py
```diff
@@ -0,0 +1,95 @@
...
+    def on_enter_get_ball(self):
+        self.remove_all_subbhevaiors()
+        #Capture the ball and move robot up
+        self.add_subbehavior(skills.capture.Capture(),
+                            'Capture ball',
+                            required = True)
+        
+        count = 0
+        for i in points
+            count ++
```
kylestach: Uh, this isn't valid Python.

soccer/gameplay/plays/offense/clear.py
kylestach: Also, use_chipper doesn't currently actually exist on CoordinatedPass. You'll probably need to add that functionality.

soccer/gameplay/plays/offense/clear.py
```diff
@@ -0,0 +1,95 @@
+import main
+import robocup
+import behavior
+import constants
+import enum
+import standard_play
+import evaluation
+import tactics.coordinated_pass
```
Jason27chan, line 9: please separate your imports like [adaptive formation](https://github.com/RoboJackets/robocup-software/blob/staging/soccer/gameplay/plays/offense/adaptive_formation.py)

soccer/gameplay/plays/offense/clear.py
```diff
@@ -0,0 +1,95 @@
...
+import skills.move
+import skills.capture
+import random
+
+#Based on clear sitation
```
Jason27chan, line 14: Remove comments before merging

soccer/gameplay/plays/offense/clear.py
```diff
@@ -0,0 +1,95 @@
...
+#    Have ball in defense
+#   Want to Kick ball up field and get other robots in position
+#
+
+class Clear(standard_play.StandardPlay):
```
Jason27chan, line 19: Please leave comments at the top of the class describing what it does.

soccer/gameplay/plays/offense/clear.py
```diff
@@ -0,0 +1,95 @@
...
+        clear_ball = 2, 'Kick the ball upfield'
+
+    def __init__(self):
+        super().__init__(continuous = False)
+
+        for s in Clear.State:
+            self.add_state(s, behavior.Behavior.State.running)
+
+        #Points to move up offense bots to
+        self.offense_point_1 = robocup.Point(0.40*constants.Field.Width,
```
Jason27chan, line 32: What if we want different numbers of offense and midfield robots? These shouldn't be hardcoded but we should be able to alter it easily, especially if we want to move up to 11 robots eventually

soccer/gameplay/plays/offense/clear.py
```diff
@@ -0,0 +1,95 @@
...
+
+#Based on clear sitation
+#    Have ball in defense
+#   Want to Kick ball up field and get other robots in position
+#
+
+class Clear(standard_play.StandardPlay):
+
+    class State(enum.enum)
+        get_ball = 1, 'Get the ball and movee other robots up'
```
Jason27chan, line 22: does this compile lmao

soccer/gameplay/plays/offense/clear.py
Jason27chan: we shouldnt because coordinated pass would wait for the receivers to reach the point before kicking, which is not what we want.

soccer/gameplay/plays/offense/clear.py
Jason27chan: I dont think we should do a coordinated pass because that would wait for the robots to reach their receive points. We can just kick it up the field and transition to a new type of play

Jason27chan (changes requested):
Make sure to comment so that [doxygen](www.doxygen.nl/manual/docblocks.html#pythonblocks) can read the code. (Use the ## format)

soccer/gameplay/plays/offense/clear.py
JNeiger: You probably don't want these to be required either

liquidmetal9015:
Uh... So @HussainGynai, why did this get deleted from staging? It's not there anymore, and detective woodward says you deleted it with a commit called "delete clear.py" pushed directly to staging???

PR #1330, opened 2019-09-08
Motion control module

kylestach:
## Description
Refactor motion control into a module that iterates over robots and applies motion control to each one.

## Associated Issue
Issue #1329.

## Design Documents
[Link](https://docs.google.com/document/d/13LJ-HotI7wjp8G1xqDhcK5y4MAMSmWZyr55kbajW1qw/edit#heading=h.yvwkw0oatlpt)

## Steps to test
Non-functional change. As long as it doesn't break halt state or joysticks we're good.

Issue #1329, opened 2019-09-08
(Tracking issue) Refactor Processor into modules


kylestach:
# Description
Currently, Processor has a lot of extra logic to iterate over robots and apply modules to each robot, etc. We want to move logic out of processor, so we will be introducing modules. Each module will be run in sequence (Processor will contain a `vector<unique_ptr<Module>>`) and will own a context pointer (so it will know how to operate).

## Requirements

 - [x] Module base class (#1330)
 - [x] Motion control module (#1330)
 - [x] Vision filter module
 - [x] Vision receiver module (#1332)
 - [x] Radio module (#1400)
 - [x] Gameplay module (trivial)
 - [x] Joysticks (see #1325)
 - [x] Referee module (trivial)
 - [x] Logger module

## Design

[Link to design document](https://docs.google.com/document/d/13LJ-HotI7wjp8G1xqDhcK5y4MAMSmWZyr55kbajW1qw/edit#heading=h.yvwkw0oatlpt)

Also, see #1303 for information about how we plan to pass information between modules.

jpfeltracco:
This might be dumb, feel free to ignore 😉. But I got a bit confused by the term "Module" at first and I think others might too in the future. C++ has had "modules" support coming out for a while (should be slated for C++20 now).

Any other terms that might prevent overloading "modules"? "Manager", "handler", "thingythatdoesstuff"?

kylestach:
@jpfeltracco that's a fair point - thinking about it, that's going to be one of those things that causes people to curse my name forever post-2020.

I'm not sure about Manager, I don't really think that captures it. `Runnable` might work (or if we really want to go all out on the Processor theme, `Processable` :stuck_out_tongue_winking_eye: ) - although the -able suffix feels sort of Java-y to me.

oswinso:
How about.... `Node`? 😋

kylestach:
`Subsystem` maybe? A little bit odd but no conflicts with existing terminology...

kylestach:
Update: discussed offline, we're going with `Node`.

PR #1332, opened 2019-09-09
Vision receiver module

kylestach:
## Description
Refactor vision receiver into a new node and refactor to use boost::asio instead of a separate thread. ~Currently marked as draft until #1330 (which includes the Node base class) is merged.~

## Associated Issue
Issue #1329.

## Design Documents
[Link](https://docs.google.com/document/d/13LJ-HotI7wjp8G1xqDhcK5y4MAMSmWZyr55kbajW1qw/edit#heading=h.3vz8rotbuhas)

## Steps to test
This is a non-functional change. It needs QA to make sure it didn't break any of the following functionality:
 - Vision on actual field
 - Using specific vision channels on the field

### DIFF
--- a/soccer/Context.hpp
+++ b/soccer/Context.hpp
@@ -3,8 +3,8 @@
 #include "DebugDrawer.hpp"
 #include "GameState.hpp"
 #include "SystemState.hpp"
-#include "vision/VisionPacket.hpp"
 #include "WorldState.hpp"
+#include "vision/VisionPacket.hpp"
 
 struct Context {
     Context() : state(this), debug_drawer(this) {}
--- a/soccer/Context.hpp
+++ b/soccer/Context.hpp
@@ -3,6 +3,7 @@
 #include "DebugDrawer.hpp"
 #include "GameState.hpp"
 #include "SystemState.hpp"
+#include "vision/VisionPacket.hpp"
 
 struct Context {
     Context() : state(this), debug_drawer(this) {}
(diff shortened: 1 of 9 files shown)


soccer/Context.hpp
```diff
@@ -17,4 +18,6 @@ struct Context {
     SystemState state;
     GameState game_state;
     DebugDrawer debug_drawer;
+
+    std::vector<std::unique_ptr<VisionPacket>> vision_packets;
```
kylestach, line 23: @CollinAvidano, I think we discussed merging VisionReceiver into the VisionFilter module, and I'm now thinking that might be a good idea. That way we don't have to have this nasty vector of protobuf structures passed around and Context only contains "clean" data.

Whatever we do is pending moving logging stuff into a module, though.

kylestach:
This is now ready for review.

JNeiger (approved):

SimonAbrelat (approved):
The changes seem totally reasonable for the VisionReceiver, but it seems weird to add more data to the already huge Context class. All that being said, I am a noob and chances are that concern is meaningless lmao.

kylestach:
The concern is valid. Basically, context exists as a giant glob of data because we don't have proper pub/sub or similar. It's not ideal but the idea is that processor shouldn't need to know any of the particulars of the modules it calls, and they should only communicate through the Context struct.

kylestach:
@SimonAbrelat also, see my comment on that field for future plans regarding it. Going to close this now though.

Issue #1335, opened 2019-09-15
RefBox crashes sim2play


kylestach:
## Description
When we run RefBox while sim2play is running, it tries to change both instances to blue team. This breaks, because we can't bind to the same port twice. 

## Steps to reproduce
1. Run `make run-sim2play`
2. Run refbox

Result: the yellow instance crashes because it is unable to bind correctly. 

- [x] This bug has been reproduced by someone else

## Additional information
I'll fill this in as I work on it.

PR #1336, opened 2019-09-18
Fix referee receiver to not change teams when no team name matches

kylestach:
## Description
Previously, soccer would always switch to a new team color immediately when it received a ref packet, even if our team name did not match the team name sent over in the packet.

This caused conflicts when running two instances of soccer on the same machine because:
1) When refbox is run, it defaults to both team names empty
2) When we received an empty string, we would switch to the blue team
3) When there is already an instance of soccer binding to the blue team's port, a new instance cannot bind to that port.

This created a crashing situtation.

I fixed it by:
 - Ignoring empty names
 - Only switching if names match exactly

## Associated Issue
Issue #1335

## Design Documents
N/A, bugfix

### Test Case 1
1. Run refbox
2. `make run-sim2play`

Expected result: One instance should remain blue and the other should remain yellow. You should be able to control the game from refbox.

### DIFF
--- a/soccer/NewRefereeModule.cpp
+++ b/soccer/NewRefereeModule.cpp
@@ -1,5 +1,7 @@
 #include "NewRefereeModule.hpp"
 
+#include <boost/algorithm/string/predicate.hpp>
+
 #include "Constants.hpp"
 
 #include <Network.hpp>
@@ -106,11 +108,12 @@ static const int KickVerifyTime_ms = 250;
 // the ref halts/stops, make this false
 static const bool CancelBallPlaceOnHalt = true;
 
-NewRefereeModule::NewRefereeModule(Context* const context)
+NewRefereeModule::NewRefereeModule(Context* const context, bool isBlue)
     : stage(NORMAL_FIRST_HALF_PRE),
       command(HALT),
       _running(false),
-      _context(context) {}
+      _context(context),
+      _blueTeam(isBlue) {}
 
 NewRefereeModule::~NewRefereeModule() { this->stop(); }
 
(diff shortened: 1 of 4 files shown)


kylestach:
FWIW I'm planning on rewriting the `NewRefereeModule` pretty soon, and the `blueTeam` variable will exist in the `Context` struct. But this is the best solution for the time being.

soccer/NewRefereeModule.cpp
```diff
@@ -186,17 +189,35 @@ void NewRefereeModule::run() {
...
-            letter = tolower(letter);
+        // If we have no name, we're using a default config and there's no
+        // sense trying to match the referee's output (because chances are
+        // everything is just on default configuration with no names, so more
+        // than one team/soccer instance will be trying to use the same color)
+        if (Team_Name_Lower.length() > 0) {
+            // We only want to change teams if we get something that actually
+            // matches our team name (either yellow or blue).
+            // Otherwise, keep the current color.
+            if (boost::iequals(yellow_info.name, Team_Name_Lower)) {
```
JNeiger, line 200: I think this could just be `yellow_info.name == Team_Name_Lower`

kylestach: iequals is case-insensitive comparison. I don't think we have a guarantee that `yellow_info.name` from the packet will always be lower case, do we?

JNeiger: Oh rip, should have read the second sentence of the two sentence doc page that talked about it being case-insensitive. 100% better to assume not and just deal with it

JNeiger (approved):

liquidmetal9015:
So I'm noticing that if I try to set robojackets as one of the teams in either of the drop downs in refbox, one of the soccer instances still crashes, which I suppose makes sense since they are both robojackets clients. But by checking the "ignore rules" box I can make the instances play soccer against each other with the dropboxes empty, which is what we are going for. So the functionality seems good.

liquidmetal9015 (approved):
I've checked the functionality including non simulator runs, toggling external referee, and launching and closing refbox and soccer in various combinations and everything checks out. The code looks good to me as well.

liquidmetal9015:
I believe I was able to produce a situation where a non-sim soccer instance did not respond to refbox being open but tried a lot and was unable to duplicate it with a bunch of combinations so I believe that was a fluke or I was misinterpreting what was open when.

Issue #1334, opened 2019-09-13
(Tracking Issue) Add Formations into Plays


JNeiger:
# Description
Currently there are some major difficulties and roadblocks that exist as we want to move the number of plays up, some of those include controlling all the robots, transitioning between plays, and scoring all the plays. This effort is to improve that.

# Requirements
- [ ] Formation Framework
- [ ] Positions
- [ ] Role Assignment
- [ ] Play Scoring
- [ ] Play prempt / valid region
- [ ] Role Queue

# Design
[Link to design document](https://docs.google.com/document/d/1YvTT1JhMt96bAyG10OOBN2x2l5xSfB9hDgA5_lJJ20E/edit#)
[Link to original freeform design document](https://docs.google.com/document/d/1U4LDRRgShm3CJHarOaSNgqiQhvnxhOag_Xn5PAGt6a4/edit?usp=drivesdk)

JNeiger:
I'm going to build this on another staging branch I think since this will cause significant and breaking changes as I work through it. It'll be as close as possible to what is one the real staging / master branch.

PR #1349, opened 2019-10-04
Latent Behaviors in Formation

JNeiger:
## Description
Partial implementation of formations. For right now this just means the basic latent behaviors and the setup of positions.

TODO:
- [X] Clean up center code

## Associated Issue
Issue #1334 partial

## Design Documents
[Link](https://docs.google.com/document/d/1YvTT1JhMt96bAyG10OOBN2x2l5xSfB9hDgA5_lJJ20E/edit#)


JNeiger:
Not sure how to set this up as a draft while I make the final fixes

oswinso:
I think when creating the PR instead of clicking the "Create pull request" button you can click the dropdown and do "Create draft pull request" and that will do it

JNeiger:
Meh, thats too much effort to fix. Close enough

kylestach:
This is looking pretty good, I'm very excited about this. My one question involves how you pull robots out of formation - are they just a lower priority in role assignment/things should work automatically?

JNeiger:
For right now, it'll be lower priority. I want a way to do a specific robot selection if it comes out cleaning in the future though.

oswinso:
One thing I am curious about is whether "formations" based positioning or "zone" based positioning would be more effective. It seems that ZJU uses a zone based system slightly different from the proposed overhaul.

kylestach:
This supports that as a potential future enhancement, by changing the implementation of the position class.

JNeiger:
They aren't going to be constrained to a single point, but more general zones within that area that keeps the general position. It's hard to explain but it'd be a more formal version of what they do. I'm hoping they would be trying to get more open for passes and that kind of thing

kylestach:
So, it looks like there's a bunch of logic to update the center of the formation, which as far as I can tell is unused (correct me if I'm wrong).

How exactly is this information used? I assume it should be used to dynamically shift the formation, right?

JNeiger:
It's going to be reenabled once I figure the ball.x/y problem. It'll shift the formation around to stay within field boundaries as well as do something smarter than averaging the ball position like the prototype

kylestach:
@JNeiger any update on this?

JNeiger:
I fixed the ball pos thing, so I just need to fix the bug with formation center, then its good to merge. I'll probably leave it as is functionality wise. I need to mess with role requirements to get the "takeover" working.

JNeiger:
Gonna say this is good enough break right now. Right now this just does formation creation. I think I need to redo how some of the low level role assignment works to really get nonbreaking behavior on stuff. I will probably go back eventually and clean up the long names and simplify stuff, but it's close enough to show what's happening right now.

kylestach (approved):
Looks good! I'm really excited about this.

Jason27chan:
It looks good overall! I'd just like to understand better how this'll work with Chris' situation analysis

JNeiger:
It doesn't directly interact with the situation stuff, but more so help the overall goal of creating more plays. Providing default locations for robots allow you to only use the ones you want to quickly improve the speed of play creation. Additionally, this allows better transitions between plays which is really important as we start using more than one with the situations.

Jason27chan (approved):

kylestach:
@JNeiger is this mergeable as soon as mypy passes?

tjones320 (approved):
I still want to see how we are handling preemption, but I think this is a solid step forward. Super cool!

JNeiger:
Oh whoops, I forgot I was crashing mypy and needed to fix that. I'll get on it.

JNeiger:
Oof, this is annoying. I want to statically type a class function that returns a list of objects of that class. This isn't allowed in the current version so I have to use the `from __future__ import annotations` (See PEP 563). This breaks any type of python style or linter because `__future__` doesn't exist in their mind.

mypy and pylint have oneliners that I can use to disable the errors, but I think `test-python` doesn't use `python 3.7` like the others. It uses `python 3.6` so `__future__` actually doesn't exist and rightfully causes an error.

JNeiger:
`test-python` is going to break, but I don't think that's me. The version of python on ci is 3.6 when we need 3.7. @kylestach is there somewhere where I can bump that version up?

kylestach:
~Why do you need 3.7?~ (Edit: forgot to read above comment.) I'd rather not version bump considering Ubuntu 18.04 still comes with 3.6 default and I don't want to mess up the versioning on everyone's systems. I don't think it would be unreasonable to just leave that one method without type hinting, right?

JNeiger:
It was fighting me earlier when I did it without, but I'll try again. There was probably a partial type hint somewhere.

JNeiger:
asdf;lkjaweef;oiij Finally got it. Had to ignore the initial variable declaration type or it'd complain

PR #1354, opened 2019-10-06
Training Triangle Pass file

tjones320:
## Description
A skeleton file to guide people through making triangle pass

## Associated Issue
NA

## Design Documents
NA

## Steps to test
There will be no tests

Expected result: It shouldn't work until we have hard-working new members


chachmu (commented):
LGTM

Issue #1340, opened 2019-09-29
Coordinated Pass Failure to Terminate


liquidmetal9015:
## Description
When playing in simulated and real matches, we often end up in a state in which Coordinated pass should have terminated but it has not. 

1. Passing robot has fully lost the ball at any point before the pass was made
2. Passing robot fails to kick the ball, but is unable to get the ball again due to role requirements
3. The ball was intercepted in between the passing robot and the receiving robot.

All of these situations lead to major paralysis of the team.

![FrozenByCoordinatedPass](https://user-images.githubusercontent.com/8135314/65836576-63a68880-e2bf-11e9-8b29-d3fbf627d53e.png)

![CoordinatedPassNotInterrupting](https://user-images.githubusercontent.com/8135314/65836602-791bb280-e2bf-11e9-9d1a-2d95f9edd028.png)

## Steps to reproduce
1. Create one of the situations in which Coordinated pass should be failing. It will happen naturally when running sim2play

Result: Coordinated pass continues to run when neither the passer or the receiver are in a position to possess the ball.

- [x] This bug has been reproduced by someone else

## Additional information

The primary solution to this problem will probably consist of adding robust termination conditions to the various states of coordinated pass. Adding a conditions that look at the rate of approach of the ball to the receiving robot and at the possession of the ball might be good places to start.

liquidmetal9015:
There is actually another major aspect to this: When the passing path becomes invalid due to obstacles getting in the way, coordinated pass tries to pass through them. This behavior needs to be fixed as well, probably by terminating coordinated pass. It's possible that coordinated pass needs some kind of overhaul.

Here is a demonstration of that behavior: 
![NotAValidPass](https://user-images.githubusercontent.com/8135314/65837646-fcd99d00-e2c7-11e9-9c8a-9d796eab3c1f.png)

Another instance:
![PanicKick](https://user-images.githubusercontent.com/8135314/65837669-49bd7380-e2c8-11e9-881e-f2af4bdaff27.png)

liquidmetal9015:
To some degree Situational Analysis based play preemption should help with this, as the play that is calling the Coordinated Pass would be terminated once we enter the pileup state. But the general notion is that we should attempt to fix it in the current system, and that Coordinated pass failing should be an expected possible outcome when any of these things happen.

kylestach:
Let's get some quality testing of this now that #1355 is merged, and then if we don't see the bad behavior we'll call it good to go.

PR #1355, opened 2019-10-06
Coordinated pass fix

HussainGynai:
## Description
Describe your pull request.

## Associated Issue
Issue #1340 

## Design Documents
[https://docs.google.com/document/d/1e7B3HV5H5m-Zxsw3CxUtah3iCKcPfLQXctz8hElXgXc/edit#heading=h.mk658lv28bsx](link-to-design-doc)

### Test Case 1
1. Run a play that passes (Distraction or Adaptive formation)
2. Just watch it pass and see if it ever reaches a state of non-termination

Expected result: Coordinated pass should terminate as expected


soccer/gameplay/tactics/coordinated_pass.py
```diff
@@ -175,15 +189,37 @@ def use_chipper(self, value):
     def on_exit_running(self):
         self.remove_subbehavior('receiver')
 
-    def on_enter_kicking(self):
+    # Funciton to find if ball is moving towards receive point
+    def ball_heading_to_receive_point(self):
+        ball_velocity = main.ball().vel;
+        if ball_velocity.mag() > 0.1:
```
kylestach: I would prefer this to be configurable, or at least a constant pulled out somewhere. `MIN_BALL_MOVING_SPEED`?

Similar with other constants below.

soccer/gameplay/tactics/coordinated_pass.py
```diff
@@ -175,15 +189,37 @@ def use_chipper(self, value):
...
+            ball_velocity_unit = [ball_velocity.x/ball_velocity.mag(), ball_velocity.y/ball_velocity.mag()]
```
JNeiger: Can do `ball_velocity.normalized()`

soccer/gameplay/tactics/coordinated_pass.py
```diff
@@ -175,15 +189,37 @@ def use_chipper(self, value):
...
+            path_vector = main.ball().pos - self.receive_point
+            path_unit = [path_vector.x/path_vector.mag(), path_vector.y/path_vector.mag()]
+            cross_product_mag = abs(ball_velocity_unit[0] * path_unit[1] -
```
JNeiger: You'll have to add `cross` to the robocup-py.cpp, but that exists too

soccer/gameplay/tactics/coordinated_pass.py
```diff
@@ -175,15 +189,37 @@ def use_chipper(self, value):
     def on_exit_running(self):
         self.remove_subbehavior('receiver')
 
-    def on_enter_kicking(self):
+    # Funciton to find if ball is moving towards receive point
+    def ball_heading_to_receive_point(self):
+        ball_velocity = main.ball().vel;
```
JNeiger: Make sure to test this on the field. In the past with the old vision filter, there existed a problem where the ball velocity was in the opposite direction of the kick for the first couple frames causing things to fail

soccer/gameplay/tactics/coordinated_pass.py
```diff
@@ -175,15 +189,37 @@ def use_chipper(self, value):
...
+
+    # Funciton that decides if an opponent robot is too close to the kicking robot
+    def opponent_too_close(self):
+        for bot in main.their_robots():
+            distance = (main.ball().pos - bot.pos).mag()
+            if distance < 0.27:
+                return True # If the ball is too close to the kicking robot then fail
+        return False
+
+    def execute_kicking(self):
```
JNeiger: What was the reasoning to swap this from on_enter to execute?

soccer/gameplay/tactics/coordinated_pass.py
```diff
@@ -240,10 +276,12 @@ def time_remaining(self):
             return 0
         return self.prekick_timeout - (time.time() - self._preparing_start)
 
-    def on_enter_receiving(self):
+    def execute_receiving(self):
```
JNeiger: Same as above, what was the reasoning to swapping this?

soccer/gameplay/tactics/coordinated_pass.py
HussainGynai: Originally I was gonna run the new function in this so needed it to be on execute, but then realized I could do it in the transition lambda, so I can just change it back to on enter. This is the same reasoning for the receiving on execute.

soccer/gameplay/tactics/coordinated_pass.py
```diff
@@ -175,15 +189,37 @@ def use_chipper(self, value):
...
+                ball_velocity_unit[1] * path_unit[0])
+            distance = path_vector.mag()
+            if distance < 0.3:
+                return True # If the ball is already close to the receive point but not heading towards it, dont fail
+            if (distance > 0.5 and cross_product_mag > 0.1) or (distance < 0.5 and cross_product_mag > 0.4) or ball_velocity.mag() < 0.05:
+                return False # If ball is too off course or too slow fail the pass
+        return True
+
+    # Funciton that decides if an opponent robot is too close to the kicking robot
+    def opponent_too_close(self):
```
kylestach: I'm kind of sketched out by the fact that this is a thing. Seems pretty questionable, but I'm alright with it as a hotfix.

Maybe put a TODO so we know to look there later.

kylestach (changes requested):

soccer/gameplay/tactics/coordinated_pass.py
kylestach: That shouldn't be an issue assuming a sufficiently high value of MIN_BALL_MOVING_SPEED. Something to keep an eye on during the testing day, though.

kylestach (approved):
LGTM after reformat.

soccer/gameplay/tactics/coordinated_pass.py
```diff
@@ -67,6 +68,14 @@ def __init__(self,
         self.receiver_required = receiver_required
         self.kicker_required = kicker_required
 
+        self.MIN_BALL_MOVING_SPEED = 0.1
+        self.TOO_CLOSE_TO_FAIL_DISTANCE =0.3
+        self.FAR_DISTANCE = 0.5
+        self.STRICT_CROSS_THRESHOLD = 0.1
+        self.RELAXED_CROSS_THRESHOLD = 0.4
+        self.OPPONENT_DISTANCE_THRESHOLD = 0.27
+        self.BALL_STOPPED_SPEED = 0.05
```
JNeiger: We usually make these class variables. I don't know if we still want that, but yeah

JNeiger (approved):

soccer/gameplay/tactics/coordinated_pass.py
kylestach: Yeah, that's probably a better idea.

soccer/gameplay/tactics/coordinated_pass.py
Jason27chan: I disagree with stopping a pass because an opponent is too close. If we consider the coordinated pass to fail, we would transition to the next play / situation, which would probably include a kick anyway

soccer/gameplay/tactics/coordinated_pass.py
```diff
@@ -83,13 +92,19 @@ def __init__(self,
...
-                     skillreceiver.State.aligned), 'kicker and receiver ready')
+                     and (self.subbehavior_with_name('receiver').state == self.
+                     skillreceiver.State.aligned or not self.receiver_required)),
+                    'kicker and receiver ready')
 
         self.add_transition(
             CoordinatedPass.State.preparing, CoordinatedPass.State.timeout,
             self.prekick_timeout_exceeded, 'Timed out on prepare')
 
+        self.add_transition(CoordinatedPass.State.preparing,
```
Jason27chan: disagree with failing if opponent is too close

Jason27chan (commented):
Comment functions like https://github.com/RoboJackets/robocup-software/blob/staging/soccer/gameplay/evaluation/passing_positioning.py#L87

kylestach:
@HussainGynai please `make checkstyle` and `git pull origin staging` to resolve CI issues.

Issue #1342, opened 2019-09-29
Updating current plays to use situation analysis and fleshing out preemption


liquidmetal9015:
## Description

So situation analysis is currently determining situations, but there is currently no model for how plays' scoring functions to interact with situation analysis and to preempt 

## Requirements

 - Current plays become complaint with situation analysis based scoring
 - A basic play preemption system exists and functions. This is more of a prototype

## Design

Current idea is to use high score for non-relevant situations, low score for relevant situations, and flash to -1 when the play is currently running and the currentPreempt flag in situation analysis goes to true.

liquidmetal9015:
This is done to the extent talked about here.

PR #1357, opened 2019-10-06
Situational plays beta

liquidmetal9015:
## Description
Some simple changes to fix the preempt function and get the scores and preemption actually working. This will need some testing and there might be an cleaner way to do the score function.

NOTE: Situation analysis is enabled by default in this currently, we might want it to be disabled by default for merging into staging/master

## Associated Issue
Issue #1342

## Design Documents
[Link](link-to-design-doc)

### Test Case 1
1. Run soccer with both adaptive formation and basic_122 enabled

Expected result: The plays switch dynamically according to the situation lists in each play

### DIFF
--- a/soccer/gameplay/plays/offense/adaptive_formation.py
+++ b/soccer/gameplay/plays/offense/adaptive_formation.py
@@ -155,7 +155,7 @@ class AdaptiveFormation(standard_play.StandardPlay):
     def score(cls):
         score = super().score()
 
-        if(score != None):
+        if (score != None):
             return score
         else:
             if (not main.game_state().is_playing()):
@@ -163,9 +163,9 @@ class AdaptiveFormation(standard_play.StandardPlay):
             if len(main.our_robots()) < 5:
                 return float("inf")
             return 8
-    
-    
-    
+
+
+
     def should_pass_from_dribble(self):
 
         # If pass is above cutoff and we dont have a good shot
--- a/soccer/gameplay/plays/offense/adaptive_formation.py
+++ b/soccer/gameplay/plays/offense/adaptive_formation.py
@@ -14,6 +14,7 @@ import tactics.advance_zone_midfielder
 import skills.move
 import skills.capture
 import situational_play_selection
+from situational_play_selection.SituationalPlaySelector import Situation
 
 
 class AdaptiveFormation(standard_play.StandardPlay):
--- a/soccer/gameplay/plays/offense/adaptive_formation.py
+++ b/soccer/gameplay/plays/offense/adaptive_formation.py
@@ -19,7 +19,7 @@ class AdaptiveFormation(standard_play.StandardPlay):
 
 
 
-    situationList = ['attack_goal', 'defend_clear', 'offensive_scramble']
+    _situationList = ['attack_goal', 'defend_clear', 'offensive_scramble']
 
 
     # Min score to pass
--- a/soccer/gameplay/plays/offense/adaptive_formation.py
+++ b/soccer/gameplay/plays/offense/adaptive_formation.py
@@ -14,7 +14,6 @@ import tactics.advance_zone_midfielder
 import skills.move
 import skills.capture
 import situational_play_selection
-from situational_play_selection.SituationalPlaySelector import Situation
 
 
 class AdaptiveFormation(standard_play.StandardPlay):
--- a/soccer/gameplay/plays/offense/adaptive_formation.py
+++ b/soccer/gameplay/plays/offense/adaptive_formation.py
@@ -19,9 +19,12 @@ import situational_play_selection
 class AdaptiveFormation(standard_play.StandardPlay):
 
 
-    _situationList = [situational_play_selection.SituationalPlaySelector.Situation.ATTACK_GOAL,
-            situational_play_selection.SituationalPlaySelector.Situation.DEFEND_CLEAR,
-            situational_play_selection.SituationalPlaySelector.Situation.OFFENSIVE_SCRAMBLE]
+    _situationList = [
+        situational_play_selection.SituationalPlaySelector.Situation.
+        ATTACK_GOAL, situational_play_selection.SituationalPlaySelector.
+        Situation.DEFEND_CLEAR, situational_play_selection.
+        SituationalPlaySelector.Situation.OFFENSIVE_SCRAMBLE
+    ]
 
     # Min score to pass
     DRIBBLE_TO_PASS_CUTOFF = 0.1
--- a/soccer/gameplay/plays/offense/adaptive_formation.py
+++ b/soccer/gameplay/plays/offense/adaptive_formation.py
@@ -17,6 +17,11 @@ import skills.capture
 
 class AdaptiveFormation(standard_play.StandardPlay):
 
+
+
+    situationList = ['attack_goal', 'defend_clear', 'offensive_scramble']
+
+
     # Min score to pass
     DRIBBLE_TO_PASS_CUTOFF = 0.1
     # Min score to shoot
@@ -148,12 +153,19 @@ class AdaptiveFormation(standard_play.StandardPlay):
 
     @classmethod
     def score(cls):
-        if (not main.game_state().is_playing()):
-            return float("inf")
-        if len(main.our_robots()) < 5:
-            return float("inf")
-        return 8
+        score = super().score()
 
+        if(score != None):
+            return score
+        else:
+            if (not main.game_state().is_playing()):
+                return float("inf")
+            if len(main.our_robots()) < 5:
+                return float("inf")
+            return 8
+    
+    
+    
     def should_pass_from_dribble(self):
 
         # If pass is above cutoff and we dont have a good shot
--- a/soccer/gameplay/plays/offense/adaptive_formation.py
+++ b/soccer/gameplay/plays/offense/adaptive_formation.py
@@ -155,7 +155,7 @@ class AdaptiveFormation(standard_play.StandardPlay):
     def score(cls):
         score = super().score()
 
-        if (score != None):
+        if (score != float("inf")):
             return score
         else:
             if (not main.game_state().is_playing()):
--- a/soccer/gameplay/plays/offense/adaptive_formation.py
+++ b/soccer/gameplay/plays/offense/adaptive_formation.py
@@ -19,12 +19,13 @@ import situational_play_selection
 class AdaptiveFormation(standard_play.StandardPlay):
 
 
+
     _situationList = [
-        situational_play_selection.SituationalPlaySelector.Situation.
-        ATTACK_GOAL, situational_play_selection.SituationalPlaySelector.
-        Situation.DEFEND_CLEAR, situational_play_selection.
-        SituationalPlaySelector.Situation.OFFENSIVE_SCRAMBLE
-    ]
+        situational_play_selection.SituationalPlaySelector.Situation.ATTACK_GOAL,
+        situational_play_selection.SituationalPlaySelector.Situation.DEFEND_CLEAR,
+        situational_play_selection.SituationalPlaySelector.Situation.OFFENSIVE_SCRAMBLE
+    ] # yapf: disable
+
 
     # Min score to pass
     DRIBBLE_TO_PASS_CUTOFF = 0.1
--- a/soccer/gameplay/plays/offense/adaptive_formation.py
+++ b/soccer/gameplay/plays/offense/adaptive_formation.py
@@ -159,8 +159,12 @@ class AdaptiveFormation(standard_play.StandardPlay):
     def score(cls):
         score = super().score()
 
+        #If we get a valid score from the super function, then we should calculate 
+        #an offset and sum that with score and return that
         if (score != float("inf")):
-            return score
+            #currently the offset is just 0 because we haven't made one
+            scoreOffset = 0
+            return score + scoreOffset
         else:
             if (not main.game_state().is_playing()):
                 return float("inf")
--- a/soccer/gameplay/plays/offense/adaptive_formation.py
+++ b/soccer/gameplay/plays/offense/adaptive_formation.py
@@ -13,14 +13,15 @@ import tactics.simple_zone_midfielder
 import tactics.advance_zone_midfielder
 import skills.move
 import skills.capture
+import situational_play_selection
 
 
 class AdaptiveFormation(standard_play.StandardPlay):
 
 
-
-    _situationList = ['attack_goal', 'defend_clear', 'offensive_scramble']
-
+    _situationList = [situational_play_selection.SituationalPlaySelector.Situation.ATTACK_GOAL,
+            situational_play_selection.SituationalPlaySelector.Situation.DEFEND_CLEAR,
+            situational_play_selection.SituationalPlaySelector.Situation.OFFENSIVE_SCRAMBLE]
 
     # Min score to pass
     DRIBBLE_TO_PASS_CUTOFF = 0.1
--- a/soccer/gameplay/plays/offense/basic_122.py
+++ b/soccer/gameplay/plays/offense/basic_122.py
@@ -64,8 +64,11 @@ class Basic122(standard_play.StandardPlay):
     @classmethod
     def score(cls):
         score = super().score()
+
+        #If the score from the super function is valid, use that with some offset
         if (score != float("inf")):
-            return score
+            scoreOffset = 0
+            return score + scoreOffset
         else:
             return 10 if main.game_state().is_playing() else float("inf")
 
--- a/soccer/gameplay/plays/offense/basic_122.py
+++ b/soccer/gameplay/plays/offense/basic_122.py
@@ -11,6 +11,8 @@ import math
 
 class Basic122(standard_play.StandardPlay):
 
+    situationList = ['clear', 'defend_goal', 'defensive_scramble', 'offensive_pileup', 'defensive_pileup']
+
     # how far the 2 support robots should stay away from the striker
     SupportAvoidTeammateRadius = 0.5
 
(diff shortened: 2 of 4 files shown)


liquidmetal9015:
Make sure to run this and check for the correct behavior.

soccer/gameplay/plays/offense/adaptive_formation.py
```diff
@@ -148,11 +153,18 @@ def __init__(self):
 
     @classmethod
     def score(cls):
-        if (not main.game_state().is_playing()):
-            return float("inf")
-        if len(main.our_robots()) < 5:
-            return float("inf")
-        return 8
+        score = super().score()
```
tjones320, line 161: What is this doing now???

soccer/gameplay/plays/offense/basic_122.py
```diff
@@ -11,6 +11,11 @@
 
 class Basic122(standard_play.StandardPlay):
 
+    situationList = [
```
tjones320: So can super score see this?

soccer/gameplay/situational_play_selection.py
```diff
@@ -300,47 +319,67 @@ def isPileup(self):
...
-            if not found:
-                raise Exception("Passed situation " + situation + " / " + up +
-                                " is not an existing situation")
-
-        if (self.currentSituaion.name == up):
-            return True
+    #exception is it does now
+    # @param situation a situation as a string or a enum 
+    # @param check Strings will be checked if they are real if True
+    def isSituation(self, situation, check=True):
```
tjones320: add more comments

soccer/gameplay/situational_play_selection.py
```diff
@@ -300,47 +319,67 @@ def isPileup(self):
...
+    # 
+    # Update determining if we want to preempt the current play or not
+    #   
     def updatePreempt(self):
-        if (self.lastSituation != self.currentSituaion and
-                self.situationChangeTime == None):
+
+        currentPlay = main._root_play.play
+
+        if (self.lastSituation != self.currentSituation and
```
tjones320, line 355: add more comments

soccer/gameplay/standard_play.py
```diff
@@ -34,9 +38,24 @@ def use_standard_defense(self):
...
+
         self.use_standard_defense()
 
     #Since the standard_play handles defense, it will always handle the goalie
     @classmethod
     def handles_goalie(cls):
         return True
+
+    @classmethod
+    def score(cls):
```
tjones320, line 80: I really like this.

tjones320 (approved):

soccer/gameplay/plays/offense/adaptive_formation.py
```diff
@@ -17,6 +17,11 @@
 
 class AdaptiveFormation(standard_play.StandardPlay):
 
+
+
+    situationList = ['attack_goal', 'defend_clear', 'offensive_scramble']
```
JNeiger: I personally would prefer these to be enums so I don't have to worry about misspelling them

tjones320: I think that's a fair point

tjones320: I agree

kylestach: Thirded.

soccer/gameplay/plays/offense/basic_122.py
kylestach: I would prefer an `is_valid(situation)` function instead of just a public list.

soccer/gameplay/situational_play_selection.py
```diff
@@ -300,47 +319,67 @@ def isPileup(self):
...
-                                " is not an existing situation")
-
-        if (self.currentSituaion.name == up):
-            return True
+    #exception is it does now
+    # @param situation a situation as a string or a enum 
+    # @param check Strings will be checked if they are real if True
+    def isSituation(self, situation, check=True):
+
+        if (isinstance(situation, str)):
```
kylestach: Is this the expected branch? Shouldn't the situation always be an enum value or am I missing something.

soccer/gameplay/standard_play.py
kylestach: What about it, exactly?

kylestach: Also, we need to have a lot more documentation about what the overridden class method needs to do.

soccer/gameplay/situational_play_selection.py
```diff
@@ -300,47 +319,67 @@ def isPileup(self):
...
+            if (self.isSituation(g, check=check)):
+                return True
+
+        return False
+
+    ##
+    # 
+    # Update determining if we want to preempt the current play or not
+    #   
     def updatePreempt(self):
```
kylestach, line 344: I would prefer this have some interaction with the play itself. i.e. `if currentPlay.tryPreempt(self.currentSituation): # Try preempt`. But I'm open to changing my mind on this.

kylestach (commented):
Mostly looks pretty good. Needs more documentation considering this is now an important part of the plays system's API.

soccer/gameplay/plays/offense/adaptive_formation.py
liquidmetal9015: So currently both

`situationList = [ 'clear', 'defend_goal', 'defensive_scramble', 'offensive_pileup','defensive_pileup']`

and
```
situationList = [situational_play_selection.SituationalPlaySelector.Situation.CLEAR,
            situational_play_selection.SituationalPlaySelector.Situation.DEFEND_GOAL,
            situational_play_selection.SituationalPlaySelector.Situation.DEFENSIVE_SCRAMBLE,
            situational_play_selection.SituationalPlaySelector.Situation.OFFENSIVE_PILEUP,
            situational_play_selection.SituationalPlaySelector.Situation.DEFENSIVE_PILEUP ]
```

are tested and working. I can remove string support if we want; I mostly added it because I thought the enums would be long although I suppose that's fine. It does have an option to spell check the strings as a default parameter. Having to write everything out brings two things to my mind though: one should I make the enum its own file?, and also I am open to renaming some things. situational_play_selection and SituationalPlaySelector might be both too wordy and no longer accurate to what the file and class do. Situation Analysis has been what we have been calling it but I'm open to suggestions.

soccer/gameplay/plays/offense/basic_122.py
liquidmetal9015: I can make an is_valid but would still prefer making an "private" _situationList to pass to situation analysis and to make it so that all you have to do is override that variable to make everything work at a basic level and allow you to use the super score function as is. is_valid would basically just be a contains call though if I'm understanding right.

soccer/gameplay/situational_play_selection.py
liquidmetal9015: Currently planning on adding
```
    def try_preempt(self):
        self.terminate()
        return True
```
to standard play that can be overridden if you want more complex functionality, which will also remove preempt logic from execute_running and change update_preempt to be actually calling for preempts rather than just updating a value.

soccer/gameplay/plays/offense/adaptive_formation.py
liquidmetal9015: This calls the score function in standard play, if it returns none then you are supposed to use your own score function, if not it gives you the starting point for your score based on the situation list you have made for the class. I'll add more documentation before merge.

soccer/gameplay/situational_play_selection.py
kylestach: That sounds good to me.

soccer/gameplay/plays/offense/basic_122.py
kylestach: That sounds like the correct fix.

soccer/gameplay/plays/offense/adaptive_formation.py
```diff
@@ -13,10 +13,19 @@
...
 
 class AdaptiveFormation(standard_play.StandardPlay):
 
+
+    _situationList = [
+        situational_play_selection.SituationalPlaySelector.Situation.
+        ATTACK_GOAL, situational_play_selection.SituationalPlaySelector.
+        Situation.DEFEND_CLEAR, situational_play_selection.
+        SituationalPlaySelector.Situation.OFFENSIVE_SCRAMBLE
+    ]
```
liquidmetal9015: So uhh, this is how make pretty wants to do it right now. See my comment on moving the enum out of situational_play_selection? Or import in a different way?

soccer/gameplay/situational_play_selection.py
```diff
@@ -298,60 +318,62 @@ def isMidfield(self):
...
-        if (self.currentSituaion.name == up):
-            return True
+    ##Returns the current situation
+    def getSituation(self):
+        return self.currentSituation
+
+    ##
+    # 
+    # @param situation a situation as an enum
+    def isSituation(self, situation):
```
liquidmetal9015, line 329: I deprecated string support as it was confusing and adding complexity.

liquidmetal9015:
Ok, @kylestach @JNeiger @tjones320. This should be ready for another pass of review.

soccer/gameplay/plays/offense/adaptive_formation.py
kylestach: Interesting formatting. Does it do this when you manually put each entry on a separate line?

soccer/gameplay/plays/offense/adaptive_formation.py
kylestach: Honestly, I would rather have this hook into the system in a manner other than the `score` function - if you do it this way you have to properly implement the super-call.

I would _prefer_ the play selection explicitly check against `is_applicable` for each play, and only check `score` if `is_applicable` returns true.

soccer/gameplay/situational_play_selection.py
kylestach: :+1:, agree with this change.

soccer/gameplay/situational_play_selection.py
```diff
@@ -298,60 +318,67 @@ def isMidfield(self):
...
 
-        if (self.currentPreempt):
-            self.currentPreempt = False
-        elif (self.situationChangeTime != None):
-            if (abs(time.time() - self.situationChangeTime) >
-                    self.playPreemptTime):
-                self.situationChangeTime = None
-                self.currentPreempt = True
-                self.LastSituation = self.currentSituaion
+        return False
```
kylestach: This could just be `return self.currentSituation in situations`, right?

soccer/gameplay/situational_play_selection.py
```diff
@@ -298,60 +318,67 @@ def isMidfield(self):
...
+
+        #If the situation has changed and its been longer than situationChange
+        #time without a play change, try to preempt the current play
+        if (self.situationChanged and
+                abs(time.time() - self.situationChangeTime) >
+                self.preemptTime):
+            #Check to see if the play is a standard play before trying to preempt
+            if (currentPlay != None and isinstance(
+                    currentPlay, standard_play.StandardPlay) and self.enabled):
+                currentPlay.try_preempt()
```
kylestach: Shouldn't this use the result from `try_preempt()`? Or does it just use the fact that the play has now been terminated if preemption succeeded?

soccer/gameplay/standard_play.py
```diff
@@ -30,13 +39,49 @@ def use_standard_defense(self):
             if self.has_subbehavior_with_name('defense'):
                 self.remove_subbehavior('defense')
 
-    #Handles activity while the play is active. A play wishing to utilize this
-    #method in additionto having an "execute_running" method of its own must call
-    #it via super
+    ## 
+    # Handles activity while the play is active. A play wishing to utilize this
+    # method in additionto having an "execute_running" method of its own must call
```
kylestach: Addition to

soccer/gameplay/situational_play_selection.py
```diff
@@ -298,60 +318,67 @@ def isMidfield(self):
...
-                                " is not an existing situation")
-
-        if (self.currentSituaion.name == up):
-            return True
+    ##Returns the current situation
+    def getSituation(self):
+        return self.currentSituation
+
+    ##
+    #
```
kylestach, line 327: Documentation

kylestach (changes requested):

soccer/gameplay/plays/offense/adaptive_formation.py
liquidmetal9015: Yes. They started on individual lines, I just tested it again.

soccer/gameplay/situational_play_selection.py
liquidmetal9015: Current behavior is just that it will continuously call try_preempt() until the play changes without regard to the return value of try_preempt(). Which I think is fine for the time being.

kylestach: Okay. So the play is responsible for deciding if it's terminated, and then situational play selection chooses a new one if it terminates itself.

Sounds good to me.

soccer/gameplay/plays/offense/adaptive_formation.py
kylestach:
```python
_situationList = [
    situational_play_selection.SituationalPlaySelector.Situation.ATTACK_GOAL,
    situational_play_selection.SituationalPlaySelector.Situation.DEFEND_CLEAR,
    situational_play_selection.SituationalPlaySelector.Situation.OFFENSIVE_SCRAMBLE
] # yapf: disable
```
Should do what you want.

kylestach (approved):
LGTM

liquidmetal9015:
Hypothetically someone else should review this since a lot has changed. To that end I will update the how to test methodology.

## How to test
1. Set the enabled Boolean to true in situational_play_selection.py 
2. Enable both basic_122 and adaptive_formation at the same time
3. Move the ball from from and too each side of the field and changing possession

Result: The running play should change based on the valid situations for that play, you should see the scores changing in real time and plays should properly be preempted a short time after the situation changes.

(Random thought: The delay after situation change that it takes before preempting should probably have some mechanism of increasing for repeated transitions. Like if we go from a->b then b->a it should increase the delay and a->b increase it again to help reduce any waffling, might need to use some history to prevent cycles of 3 or more. That would be a future feature)

@tjones320 @JNeiger should both probably take a look at this again.

There will still be future iterations where some things will change. Modifying the score functionality in root_play to categorically exclude plays is something that is on the table. For the moment it has been decided to leave it up to the score function. Sticking for basics right now until we have a decent little pool of plays to be swapping between

soccer/gameplay/plays/offense/adaptive_formation.py
JNeiger: Just in general something to keep in mind, the less boilerplate that needs to be added to new plays the better. It's much easier to mess with and see what a play is doing if you don't have to skip over tons of repeated stuff. If something is really simple, it should look really simple functionality wise.

soccer/gameplay/situational_play_selection.py
```diff
@@ -56,8 +69,7 @@ class Situation(Enum):
 
     ##Enum for representing where the ball is on the field
     #
-    # The regions are defined in the update function
-    #
+    # The regions are defined in the update fudoablse#
```
JNeiger: `fudoablse`? Typo

soccer/gameplay/situational_play_selection.py
```diff
@@ -298,60 +318,64 @@ def isMidfield(self):
...
-            if (abs(time.time() - self.situationChangeTime) >
-                    self.playPreemptTime):
-                self.situationChangeTime = None
-                self.currentPreempt = True
-                self.LastSituation = self.currentSituaion
+    ##
+    # 
+    # Will determine if the current play has lasted too long over a situation change
+    #   and needs to be preempted, it will call try_preempt
+    def updatePreempt(self):
```
JNeiger, line 344: Correct me if I'm wrong, but this will preempt on every situation change even if the current play accounts for this situation? Is that the behavior we want? I personally would let the current play continue to run so we don't loose state info from just restarting it.

soccer/gameplay/situational_play_selection.py
```diff
@@ -298,60 +318,64 @@ def isMidfield(self):
...
+        #If the situation has changed and its been longer than situationChange
+        #time without a play change, try to preempt the current play
+        if (self.situationChanged and
+                abs(time.time() - self.situationChangeTime) >
+                self.preemptTime):
+            #Check to see if the play is a standard play before trying to preempt
+            if (currentPlay != None and isinstance(
+                    currentPlay, standard_play.StandardPlay) and self.enabled):
+                if (currentPlay.try_preempt()):
+                    self.situationChanged = False
```
JNeiger, line 375: This only asks the play if it should try to preempt once on situation change. Should we continuously do it? I imagine we might want to put off a preempt if a pass is happening or something, but will it be up to the play to then quit after that happens or should we just continue polling until that pass is finished

soccer/gameplay/standard_play.py
```diff
@@ -30,13 +39,49 @@ def use_standard_defense(self):
...
+    #
+    # The expected interaction is to override this function can call super()
+    # to get a baseline score, then to modify that score based on situational
+    # factors. If you get a float("inf") back you know that situation analysis
+    # is not running and you will need to overwrite that with some other score
+    #
+    @classmethod
+    def score(cls):
+        if (not main.situationAnalysis.enabled):
+            return float('inf')
```
JNeiger, line 82: If you want to simplify the score for the time being, it might be easier to just return `0` here, and then in the plays, always return score + whatever was there before. This way you don't have that super ugly if statement in the score function of plays. If you want to keep it separated due to clarity about what's being executed when situationAnalysis is off or on thats very reasonable too

JNeiger (approved):
Functionality looks good, just some clarification on minut details

soccer/gameplay/situational_play_selection.py
liquidmetal9015: So I had it preempting continuously up until a few days ago and I changed it to only do it a single time to as a quick fix to concerns I had about not detecting when a play is preempted and restarts, and continuously preempting it due to that (because we only check root play equality, which will still be true after switching). I plan to switch back to continuously preempting in the next iteration alongside a more robust detection for changes. This is just a beta that is disabled by default.

soccer/gameplay/situational_play_selection.py
liquidmetal9015: So I think that on one hand if we are running in a gimped situation where we have one play covering many adjacent situations and are using our current plays, we will get a performance benefit from not preempting switches from and to the same play. But I think at this point it's more valuable to preempt them and see how everything handles it since that will be the expectation when we have good play coverage, our plays should be able to handle being started and stopped because that is how it's going to be with more plays in the playbook. For that reason I have kept it preempting switches between the same plays.

soccer/gameplay/standard_play.py
liquidmetal9015: I'm still thinking about how to make the score functions cleaner, probably by making some changes in how play selection works and changing the standard_play score function. Right now I think this is the safest way to go since if it is an no-run by default.

PR #1358, opened 2019-10-08
Replace custom clamp method with `std::clamp`

ruiyi-gao:
Replace the method with std::clamp

## Description
Describe your pull request.

## Associated Issue
Issue #???

## Design Documents
[Link](link-to-design-doc)

### Test Case 1
1. Step 1
2. Step 2
3. Step 3

Expected result: ???

- [ ] Step 1
- [ ] Step 2

### DIFF
--- a/common/Geometry2d/Util.hpp
+++ b/common/Geometry2d/Util.hpp
@@ -34,9 +34,4 @@ static bool nearlyEqual(float a, float b) {
     return std::fabs(a - b) < FLOAT_EPSILON;
 }
 
-/** Returns @value if it is in bounds, otherwise returns the bound it is closest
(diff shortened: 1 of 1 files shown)


common/Geometry2d/Util.hpp
```diff
@@ -34,9 +34,4 @@ static bool nearlyEqual(float a, float b) {
     return std::fabs(a - b) < FLOAT_EPSILON;
 }
 
-/** Returns @value if it is in bounds, otherwise returns the bound it is closest
```
kylestach, line 37: I would just explicitly specify `std::clamp` in the place that it's used.

Issue #1326, opened 2019-09-03
Clean up utilities


kylestach:
## Description
We currently have functions all over the (C++) codebase that are useful in many places. Some are already in a common area, some are reimplemented in several places, and some could be replaced with `std::` equivalents. Here is a running list:
 - [x] https://github.com/RoboJackets/robocup-software/blob/staging/soccer/optimization/GradientAscent1D.cpp#L6
 - [x] https://github.com/RoboJackets/robocup-software/blob/staging/common/Geometry2d/Util.hpp#L40
 - [x] https://github.com/RoboJackets/robocup-software/blob/staging/soccer/vision/ball/BallBounce.cpp#L25
 - [x] https://github.com/RoboJackets/robocup-software/blob/staging/soccer/WindowEvaluator.cpp#L225
    - Replace with https://en.cppreference.com/w/cpp/numeric/math/erfc
 - [ ] We do `lerp` in a bunch of different places...
    - [ ] https://github.com/RoboJackets/robocup-software/blob/staging/soccer/planning/CollectPathPlanner.cpp#L129
    - [ ] https://github.com/RoboJackets/robocup-software/blob/staging/soccer/planning/CollectPathPlanner.cpp#L443
    - [ ] https://github.com/RoboJackets/robocup-software/blob/staging/soccer/planning/SettlePathPlanner.cpp#L582
 - [x] If we have lerp, it would also be nice to have cubic interpolation considering we're starting to use it in a few different places (RRTPlanner uses it, InterpolatedPath will use it too with #1324).

We should move all of these functions into a new file in `common/util`, or similar, that is used everywhere.

This would be a great first issue for new members on the C++ side.

## Requirements

 - All functions listed above are moved to the relevant place in `common/`
 - All functions have associated tests

## Design
(N/A, small feature)

PR #1359, opened 2019-10-08
Switching sign to std::signbit

SimonAbrelat:
## Description
We switched an inline sign to std::signbit in Gradient Ascent 1D

## Associated Issue
Issue #1326


soccer/optimization/GradientAscent1D.cpp
```diff
@@ -33,8 +31,8 @@ bool GradientAscent1D::singleStep() {
     currentdx = std::get<1>(funcOutput);
     currentVal = std::get<0>(funcOutput);
 
-    // Decrease temperature when derivative flips sign
-    if (sign(previousdx) == -1 * sign(currentdx)) {
+    // Decrease temperature when derivative flips signbit
+    if (signbit(previousdx) == -1 * signbit(currentdx)) {
```
JNeiger: `signbit` returns a bool depending on whether the number is negative or not. I don't think this if statement has the same behavior that the original statement had. The original `sign` function returned `-1` when negative, `0` when 0, and `1` when positive. 

https://stackoverflow.com/a/4609795
https://en.cppreference.com/w/cpp/numeric/math/signbit
https://onlinegdb.com/SkYtzoqdr

SimonAbrelat: Yes you are totally right. I was just very lazy and hoped the bools would be auto-converted to ints and didnt test it. A find and replace wasn't quiet enough.

kylestach:
Yup, Joe's right. We actually want to move the sign function to `common/Util.h`. I think this was my bad on the instructions.

soccer/optimization/GradientAscent1D.cpp
JNeiger: FWIW, I probably would have done `signbit(previousdx) != signbit(currentdx)` to make it clear that we want to make sure the signs aren't equal.

This whole section of code optimizes a math function to find the local max. It's basically moves up the slope proportional to the steepness of the slope. The steeper it is, the further it moves. The extra piece here is that the derivative is not smooth so there are sharp corners. The sharp corners cause problems because the derivative is not guaranteed to go to zero. We deal with this by introducing the idea of temperature from simulated annealing. Every time the sign of the derivative changes (AKA we jump over one of these corners), we decrease the temperature. This temperature also controls how far it jumps. Eventually, after crossing a corner enough, the optimization "stops" at the top of the corner and returns the local max.

common/Utils.hpp
```diff
@@ -16,6 +16,9 @@
 
 const static bool THROW_DEBUG_EXCEPTIONS = true;
 
+template<typename T>
+inline int signum(T val) { return (0 < val) - (val <= 0); }
```
ghost: I feel like it would be more clear to say `return val > 0 ? 1 : -1`.

ghost (approved):
looks good to me

common/Utils.hpp
kylestach: I think it should actually be `(0 < val) - (val < 0)`, so that signum(0) = 0

JNeiger:
Fwiw, the ball bounce one was specifically chosen to be -1 or 1 with the "break" not at zero. I'll have to go through the math again to figure out why, but it might be significant. The other one is fine to change to this.

Edit: Can confirm, the sign function being used in ball bounce is specifically chosen to be -1 or 1. I'm doing a circle line intersection following [this](http://mathworld.wolfram.com/Circle-LineIntersection.html) and the `sgn*` is -1 for x < 0, 1 otherwise. The one inside `CalcBallBounce` doesn't matter too much as it'll just assume create no "damping" on the ball bounce

JNeiger (approved):

kylestach:
Fix `mypy` issues by pulling in latest changes from staging.

Fix `style` issues by running `make pretty-lines`.

PR #1361, opened 2019-10-09
Revert triangle pass code

kylestach:
## Description
#1354 and #1356 probably shouldn't have been merged. At any rate, they're now breaking CI. This reverts them (and should get us back to working builds).


JNeiger:
Is the code that's in plays/skel not what you want? Iirc anything that is in that folder without the `__init__.py` is not seen by soccer so you don't have to worry about it not compiling or anything

kylestach:
Yes, the plays/skel code is correct. This was added by mistake, so I'm reverting the PRs here.

liquidmetal9015:
So I believe that the new file was supposed to be harder than the one in skel, having less of the lines already filled in, It should probably just be in skel rather than in a directory with __init__.py in it. I also think that that is the next assignment for the people in the plays group.

kylestach:
@liquidmetal9015 Hmm, you might be correct. Regardless, this should not be in the main repository (I believe Tanner put it on a branch, which is what will be used for training).

JNeiger (approved):

Zolkin1 (approved):
lgtm

PR #1363, opened 2019-10-19
Celebrate Good Times, Come on

tjones320:
## Description
Created a play to put all our enemies to shame and to show off in front of our biggest fans, its no coach play but it is amazingly epic.

## Associated Issue
Not enough pride and joy

## Design Documents
See how to be fun

### Test Case 1
1. check test_celibrate

Expected result: run around and have fun and maybe spin


chachmu (approved):
LGTM

soccer/gameplay/tactics/positions/celebration.py
```diff
@@ -0,0 +1,77 @@
...
+
+	def execute_spin(self):
+		angle = self.robot.angle
+		facing_point = robocup.Point.direction(angle) + self.robot.pos
+		facing_point.rotate(self.robot.pos, Celebration.SpinPerTick)
+		self.spin_angle += Celebration.SpinPerTick
+		self.robot.face(facing_point)
+
+	def im_dizzy(self):
+		return time.time() - self.start_time > 8 and time.time() - self.start_time < 20
```
kylestach: If we've been running for 20+ seconds should we just spin forever, dizziness be damned?

kylestach (approved):
I am very glad we decided to add this important tactic to our arsenal. I think it'll really round out our lineup.

soccer/gameplay/tactics/positions/celebration.py
```diff
@@ -0,0 +1,77 @@
...
+import robocup
+import constants
+import single_robot_composite_behavior
+import main
+import enum
+import skills
+import random
+import time
+
+class Celebration(single_robot_composite_behavior.SingleRobotCompositeBehavior):
```
JNeiger: You don't actually use any subbehaviors so I would just make this a single robot behavior instead of composite

soccer/gameplay/tactics/positions/celebration.py
```diff
@@ -0,0 +1,77 @@
...
+						robocup.Point(-constants.Field.Width/2 + r, constants.Field.Length - r),
+						robocup.Point(0, constants.Field.Length/2)]
+
+		self.current_corner = 0
+		self.start_time = time.time()
+
+
+	def on_enter_run_around(self):
+		self.current_corner = random.randint(0,4)
+		self.robot.move_to_direct(self.corners[0])
```
JNeiger: This feels wrong to force to corner zero even though you have a variable called current corner. It doesn't make a difference since running is called the same frame, but you shouldn't rely on that

soccer/gameplay/tactics/positions/celebration.py
```diff
@@ -0,0 +1,77 @@
...
+
+	def execute_run_around(self):
+		if (self.robot.pos - self.corners[self.current_corner]).mag() <= .05 :
+			if (self.current_corner == 4) :
+				self.spin_time = True
+				self.current_corner = random.randint(0,3)
+			else :
+				self.current_corner = random.randint(0,4)
+
+		if (self.current_corner < 5):
```
JNeiger: It should always be less than 5?

soccer/gameplay/tactics/positions/celebration.py
```diff
@@ -0,0 +1,77 @@
...
+			self.robot.move_to_direct(self.corners[self.current_corner])
+
+
+	def on_enter_spint(self):
+		self.start_time = time.time()
+
+	def execute_spin(self):
+		angle = self.robot.angle
+		facing_point = robocup.Point.direction(angle) + self.robot.pos
+		facing_point.rotate(self.robot.pos, Celebration.SpinPerTick)
```
JNeiger: 1 radian per tick seems high unless it's actually in degrees which would surprise me

soccer/gameplay/tactics/positions/celebration.py
```diff
@@ -0,0 +1,77 @@
...
+		self.spin_angle += Celebration.SpinPerTick
```
JNeiger: You don't use spin angle

PR #1366, opened 2019-10-21
Update requirements3.txt

kylestach:
## Description
Don't specify versions in requirements3.txt.

We usually just want to be using the latest version. And I think old versions broke somewhere down the line.

## Associated Issue
(No issue, CI is failing for all builds)

## Design Documents
N/A

## Steps to test
CI Passes

### DIFF
--- a/util/requirements3.txt
+++ b/util/requirements3.txt
@@ -1,10 +1,10 @@
 # python requirements
 
-graphviz>=0.4.2 # make pretty graphs/diagrams
+graphviz # make pretty graphs/diagrams
 watchdog # file-system event notifications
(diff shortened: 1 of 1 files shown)


kylestach:
Open to discussion on whether or not we should actually pin specific versions of python packages - this is a test to make sure it actually resolves the issue. I honestly don't think we _need_ to pin specific versions (especially of tools like mypy, pylint, etc.) and it's more overhead to maintain, but I'm open to changing my mind if there's a convincing argument otherwise.

JNeiger (approved):

SimonAbrelat (approved):

PR #1367, opened 2019-10-21
Robot Intent

ghost:
## Description
Part of the SystemState/Context refactor. This implements RobotIntent, which holds data for an individual robot. this data is passed around rather than passing around the raw protobuf packets. the protobuf packets are now generated/updated in the radio code instead.

## Associated Issue
Issue #1301

### DIFF
--- a/soccer/Context.hpp
+++ b/soccer/Context.hpp
@@ -8,6 +8,7 @@
 #include "WorldState.hpp"
 #include "motion/MotionSetpoint.hpp"
 #include "vision/VisionPacket.hpp"
+#include <set>
 
 struct Context {
     Context() : state(this), debug_drawer(this) {}
@@ -19,8 +20,9 @@ struct Context {
     Context(Context&&) = delete;
     Context& operator=(Context&&) = delete;
 
-    std::array<RobotIntent, Num_Shells> robotIntents;
-    std::array<MotionSetpoint, Num_Shells> motionSetpoints;
+    std::array<RobotIntent, Num_Shells> robot_intents;
+    std::array<MotionSetpoint, Num_Shells> motion_setpoints;
+    std::set<int> active_robots;
 
     SystemState state;
     GameState game_state;
--- a/soccer/Context.hpp
+++ b/soccer/Context.hpp
@@ -19,15 +19,13 @@ struct Context {
     Context(Context&&) = delete;
     Context& operator=(Context&&) = delete;
 
+    std::array<RobotIntent, Num_Shells> robotIntents;
+    std::array<MotionSetpoint, Num_Shells> motionSetpoints;
+
     SystemState state;
     GameState game_state;
     DebugDrawer debug_drawer;
 
     std::vector<std::unique_ptr<VisionPacket>> vision_packets;
     WorldState world_state;
-
-    // construct these right away so they can be used from the SystemState
-    // constructor
-    std::array<RobotIntent, Num_Shells> robotIntents;
-    std::array<MotionSetpoint, Num_Shells> motionSetpoints;
 };
--- a/soccer/Context.hpp
+++ b/soccer/Context.hpp
@@ -6,9 +6,11 @@
 #include "WorldState.hpp"
 #include "vision/VisionPacket.hpp"
 #include "RobotIntent.hpp"
+#include <Constants.hpp>
 
 struct Context {
-    Context() : state(this), debug_drawer(this) {}
+    Context() : state(this), debug_drawer(this), robotIntents(Num_Shells) {
+    }
 
     // Delete copy, copy-assign, move, and move-assign because
     // many places are expected to hold Context* pointers.
--- a/soccer/Context.hpp
+++ b/soccer/Context.hpp
@@ -1,16 +1,16 @@
 #pragma once
 
 #include <Constants.hpp>
-#include "motion/MotionSetpoint.hpp"
 #include "DebugDrawer.hpp"
 #include "GameState.hpp"
 #include "RobotIntent.hpp"
 #include "SystemState.hpp"
 #include "WorldState.hpp"
+#include "motion/MotionSetpoint.hpp"
 #include "vision/VisionPacket.hpp"
 
 ////stores the outputs published by MotionControl
-//struct MotionSetpoint {
+// struct MotionSetpoint {
 //    float xvelocity;
 //    float yvelocity;
 //    float avelocity;
--- a/soccer/Context.hpp
+++ b/soccer/Context.hpp
@@ -22,7 +22,6 @@ struct Context {
 
     std::array<RobotIntent, Num_Shells> robot_intents;
     std::array<MotionSetpoint, Num_Shells> motion_setpoints;
-    std::set<int> active_robots;
 
     SystemState state;
     GameState game_state;
--- a/soccer/Context.hpp
+++ b/soccer/Context.hpp
@@ -5,6 +5,7 @@
 #include "SystemState.hpp"
 #include "WorldState.hpp"
 #include "vision/VisionPacket.hpp"
+#include "RobotIntent.hpp"
 
 struct Context {
     Context() : state(this), debug_drawer(this) {}
@@ -22,4 +23,6 @@ struct Context {
 
     std::vector<std::unique_ptr<VisionPacket>> vision_packets;
     WorldState world_state;
+
+    std::vector<RobotIntent> robotIntents;
 };
--- a/soccer/Context.hpp
+++ b/soccer/Context.hpp
@@ -1,6 +1,7 @@
 #pragma once
 
 #include <Constants.hpp>
+#include <set>
 #include "DebugDrawer.hpp"
 #include "GameState.hpp"
 #include "RobotIntent.hpp"
@@ -8,7 +9,6 @@
 #include "WorldState.hpp"
 #include "motion/MotionSetpoint.hpp"
 #include "vision/VisionPacket.hpp"
-#include <set>
 
 struct Context {
     Context() : state(this), debug_drawer(this) {}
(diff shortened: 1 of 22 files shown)


soccer/Robot.cpp
```diff
@@ -97,14 +94,14 @@ bool OurRobot::avoidOpponents() const {
     // checks for avoiding all opponents
     for (size_t i = 0; i < Num_Shells; ++i) {
         if (_context->state.opp[i] && _context->state.opp[i]->visible() &&
-            _opp_avoid_mask[i] < 0.1)
+            _context->robotIntents[shell()]._opp_avoid_mask[i] < 0.1)
```
kylestach: You repeat `_context->robotIntents[shell()]` a lot - maybe make a private helper function?

soccer/motion/MotionControl.cpp
```diff
@@ -261,8 +259,9 @@ void MotionControl::_targetBodyVel(Point targetVel) {
     }
 
     // set control values
-    _robot->control->set_xvelocity(targetVel.x() * _x_multiplier->value());
-    _robot->control->set_yvelocity(targetVel.y());
+    _context->robotIntents[_robot->shell()].setpoints.xvelocity =
+        targetVel.x() * _x_multiplier->value();
+    _context->robotIntents[_robot->shell()].setpoints.yvelocity = targetVel.y();
```
kylestach: This is my only concern with this PR. I don't want both gameplay and motion control to be publishing _partial_ structs, because a large part of the point of the refactor is allowing us to transition smoothly to a pub/sub where you're publishing an entire message at a time (each message needs to be generated _in whole_ before it is "published")

Maybe have a separate struct for `RobotMotionSetpoints` that's published by motion control, instead of writing back into the intent? I'm just spitballing, this is one of the annoying details of this change that I don't think I had thought through entirely, so I'd love to hear ideas.

soccer/radio/PacketConvert.cpp
```diff
@@ -72,6 +72,57 @@ Packet::RadioRx convert_rx_rtp_to_proto(const rtp::RobotStatusMessage& msg) {
     return packet;
 }
 
+void construct_tx_proto(Packet::RadioTx& radioTx,
+                        const std::array<RobotIntent, Num_Shells>& intents) {
+    // I'm assuming this is necessary to do logging. idk
```
kylestach: This constructs the `RadioTx` from the intents, right?

soccer/radio/Radio.hpp
```diff
@@ -17,7 +17,8 @@ class Radio {
     Radio() { _channel = 0; }
 
     virtual bool isOpen() const = 0;
-    virtual void send(Packet::RadioTx& packet) = 0;
+    virtual void send(Packet::RadioTx& radioTx,
+                      const std::array<RobotIntent, Num_Shells>& intents) = 0;
```
kylestach: I don't know if the cleanest way to do this logging is by passing it out like this...I need to think some more about how we want to do logging, I think, but this should be fine for this PR.

kylestach (changes requested):
Overall, this looks really good! I'm really happy with it, mostly.

My one concern is how we're handling setpoints from motion control. I'll repeat one of my comments here: 

> I don't want both gameplay and motion control to be publishing partial structs, because a large part of the point of the refactor is allowing us to transition smoothly to a pub/sub where you're publishing an entire message at a time (each message needs to be generated in whole before it is "published")

I have some more thoughts in that comment - any ideas on how to tie this up?

soccer/RobotIntent.hpp
```diff
@@ -0,0 +1,40 @@
...
+
+    std::unique_ptr<Planning::MotionCommand> _motionCommand;
+    std::unique_ptr<Planning::RotationCommand> _rotationCommand;
+
+    /// set of obstacles added by plays
+    Geometry2d::ShapeSet _local_obstacles;
+
+    /// masks for obstacle avoidance
+    RobotMask _opp_avoid_mask;
+    float _avoidBallRadius;  /// radius of ball obstacle
```
kylestach: This is a struct, so you don't need to start things with _. Also, all entries should be consistently snake_case.

soccer/Context.hpp
```diff
@@ -22,4 +25,9 @@ struct Context {
 
     std::vector<std::unique_ptr<VisionPacket>> vision_packets;
     WorldState world_state;
+
+    // construct these right away so they can be used from the SystemState
+    // constructor
```
kylestach: If you want them to be constructed before the systemstate constructor they should be placed before the systemstate in the class declaration (so move this line up)

kylestach (changes requested):
This is looking great! I'll give another final in depth review tomorrow, but I think it's pretty much there.

soccer/radio/NetworkRadio.cpp
```diff
@@ -24,11 +24,16 @@ void NetworkRadio::startReceive() {
...
+    const std::array<MotionSetpoint, Num_Shells>& setpoints) {
+    construct_tx_proto(radioTx, intents, setpoints);
+
     // Get a list of all the IP addresses this packet needs to be sent to
-    for (int robot_idx = 0; robot_idx < packet.robots_size(); robot_idx++) {
-        const Packet::Control& control = packet.robots(robot_idx).control();
-        uint32_t robot_id = packet.robots(robot_idx).uid();
+    for (int robot_idx = 0; robot_idx < radioTx.robots_size(); robot_idx++) {
+        const Packet::Control& control = radioTx.robots(robot_idx).control();
+        uint32_t robot_id = radioTx.robots(robot_idx).uid();
```
JNeiger, line 31: The whole robot ID being different from index in radioTx array is obe if we are just going to create a `robot` for every single ID. Probably not in scope for this, but it's something to be aware of for future refactoring efforts

soccer/radio/PacketConvert.hpp
```diff
@@ -3,6 +3,8 @@
 #include <protobuf/RadioRx.pb.h>
 #include <protobuf/RadioTx.pb.h>
 #include <protobuf/Robot.pb.h>
+#include <RobotIntent.hpp>
+#include <motion/MotionControl.hpp>
```
JNeiger: What is needed from the MotionControl header?

soccer/radio/PacketConvert.cpp
```diff
@@ -72,6 +73,59 @@ Packet::RadioRx convert_rx_rtp_to_proto(const rtp::RobotStatusMessage& msg) {
     return packet;
 }
 
+void construct_tx_proto(
+    Packet::RadioTx& radioTx,
```
kylestach, line 77: You know, on second thought I'd actually rather we take out all of the protobuf stuff from the radio interface and do the log struct filling in Processor.

soccer/radio/SimRadio.cpp
```diff
@@ -24,11 +25,14 @@ SimRadio::SimRadio(Context* const context, bool blueTeam)
...
+void SimRadio::send(Packet::RadioTx& radioTx,
+                    const std::array<RobotIntent, Num_Shells>& intents,
+                    const std::array<MotionSetpoint, Num_Shells>& setpoints) {
+    construct_tx_proto(radioTx, intents, setpoints);
     grSim_Packet simPacket;
     grSim_Commands* simRobotCommands = simPacket.mutable_commands();
-    for (int i = 0; i < packet.robots_size(); i++) {
-        const Packet::Robot& robot = packet.robots(i);
+    for (int i = 0; i < radioTx.robots_size(); i++) {
+        const Packet::Robot& robot = radioTx.robots(i);
```
JNeiger, line 32: When you test this, make sure that grSim handles the extra "empty" robots being sent nicely. Based on the [source](https://github.com/RoboCup-SSL/grSim/blob/master/src/sslworld.cpp#L549), it /should/ be fine, but its very easy to miss while testing if you don't sepcifcally try it. I think just settting a nonzero default command should make it obvious

JNeiger (approved):

soccer/radio/PacketConvert.hpp
JNeiger: It also exists in a few other network .hpp files

soccer/radio/NetworkRadio.cpp
kylestach: True. Especially since we're moving towards this type of interface (vector of objects indexed by robot id) I would prefer we fix that in RadioTx. Probably not in this PR, though.

soccer/radio/SimRadio.cpp
ghost: I may be misunderstanding, but I don't think it's sending empty robots because construct_tx_proto() should initialize all the robots before Radio::send().

soccer/radio/PacketConvert.cpp
ghost: Yeah that's better. I agree

soccer/radio/PacketConvert.cpp
ghost: yep

soccer/radio/SimRadio.cpp
JNeiger: Sorry, I wasn't super clear. Also, I'm not 100% sure on this info, based on the code I think it worked this way.

I think before we would only create the `Robot` class when we got radio communication or saw it on camera. In sim, this would mean 6 `Robot` instantiations, each with their own control packet. Only those 6 would be sent to grSim. Based on my reading of the new code, 16 control packets would be sent, not just the 6 seen/activily controlled. The other 10 will be filled with zeros so it shouldn't cause any crashing on our end, but it's very possible that on their end, the extra 10 may cause some weird effects, including crashing, or controlling the oponent robots on accident.

kylestach:
I've tested with grsim and changed robot counts, and this holds up fine - so it looks good to me. @egordon9dev please pull staging into this to resolve CI issues, and then it should be good to merge!

kylestach (approved):

JNeiger (approved):

Issue #1350, opened 2019-10-04
main.ball().pos.x/y is not readonly


JNeiger:
## Description
Consider the following code.

```
main.ball().pos.x = .1
main.ball().pos.y = .1
```

This actually causes the position of the ball to change.

## Steps to reproduce
I'll upload the latest commit this happens with a screenshot tonight.

1. Step 1
2. Step 2

Result: ???

- [ ] This bug has been reproduced by someone else

JNeiger:
Can confirm this happened with the old code as well. Just that a copy was made between the C++/python layer such that the ball position was only changed on the python side. Now, the change filters back to the C++ side and causes the drawing of the ball to change as well since it's a pointer/reference. I'm going to see about forcing deep copies or writing a custom shallow copy function so we dont have to worry about this

PR #1368, opened 2019-10-22
Copy ball pos/vel when reading from python instead of by ref

JNeiger:
## Description
It was returning a reference to the position object, which then allowed direct changing on the x and y member variables. Since it was read only, you could not change the location of where the position object point thought, so that worked as intended. There were a few different ways to do this, but all required a function wrapper. This just treats the pos as a getter only that passes a copy back to python. Conviently, while making sure the robot class doesn't have this problem as well, I came across this commit [b78cefa](https://github.com/RoboJackets/robocup-software/commit/b78cefa588b0aa888c0304ab53e9844fdb26a042)

## Associated Issue
Issue #1350

## Design Documents
[Link](link-to-design-doc)

### Test Case 1
1. Try to set .x and .y of ball pos object
2. See if ball moves
3. ???

Expected result: It doesn't

### DIFF
--- a/soccer/gameplay/robocup-py.cpp
+++ b/soccer/gameplay/robocup-py.cpp
@@ -90,8 +90,6 @@ std::string Point_repr(Geometry2d::Point* self) { return self->toString(); }
 
 std::string Robot_repr(Robot* self) { return self->toString(); }
 
-Geometry2d::Point Robot_pos(Robot* self) { return self->pos; }
-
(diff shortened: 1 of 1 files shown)


kylestach (approved):
LGTM

kylestach:
@JNeiger CI is pending merge conflicts, it looks like.

chachmu (approved):
LGTM

PR #1369, opened 2019-10-22
using std::clamp

ghost:
this uses std::clamp instead of our function which did the same thing

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,7 +6,7 @@ MAINTAINER Jay Kamat jaygkamat@gmail.com
 ENV DEBIAN_FRONTEND noninteractive
 
 # setup apt tools and other goodies we want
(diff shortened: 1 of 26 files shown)


kylestach (approved):
LGTM

JNeiger:
I think clamp is only used in one spot in `PacketConvert`, would it be better to just replace that instead of doing this using thing?

tcontis (approved):
LGTM

kylestach:
This is failing because of mypy issues fixed in #1366. Please pull in changes from `staging` to resolve CI issues.

kylestach:
@JNeiger I think the latest changes address your comments.

tcontis (approved):

kylestach:
Merging this. Technically @tcontis' review isn't "authorized" as an approval but this was a new member project in which part of the exercise was to do code review.

PR #1370, opened 2019-10-23
replaced phi

rashmiathavale:
## Description
replaced phi

## Associated Issue
Issue #1326


soccer/WindowEvaluator.cpp
```diff
@@ -224,25 +224,7 @@ WindowingResult WindowEvaluator::eval_pt_to_seg(Point origin, Segment target) {
...
-    if (x < 0)
-        sign = -1;
-    x = fabs(x)/sqrt(2.0);
-
-    // A&S formula 7.1.26
-    double t = 1.0/(1.0 + p*x);
-    double y = 1.0 - (((((a5*t + a4)*t) + a3)*t + a2)*t + a1)*t*exp(-x*x);
-
-    return 0.5*(1.0 + sign*y);
+    return std::erf(x);
```
JNeiger: `phi` in this was trying to estimate the [CDF of a normal distribution](https://en.wikipedia.org/wiki/Normal_distribution#Numerical_approximations_for_the_normal_CDF). The overall goal was to figure out how likely a shot on goal will actually score. If we aim at the center of the goal, various noise (both mechanical and electrical) cause the ball to not always hit that target consistently. If you try 100 times, we assume itll be a normal distribution around that target location. 

If you're near the corner of the field, the "width" of the goal that the robot can hit and score is very small. You can't just assume that the entire normal distribution is "within" the goal. Putting it into math terms you are trying to find `P(Z < X < Y)`, where Z and Y are the bounds of the goal. You can convert this to the CDF and end up with `F(Y) - F(Z)`. 

This can also apply to passing too with some small changes.

Anyways, to estimate a normal CDF, you can actually use the erf/erfc functions to estimate it. The previous code did that using a numerical version of the erf function estimate using a polynomial.

[Wikipedia ](https://en.wikipedia.org/wiki/Error_function#Cumulative_distribution_function)has the formula you should use.
image

tjones320: What's the progress on this?

kylestach:
I finished this up because I'd like to get it merged through to clear up the PRs section. @JNeiger @tjones320 could you re-review?

JNeiger (approved):

kylestach (approved):

PR #1373, opened 2019-10-28
Enable ball sense -- Use OurRobot.has_ball() over evaluation.ball.robot_has_ball(OurRobot) where possible

liquidmetal9015:
I've gone in and replaced everywhere that I could find that uses evaluation.ball.robot_has_ball on our robots with robot.has_ball, which will rely on ball sense. This should mostly be undoing what was done as a band-aid before comp since our ball sense was broken although its possible I might have changed it in a few more places and we should look to see if it makes sense in those 

I think that in the future we should possibly have a centralized way to change which is used. Possibly by using an evaluation function everywhere and checking to see if the robot is ours, combined with possibly a dict of which robots have functioning ball sense so that we can quickly disable one if we have a ball sense wire get ripped out at comp. Realistically it shouldn't be that hard to have an auto-disable on some basic logic. Although I think that is something to discuss later.

For right now this just re-enables ball sense, I can add that functionality or make an issue in a later pull request.

### DIFF
--- a/soccer/gameplay/plays/offense/adaptive_formation.py
+++ b/soccer/gameplay/plays/offense/adaptive_formation.py
@@ -210,8 +210,7 @@ class AdaptiveFormation(standard_play.StandardPlay):
         return True
 
     def dribbler_has_ball(self):
-        return any(
-            r.has_ball() for r in main.our_robots())
+        return any(r.has_ball() for r in main.our_robots())
 
     def on_enter_collecting(self):
         self.remove_all_subbehaviors()
--- a/soccer/gameplay/plays/offense/adaptive_formation.py
+++ b/soccer/gameplay/plays/offense/adaptive_formation.py
@@ -211,7 +211,7 @@ class AdaptiveFormation(standard_play.StandardPlay):
 
     def dribbler_has_ball(self):
         return any(
-            robot.has_ball() for r in main.our_robots())
+            r.has_ball() for r in main.our_robots())
 
     def on_enter_collecting(self):
         self.remove_all_subbehaviors()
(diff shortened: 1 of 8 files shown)


JNeiger:
@future thing. I've been of the opinion we have the has ball in the c++ robot class, and OurRobot overrides that with ball sense. Then if ball sense is bad, we just return the parent function instead of overriding it. It keeps everything the same in Python for both versions and the code is where it should be

JNeiger (approved):
Biggest thing is to make sure that we don't try to call this on any opponent robots otherwise we'll crash. It looks good to me

soccer/gameplay/plays/offense/adaptive_formation.py
```diff
@@ -210,8 +210,7 @@ def should_clear_from_dribble(self):
         return True
 
     def dribbler_has_ball(self):
-        return any(
-            evaluation.ball.robot_has_ball(r) for r in main.our_robots())
+        return any(r.has_ball() for r in main.our_robots())
```
kylestach, line 213: Probably this should check the robot to which the relevant tactic (?) is assigned, but this PR keeps the old behavior so I'm fine with that.

soccer/gameplay/skills/angle_receive.py
```diff
@@ -144,8 +144,7 @@ def execute_receiving(self):
 
         self.ball_kicked = True
         # Kick the ball!
-        if (self.robot is not None and evaluation.ball.robot_has_ball(
-                self.robot)):  #self.robot.has_ball()): #
+        if (self.robot is not None and self.robot.has_ball()):
```
kylestach, line 147: No parentheses needed.

kylestach (approved):
LGTM

soccer/gameplay/skills/angle_receive.py
oswinso: oof

Issue #1328, opened 2019-09-04
Add automated testing framework


kylestach:
## Description
A system to automatically run tests for plays. The GUI will allow users to select sets of tests to run. The system will then automatically place robots and the ball in the locations specified by the tests, run the tests, and report the results.

## Requirements

 - A user should be able to specify a list of robot positions and velocities, ball position and velocity, and game state. 
 - Soccer can be launched in “testing mode” via a command-line argument specifying a python script that specifies above data. 
 - Soccer will then proceed through all testing scripts, running each situation until time limit. 

## Design

[Link to design document](https://docs.google.com/document/d/1jem3jhoX-xXPAJe7aBImg3YXuQ82eKvUvw6GmLgNdgg/edit?usp=drivesdk)

kylestach:
@chachmu Is this done?

liquidmetal9015:
I think this just needs the time limit and the command line launching. Of course we don't really have the "report the results" part as a part of the current system. The current system just sets up the game state and I think there might be some small issues of things that need to be set. Running it in sim2play mode may also not work great. Other than that we need to actually create testing scenarios.

PR #1375, opened 2019-11-04
Add Automated Testing Framework

chachmu:
## Description
https://docs.google.com/document/d/1jem3jhoX-xXPAJe7aBImg3YXuQ82eKvUvw6GmLgNdgg/edit?usp=drive_web&ouid=107140160100665177019

## Associated Issue
Issue #1328 

## Design Documents
[Link](https://docs.google.com/document/d/1jem3jhoX-xXPAJe7aBImg3YXuQ82eKvUvw6GmLgNdgg/edit?usp=drive_web&ouid=107140160100665177019)

### Test Case 1
1. Select the Testing Tab
2. Click the checkmark next to the example test
3. Click add
4. Click Run or Next

Expected result: 2 blue robots and 2 yellow robots will appear on the opposite side of the field as the rest of the team and the ball will move to a new location with a significant velocity assigned. Additionally, Basic 122 will begin running and test pivot kick will also be selected.

### DIFF
--- a/soccer/CMakeLists.txt
+++ b/soccer/CMakeLists.txt
@@ -63,6 +63,7 @@ set(ROBOCUP_LIB_SRC
     "ui/SimFieldView.cpp"
     "ui/StripChart.cpp"
     "ui/StyleSheetManager.cpp"
+    "grSimCom.cpp"
     "VisionReceiver.cpp"
     "vision/ball/BallBounce.cpp"
     "vision/ball/CameraBall.cpp"
--- a/soccer/CMakeLists.txt
+++ b/soccer/CMakeLists.txt
@@ -64,7 +64,7 @@ set(ROBOCUP_LIB_SRC
     "ui/SimFieldView.cpp"
     "ui/StripChart.cpp"
     "ui/StyleSheetManager.cpp"
-    "grSimCommunicator.cpp"
+    "GrSimCommunicator.cpp"
     "VisionReceiver.cpp"
     "vision/ball/BallBounce.cpp"
     "vision/ball/CameraBall.cpp"
--- a/soccer/CMakeLists.txt
+++ b/soccer/CMakeLists.txt
@@ -64,7 +64,7 @@ set(ROBOCUP_LIB_SRC
     "ui/SimFieldView.cpp"
     "ui/StripChart.cpp"
     "ui/StyleSheetManager.cpp"
-    "grSimCom.cpp"
+    "grSimCommunicator.cpp"
     "VisionReceiver.cpp"
     "vision/ball/BallBounce.cpp"
     "vision/ball/CameraBall.cpp"
--- a/soccer/CMakeLists.txt
+++ b/soccer/CMakeLists.txt
@@ -64,6 +64,7 @@ set(ROBOCUP_LIB_SRC
     "ui/SimFieldView.cpp"
     "ui/StripChart.cpp"
     "ui/StyleSheetManager.cpp"
+    "grSimCom.cpp"
     "VisionReceiver.cpp"
     "vision/ball/BallBounce.cpp"
     "vision/ball/CameraBall.cpp"
--- a/soccer/Context.hpp
+++ b/soccer/Context.hpp
@@ -34,5 +34,5 @@ struct Context {
     std::optional<grSim_Packet> grsim_command;
 
     std::optional<QPointF> ball_command;
-    std::optional<Geometry2d::TransformMatrix> screen_to_world;
+    std::optional<Geometry2d::TransformMatrix> screen_to_world_command;
 };
--- a/soccer/Context.hpp
+++ b/soccer/Context.hpp
@@ -1,5 +1,6 @@
 #pragma once
 
+#include <protobuf/grSim_Packet.pb.h>
 #include <Constants.hpp>
 #include <set>
 #include "DebugDrawer.hpp"
@@ -29,4 +30,9 @@ struct Context {
 
     std::vector<std::unique_ptr<VisionPacket>> vision_packets;
     WorldState world_state;
+
+    std::optional<grSim_Packet> grsim_command;
+
+    std::optional<QPointF> ball_command;
+    std::optional<Geometry2d::TransformMatrix> screen_to_world;
 };
--- a/soccer/Processor.cpp
+++ b/soccer/Processor.cpp
@@ -84,7 +84,8 @@ Processor::Processor(bool sim, bool defendPlus, VisionChannel visionChannel,
     _vision = std::make_shared<VisionFilter>();
     _refereeModule = std::make_shared<NewRefereeModule>(&_context);
     _refereeModule->start();
-    _gameplayModule = std::make_shared<Gameplay::GameplayModule>(&_context, _refereeModule.get(), &_grCom);
+    _gameplayModule = std::make_shared<Gameplay::GameplayModule>(
+        &_context, _refereeModule.get(), &_grCom);
     _pathPlanner = std::unique_ptr<Planning::MultiRobotPathPlanner>(
         new Planning::IndependentMultiRobotPathPlanner());
 
--- a/soccer/Processor.cpp
+++ b/soccer/Processor.cpp
@@ -85,12 +85,13 @@ Processor::Processor(bool sim, bool defendPlus, VisionChannel visionChannel,
     _refereeModule = std::make_shared<NewRefereeModule>(&_context, _blueTeam);
     _refereeModule->start();
     _gameplayModule = std::make_shared<Gameplay::GameplayModule>(
-        &_context, _refereeModule.get(), &_grCom);
+        &_context, _refereeModule.get());
     _pathPlanner = std::unique_ptr<Planning::MultiRobotPathPlanner>(
         new Planning::IndependentMultiRobotPathPlanner());
     _motionControl = std::make_unique<MotionControlNode>(&_context);
     _visionReceiver = std::make_unique<VisionReceiver>(
         &_context, sim, sim ? SimVisionPort : SharedVisionPortSinglePrimary);
+    _grSimCom = std::make_unique<GrSimCommunicator>(&_context);
 
     _visionChannel = visionChannel;
 
@@ -107,6 +108,7 @@ Processor::Processor(bool sim, bool defendPlus, VisionChannel visionChannel,
 
     _nodes.push_back(_visionReceiver.get());
     _nodes.push_back(_motionControl.get());
+    _nodes.push_back(_grSimCom.get());
 }
 
 Processor::~Processor() {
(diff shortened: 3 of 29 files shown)


soccer/grSimCom.hpp
```diff
@@ -0,0 +1,23 @@
...
+#pragma once
+
+#include <protobuf/grSim_Commands.pb.h>
+#include <protobuf/grSim_Packet.pb.h>
+#include <protobuf/grSim_Replacement.pb.h>
+
+#include <QUdpSocket>
+#include <Geometry2d/TransformMatrix.hpp>
+
+class grSimCom {
```
kylestach: `GrSimCommunicator`

soccer/gameplay/GameplayModule.cpp
```diff
@@ -378,6 +385,19 @@ void Gameplay::GameplayModule::run() {
 
             getMainModule().attr("set_context")(&_context);
 
+            // Handle Tests
+            if (runningTests) {
+
+              // I could add a bool to check if this needs to run or not if this is too inefficient
+                object rtrn(handle<>( PyRun_String("ui.main._tests.getNextCommand()", Py_eval_input, _mainPyNamespace.ptr(), _mainPyNamespace.ptr()) ));
```
kylestach: :cry: 

This is super sketchy and I hate it...but if it's the only way I'm fine.

Correct me if I'm wrong - my understanding of the situation is that...
 - We can't call `_tests.getNextCommand()` from a function in main, because `_tests` also depends on main.
 - We can't bind a C++ function to a function in `_tests` without duplicating our setup for using `main`

soccer/gameplay/GameplayModule.cpp
```diff
@@ -455,3 +475,131 @@ void Gameplay::GameplayModule::updateFieldDimensions() {
...
     PyGILState_Release(state);
 }
+
+void Gameplay::GameplayModule::addTests() {
+    PyGILState_STATE state = PyGILState_Ensure();
+    {
+        try {
+          handle<> ignored3(
+                            (PyRun_String("import ui.main; ui.main._tests.addTests()", Py_file_input,
+                                          _mainPyNamespace.ptr(), _mainPyNamespace.ptr())));
```
kylestach: If at first you don't succeed: cry, cry again.

soccer/gameplay/gameplay_test.py
```diff
@@ -0,0 +1,62 @@
...
+        self.ballVelocity = robocup.Point(0, 0)
+
+        # List of plays to enable e.g. ["stopped", "offense/basic_122"]
+        self.play_list = []
+        # List of referee commands to run at the start (will run 1 each frame)
+        # Default value is seen here
+        self.start_commands = [
+            robocup.Command.halt,
+            robocup.Command.force_start
+        ]
```
chachmu: Make pretty-lines really wants to make this 
```
self.start_commands = [
       robocup.Command.halt, robocup.Command.force_start
]
```

But I feel pretty strongly against that

soccer/Processor.cpp
```diff
@@ -84,7 +84,8 @@ Processor::Processor(bool sim, bool defendPlus, VisionChannel visionChannel,
     _vision = std::make_shared<VisionFilter>();
     _refereeModule = std::make_shared<NewRefereeModule>(&_context);
     _refereeModule->start();
-    _gameplayModule = std::make_shared<Gameplay::GameplayModule>(&_context);
+    _gameplayModule = std::make_shared<Gameplay::GameplayModule>(
+        &_context, _refereeModule.get(), &_grCom);
```
kylestach: I still don't love this. IMO it would be significantly preferable to pass around in `Context` a `optional<BallMovementRequest>` that gets reset to `nullopt` by the GrSim communicator after it has been serviced.

Come to think of it, doing the same with a `GameStateOverrideRequest` might solve some of the python nastiness. I'll need to think about this more, because on one hand it's basically baking an RPC framework into our codebase, but to be fair once we're moving over to ROS that's what we'll be using anyway.

soccer/gameplay/gameplay_test.py
kylestach: https://github.com/google/yapf#id12

kylestach:
```python
self.start_commands = [
    robocup.Command.halt,
    robocup.Command.force_start
] # yapf: disable
```

soccer/gameplay/test_list.py
```diff
@@ -0,0 +1,91 @@
...
+        return len(self.tList)
+
+    def data(self, index, role=None):
+        if not index.isValid():
+            return None
+        node = index.internalPointer()
+
+        # This if is annoyingly necessary
+        # QT somehow calls this on garbage data
+        # during the first iteration after a remove
```
kylestach, line 59: ...excuse me what?

Okay sure seems reasonable, have a nice day.

soccer/ui/MainWindow.cpp
```diff
@@ -168,6 +169,9 @@ MainWindow::MainWindow(Processor* processor, QWidget* parent)
     if (!_processor->simulation()) {
         _ui.menu_Simulator->setEnabled(false);
     } else {
+        // Pass grCom into the simFieldView
+        _ui.fieldView->setGrCom(grCom);
```
kylestach: Yuck okay.

I really wish this whole file operated off of Context instead of directly calling out to things but oh well.

kylestach (changes requested):
I legitimately am not sure how you got through this PR without being driven mad by PyEval.

But looks mostly good, I haven't tested it though. The example file looks reasonable which is a good sign.

Probably would be good to have programmable termination conditions for tests but that can be a later PR.

soccer/gameplay/GameplayModule.cpp
chachmu: Pretty much. Referee commands can only be sent from C++ and this is really the only place to put it without passing the referee or gameplay modules through the robocup-py file which would also be pretty terrible

soccer/gameplay/test_list.py
chachmu: Something with the way QT handles the resize in the qAbstractList causes it to try to index role_requirements objects and a few other weird things through the data function. I did some testing but I was never able to figure out why :man_shrugging:

kylestach:
It looks like this doesn't properly reset the robots when you click "next" (only when a test is added and removed), but the ball is reset to the correct location. This is, as far as I know, incorrect behavior.

JNeiger:
What was the though process behind putting the GUI code in python?

chachmu:
@kylestach Are you just saying that the robots don't get reset to a default start state if you don't specify positions in the test? Otherwise I would need some specific steps to recreate the issue because I tried having 3 different tests with different robot positions and it seemed to work through all iterations of running and removing tests.

@JNeiger This is mostly because the tests are treated similarly to plays and are loaded as python scripts into the system so I chose to handle them from the python side the same way that the plays are. I guess I could theoretically try to load them through the boost python interface although that seems like it would actually greatly increase the amount of calls going through boost.

JNeiger:
I really like the feature and the way you have the tests set up in Python. I also really appreciate the effort due to the boost python funk. It's long overdue and been a want for a few years now. 

My one concern is that this kinda feels a step sideways or backwards for the other goal of improving data flow and the clean framework of soccer itself. Feel free to overrule me since I haven't fully been in the loop for this. It can be either in this pr or a subsequent refactor, but I feel this can be improved by becoming more consistent with our current implementations. I'd be more than happy to help with that since I brought it up if that's done we would like to pursue. 

I would need to look closer at everything, but I would like to create a stronge interface to soccer and our ref module, maybe move the GUI to c++ with the others with a clean command and control pipeline, and normalize the execution interface in our gameplay module. The amount of work that would take and the interface I'm not exactly sure yet, but I feel that would be a little more consistent with our current framework and our eventual framework goals. 

Again, feel free to override this

kylestach:
@chachmu my issue can be reproduced when you have only one test running. It should repeat it from the beginning but instead it only resets balls. 

@JNeiger I agree that there is a lot of nastiness here, and Woodward had to do a lot of BS to interface with our current system. My understanding is that it was a result of the way the gameplay and ui modules are set up but if you have ideas on how we can improve I am all ears. I'm not as knowledgeable as I should be on this part of the code base unfortunately.

JNeiger:
Give me a few days to think about a good solution and look at code some more. Im going out of town this weekend.

I'm very much of the opinion that guis should do pubs for options such that you could literally throw it away and do a full cli interface with no internal changes if you wanted. It should sub to any output data it wants to visualize.

I want to see if there's an easier way to keep the bare minimum number of calls from c++ to python code. It's very hard to follow across the barrier so keeping all the logic on one side or the other is very important in my mind. I would like to only communicate through single producer single consumer objects in system state if possible.

A lot of the functionality stay the same, the communication will just change to be much more structured and diliberate so data contracts, for lack of a better word, will be explicitly defined

chachmu:
@JNeiger I agree but I think that given the pending switch to ROS2 in the summer, spending time breaking only this part out into a message architecture would be unnecessary when we could instead convert all of the pyEval calls to a more ROS friendly approach when we can actually go all the way. I don't think making a half step here is worth the effort.

JNeiger:
I didn't realize the timeline was that short. It's probably fine then

soccer/CMakeLists.txt
```diff
@@ -64,6 +64,7 @@ set(ROBOCUP_LIB_SRC
     "ui/SimFieldView.cpp"
     "ui/StripChart.cpp"
     "ui/StyleSheetManager.cpp"
+    "grSimCommunicator.cpp"
```
kylestach: File name.

`GrSimCommunicator.cpp` <3

soccer/Processor.hpp
```diff
@@ -184,6 +185,7 @@ class Processor : public QThread {
     std::optional<RJ::Time> firstLogTime;
 
     Context* context() { return &_context; }
+    grSimCommunicator* grCom() { return &_grCom; }
```
kylestach: `GrSimCommunicator`

kylestach (changes requested):

soccer/Context.hpp
```diff
@@ -29,4 +30,9 @@ struct Context {
 
     std::vector<std::unique_ptr<VisionPacket>> vision_packets;
     WorldState world_state;
+
+    std::optional<grSim_Packet> grsim_command;
+
+    std::optional<QPointF> ball_command;
+    std::optional<Geometry2d::TransformMatrix> screen_to_world;
```
kylestach: This one should not be optional, right?

kylestach (approved):
LGTM :tada:

soccer/Context.hpp
chachmu: It gets passed in at the same time as the ball command and depends on the current orientation of the field view so I think it should be handled in the same way as the ball command

kylestach: Hmm, okay. I'm fine with this. Maybe call it `screen_to_world_command` or something to indicate that it shouldn't just be used as a global screen to world transform?

JNeiger (approved):

Issue #1086, opened 2017-12-20
Arch setup is missing flann package


JNeiger:

jgkamat:
Hmm, this seems to not be in the arch packages, so we might have to pull it from the AUR, does this package work for you: https://aur.archlinux.org/packages/flann/

JNeiger:
Yep, that one works!

PR #1385, opened 2019-11-23
Update package list for Arch/Manjaro

kylestach:
## Description
Update packages.txt for use with Arch Linux (and Manjaro).

BTW I use ~Arch~ Manjaro

## Associated Issue
Issue #1086, among others.

## Design Documents
N/A

### Test Case 1
1. Use a fresh copy of arch/Manjaro
2. Clone this repository
3. Run util/arch-setup
4. `make run-sim`

Expected result: Soccer runs out-of-the-box


liquidmetal9015:
Ok, so I've gotten from a clean Manjaro install to getting soccer running and here is what I've gathered needs to happen for Manjaro specifically. I'm not sure how this differs from arch or if we will need a separate file or to detect OS or whatnot.

Here are the things in order of my discovery:

1. If there is anything that pacman can not find in the arch-packages.txt, it won't install any of them. Currently on manjaro "qwtplot3d" and "flann" (which are both in arch-packages.txt on this branch) are not found and so nothing installs. I believe commenting them out makes the rest of the packages install.

2. The aur helper yay will need to be installed in order to install the aur packages that will be needed. 

3. A package called "base-devel" needs to be installed so that yay can emulate root privileges to actually install things.

4. flann and qwtplot3d need to be installed using yay (or just removed from the arch-packages.txt and added to the arch-packages-aur.txt), there are multiple candidates upon searching but it was relatively clear which one was the correct one. 

5. You will need a package called qtquickcontrols, which for me installed qt5-declarative-5.13.2-1   and qt5-quickcontrols-5.13.2-1.

With that and running git submodule update --init --recursive soccer is running for me. I am new to manjaro/arch so I kind of bumbled into it but I think it should be relatively simple to just put it into the setup file. We should possibly make yay the default aur helper?

liquidmetal9015:
I have also noticed that my make pretty does not work with make pretty-lines returning "clang-format-diff: command not found". I can use clang and clang-format in the command line but not clang-format-diff so that might be something else to add to the file.

liquidmetal9015:
Mine broke and I have had to downgrade boost and boost-libs to 1.71.0-2 from 1.71.0-4. I don't know exactly why. It also claimed that a couple python libraries that used to be installed weren't so I had to install them again using pip. (graphviz and munkres, no idea why, I thought I installed them before when everything worked)

liquidmetal9015:
Addition, yapf needs to be installed with pip in order for make pretty lines. (Along with the PR that Kyle just merged)

kylestach:
@liquidmetal9015 I added yapf, is it alright if I merge this?

JNeiger (approved):
Everything looks good, though I haven't tried it or used arch in almost 2 years so no telling if it works

liquidmetal9015 (approved):
That's fine

PR #1388, opened 2019-11-24
Situation Analysis actual enable

liquidmetal9015:
## Description
This pull request does a variety of things to get the minimum viable product of situation analysis.

1. Moves all plays into directories based on their primary situation
2. Modified situation analysis and a number of plays for scoring and generally smoothing out functionality
3. Created play variants to cover situations, including adaptive formation.

### Test Case 1
1. Load the Situation Analysis playbook
2. Observe game play

Note that playbook changes and situation changes will be needed before 

Expected result: We switch plays appropriately and it's relatively smooth and good

## General

This is a pretty big change and I would like a lot of people to look at it. I'm kind of just throwing this out there to get feedback. Please take a look at it and run it, its really exciting to watch the play switching.

I also can't make pretty on my new laptop, so that is a thing. I expect there to be changes.


soccer/gameplay/standard_play.py
```diff
@@ -82,6 +83,6 @@ def score(cls):
             return float('inf')
         else:
             if (cls.is_valid(main.situationAnalysis.getSituation())):
-                return main.situationAnalysis.inSituationScore
+                return main.situationAnalysis.inSituationScore + random.random()
```
liquidmetal9015: This is planned to be temporary as it WILL give you eye cancer if you look at the scores. A more formal way to inject randomness into play selection is pending but this won't affect anything when situational analysis is turned off.

soccer/gameplay/playbooks/SituationalTestingBook.pbk
```diff
@@ -0,0 +1,8 @@
+Attack_Goal/adaptive_attack
```
liquidmetal9015, line 1: This is just a playbook with all our "Situational Plays" added to it.

soccer/gameplay/plays/Attack_Goal/adaptive_attack.py
```diff
@@ -0,0 +1,292 @@
...
+import tactics.coordinated_pass
+import tactics.defensive_forward
+import tactics.simple_zone_midfielder
+import tactics.advance_zone_midfielder
+import skills.move
+import skills.capture
+import situational_play_selection
+
+
+class AdaptiveAttack(standard_play.StandardPlay):
```
liquidmetal9015, line 23: Adaptive attack is adaptive formation with all clearing functionality removed. I should probably add that as a comment. I've created new files for the variants to keep them separate from the known working and all situation covering adaptive formation, which has been moved to the Legacy Folder.

soccer/gameplay/plays/Attack_Goal/basic_122_attack.py
```diff
@@ -0,0 +1,198 @@
...
+import tactics
+import main
+import robocup
+import evaluation
+import constants
+import math
+import situational_play_selection
+
+
+class Basic122(standard_play.StandardPlay):
```
liquidmetal9015, line 17: I was thinking about changing the name of the ATTACK_GOAL basic_122 to be something like basic_122 attack, but didn't go through with it. I think it should be noted that this version has a narrower scope than the original but it seemed dumb to add "attack" to everything.

soccer/gameplay/plays/Attack_Goal/distraction_attack.py
```diff
@@ -257,3 +265,20 @@ def on_enter_shoot(self):
             priority=10)
         self.add_subbehavior(
             skills.pivot_kick.PivotKick(), 'shooting', required=True)
+
+
+
+    @classmethod
+    def score(cls):
```
liquidmetal9015, line 274: I've just added situational code to distraction, although it probably doesn't really need a score function for right now as the score function isn't really doing anything since there is no offset.

soccer/gameplay/plays/Attack_Goal/two_sided_attack.py
```diff
@@ -0,0 +1,161 @@
...
+import main
+import skills.move
+import skills.capture
+import enum
+import evaluation
+import tactics.coordinated_pass
+import tactics.defense
+import situational_play_selection
+
+class TwoSideAttack(standard_play.StandardPlay):
```
liquidmetal9015, line 21: Just a straight copy of two side attack

soccer/gameplay/plays/Clear/adaptive_clear.py
```diff
@@ -0,0 +1,280 @@
...
+import tactics.advance_zone_midfielder
+import skills.move
+import skills.capture
+import situational_play_selection
+
+
+
+#This is essentially a copy of adaptive formation, found in the legacy folder,
+#With the shooting option removed. It is intended to be used as a clearing play
+class AdaptiveClear(standard_play.StandardPlay):
```
liquidmetal9015, line 22: Adaptive clear is a copy of adaptive formation, with the shooting functionality and shot evaluation removed. We are also waiting on @HussainGynai clearing play which was accidentally removed.

soccer/gameplay/situational_play_selection.py
```diff
@@ -247,7 +247,7 @@ def closestReciever(self):
     def possesses_the_ball(self,
                            ballPos,
                            robot,
-                           distThresh=0.14,
+                           distThresh=0.2,
```
liquidmetal9015, line 250: There was some flip flopping in and out of possession due to our poor capture performance so I made the thresholds a bit more forgiving, changing the preempt delay is also an option.

soccer/gameplay/situational_play_selection.py
```diff
@@ -372,6 +372,7 @@ def updatePreempt(self):
             if (currentPlay != None and isinstance(
                     currentPlay, standard_play.StandardPlay) and self.enabled):
                 if (currentPlay.try_preempt()):
+                    print("Play has been preempted!")
```
liquidmetal9015, line 375: This should probably be a better message but I do think it should exist.

soccer/gameplay/tactics/advance_zone_midfielder.py
```diff
@@ -12,7 +12,7 @@
 import evaluation.passing
 import evaluation.shooting
 import functools
-import plays.offense.adaptive_formation
+import plays.Legacy.adaptive_formation
```
liquidmetal9015, line 15: Had to change an import due to the move of adaptive formation. This should probably be looked at to make sure it doesn't break anything. It feels a bit circular/bad that this requires the play that uses it?

liquidmetal9015:
It's been brought up that the directories should possibly be lower case because our other directories are lower case, although I do like something to differentiate them and it makes them look more similar to the all caps enums. So idk.

JNeiger:
I like lowercase just because it fits the rest of the code.

Where would a play go that is applied to multiple situations?

liquidmetal9015:
Most of the plays are covering multiple situations at this point and I have just put them in the situation directory "that they aspire to be", or essentially their primary situation. We do essentially aim to have each play cover only a single situation even if they cover multiple currently.

soccer/gameplay/tactics/advance_zone_midfielder.py
tjones320: The reason why we did this is that it needs to pull some data from it. I can modify it, but I may need to re-think how I want to do it to make it general.

soccer/gameplay/standard_play.py
tjones320: What is the proportion of inSituationScore to randomness factor? Are we only using it for tie-breaker or to create a kind of discovery factor for plays?

soccer/gameplay/situational_play_selection.py
```diff
@@ -372,6 +372,7 @@ def updatePreempt(self):
             if (currentPlay != None and isinstance(
                     currentPlay, standard_play.StandardPlay) and self.enabled):
                 if (currentPlay.try_preempt()):
```
tjones320, line 374: How good results have you gotten from the current preemption software?

soccer/gameplay/situational_play_selection.py
```diff
@@ -247,7 +247,7 @@ def closestReciever(self):
     def possesses_the_ball(self,
```
tjones320, line 247: Is this simple formula what we want to calculate possession? Just curious if there is a better way to do this?

soccer/gameplay/plays/Defend_Goal/basic_defend_goal.py
```diff
@@ -45,6 +51,9 @@ def __init__(self, num_defenders=3, num_wingers=2):
             self.add_subbehavior(wing_defender.WingDefender(),
                                  'winger' + str(i))
 
+    def use_standard_defense(self):
```
tjones320, line 55: Is this supposed to return false?

soccer/gameplay/plays/Attack_Goal/basic_122_attack.py
tjones320: I think if it's in the folder it should be sufficient to tell what its purpose is.

tjones320 (changes requested):
I think this is generally good! I like the new organization structure, and I think this will be really helpful going forward. I don't know how I feel about the randomness factor quite yet, but I think this is the right direction. Also, good job getting your line count up!

tjones320:
Also, make it so you pass style in CircleCI

soccer/gameplay/plays/Attack_Goal/adaptive_attack.py
Jason27chan: Bump for adding the comment to the code

soccer/gameplay/plays/Attack_Goal/basic_122_attack.py
```diff
@@ -0,0 +1,198 @@
...
+import skills
+import tactics
+import main
+import robocup
+import evaluation
+import constants
+import math
+import situational_play_selection
+
+
```
Jason27chan, line 12: Add a comment at the top of the class telling people what it does

Jason27chan (approved):
I think it looks good, but I'd like to take another look at it when I'm not dead tired

soccer/gameplay/standard_play.py
liquidmetal9015: The random factor is -1 to 1, although right now all the scores are the same the expected result of this is to only use it as a tie breaker. I think in the future a higher degree of randomness may be required depending on how good or at least chaotic our sub-scoring is. May want to sample a normal distribution or something like that with a bigger factor like -10 to 10. That will be a part of play balancing which will have to be done after we devise a playbook. (Thinking about this makes me start to think about if we should try to get balancing factors baked into the playbooks themselves, so we could quickly switch play balance by that mechanism or by some other mechanism. Something to think about.)

soccer/gameplay/situational_play_selection.py
liquidmetal9015: It's been pretty solid in that it preempts when it should and I haven't really noticed any major quirks aside from our normal behavior switching issues.

soccer/gameplay/plays/Defend_Goal/basic_defend_goal.py
liquidmetal9015: Looking at standard_play.py, no, this function never returns a boolean although you'd think it would based on the title. It possibly should work that way but it does not.

liquidmetal9015:
> Also, make it so you pass style in CircleCI

So I can't currently run style on my system, I will work on that. Someone else can run style for me if it comes to that.

soccer/gameplay/plays/Attack_Goal/adaptive_attack.py
```diff
@@ -0,0 +1,292 @@
...
+                main.our_robots(), AdaptiveAttack.MIN_PASS_DIST,
+                AdaptiveAttack.FIELD_POS_WEIGHTS,
+                AdaptiveAttack.NELDER_MEAD_ARGS,
+                AdaptiveAttack.DRIBBLING_WEIGHTS)
+
+        # TODO: Get list of top X pass positions and have robots in good positions to reach them
+        # Good positions can be definied by offensive / defensive costs
+        # Offensive positions move onto the ball in the direction of the goal
+        # Defensive cover the center of the field
+
```
HussainGynai, line 256: There are a few TODOs in these files are these yours, or are they just there from before? And if they're there from before do we still need them for these narrower plays?

HussainGynai (commented):
I think it looks pretty good, just take a look at the TODOs, since Im not sure if they're yours or there from before.

soccer/gameplay/plays/Attack_Goal/adaptive_attack.py
JNeiger: These are from when I initially created this. They're ideas to improve adaptive in the future. I think that only the last line 257 is changed with these narrower plays. Adaptive could still move all the offense robots to be even more agressive based on the pass position cost functions instead of just the basic positions

soccer/gameplay/plays/Clear/adaptive_clear.py
JNeiger, line 22: FWIW, this has both passing and "clearing" from adaptive. It doesn't make much of a difference, but yeah.

soccer/gameplay/plays/Defend_Goal/basic_defend_goal.py
JNeiger, line 55: In the future we should rework this to make it a variable each play initializes instead of doing this weird override. Not applicable to this PR though.

soccer/gameplay/situational_play_selection.py
JNeiger: Depending on what data you have, you should specify from what situation to what situation transition happened if it's not already there. It will be useful for debugging in the future if someone makes a play that triggers these transitions consistently

soccer/gameplay/standard_play.py
JNeiger: We can shrink the randomness and force the display in the table to have less sigfigs than the randomness to stop the cancer

soccer/gameplay/tactics/advance_zone_midfielder.py
JNeiger: We should definitely try to transition it to only a 1 way data flow, from parent to child play. Otherwise you can get some real funk down the rode when we change stuff

JNeiger (approved):

JNeiger:
Mostly file movement, but everything looks pretty good

kylestach:
It's currently trying to stylize clang-format-diff so that gets an override from me. @liquidmetal9015 this is good to go.

liquidmetal9015:
@tjones320 Do you know what your requested changes are? Because I can't seem to find them atm. If you can take another look that would be great.

tjones320 (approved):

PR #1393, opened 2019-12-24
Change play selection for gameplay tests

liquidmetal9015:
## Description
I want to be able to run tests with more flexibility as to which plays are activated without having to go through and change the play list in each of the tests I want to run. So I've done two things here. One, made it so that running the test only changes which plays are enabled if there is at least one play in the play_list. Two, made it so that you can add a .pbk file to the list of plays and they will all be added. This will help with making sure the tests are running with the correct plays and allow a little bit more flexibility if you just want to note out the play_list line to run with a custom combination of plays.

Eventually we want to do this with a UI element checkbox that will toggle using the current plays or the plays listed in the test file.

## Steps to test
1. Set different play and playbook combinations in example.py
2. See if the expected plays are enabled correctly when the test is ran

Expected result: When the list is empty, the play list remains as it was before the test was run. When a playbook file is added to list, all the plays from it are activated in addition to any other plays in the list. Multiple overlapping playbooks should work as well.

### DIFF
--- a/soccer/gameplay/gameplay_test.py
+++ b/soccer/gameplay/gameplay_test.py
@@ -53,7 +53,12 @@ class GameplayTest():
         self.ballVelocity = robocup.Point(0, 0)
 
         # List of plays to enable e.g. ["stopped", "offense/basic_122"]
+        # You can also add the name of a playbook file i.e. 
+        # ["comp2019.pbk", "testing/test_pivot_kick"]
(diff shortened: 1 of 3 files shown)


liquidmetal9015:
Oh yeah, I still can't run style. So someone else will have to run style for this to pass it unless I can fix it on my computer.

liquidmetal9015:
I'm just going to close these and make one big one when its ready.

PR #1394, opened 2019-12-26
Change test robots to use points

liquidmetal9015:
## Description
This should be considered after #1393, as it is branched off of that branch and is a minor change after that. I'm not exactly sure how to handle dependent/sequential pull requests as it seems there is really no good solution for it aside from just adding changes to the previous pull request. I guess I'm just doing this for now.

Currently the testRobot object the gameplay tests system uses x and y as values, where it is seeming like using robocup.Point to be more ergonomic for quickly defining points and offsets

## Steps to test
Basically just make sure that the gameplay test system works as intended.

### DIFF
--- a/soccer/gameplay/gameplay_test.py
+++ b/soccer/gameplay/gameplay_test.py
@@ -5,9 +5,8 @@ import robocup
 
 # Class to define robot locations/velocities
 class testRobot():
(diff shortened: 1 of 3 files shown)


liquidmetal9015:
I'm going to just close these and make one big one when it's ready.

PR #1395, opened 2020-01-08
Make changes to the testing framework (Draft)

liquidmetal9015:
## Description
When making tests I realized that there are some more features that I would like for the testing system to have, and some nitpicks I would like to fix. This PR should fix some of those things. I'm going to 
 
(this is a draft, I'm making a pr before I forget and to get feedback from people)

## Things that this does
1. Changes test robots to use points instead of x and y
2. Changes test coordinate system to be consistent with python coordinate system in a debatable jank way.
3. Adds support for using playbooks in the play list
4. Changes empty playbook behavior to not change active plays

### Test Case 1
1. Step 1
2. Step 2
3. Step 3

Expected result:  it worko

### DIFF
--- a/soccer/gameplay/GameplayModule.cpp
+++ b/soccer/gameplay/GameplayModule.cpp
@@ -576,7 +576,11 @@ void Gameplay::GameplayModule::loadTest() {
                         float x = extract<float>(robot[0]);
                         float y = extract<float>(robot[1]);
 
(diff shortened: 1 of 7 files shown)


JNeiger (approved):
FWIW I think there's a Geometry2d transform matrix that you can build that does the team->world transform with just a multiplication, but you would have to build the C++->Python bridge for it.

Other than that, it's straight forward and should work fine

chachmu (approved):
LGTM

Issue #1347, opened 2019-10-02
There are currently no plays for the "Offensive Scramble" or "Defensive Scramble" situations


liquidmetal9015:
## Description

Normally you would design a play of whatever situation in a design document, then create an issue for you or someone else to make it. But since we are currently lacking any plays at all for "Offensive and Defensive Scramble" situations, we are creating an issue to create a play or plays for those situations at all. In the longer run we want to have at least a couple of plays for each situation, but for this issue just create a play or plays for "Offensive and Defensive Scramble" using whatever methodology just so we have a baseline for those situations.

Eventually this play will need to interface with the situation analysis module for score determination, but that functionality is still in development so for right now the play will be developed as a general play and that integration will be added in later on. The integration should be pretty seamless with only the score function being affected, and even then a regular score function will still be the default if we decide to disable situation analysis for any reason.

In the new situational plays paradigm, plays will only be active during a specific subset of game play. So you will only need to handle what happens inside that situation and should assume that your play will be artificially terminated when game play leaves that situation. (You may still want to have your play terminate itself on certain triggers however). Below I will describe the situation that this play needs to cover. 

## Scramble Situations:

Noting our lack of responsiveness to scramble situations was one of the driving factors behind creating the situation analysis system and they are a very important case to have covered. 

An scramble situation is simply when neither team has the ball, it is free floating on the field and is up for grabs to anyone who can get to it. An "Offensive Scramble" is a scramble when, if we get the ball, we will be within shooting range (i.e. we are on our opponents side of the field). A "Defensive Scramble" is the same situation but when our opponents will be in shooting range if they get the ball. This situation happens A LOT, especially in division B, and making sure we are responsive during this situation is critical.

## Existing code and ideas:

We don't really have all that much for this situation so far. Ideas based around body blocking other robots and pre-positioning robots for defensive or offensive purposes are key ideas. We want to make sure that we have the best possible change to get the ball and to make the following play go as good for us as possible.

The reason this is one issue is that naturally the two situations are fairly related. We aren't sure how we want to tackle related plays yet but using an inheritance structure is one possibility. Its also possible that for right now we just develop a single scramble play and use that play in all situations. In fact, that should almost certainly be the starting point for development on this.

## Notes about design:

[...]

PR #1396, opened 2020-01-12
Basic defensive scramble

Jason27chan:
Dependent on [situational analysis actual enable](https://github.com/RoboJackets/robocup-software/pull/1388)

## Description
use standard defense with two wing defenders with one robot capturing to transition to clear situation

## Associated Issue
Issue #1347 

## Design Documents
N/A

### Test Case 1
1. Load situational analysis playbook
2. Play a game and get to the situation where the ball is on our side of the field and no team has possession

Expected result: See in the behavior tree that our team runs BasicDefensiveScramble


soccer/gameplay/plays/Defensive_Scramble/basic_defensive_scramble.py
```diff
@@ -0,0 +1,77 @@
...
+        self.add_subbehavior(capture.Capture(), 'ball capturer')
+
+        # Keep track of which robots are currently being defended
+        self.defended = {}
+        for i in range(len(main.their_robots())):
+            self.defended[i] = False
+
+        for i in range(self.num_wingers):
+            self.add_subbehavior(wing_defender.WingDefender(),
+                                 'winger' + str(i))
```
HussainGynai, line 49: Shouldn't wing defender do all this stuff on its own? I feel ideally you would just call WingDefender and it automatically checks defended bots.

HussainGynai (changes requested):
Looks good, but I think WingDefender needs to be changed so its not such a hassle to use.

soccer/gameplay/plays/Defensive_Scramble/basic_defensive_scramble.py
```diff
@@ -0,0 +1,77 @@
...
+        super().__init__(continuous=True)
+
+        self.num_defenders = num_defenders
+        self.num_wingers = num_wingers
+
+        self.add_transition(behavior.Behavior.State.start,
+                            behavior.Behavior.State.running, lambda: True,
+                            'Immediately')
+
+        self.remove_all_subbehaviors()
```
JNeiger: This is due to the standard play defense stuff right?

soccer/gameplay/plays/Defensive_Scramble/basic_defensive_scramble.py
```diff
@@ -0,0 +1,77 @@
...
+import tactics.defense
+import situational_play_selection
+
+## Play that uses submissive defenders and wingers to defend
+#  an attack close to our goal and cause one robot to capture.
+#  
+#  By default, we will use standard defense (two submissive 
+#  defenders, one goalie) and two wing defense robots. The
+#  remaining robot will mark the highest threat robot. 
+class BasicDefensiveScramble(standard_play.StandardPlay):
```
JNeiger, line 24: For what it's worth, I think the whole standard play thing has kinda blown up. It was just used cause we didn't want to add a defense to every play we make manually. If you want to control the defense, I think it would be better to pull from `play.Play` and do the correct overrides. @liquidmetal9015 will have to make some changes, but I think `play.Play` was supposed to be the base play class that everything inherits from and `standardPlay` was moreso for the lazy people who didn't want to add a defense and goalie every single time for things like adaptive, or basic122

soccer/gameplay/plays/Defensive_Scramble/basic_defensive_scramble.py
JNeiger: Wing defender was built to be basically mark, but marking in a specific location that kinda blocks the shot and pass at the same time. It relies on something higher up to decide which robots to mark so you can abstract things away to build up the wall and other things like that

soccer/gameplay/plays/Defensive_Scramble/basic_defensive_scramble.py
Jason27chan: Thanks for letting me know. I think I got a bit weird in this situation, but I do want to use standard defense in this situation, so standard play is the way to go.

soccer/gameplay/plays/Defensive_Scramble/basic_defensive_scramble.py
Jason27chan: This is probably extra code. I'll make the changes

HussainGynai (approved):

tjones320 (approved):
Very simple but this should work

PR #1398, opened 2020-01-14
tcontis/radio module refactor

tcontis:
## Description
Radio module refactor

My first real PR, probably has some errors in it that I can't spot.

## Associated Issue
#1329

## Design Documents
[Link](link-to-design-doc)


soccer/Context.hpp
```diff
@@ -35,4 +35,14 @@ struct Context {
 
     std::optional<QPointF> ball_command;
     std::optional<Geometry2d::TransformMatrix> screen_to_world_command;
+
+    //"Processor Status"
+    RJ::Time lastLoopTime;
+    RJ::Time lastVisionTime;
+    RJ::Time lastRefereeTime;
+    RJ::Time lastRadioRxTime;
```
JNeiger, line 43: What's the reasoning for shifting these outside Processor?

soccer/Processor.hpp
```diff
@@ -261,7 +245,6 @@ class Processor : public QThread {
     // This is used by the GUI to indicate status of the processing loop and
     // network
     QMutex _statusMutex;
```
JNeiger, line 247: This mutex still exists. I haven't looked too far into the code, but I imagine that the gui is in a seperate thread than this code, so if both are sharing the fps and networking status, then there can be a conflict

soccer/Processor.cpp
```diff
@@ -92,14 +91,10 @@ Processor::Processor(bool sim, bool defendPlus, VisionChannel visionChannel,
     _visionReceiver = std::make_unique<VisionReceiver>(
         &_context, sim, sim ? SimVisionPort : SharedVisionPortSinglePrimary);
     _grSimCom = std::make_unique<GrSimCommunicator>(&_context);
+    _radio = std::make_unique<Radio>(&_context, _simulation, _blueTeam, _multipleManual, _manualID);
```
JNeiger: In the hpp you should change the _radio object to be a unique pointer instead of a raw pointer. There might end up being some ownership mishaps when the smart pointer gets thrown away

soccer/radio/Radio.hpp
```diff
@@ -12,9 +14,9 @@
  * @details This is the abstract superclass for USBRadio and SimRadio, which do
  * the actual work - this just declares the interface.
  */
-class Radio {
+class Radio : public Node {
```
JNeiger, line 17: I think the best way to handle this is to leave the original `Radio` class as the interface for the two radios we have and then have a second class named RadioNode or something that wraps the `_radio` and deals with all the state info. It'll clean up the inheritance and split up any driver specific code away from the node itself.

soccer/Processor.cpp
```diff
@@ -191,7 +186,7 @@ void Processor::blueTeam(bool value) {
 
     if (_blueTeam != value) {
         _blueTeam = value;
-        if (_radio) _radio->switchTeam(_blueTeam);
+        if (_radio->_radio) _radio->_radio->switchTeam(_blueTeam);
```
JNeiger, line 191: It might be a good idea to have the RadioNode type wrapper have a `switchTeam` call so that there will be no implementation specific code elsewhere outside the RadioNode class. It makes changing code later on super duper easy since you don't have to just hope nothing is messed with outside the normal control flow

JNeiger (commented):
Looking good so far! Just a could comments on things to watch out for and some suggestions on how to split up the code even more into nodes

tcontis:
@JNeiger Thank you for the advice! I was originally thinking of making two separate classes, but wasn't sure. The commits are all a mess, so I will reset the branch and work from there.

soccer/Context.hpp
tcontis: To update lastRadioRxTime on radio's side of things

PR #1403, opened 2020-01-22
Fix segfault when clicking on field when running real robots

chachmu:
## Description
Fixes the segfault when clicking on the field while running for real robots

### Test Case 1
1. `make run`
2. Click the field
3. ???
4. Profit

Expected result: Soccer should not segfault

### DIFF
--- a/soccer/Context.hpp
+++ b/soccer/Context.hpp
@@ -35,4 +35,6 @@ struct Context {
 
     std::optional<QPointF> ball_command;
     std::optional<Geometry2d::TransformMatrix> screen_to_world_command;
+
+    bool is_simulation;
 };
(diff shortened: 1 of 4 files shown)


soccer/Context.hpp
```diff
@@ -35,4 +35,6 @@ struct Context {
 
     std::optional<QPointF> ball_command;
     std::optional<Geometry2d::TransformMatrix> screen_to_world_command;
+
+    bool is_simulation;
```
chachmu, line 39: @kylestach should this be an optional or is a bool fine

kylestach: Boolean is fine. We are always either in sim or not in sim.

kylestach (approved):
LGTM

Jason27chan (approved):

PR #1415, opened 2020-02-19
Consolidate CI Jobs and Disable Caching

kylestach:
## Description
Caching wasn't working properly, so the builds were taking 10m each (so >20m total). I just added the `test-python` and `test-soccer` targets in the regular `compile` job. I also made style not depend on the rest of the targets.

If anyone knows what was wrong with caching, I'd love to hear thoughts. The ccache was being generated properly, the compiler just wasn't getting a speedup from it (so I assume it was being ignored). I also tried setting `CMAKE_CXX_COMPILER=/usr/lib/ccache/g++` to no avail.

Regardless, as a temporary solution I'm fine with just disabling it.

## Associated Issue
N/A

## Design Documents
N/A

### CI
1. Run CI

Expected result: CI doesn't take 20+ minutes to run.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -16,27 +16,9 @@ defaults_image: &image
 defaults_deps: &install_deps
   run: ./util/ubuntu-setup --yes && sudo ccache -M 100M
 
-defaults_ssc: &save_src_cache
-  save_cache:
-    key: source-v1-{{ .Branch }}-{{ .Revision }}
-    paths:
-      - ".git"
-
-defaults_lsc: &load_src_cache
-  restore_cache:
-    keys:
-      - source-v1-{{ .Branch }}-{{ .Revision }}
-      - source-v1-{{ .Branch }}-
-      - source-v1-
 defaults_scc: &save_compile_cache
   save_cache:
     key: ccache-{{ arch }}-{{ .Branch }}-{{ .Revision }}
-    paths:
-      - ~/.ccache
-
-defaults_sccc: &save_compile_coverage_cache
-  save_cache:
-    key: ccache-cov-{{ arch }}-{{ .Branch }}-{{ .Revision }}
     paths:
       - ~/.ccache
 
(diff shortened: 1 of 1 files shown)


kylestach:
Ignore the expected CI checks, those just need to be updated.

FWIW this speeds up CI by roughly 2x, from about 20m to 10m.

jgkamat:
Kyle Stachowicz writes:

> If anyone knows what was wrong with caching, I&#39;d love to hear thoughts. 
> The ccache was being generated properly, the compiler just wasn&#39;t 
> getting a speedup from it (so I assume it was being ignored). I also tried 
> setting `CMAKE_CXX_COMPILER=/usr/lib/ccache/g++` to no avail.

It's possible that circleci wasn't preserving some attribute about the cached 
files (eg: the timestamp or extended attributes), which caused the cache to 
get invalidated. If you don't mind a 10min cycle it should be fine to remove 
this all though. IIRC the build times were ~2/3 min before: 
https://circleci.com/gh/RoboJackets/robocup-software/637

If you cared about getting it down to that time, you could try to find the 
build that started the slowdown and see what changed (but it might be 
something in ubuntu or similar, so I would be a bit careful). Unfortuantely 
circleci's build time graphs don't go far back enough to help you at all in 
this process :(

jgkamat (approved):
Under the assumption that 10min is good enough, this seems fine to me. It'll be easier/less confusing to maintain in the future as well :)

jgkamat:
Also, if you are planning on removing the coverage task (not sure if it was working before or not), you should remove it from the README as well

kylestach:
It's not a CircleCI cache issue - I ssh'd into the build server and set up ccache (with a fairly large cache) and clean-built, then rm'd the build dir and built again, and got no speed-up. I'm not sure what is up exactly.

jgkamat:
Kyle Stachowicz writes:

> It's not a CircleCI cache issue - I ssh'd into the build server and set up 
> ccache (with a fairly large cache) and clean-built, then rm'd the build dir 
> and built again, and got no speed-up. I'm not sure what is up exactly.

Shot in the dark, but maybe the max cache size isn't large enough anymore (I 
set it to 100M). I didn't want the cache to keep growing forever (as the 
larger the size is, the longer it takes to restore). But if the cache is too 
small, it might never hit (?). That's (afaik) the only difference between 
ccache on CI and normal machines

kylestach:
Yeah, it was hitting it so I had also tried bumping up cache size to 500M, but it was still compiling everything from scratch.

chachmu (approved):
LGTM :+1:

Issue #1411, opened 2020-02-16
Remove Gap in Wall Tactic


Jason27chan:
## Description
When we use the wall tactic, the wall may prioritize the sides of the walls instead of the middle, which leaves an open shot even if the wall tactic is active.
Also it crashes sometimes

-  There is a crash in the wall tactic #1407 [ ] 
-  Prioritize movement so no gaps in the wall [ ]

## Requirements
- When you run testWall, it doesnt error.
- When wall is formed, wall always forms with the center robot first, no situation with gap in wall

## Design Document
[link](https://docs.google.com/document/d/1Q15LMDUtM8QfPHnwtErewLPfjTT3usocLNChK5rKQLc/edit?usp=drive_web&ouid=107561101778230627083)

PR #1417, opened 2020-02-23
Remove gap in wall tactics

Jason27chan:
## Description
See Title

## Associated Issue
Issue #1411 

## Design Documents
See issue

### Test Case 1
1. Run TestWall
2. Move Ball to various spots

Expected result: Wall does not error at any point

### Test Case 2
1. Put robots and wall in the following configuration
Start config
2. Run TestWall

Expected Result: No gap forms. The wall should block the direct shot

### DIFF
--- a/soccer/gameplay/tactics/wall.py
+++ b/soccer/gameplay/tactics/wall.py
@@ -16,23 +16,25 @@ class Wall(composite_behavior.CompositeBehavior):
         shot = 2
         scramble = 3
 
+    ## 
+    # @param num_defenders: number of robots in the wall (default 3)
+    # @param curvature: 'curvature' (in radians) of the wall
+    # @param mark_point: what point are we defending against (default: ball) 
+    # @param defender_point: what point are we defending (default: goal)
+    # @param defender_spacing: number of robot radii between the centers of the
+    #   defenders in the wall
+    # @param dist_from_mark: distance from the mark point we want to build the wall
     def __init__(
             self,
-            num_defenders=3,  # number of defenders we're making the wall with (default 3)
-            curvature=.3,  # 'curvature' (in radians) of the wall 
-            mark_point=None,  # what point we are defending against (default is ball)
+            num_defenders=3,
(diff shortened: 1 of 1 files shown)


HussainGynai (approved):

liquidmetal9015:
Does this fix the crash? Because the issue that it links links the issue that is the wall crashing.

soccer/gameplay/tactics/wall.py
```diff
@@ -16,24 +16,25 @@ class State(enum.Enum):
         shot = 2
         scramble = 3
 
+    ##
+    # @param num_defenders: number of robots in the wall (default 3)
+    # @param curvature: 'curvature' (in radians) of the wall
```
kylestach, line 21: Could you explain what this actually means? Curvature should be in 1 / m

soccer/gameplay/tactics/wall.py
```diff
@@ -196,3 +146,15 @@ def _update_wall(self):
             if self.has_subbehavior_with_name("robot" + str(i)):
                 behavior = self.subbehavior_with_name("robot" + str(i))
                 behavior.pos = self.calculate_destination(i)
+
+
+    ## Move behavior that prioritizes the center point of the wall
```
kylestach, line 151: What does "prioritizes the center point of the wall" mean?

kylestach (approved):
Tested, seems to fix gaps. It seems a little bit aggressive to cluster all of our robots in one spot, but it does what it says on the tin I guess.

LGTM

soccer/gameplay/tactics/wall.py
Jason27chan: role requirements is determined by a robot's distance from the center part of the wall. The closest robot should always go to the center point of the wall

soccer/gameplay/tactics/wall.py
Jason27chan: *shrug*

PR #1420, opened 2020-03-04
Distraction fix

HussainGynai:
## Description
A fix for distraction

### Test Case 1
1. Set up game to be in attack goal situation
2. Run game with distraction
Expected result:  Passes shouldn't be too close and the pass locations shouldn't be the same every time.


chachmu (approved):
:+1: LGTM, not sure if this is the place to stylize clear.py but I don't think it really matters

soccer/gameplay/plays/Attack_Goal/distraction_attack.py
```diff
@@ -230,11 +230,12 @@ def on_enter_passing(self):
     def on_enter_cross(self):
         self.remove_all_subbehaviors()
         #if the ball is passed to the distractor the ball is passed to the striker, as the third robot moves to the right to distract more
-        #add chip here if pass chance is low
-        self.add_subbehavior(
-            tactics.coordinated_pass.CoordinatedPass(self.striker_point),
-            'pass to striker',
-            required=True)
+        ball = main.ball().pos
+        self.striker_point.x = -(ball.x / 2 + constants.Field.Width / 4)
```
tjones320, line 234: Do you always want this to happen on the right side of the field? You may want to reflect the play depending on robot positions.

tjones320 (changes requested):

soccer/gameplay/plays/Attack_Goal/distraction_attack.py
```diff
@@ -230,11 +230,12 @@ def on_enter_passing(self):
...
+        self.add_subbehavior(tactics.coordinated_pass.CoordinatedPass(
+            self.striker_point, use_chipper=True),
```
liquidmetal9015, line 236: Are we confident that using the chipper for passes is going to work here? I guess it should if they are always far enough apart I just notice that the old pass was a grounded pass.

soccer/gameplay/plays/Attack_Goal/distraction_attack.py
kylestach: I would suggest some randomness here, but at the very least pick which side to go on in an intelligent fashion. Although, we should really integrate pass optimization here in the long run.

kylestach (approved):

liquidmetal9015 (approved):

Issue #1348, opened 2019-10-02
There are currently no plays for the "Offensive Pileup" or "Defensive Pileup" situations


liquidmetal9015:
## Description

Normally you would design a play of whatever situation in a design document, then create an issue for you or someone else to make it. But since we are currently lacking any plays at all for "Offensive and Defensive Pileup" situations, we are creating an issue to create a play or plays for those situations at all. In the longer run we want to have at least a couple of plays for each situation, but for this issue just create a play or plays for "Offensive and Defensive Pileup" using whatever methodology just so we have a baseline for those situations.

Eventually this play will need to interface with the situation analysis module for score determination, but that functionality is still in development so for right now the play will be developed as a general play and that integration will be added in later on. The integration should be pretty seamless with only the score function being affected, and even then a regular score function will still be the default if we decide to disable situation analysis for any reason.

In the new situational plays paradigm, plays will only be active during a specific subset of game play. So you will only need to handle what happens inside that situation and should assume that your play will be artificially terminated when game play leaves that situation. (You may still want to have your play terminate itself on certain triggers however). Below I will describe the situation that this play needs to cover. 

## Pileup Situations:

The pileup situation is fairly unique and it tends to cover the most chaotic segments of game play where the result is unpredictable. Goals are very often scored in the moments following a pile up, especially in division B.

A pile up situation occurs when robots from two separate teams both have the ball in their mouths at the same time, or there are 3 robots from a mixture of teams in very close proximity to the ball at the same time. Basically there has been a "pile up" of robots around the ball. An "Offensive Pileup" is just a pileup where the ball is on our opponents side of the field and a "Defensive Pileup" is a pileup that occurs on our side of the field.

## Existing code and ideas:

We don't really have all that much for this situation so far. In fact this play will be particularly challenging while #1341 remains unsolved, and that problem is a challenging one. In the longer term we will probably want to have custom low level maneuvering for these situations that bypass the path planner to attempt to wrestle the ball out of the pile up. For now we are mostly want to robustly position robots around the pileup to address what comes next. Those positions could be based on some new evaluation as to what the result of the pileup is going to be.

We have defiantly considered creating behaviors that force a pileup on purpose if we think we can win, or want to slow down game play. That is if these types of plays end up working well. 

[...]

PR #1422, opened 2020-03-08
Basic offensive pileup

HussainGynai:
## Description
Basic Offensive Pileup play. Has one robot capture the ball. One robot goes to a dropback point behind the pileup. One robot goes to the side of the pileup

## Associated Issue
Issue #1348 

### Test Case 1
1. Run sim
2. Select basic_offensive_pileup
3. run to see if it works as intended.


liquidmetal9015 (commented):
The new pileup directories need __init__.py files in order to be seen by soccer.

liquidmetal9015 (commented):
Why is a distraction offensive play deleted by this PR? Isn't this being worked on currently in another PR?

liquidmetal9015:
Other than the lack of an __init__.py file, the distraction play being deleted, and style not passing, the play seems to function pretty decently for me.

HussainGynai:
> Other than the lack of an **init**.py file, the distraction play being deleted, and style not passing, the play seems to function pretty decently for me.

Yeah I'm not sure why distraction was added to the pull request the same thing keeps happening with clear, where I only git add offensive pileup but clear and distraction are in the pull request. Then I tried to delete distraction from the pull request, but it appears that this means that it will distraction from the repository, instead of just the pull request.

liquidmetal9015 (approved):
Looks good to me.

soccer/gameplay/plays/Offensive_Pileup/basic_offensive_pileup.py
```diff
@@ -0,0 +1,50 @@
...
+    def on_enter_running(self):
+        ball = main.ball().pos
+        dropback_point = robocup.Point(
+            ball.x, 2 * ball.y / 3
+        )  #Since we have to divide by ball.x, if its zero, make it slightly positive
+        if (ball.x == 0):
+            ball.x += 0.01
+        side_point = robocup.Point(
+            (-(abs(ball.x) / ball.x) * 0.4 * constants.Field.Width) + ball.x,
+            ball.y)
```
kylestach, line 47: A better way to pick this would be copysign: https://docs.python.org/3/library/math.html#math.copysign

```python
math.copysign(0.4 * constants.Field.Width, ball.x)
```

kylestach (approved):

soccer/gameplay/plays/Offensive_Pileup/basic_offensive_pileup.py
oswinso: whats the magic number 0.4 for?

HussainGynai: Its 0.4 of the field Width, since we are meant to have all points be relative to the field size since it isn't constant between our field and sim and comp fields. You'll probably see this quite a bit in the plays side, where its (some_constant * constants.Field.Width/Length)

PR #1423, opened 2020-03-21
Added init files to the two directories that don't have them

liquidmetal9015:
## Description
Two play directories on staging have plays in them but do not have the __init__.py file in them resulting in soccer not identifying them and the plays in them.

This should fix that.


HussainGynai (approved):

kylestach (approved):
LGTM

Issue #1424, opened 2020-03-27
Clean up all nodes in prep for pub-sub


kylestach:
## Description
As preparation for the ROS port, we want to clean up all of our code to make it more conducive to a pub-sub framework. This means making all of our nodes do something like the following:
```cpp
class SomeNode : public Node {
public:
  void run() override {
    runNode(_context->world_state, &_context->robot_intents);
  }

private:
  Context* _context;
  void runNode(const WorldState& state, RobotIntent* intent);
};
```
Basically, `runNode` (or whatever it is called) should only take in the parts of Context that it actually reads/writes. Nothing else from Context should be directly accessed with the exception of debug-drawing facilities. This enforces that the node's application logic only uses parts of Context that it has declared as dependencies.

## Requirements

The following nodes should not use Context other than in their top-level interface:
 - [x] Motion control (@kylestach)
 - [x] GUI (@kylestach @chachmu)
 - [x] Radio (@HussainGynai)
 - [x] Vision receiver
 - [x] Vision filter (pretty much already done)
 - [x] Manual control (@CollinAvidano)
 - [x] Referee (@oswinso)
 - [x] GrSim (this is already fine)
 - [x] Planning (@egordon9dev, waiting on refactor to be complete)
For now, gameplay will not be changed (the interface to context will need to be rewritten anyway).

Deadline targets are listed in the below document.

## Design

[Roadmap meeting](https://docs.google.com/document/d/1wix-ZK9NPBY0zC8T9LyUihT1ONubXMKxHvb8KYmJkNQ/edit#)

PR #1425, opened 2020-03-27
Motion control node

kylestach:
## Description
Clean up the motion control node in preparation for moving to a pub-sub.

## Associated Issue
Issue #1424 

## Design Documents
[ROS Roadmap](https://docs.google.com/document/d/1wix-ZK9NPBY0zC8T9LyUihT1ONubXMKxHvb8KYmJkNQ/edit#)

## Steps to test
Motion control should experience no regressions. Run adaptive formation to test.

### DIFF
--- a/logging/CMakeLists.txt
+++ b/logging/CMakeLists.txt
@@ -1,4 +1,3 @@
-
 include_directories("${PROJECT_SOURCE_DIR}/common")
 include_directories("${PROJECT_SOURCE_DIR}/soccer")
 
@@ -8,3 +7,7 @@ target_link_libraries(simple_logger common rc-fshare)
 
 add_executable(log_converter log_converter.cpp)
 target_link_libraries(log_converter common rc-fshare)
+
+# We need XML because Context includes configurables,
+# which uses XML to represent the tree.
+qt5_use_modules(log_converter Xml Core)
--- a/soccer/Processor.cpp
+++ b/soccer/Processor.cpp
@@ -36,24 +36,21 @@ using namespace google::protobuf;
 
 static const auto Command_Latency = 0ms;
 
-RobotConfig* Processor::robotConfig2008;
-RobotConfig* Processor::robotConfig2011;
-RobotConfig* Processor::robotConfig2015;
-
-// TODO: verify that this is correct
-std::vector<RobotStatus*> Processor::robotStatuses;
-
 // TODO: Remove this and just use the one in Context.
 Field_Dimensions* currentDimensions = &Field_Dimensions::Current_Dimensions;
 
+// A temporary place to store RobotStatus/RobotConfig variables as we create them.
+// They are initialized in createConfiguration, before the Processor class is
+// initialized, so we need to temporarily store them somewhere.
+std::vector<RobotStatus> robot_status_init;
+std::unique_ptr<RobotConfig> Processor::robot_config_init;
+
 void Processor::createConfiguration(Configuration* cfg) {
-    robotConfig2008 = new RobotConfig(cfg, "Rev2008");
-    robotConfig2011 = new RobotConfig(cfg, "Rev2011");
-    robotConfig2015 = new RobotConfig(cfg, "Rev2015");
+    robot_config_init = std::make_unique<RobotConfig>(cfg, "Rev2015");
 
     for (size_t s = 0; s < Num_Shells; ++s) {
-        robotStatuses.push_back(
-            new RobotStatus(cfg, QString("Robot Statuses/Robot %1").arg(s)));
+        robot_status_init.emplace_back(
+                cfg, QString("Robot Statuses/Robot %1").arg(s));
     }
 }
 
@@ -76,6 +73,14 @@ Processor::Processor(bool sim, bool defendPlus, VisionChannel visionChannel,
     _dampedRotation = true;
 
     _kickOnBreakBeam = false;
+
+    // Configuration-time variables.
+    _context.robot_config = std::move(robot_config_init);
+    for (int i = Num_Shells - 1; i >= 0; i--) {
+        // Set up fields in Context
+        _context.robot_status[i] = std::move(robot_status_init.back());
+        robot_status_init.pop_back();
+    }
 
     // Initialize team-space transformation
     defendPlusX(defendPlus);
@@ -302,28 +307,6 @@ void Processor::run() {
             logConfig->set_simulation(_simulation);
         }
 
-        for (OurRobot* robot : _context.state.self) {
-            // overall robot config
-            switch (robot->hardwareVersion()) {
-                case Packet::RJ2008:
-                    robot->config = robotConfig2008;
-                    break;
-                case Packet::RJ2011:
-                    robot->config = robotConfig2011;
-                    break;
-                case Packet::RJ2015:
-                    robot->config = robotConfig2015;
-                    break;
-                case Packet::Unknown:
-                    robot->config =
-                        robotConfig2011;  // FIXME: defaults to 2011 robots
-                    break;
-            }
-
-            // per-robot configs
-            robot->status = robotStatuses.at(robot->shell());
-        }
-
         ////////////////
         // Inputs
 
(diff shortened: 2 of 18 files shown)


soccer/motion/MotionControl.cpp
```diff
@@ -35,244 +30,150 @@ void MotionControl::createConfiguration(Configuration* cfg) {
...
-            angleVel > 0 ? minEffectiveAngularSpeed : -minEffectiveAngularSpeed;
+    // Debug drawing
+    {
+        if (at_end) {
+            _drawer->drawCircle(maybe_target->motion.pos, .15, Qt::red,
+                                "Planning");
+        } else if (maybe_target) {
+            _drawer->drawCircle(maybe_target->motion.pos, .15, Qt::green,
+                                "Planning");
+        }
```
JNeiger, line 129: You lost the velocity debug drawer

soccer/motion/MotionControl.cpp
```diff
@@ -35,244 +30,150 @@ void MotionControl::createConfiguration(Configuration* cfg) {
...
-    //} else {
-    if (optTarget) {
-        if (optTarget->angle) {
-            if (optTarget->angle->angle) {
-                targetAngleFinal = *optTarget->angle->angle;
-            }
-        }
-    }
-    //}
+    // TODO: Reset and return if robot is joystick-controlled.
```
JNeiger: I would make sure this is added before actually merging. It would be a real surprise to whoever tests next if you dont lol

soccer/motion/MotionControlNode.cpp
```diff
@@ -1,28 +1,30 @@
...
+}
 
+void MotionControlNode::runMotion(
+    const WorldState& world_state, const GameState& game_state,
+    const std::array<Planning::AngleFunctionPath, Num_Shells>& paths,
+    std::array<MotionSetpoint, Num_Shells>* setpoints) {
+    bool force_stop = game_state.state == GameState::State::Halt;
+    for (int i = 0; i < Num_Shells; i++) {
+        MotionControl& controller = _controllers[i];
+        MotionSetpoint* setpoint = &(*setpoints)[i];
```
JNeiger, line 24: This looks a little jank from a syntax pov, but there's not really a way to fix it :/

logging/CMakeLists.txt
```diff
@@ -8,3 +7,7 @@ target_link_libraries(simple_logger common rc-fshare)
 
 add_executable(log_converter log_converter.cpp)
 target_link_libraries(log_converter common rc-fshare)
+
+# We need XML because Context includes configurables,
+# which uses XML to represent the tree.
+qt5_use_modules(log_converter Xml Core)
```
JNeiger, line 13: This feels unrelated to this pr. Is this part of the process to save context every frame as a data logging solution or something?

soccer/RobotConfig.hpp
```diff
@@ -78,11 +84,14 @@ class RobotConfig {
...
+    RobotStatus() {}
     RobotStatus(Configuration* config, QString prefix);
-    ~RobotStatus() {}
 
-    ConfigBool* chipper_enabled;
-    ConfigBool* kicker_enabled;
-    ConfigBool* ball_sense_enabled;
-    ConfigBool* dribbler_enabled;
+    RobotStatus(RobotStatus&&) = default;
+    RobotStatus& operator=(RobotStatus&&) = default;
```
JNeiger, line 91: Are we sure we want to allow copies with all the unique ptrs here?

soccer/motion/MotionControlNode.cpp
kylestach: I mean, it's not really that weird. It's "address of the value at index i in the array pointed to by setpoints", which is a pretty reasonable thing to express. Either way not much we can do about it.

logging/CMakeLists.txt
kylestach: No, this is because the per-robot configurables got added into context, which log_converter depends upon. When this isn't here log_converter will (eventually) include context, which will include some Qt files from the Xml package, so it will fail. The "proper" fix is that `qt5_use_modules` should transitively allow the include directories (the macro is kind of broken, but it's not our code) - however, this is a reasonable workaround.

soccer/motion/MotionControl.cpp
```diff
@@ -35,244 +30,150 @@ void MotionControl::createConfiguration(Configuration* cfg) {
...
-
-    // Clamp World Acceleration
-    auto dt = RJ::Seconds(RJ::now() - _lastCmdTime);
-    Point targetAccel = (target.vel - _lastWorldVelCmd) / dt.count();
-    targetAccel.clamp(*_max_acceleration);
-
-    target.vel = _lastWorldVelCmd + targetAccel * dt.count();
-
-    _lastWorldVelCmd = target.vel;
-    _lastCmdTime = RJ::now();
```
JNeiger, line 206: Just to throw this out there, it's more correct without all the motion control hacks, so removing them is the "correct" thing to do, but it did bite us in the ass when I did that on firmware.  It might be prudent to leave them in until tuning phase, and then remove them if they're not needed.

kylestach: That's true. However, the minimum-velocity hacks actually significantly (negatively) impact sim performance, and so I think they belong on the robot side of motion control regardless. I'd honestly rather optimize this part for sim anyways, considering that we are who knows how far from seeing an actual robot, and the first thing we do with new robots is motion control work where I'd rather have a clean(er) slate to work with anyway.

Acceleration _feedthrough_ is something that I think probably does belong here. I'll add it in tomorrow. Acceleration limiting is more questionable  - there's an argument to be said that it belongs on the robot, or that it belongs here, but either way I don't think it will be making much difference considering the motion planning already acceleration limits the goal (so it only matters here if feedback is large).

soccer/RobotConfig.hpp
kylestach: That's a move. Moving is okay, copying isn't.

soccer/motion/MotionControl.cpp
kylestach: Yeah, I'll add that in tomorrow. I just couldn't think of a good place for the "is joystick controlled" field to go in Context.

soccer/motion/MotionControl.cpp
kylestach: Fixed

soccer/RobotConfig.hpp
JNeiger: Oof, super misread that

JNeiger (approved):
LGTM

soccer/Processor.cpp
```diff
@@ -121,6 +135,17 @@ Processor::~Processor() {
     // DEBUG - This is unnecessary, but lets us determine which one breaks.
     //_refereeModule.reset();
     _gameplayModule.reset();
+
+    // Put back configurables where we found them.
+    // This is kind of a hack, but if we don't do that they get destructed
+    // when Processor dies. That normally isn't a problem, but in unit tests,
+    // we create and destroy multiple instances of Processor for each test.
+    robot_config_init = std::move(_context.robot_config);
+    std::cout << robot_config_init.get() << std::endl;
```
oswinso: debug?

soccer/motion/MotionSetpoint.hpp
```diff
@@ -1,9 +1,9 @@
 #pragma once
 // stores the outputs published by MotionControl
 struct MotionSetpoint {
-    float xvelocity;
-    float yvelocity;
-    float avelocity;
+    double xvelocity;
+    double yvelocity;
+    double avelocity;
```
oswinso, line 6: is there a reason for this change? Aren't floats used everywhere though?

oswinso (commented):
two comments

soccer/motion/MotionSetpoint.hpp
kylestach: I mean, yes they are in a lot of places, but that doesn't really mean they should be. There's really no reason not to use doubles. I changed some motion control code to use double, so I figured I might as well change this too.

oswinso: I guess we should make an issue to just straight up change all floats to doubles then

kylestach: Probably

oswinso (approved):
lgtm

PR #1426, opened 2020-03-28
Refactor Referee Module for pub/sub architecture + small refactor

oswinso:
## Description
- Clean up the referee node in preparation for moving to a pub-sub
- Performs a small refactor of the business logic inside of the referee node

## Associated Issue
Issue #1424 

## Design Documents
[ROS Roadmap](https://docs.google.com/document/d/1wix-ZK9NPBY0zC8T9LyUihT1ONubXMKxHvb8KYmJkNQ/edit#)

## Steps to test
Referee should still work? I can't compile grsim on my computer because of some conflicts between the version of Qt4 and boost on my ubuntu 18.04, so I can't test.

### DIFF
--- a/soccer/Referee.cpp
+++ b/soccer/Referee.cpp
@@ -1,18 +1,18 @@
-#include "NewRefereeModule.hpp"
+#include "Referee.hpp"
 
+#include <unistd.h>
+
+#include <Network.hpp>
+#include <QMutexLocker>
+#include <QUdpSocket>
+#include <Utils.hpp>
 #include <boost/algorithm/string/predicate.hpp>
+#include <multicast.hpp>
+#include <stdexcept>
 
 #include "Constants.hpp"
 
-#include <Network.hpp>
-#include <multicast.hpp>
-#include <Utils.hpp>
-#include <unistd.h>
-#include <QMutexLocker>
-#include <QUdpSocket>
-#include <stdexcept>
-
-namespace NewRefereeModuleEnums {
+namespace RefreeModuleEnums {
 std::string stringFromStage(Stage s) {
     switch (s) {
         case NORMAL_FIRST_HALF_PRE:
@@ -90,10 +90,9 @@ std::string stringFromCommand(Command c) {
             return "";
     }
 }
-}
+}  // namespace RefreeModuleEnums
 
-using namespace std;
-using namespace NewRefereeModuleEnums;
+using namespace RefreeModuleEnums;
 
 /// Distance in meters that the ball must travel for a kick to be detected
 static const float KickThreshold = Ball_Radius * 3;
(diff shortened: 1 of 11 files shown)


soccer/Referee.cpp
```diff
@@ -108,32 +107,32 @@ static const int KickVerifyTime_ms = 250;
...
+void Referee::getPackets(std::vector<RefereePacket*>& packets) {
     _mutex.lock();
     packets = _packets;
     _packets.clear();
     _mutex.unlock();
 }
 
-void NewRefereeModule::run() {
+void Referee::run() {
     QUdpSocket socket;
```
JNeiger, line 131: @kylestach Did we want to switch to boost::asio instead of qt?

oswinso: yeah I think we did, but I think that should be a separate PR as it's like a different concern?

JNeiger: That works, I just vaguely remembered something about that and just wanted to make sure.

kylestach: Yeah, that can happen later if we want. I'm just annoyed by how Qt tries to be literally everything (QMutex, QString, QUdpSocket, ...).

soccer/Referee.hpp
```diff
@@ -194,27 +195,27 @@ class NewRefereeModule : public QThread {
     TeamInfo yellow_info;
     TeamInfo blue_info;
 
-    void updateGameState(bool blueTeam);
+    void update();
+    GameState updateGameState(const GameState& game_state) const;
 
-    void spinKickWatcher();
+    void spin();
```
kylestach, line 205: Why not combine `spin()` and `update()`? Is there ever a case where one is called without the other?

soccer/Referee.cpp
```diff
@@ -262,167 +263,182 @@ void NewRefereeModule::spinKickWatcher() {
...
+        std::cout << "REFEREE: Stage = " << stringFromStage(stage) << std::endl;
+        prev_stage = stage;
+    }
+}
+
+GameState Referee::updateGameState(const GameState& game_state) const {
+    using namespace RefreeModuleEnums;
+
+    const GameState::Period period = [stage =
+                                          this->stage]() -> GameState::Period {
```
kylestach: Yikes this is some wonky looking formatting. I guess it's fine though.

soccer/Referee.cpp
```diff
@@ -212,13 +211,15 @@ void NewRefereeModule::run() {
     }
 }
 
-void NewRefereeModule::overrideTeam(bool isBlue) {
+void Referee::overrideTeam(bool isBlue) {
     if (!_isRefControlled) {
         blueTeam(isBlue);
     }
 }
```
kylestach: Shouldn't it just pull this information from the game state, and put it into the game state if we're getting it from the ref box?

That said, I think we need to discuss more how the referee module interacts with the GUI. There's some annoying override logic like this spread out in a lot of different places, and I'd like to figure out how best to consolidate/get rid of it (@chachmu).

soccer/Referee.cpp
```diff
@@ -262,167 +263,182 @@ void NewRefereeModule::spinKickWatcher() {
...
+            case Stage::PENALTY_SHOOTOUT_BREAK:
+                return GameState::PenaltyShootout;
+            case Stage::PENALTY_SHOOTOUT:
+                return GameState::PenaltyShootout;
+            case Stage::POST_GAME:
+                return GameState::Overtime2;
+            default:
+                return GameState::FirstHalf;
+        };
+    }();
```
kylestach, line 334: Do you miss Rust yet?

kylestach (commented):
Tested, and I think this is good cleanup for the most part. We can move off of Qt later.

soccer/Referee.cpp
oswinso: Idk I just asked `clang-format`

soccer/Referee.hpp
oswinso: Oh I didn't actually see, I just assumed that it needed to be two separate functions

soccer/Referee.cpp
oswinso: it's actually not that different

oswinso: It definitely is a rust pattern though

oswinso:
@kylestach @JNeiger could you guys check this again?

soccer/Referee.cpp
```diff
@@ -262,167 +263,184 @@ void NewRefereeModule::spinKickWatcher() {
...
-        case Stage::PENALTY_SHOOTOUT_BREAK:
-            _context->game_state.period = GameState::PenaltyShootout;
-            break;
-        case Stage::PENALTY_SHOOTOUT:
-            _context->game_state.period = GameState::PenaltyShootout;
-            break;
-        case Stage::POST_GAME:
-            _context->game_state.period = GameState::Overtime2;
+void Referee::update() {
+    _context->game_state = updateGameState(_context->game_state);
```
JNeiger: I'll have to double check, but nothing else should update game state right? I can't find anything when I search and I think gui changes `command`.

If that's the case, I think it would fit this module structure better to have a private `game_state` variable in this class that we just keep updated and then push into context every frame. It'll mirror the pub/sub quite a bit better

oswinso: Yeah, this (and `overrideTeam` but that works with what you proposed) is the only place where it updates game state. I think it's cleaner as well, I'll make that change.

JNeiger (approved):
LGTM

kylestach (approved):
Looks good to me. Merging.

Issue #1184, opened 2018-07-14
Path Planner doesn't move in the correct direction at the beginning of the path


JNeiger:
When a path planner repeatedly replans the path while the robot is require to do some sort turn early in on the path, it will never actually converge to a solution for some time. Even a small turn such as this will cause problems.

![image](https://user-images.githubusercontent.com/5797304/42720504-80f1e2ce-86ed-11e8-9048-280e996b8729.png)

In some of the worst case scenarios, we will just move the complete opposite way slowly while trying to move to a target location behind us.

Right now we are only replaning occasionally so it's not a obvious problem, but this solution isn't perfect since you may get multiple replans in a row causing the problem anyways.

I think the cause is the first waypoint isn't actually far enough away from our current velocity vector so when do the linear interpolation for the first frame, it's basically a the same speed and direction we are currently going. I think just getting some of the turn happening in the first waypoint will solve this problem.

@ashaw596 Thoughts?

ashaw596:
Hmm. Do you have a good way to replicate this behavior for testing?

JNeiger:
I don't have any code offhand, but I think just having the robot velocity not being in line with the target point it's trying to move to will cause it. It should just slowly move in the opposite direction of the target. I'll see if I can edit some of the code in #1174 to cause it consistently

PR #1429, opened 2020-04-01
Motion planning rewrite clean

### LINKED ISSUE #813: Refactor Motion and Angular Command classes to use boost::variant.
It'll simplify the code and design alot.

ghost:
## Description
Motion planning refactor/rewrite. should roughly behave the same as before. I did modify some of the configuration constants because it seemed to improve performance of the robots in sim.

## Associated Issue
Issue #1329
Issue #1323
- split up big files into smaller files (e.g. RRTPlanner.cpp), big functions into smaller functions, reuse code instead of copy and pasting, reduce number of magic numbers, etc.
- prepared some functions for angle profiling
- added debug layers for planning and angle planning
- added omp for loop for settle brute force

### Test Case 1
1. make rs
2. observe robots

Expected result:  robots behave the same as they used to or better


soccer/planning/planner/MotionCommand.hpp
```diff
@@ -0,0 +1,71 @@
...
+#include <Geometry2d/Point.hpp>
+#include <Geometry2d/Pose.hpp>
+#include <WorldState.hpp>
+#include "planning/trajectory/Trajectory.hpp"
+#include "planning/MotionInstant.hpp"
+#include "Utils.hpp"
+
+namespace Planning {
+//todo(Ethan) discuss whether we want to have settle with bounce
+struct SettleCommand{};
```
ghost: I didn't include the option for robot.settle_w_bounce() in this rewrite because it seemed like it wasn't used. I'm not sure whether we were planning on using that soon. so I can add that back if needed.

soccer/Context.hpp
```diff
@@ -12,7 +12,7 @@
 #include "vision/VisionPacket.hpp"
 
 struct Context {
-    Context() : state(this), debug_drawer(this) {}
+    Context() : state(this), debug_drawer(this), goalie_id(1337) {}
```
kylestach, line 15: I think goalie_id should be in game_state already.

soccer/Processor.cpp
```diff
@@ -381,79 +379,9 @@ void Processor::run() {
...
-        }
-
-        // Visualize obstacles
-        for (auto& shape : globalObstacles.shapes()) {
-            _context.debug_drawer.drawShape(shape, Qt::black,
-                                            "Global Obstacles");
-        }
+        // Run the path planner.
+        // TODO(Kyle): Remove when everything is in modules.
+        _pathPlanner->run();
```
kylestach, line 384: :smile:

soccer/RobotIntent.hpp
```diff
@@ -13,7 +12,7 @@ struct RobotIntent {
     enum class Song { STOP, CONTINUE, FIGHT_SONG };
 
     std::unique_ptr<Planning::MotionCommand> motion_command;
-    std::unique_ptr<Planning::RotationCommand> rotation_command;
+    std::optional<double> angle_override;
```
kylestach, line 15: Could you add a comment describing what this field does? I assume it's a angle function replacement but I could be wrong.

soccer/motion/MotionControl.cpp
```diff
@@ -75,46 +75,62 @@ void MotionControl::run() {
     _angleController.kd = *_robot->config->rotation.d;
```
kylestach, line 75: Just FYI, everything in this file just got rewritten, so this stuff will need to be ported over. I'll deal with the merge as soon as this PR is through review.

soccer/planning/DynamicObstacle.hpp
```diff
@@ -1,54 +1,14 @@
...
-    std::shared_ptr<Geometry2d::Circle> getStaticObstacle() const {
-        return staticObstacle;
-    }
+class Trajectory;
+struct DynamicObstacle {
+    DynamicObstacle(const std::shared_ptr<Geometry2d::Circle>& circ,
+                    const Trajectory* pAth)
+        : circle(circ), path(pAth) {}
+    std::shared_ptr<Geometry2d::Circle> circle;
+    const Trajectory* path;
```
kylestach, line 12: It's unclear, who has ownership over this trajectory? Or, I guess, what guarantees do you need to provide about it?

soccer/planning/MotionInstant.hpp
```diff
@@ -25,48 +25,4 @@ struct MotionInstant {
     }
```
kylestach, line 25: Do you actually use `MotionInstant` anywhere?

soccer/planning/PlannerNode.cpp
```diff
@@ -0,0 +1,147 @@
...
+
+        // create and visualize obstacles
+        ShapeSet staticObstacles = robot->collectStaticObstacles(
+            globalObstaclesForBot, !robotIgnoreGoalZone);
+        for (OurRobot* r2 : context_->state.self) {
+            if (robot == r2) {
+                continue;
+            }
+            staticObstacles.add(
+                std::make_shared<Circle>(r2->pos(), Robot_Radius));
```
kylestach, line 64: Style nit here, I'd prefer:
```cpp
if (robot != r2) {
    staticObstacles.add(...);
}
```

soccer/planning/PlannerNode.cpp
```diff
@@ -0,0 +1,147 @@
...
+    globalObstaclesWithGoalZones.add(context_->goalZoneObstacles);
+
+    std::vector<PlanRequest> requests;
+    for (OurRobot* robot : context_->state.self) {
+        if (!robot) {
+            continue;
+        }
+
+        if (!robot->visible() ||
+            context_->game_state.state == GameState::Halt) {
```
kylestach, line 36: This is already compound, I think it should be fine to combine it with the above statement:
```cpp
if (!robot ||
    !robot->visible ||
    context_->game_state.state == GameState::Halt) {
    ...
}
```

soccer/planning/PlannerNode.cpp
```diff
@@ -0,0 +1,147 @@
...
+
+        // draw debug info
+        robot->path().draw(&context_->debug_drawer,
+                           robot->pos() + Point(.1, 0));
+        context_->debug_drawer.drawText(
+            (const char* []){
+                "EmptyCommand", "PathTargetCommand", "WorldVelCommand",
+                "PivotCommand", "SettleCommand", "CollectCommand",
+                "LineKickCommand",
+                "InterceptCommand"}[robot->motionCommand()->index()],
```
kylestach, line 105: Thanks C++, very cool.

soccer/planning/PlannerNode.cpp
```diff
@@ -0,0 +1,147 @@
...
+                *robot->motionCommand(), robot->robotConstraints(),
+                robot->path_movable(), std::move(staticObstacles),
+                std::vector<DynamicObstacle>{}, robot->shell(),
+                robot->getPlanningPriority());
+        }
+    }
+    std::sort(requests.begin(), requests.end(),
+              [](const PlanRequest& pr1, const PlanRequest& pr2) {
+                  return pr1.priority > pr2.priority;
+              });
```
kylestach, line 83: Do we even use priorities anyway? I'd almost be tempted to drop that feature, but I guess there's no harm keeping it around for the time being.

soccer/planning/PlannerNode.cpp
```diff
@@ -0,0 +1,147 @@
...
+    // This gives the planners a sort of "priority" - this makes sense, because
+    // the empty planner is always last.
+    DebugDrawer* drawer = &request.context->debug_drawer;
+    for (int i = 0; i < planners_.size(); i++) {
+        if (planners_[i]->isApplicable(request.motionCommand)) {
+            // clear path when planner changes
+            if (plannerIdx[request.shellID] != i) {
+                plannerIdx[request.shellID] = i;
+                request.prevTrajectory = Trajectory{{}};
+            }
```
kylestach, line 128: What happens when there was no valid planner? Should this be set to -1 or something?

soccer/planning/planner/CollectPlanner.cpp
```diff
@@ -0,0 +1,139 @@
...
+        new ConfigDouble(cfg, "Capture/Collect/bufferDistAfterContact");
+    _ballSpeedApproachDirectionCutoff = new ConfigDouble(
+        cfg, "Capture/Collect/ballSpeedApproachDirectionCutoff");
+    _touchDeltaSpeed = new ConfigDouble(cfg, "Capture/Collect/touchDeltaSpeed");
+    _targetChangeThreshold =
+        new ConfigDouble(cfg, "Capture/Collect/targetChangeThreshold");
+    _controlAccelScalePercent =
+        new ConfigDouble(cfg, "Capture/Collect/controlAccelScalePercent");
+    _distCutoffToControl =
+        new ConfigDouble(cfg, "Capture/Collect/distCutoffToControl");
```
kylestach, line 28: Might be worth providing a description for these (optional third argument) given that they're kind of esoteric.

soccer/planning/planner/CollectPlanner.cpp
```diff
@@ -0,0 +1,139 @@
...
+}
+
+using namespace Geometry2d;
+Trajectory CollectPlanner::plan(PlanRequest&& request) {
+    processStateTransitions(request);
+    switch (_collectStates[request.shellID]) {
+        case CollectState::Course:
+            return course(std::move(request));
```
kylestach, line 36: course -> coarse?

soccer/planning/planner/CollectPlanner.cpp
```diff
@@ -0,0 +1,139 @@
...
+        targetInstant = RobotInstant{Pose{noisyTargetPoint, 0},
+                                     Twist{noisyTargetVel, 0}, RJ::Time{0s}};
+    }
+    request.motionCommand = PathTargetCommand{*targetInstant};
+    request.static_obstacles.add(
+        std::make_shared<Circle>(ball.pos, Ball_Radius));
+    Trajectory path = _pathTargetPlanner.plan(std::move(request));
+    PlanAngles(path, startInstant, AngleFns::faceAngle(approachDir.angle()),
+               rotationConstraints);
+    return std::move(path);
```
kylestach, line 67: I know none of the logic here is insanely complicated, but right now it is kind of a wall of text with no comments and it's really hard to parse. Near as I can tell the steps are:
 - Find a rough direction from which to approach the ball
 - Move that out a bit from the ball's position and call it the "target point"
 - Apply hysteresis to that to make it less noisy
 - Use that to plan a path regularly, facing at the angle towards the ball
Is this on the right track?

soccer/planning/planner/CollectPlanner.cpp
```diff
@@ -0,0 +1,139 @@
...
+    request.constraints.mot.maxSpeed =
+        ball.vel.dot(approachDir) + *_touchDeltaSpeed;
+    Point targetVel = request.constraints.mot.maxSpeed * approachDir;
+    RobotInstant targetInstant{Pose{targetPoint, 0}, Twist{targetVel, 0},
+                               RJ::Time{0s}};
+    request.motionCommand = PathTargetCommand{targetInstant};
+    Trajectory path = _pathTargetPlanner.plan(std::move(request));
+    PlanAngles(path, startInstant, AngleFns::faceAngle(approachDir.angle()),
+               rotationConstraints);
+    return std::move(path);
```
kylestach, line 85: Similarly difficult to read. Looks like the logic is way simpler, but it could still really use being broken up and commented.

soccer/planning/planner/CollectPlanner.cpp
```diff
@@ -0,0 +1,139 @@
...
+        startInstant.pose.position() + stoppingDist * approachDir;
+    RobotInstant targetInstant{Pose{targetPoint, 0}, {}, RJ::Time{0s}};
+    request.motionCommand = PathTargetCommand{targetInstant};
+    request.constraints.mot.maxSpeed = std::min(
+        startInstant.velocity.linear().mag(), request.constraints.mot.maxSpeed);
+    request.constraints.mot.maxAcceleration = accel;
+    Trajectory path = _pathTargetPlanner.plan(std::move(request));
+    PlanAngles(path, startInstant, AngleFns::faceAngle(approachDir.angle()),
+               rotationConstraints);
+    return std::move(path);
```
kylestach, line 106: Again, more comments please.

soccer/planning/planner/CollectPlanner.cpp
```diff
@@ -0,0 +1,139 @@
...
+}
+void CollectPlanner::processStateTransitions(const PlanRequest& request) {
+    const Ball& ball = request.context->state.ball;
+    RobotInstant startInstant = request.start;
+    double distToBall = (startInstant.pose.position() - ball.pos).mag();
+    double distToBallAtContact = Ball_Radius + Robot_MouthRadius;
+    CollectState& collectState = _collectStates[request.shellID];
+    const double distCutoffToFine = *_bufferDistBeforeContact + 0.0765;
```
kylestach, line 114: Magic number.

soccer/planning/planner/CollectPlanner.cpp
```diff
@@ -0,0 +1,139 @@
...
+            collectState = CollectState::Control;
+        }
+    }
+    // backward transitions
+    bool fineToCourse =
+        collectState == CollectState::Fine &&
+        distToBall > distCutoffToFine + distToBallAtContact + 0.2;
+    bool controlToCourse =
+        collectState == CollectState::Control &&
+        distToBall > *_distCutoffToControl + distToBallAtContact + 0.072;
```
kylestach, line 133: More magic numbers, 0.2 and 0.072.

soccer/planning/planner/CollectPlanner.cpp
```diff
@@ -0,0 +1,139 @@
+#include "CollectPlanner.hpp"
+#include "planning/planner/PlanRequest.hpp"
+#include "planning/trajectory/RRTUtil.hpp"
+namespace Planning {
+ConfigDouble* CollectPlanner::_bufferDistBeforeContact;
+ConfigDouble* CollectPlanner::_bufferDistAfterContact;
```
kylestach, line 6: It looks like this is unused, please remove it.

soccer/planning/planner/CollectPlanner.hpp
```diff
@@ -0,0 +1,75 @@
...
+        const Ball& ball = request.context->state.ball;
+        return ball.vel.mag() < *_ballSpeedApproachDirectionCutoff
+                   ? (ball.pos - request.start.pose.position()).norm()
+                   : ball.vel.norm();
+    }
+
+    // At what speed should we be when we touch the ball (Well, the difference
+    // in speed between the ball and robot) Should be as low as possible where
+    // we still are able to touch the ball and control it If we are slamming
+    // into the ball decrease this number If we aren't even touching it to the
```
kylestach, line 38: Punctuation please :slightly_smiling_face:

soccer/planning/planner/CollectPlanner.hpp
```diff
@@ -0,0 +1,75 @@
...
+    //
+    // Low number indicates that it will always try to choose a point for the
+    // robot behind the velocity vector
+    //
+    // High number indicates that it will always try to choose a point nearest
+    // to the current robot position
+    static ConfigDouble* _ballSpeedApproachDirectionCutoff;  // m/s
+
+    // the amount the target needs to change before replanning
+    static ConfigDouble* _targetChangeThreshold;
```
kylestach, line 68: Okay, now I see the documentation for these config variables. Putting it in the description would be really nice still, but not necessary.

soccer/planning/planner/LineKickPlanner.cpp
```diff
@@ -0,0 +1,204 @@
...
+    }
+
+    if (!prevTrajectory.empty() && _targetKickPos) {
+        RJ::Seconds timeInto = RJ::now() - prevTrajectory.begin_time();
+        RJ::Seconds timeLeft = prevTrajectory.duration() - timeInto;
+
+        Point targetPos;
+        RJ::Time time = ball.estimateTimeTo(*_targetKickPos, &targetPos);
+        RJ::Seconds timeToHit = time - curTime;
+        if (timeLeft < RJ::Seconds(1.0)) {
```
kylestach, line 47: Magic numbers :(

What case is this condition supposed to actually handle?

soccer/planning/planner/LineKickPlanner.cpp
```diff
@@ -0,0 +1,204 @@
...
+            Point targetVel =
+                (aimTarget - targetPos).normalized(*_approachSpeed);
+            targetPos -= targetVel.normalized(Robot_Radius + Ball_Radius * 2);
```
kylestach, line 50: Why two ball radii away?

soccer/planning/trajectory/Trajectory.hpp
```diff
@@ -0,0 +1,348 @@
...
+
+    /**
+     * Returns true if the path hits an obstacle
+     *
+     * @param[in]	shape The obstacles on the field
+     * @param[out]  hitTime the approximate time when the path hits an obstacle.
+     * If no obstacles are hit, this value is not modified
+     * @param[in] 	startTimeIntoPath The time on the path to start checking
+     *from
+     * @return 		true if it hits an obstacle, otherwise false
```
kylestach, line 112: Looks like the formatting on this got screwed up.

soccer/planning/trajectory/Trajectory.hpp
```diff
@@ -0,0 +1,348 @@
...
+
+    /**
+     * get the time this trajectory was created
+     * @return time stamp
+     */
+    RJ::Time timeCreated() { return stamp; }
+
+private:
+    // A sorted array of RobotInstants (by timestamp)
+    std::list<RobotInstant> instants_;
```
kylestach, line 303: I'm still very skeptical that a list is really helping you here considering how much you iterate over it. Vectors are really really fast even if you're copying data around - I could be wrong but my hunch is that cache locality is basically going to beat out everything else in terms of performance. cc @oswinso @JNeiger for opinions.

soccer/planning/trajectory/VelocityProfiling.cpp
```diff
@@ -0,0 +1,235 @@
...
+    }
+    //    //if derives1 = 0 at the at the endpoints curvature gets undefined
+    //    if(num_points > 1) {
+    //        if(derivs1[0].mag() < 1e-6) {
+    //            curvature[0] = curvature[1];
+    //        }
+    //        if(derivs1[num_points-1].mag() < 1e-6) {
+    //            curvature[num_points-1] = curvature[num_points-2];
+    //        }
+    //    }
```
kylestach, line 93: Can we get rid of this?

soccer/planning/trajectory/VelocityProfiling.cpp
```diff
@@ -0,0 +1,235 @@
...
+        //        }
+        double vbar = (speed[n] + speed[n - 1]) / 2;
+        assert(vbar != 0);
+        double t_sec = distance / vbar;
+        assert(t_sec > 1e-12);
+        RJ::Time current_time = trajectory.last().stamp + RJ::Seconds(t_sec);
+        // Add point n in
+        trajectory.AppendInstant(RobotInstant{
+            Pose(points[n], 0), Twist(derivs1[n].normalized(speed[n]), 0),
+            current_time});
```
kylestach, line 153: Comment that this loop is transcribing the list to a trajectory, and probably delete the if/else stuff.

soccer/planning/trajectory/VelocityProfiling.cpp
```diff
@@ -0,0 +1,235 @@
...
+    assert(maxAccel > 1e-12);
+    assert(std::abs(angleVels.back()) > 1e-12);
+    if (std::abs(angleVels.back()) > 1e-12) {
+        double accel = maxAccel * (angleVels.back() > 0 ? -1 : 1);
+        double stoppingAngle = -std::pow(angleVels.back(), 2) / (2 * accel);
+        angles.push_back(angles.back() + stoppingAngle);
+        angleVels.push_back(0);
+    }
+}
+// todo: do velocity profile angles
```
kylestach, line 209: Is this still an issue?

soccer/planning/trajectory/VelocityProfiling.cpp
```diff
@@ -0,0 +1,235 @@
...
+    for (int i = 1; instants_it != trajectory.instants_end(); i++) {
+        RobotInstant& instant_after = *instants_it;
+        double deltaTime =
+            RJ::Seconds(instant_after.stamp - instant_before.stamp).count();
+        instant_after.pose.heading() = angle_function(instant_before);
+        double deltaAngle = fixAngleRadians(instant_after.pose.heading() -
+                                            instant_before.pose.heading());
+        instant_after.velocity.angular() = deltaAngle / deltaTime;
+        instant_before = instant_after;
+        ++instants_it;
```
kylestach, line 232: This can be part of the for loop.

soccer/planning/trajectory/VelocityProfiling.cpp
```diff
@@ -0,0 +1,235 @@
...
+    }
```
kylestach, line 233: Comments for this function.

soccer/planning/trajectory/VelocityProfiling.hpp
```diff
@@ -0,0 +1,80 @@
+#pragma once
+
+#include <planning/RotationConstraints.hpp>
+#include "PathSmoothing.hpp"
+#include "Trajectory.hpp"
+
+namespace Planning {
+
+/**
+ * Returns target angle from (position, linear velocity, previous angle)
```
kylestach, line 10: And angular velocity?

soccer/ui/MainWindow.cpp
```diff
@@ -40,9 +40,10 @@ using namespace Packet;
 using namespace Eigen;
 
 static const std::vector<QString> defaultHiddenLayers{
-    "MotionControl", "Global Obstacles", "Local Obstacles",
-    "Planning0",     "Planning1",        "Planning2",
-    "Planning3",     "Planning4",        "Planning5"};
+    "MotionControl",  "Global Obstacles",  "Local Obstacles", "Planning0",
+    "Planning1",      "Planning2",         "Planning3",       "Planning4",
+    "Planning5",      "PlanningDebugText", "MotionCommands",  "Planning",
+    "Angle Planning", "Settle Brute Force"};
```
kylestach, line 46: Why do we have all of the `Planning<N>` layers?

soccer/planning/trajectory/RRTUtil.cpp
```diff
@@ -0,0 +1,196 @@
...
+                    const ShapeSet& static_obstacles,
+                    const vector<DynamicObstacle>& dynamic_obstacles,
+                    const vector<Point>& biasWaypoints) {
+    Trajectory rrtPath =
+        CreatePath::rrt(start, goal, motionConstraints, static_obstacles,
+                        dynamic_obstacles, biasWaypoints);
+    if (!rrtPath.empty()) {
+        return std::move(rrtPath);
+    }
+    return CreatePath::simple(start, goal, motionConstraints);
```
kylestach, line 193: Shouldn't we actually just fail instead of producing a wrong path?

soccer/planning/trajectory/RRTUtil.cpp
```diff
@@ -0,0 +1,196 @@
...
+    }
+    vector<Point> points = biRRT.getPath();
+    RRT::SmoothPath(points, *state_space);
+    return std::move(points);
+}
+
+vector<Point> GenerateRRT(Point start, Point goal, const ShapeSet& obstacles,
+                          const vector<Point>& waypoints) {
+    // note: we could just use state_space.transitionValid() for the straight
+    // line test, but this runs quicker
```
kylestach, line 114: Wait, seriously? Weird...

soccer/planning/trajectory/RRTUtil.cpp
```diff
@@ -0,0 +1,196 @@
...
+    if (start.pose.position().distTo(goal.pose.position()) < Robot_Radius ||
+        !straightTrajectory.hit(static_obstacles, 0s) &&
+            !straightTrajectory.intersects(dynamic_obstacles, start.stamp)) {
+        return std::move(straightTrajectory);
+    }
+
+    ShapeSet obstacles = static_obstacles;
+    Trajectory path{{}};
+    constexpr int attemptsToAvoidDynamics = 10;
+    for (int i = 0; i < attemptsToAvoidDynamics; i++) {
```
kylestach, line 159: I kind of worry about this, especially considering that I think (correct me if I'm wrong) this gets run in i.e. capture's brute-force. At the very least `attemptsToAvoidDynamics` should be configurable, and probably a parameter to this function (that gets set to 1 when doing brute-force search).

soccer/planning/tests/TrajectoryTest.cpp
```diff
@@ -0,0 +1,423 @@
...
+TEST(Trajectory, createRRTSmall) {
+    RobotInstant start{Pose{{}, .1}, Twist{}, RJ::now()};
+    Trajectory a = CreatePath::rrt(start, start, MotionConstraints{}, {});
+    ASSERT_FALSE(a.empty());
+    ASSERT_TRUE(a.num_instants() == 1);
+    ASSERT_NEAR(a.duration().count(), 0.0, 1e-6);
+}
+
+namespace Planning {
+double clampAccel(double v1, double v2, double deltaX, double maxAccel);
```
kylestach, line 222: Why not just include the relevant header file?

soccer/planning/tests/TrajectoryTest.cpp
```diff
@@ -0,0 +1,423 @@
...
+    ASSERT_NEAR(evalHalfway->velocity.angular(), halfwayVel, 1e-6);
+
+    assertPathContinuous(out, constraints);
+}
+
+/*
+ * Note:
+ * These tests are for future use when we actually do angle profiling
+ * For now, they fail
+ */
```
kylestach, line 407: Are they okay now?

soccer/planning/trajectory/Trajectory.cpp
```diff
@@ -0,0 +1,370 @@
...
+    }
+    if (empty() || startTimeIntoPath > duration()) {
+        return false;
+    }
+    // this fixes the edge case where we increment by a duration of 0s and
+    // end up looping forever TODO do this in a more readable way
+    constexpr int max_iterations = 100;
+    TrajectoryIterator it =
+        iterator(begin_time() + startTimeIntoPath,
+                 (duration() - startTimeIntoPath) / (double)max_iterations);
```
kylestach, line 153: Uh, what? How does this fix that?

soccer/planning/trajectory/Trajectory.cpp
```diff
@@ -0,0 +1,370 @@
...
+    }
+    return false;
+}
+bool Trajectory::intersects(const std::vector<DynamicObstacle>& obstacles,
+                            RJ::Time startTime, Geometry2d::Point* hitLocation,
+                            RJ::Seconds* hitTime) const {
+    if (empty()) {
+        return false;
+    }
+    constexpr RJ::Seconds deltaT = 0.05s;
```
kylestach, line 192: This being at a fixed step size and the above one not being at a fixed size kind of sketches me out. Is there a reason for that?

soccer/planning/trajectory/Trajectory.cpp
```diff
@@ -0,0 +1,370 @@
...
+                    if (hitLocation) {
+                        *hitLocation = obsPoint;
+                    }
+                    return true;
+                }
+            }
+        }
+    }
+    return false;
+}
```
kylestach, line 225: I guess some of this code is sketch but unavoidable.

soccer/CMakeLists.txt
```diff
@@ -5,6 +5,9 @@ set(CMAKE_AUTOMOC ON)
 # include Eigen3 for linear algebra
 include_directories(SYSTEM ${EIGEN_INCLUDE_DIR})
 
+set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -O3 -fopenmp")
+set(CMAKE_C_FLAGS_DEBUG "${CMAKE_C_FLAGS_DEBUG} -O3 -fopenmp")
```
JNeiger, line 9: O3 in debug mode may make it quite annoying to debug due to all the optimization. I'm not the most experienced with this, but -Og is the one I usually see in debug. Someone else may have better input here

soccer/planning/trajectory/Trajectory.cpp
```diff
@@ -0,0 +1,370 @@
...
+    other.stamp = RJ::Time{0s};
+    return *this;
+}
+
+Trajectory::Trajectory(const Trajectory& a, const Trajectory& b)
+    : Trajectory(Trajectory{a}, Trajectory{b}) {}
+
+Trajectory::Trajectory(Trajectory&& a, Trajectory&& b) : stamp(b.stamp) {
+    instants_ = std::move(a.instants_);
+    if (!b.empty()) b.instants_.pop_front();
```
kylestach, line 41: Why pop the front? You don't _always_ want to do that, right?

soccer/planning/tests/TestingUtils.cpp
```diff
@@ -0,0 +1,79 @@
...
+        //            cur.pose.position().distTo(nxt.pose.position()); double
+        //            distAngle = std::abs(
+        //                    fixAngleRadians(cur.pose.heading() -
+        //                    nxt.pose.heading()));
+        //            double maxJump = std::max((double) Robot_Radius,
+        //                                      path.last().pose.position().distTo(
+        //                                              path.first().pose.position())
+        //                                              *
+        //                                      0.1);
+        //            ASSERT_LT(dist, maxJump);
```
kylestach, line 36: They should fail by like a small, right? I guess something you could do is expect that the change in position will be bounded by the max(speed at beginning, speed in middle, speed at end) times ds times some fudge factor. I think that's guaranteed to be correct for some value of the fudge factor, for third order splines like we're using. (If the derivative of a cubic is small at three separate places in an interval it has to be small everywhere within the interval, I think).

soccer/planning/planner/PlanRequest.hpp
```diff
@@ -0,0 +1,88 @@
...
+    PlanRequest copyNoHistory() const {
+        return PlanRequest{context,           start,          motionCommand,
+                           constraints,       Trajectory{{}}, static_obstacles,
+                           dynamic_obstacles, shellID,        priority};
+    }
+
+    /**
+     * The system context. Used for debug drawing and robot state information.
+     */
+    Context* context;
```
kylestach, line 49: In theory, the next step will be to remove context from this. That's going to be a lot of work so we'll save it for later though.

soccer/planning/planner/Planner.hpp
```diff
@@ -0,0 +1,61 @@
...
+     *
+     * @param request The request to plan.
+     * @return A trajectory for the robot to follow.
+     */
+    virtual Trajectory plan(PlanRequest&& request) = 0;
+
+    /**
+     * reuse previous path
+     */
+    static Trajectory reuse(PlanRequest&& request);
```
kylestach, line 35: I would suggest explaining exactly what this does in the documentation here, considering the other functions here have nice doxygen.

soccer/planning/planner/PathTargetPlanner.hpp
```diff
@@ -0,0 +1,46 @@
...
+    RobotInstant getGoalInstant(const PlanRequest& request) const;
+    bool veeredOffPath(const PlanRequest& request) const;
+    bool goalChanged(const RobotInstant& prevGoal,
+                     const RobotInstant& goal) const;
+
+    Trajectory partialPath(const Trajectory& prevTrajectory) {
+        return prevTrajectory.subTrajectory(
+            0s, (RJ::now() - prevTrajectory.begin_time()) +
+                    RJ::Seconds{*_partialReplanLeadTime});
+    }
```
kylestach, line 38: Documentation for what all of these do please.

soccer/planning/planner/LineKickPlanner.cpp
```diff
@@ -0,0 +1,204 @@
...
+    RobotInstant startInstant = request.start;
+    Ball& ball = request.context->state.ball;
+    RotationConstraints rotationConstraints = request.constraints.rot;
+    const RJ::Time curTime = RJ::now();
+    Trajectory& prevTrajectory = request.prevTrajectory;
+    if (prevTrajectory.empty()) {
+        _finalApproach = false;
+    }
+
+    // Fine segment: in the final part of the trajectory we don't replan
```
kylestach, line 24: Can you break this out into a state machine like you did with collect? I think that really cleaned things up.

soccer/Robot.cpp
```diff
@@ -190,58 +195,56 @@ void OurRobot::move(Geometry2d::Point goal, Geometry2d::Point endVelocity) {
...
     *_cmdText << "move(" << goal.x() << ", " << goal.y() << ")" << endl;
     *_cmdText << "endVelocity(" << endVelocity.x() << ", " << endVelocity.y()
               << ")" << endl;
 }
 
-void OurRobot::settle(std::optional<Point> target) {
+void OurRobot::settle() {
     if (!visible()) return;
-
-    intent().motion_command = std::make_unique<Planning::SettleCommand>(target);
```
JNeiger, line 204: FWIW, settle with target was filling two roles in my mind when I made it. A first run of a one touch command results in almost the exact same code in terms of finding where to intercept, getting there, and then stopping. the only difference is you want to face a certain direction and kick on breakbeam. The second role is that if you want to two touch settle a ball, you would settle at an angle and allow the not fully damped ball bounce in the direction of the next shot/pass and one touch the slow moving ball.

I haven't looked at the rest of the code to make sure one touch still worked in this new system.

soccer/planning/planner/SettlePlanner.cpp
```diff
@@ -0,0 +1,177 @@
...
+    RJ::Time interceptTime =
+        ball.estimateTimeTo(ball.pos + ball.vel.normalized(endDist));
+    _pathTargetPlanner.drawRadius = Ball_Radius;
+    _pathTargetPlanner.drawColor = Qt::white;
+    _pathTargetPlanner.drawLayer = "Settle Brute Force";
+#pragma omp parallel for default(none) \
+    shared(request, dists, ball, interceptTime)
+    for (int i = 0; i < dists.size(); i++) {
+        RJ::Time contactTime =
+            ball.estimateTimeTo(ball.pos + ball.vel.normalized(dists[i]));
```
kylestach, line 56: I'd expect that a surprisingly heft amount of the time here is being taken by `estimateTimeTo`. You could probably improve this by searching over intercept _times_ instead of intercept _positions_, and then doing forward-lookups instead of reverse lookups like this.

soccer/SystemState.cpp
```diff
@@ -189,3 +140,18 @@ std::vector<int> SystemState::ourValidIds() {
...
+    for (int i = 0; i < interpolations; i++) {
+        double percent = static_cast<double>(i) / (interpolations - 1);
+        RJ::Time stamp = RJ::now() + RJ::Seconds{totalTime * percent};
+        MotionInstant ballInstant = predict(stamp);
+        instants.emplace_back(Pose{ballInstant.pos, 0},
+                              Twist{ballInstant.vel, 0}, stamp);
+    }
+    _path = Planning::Trajectory{std::move(instants)};
+    return DynamicObstacle{std::make_shared<Circle>(pos, Ball_Radius), &_path};
+}
```
kylestach, line 157: I'm not sure that any of the ball stuff belongs in SystemState anymore. Ideally it shold be moved to its own file.

soccer/planning/planner/LineKickPlanner.cpp
```diff
@@ -0,0 +1,204 @@
...
+    request.motionCommand = PathTargetCommand{targetInstant};
+    Trajectory path = _pathTargetPlanner.plan(std::move(request));
+    path.setDebugText("Gives ups");
+    _targetKickPos = std::nullopt;
+    PlanAngles(path, startInstant, AngleFns::facePoint(aimTarget),
+               rotationConstraints);
+    return std::move(path);
+}
+
+Trajectory LineKickPlanner::planForSlowMovingBall(PlanRequest&& request) {
```
kylestach, line 99: Does line kick even work on a fast moving ball?

soccer/planning/planner/LineKickPlanner.cpp
```diff
@@ -0,0 +1,204 @@
...
+                               RJ::Time{0s}};
+    request.motionCommand = PathTargetCommand{targetInstant};
+    Trajectory path = _pathTargetPlanner.plan(std::move(request));
+    PlanAngles(path, startInstant, AngleFns::facePoint(aimTarget),
+               rotationConstraints);
+    path.setDebugText("Slow Moving Ball");
+    return std::move(path);
+}
+
+std::optional<Trajectory> LineKickPlanner::attemptBruteForce(
```
kylestach, line 146: This brute-force logic is close to common between the planners, right? Could we somehow combine it?

(Also, comments pls)

soccer/planning/planner/LineKickPlanner.cpp
```diff
@@ -0,0 +1,204 @@
...
+    constexpr double ballAvoidDistance = 0.05;
+    Point aimTarget = std::get<LineKickCommand>(request.motionCommand).target;
+    const Ball& ball = request.context->state.ball;
+    Trajectory& prevTrajectory = request.prevTrajectory;
+    RobotInstant startInstant = request.start;
+    RotationConstraints rotationConstraints = request.constraints.rot;
+    Point targetVel = (aimTarget - ball.pos).normalized(*_approachSpeed);
+    Point targetPos;
+    if (std::abs(targetVel.angleBetween(
+            ball.pos - startInstant.pose.position())) > 10 * M_PI / 180.0) {
```
kylestach, line 109: So, if our target velocity is more than 10 degrees off from the direction to the goal? Then we should pick a new target point off of that? Honestly a lot of the intuition behind these loses me so more comments would be nice (I know I sound like a broken record at this point).

soccer/planning/planner/SettlePlanner.cpp
```diff
@@ -0,0 +1,177 @@
...
+    _targetPointGain = new ConfigDouble(cfg, "Capture/Settle/targetPointGain");
+    _searchStartDist = new ConfigDouble(cfg, "Capture/Settle/searchStartDist");
+    _searchEndDist = new ConfigDouble(cfg, "Capture/Settle/searchEndDist");
+    _searchIncDist = new ConfigDouble(cfg, "Capture/Settle/searchIncDist");
+    _targetChangeThreshold =
+        new ConfigDouble(cfg, "Capture/Settle/targetChangeThreshold");
+    _shortcutDist = new ConfigDouble(cfg, "Capture/Settle/shortcutDist");
+}
+using namespace Geometry2d;
+// todo(motion planning) find a better way to handle moving targets
```
kylestach, line 30: :'(

soccer/gameplay/skills/angle_receive.py
```diff
@@ -16,24 +16,12 @@
 # Kick is a single_robot_behavior, so no need to import both
 class AngleReceive(skills.pass_receive.PassReceive):
```
JNeiger, line 17: Angle receive captures the ball at a heading offset. This is used in one touch passes as the receiver. With the current changes, I don't think this would work in it's current state. It's ok to push this off as an issue, but we should just be specific that this is the decision being made so the AI can be updated to remove one touches.

soccer/gameplay/skills/settle.py
```diff
@@ -10,10 +10,9 @@
 # while facing the target to either bounce the ball in that
 # direction or kick on breakbeam in that direction
```
JNeiger, line 11: This comment should be updated if we're removing that functionality

kylestach (changes requested):
First of all, really nice work! This is a significant step forwards in terms of cleanliness, and should enable us to do some much more fun things very soon.

Second: I'm sorry for the barrage of comments. At some point I realized that half of them were just "more comments", which you probably knew already, so not super helpful. That said, I've highlighted some of the areas I had trouble understanding immediately but when I got to the actual planners I pretty much gave up and figured I'd wait for comments.

Third: there are definitely some issues that I've run into trying to use this in actual soccer. We seem to get stuck a lot when running the old adaptive code, especially when we're near the ball. In one of these bugs, processing is actually hung (this indicates the code is freezing up and getting stuck in a long loop here, although it is eventually able to break out of it once I try to move the ball, so it's not infinite). I'll do more investigation into what's causing it, but I was able to trigger it relatively frequently by just running adaptive normally. It doesn't get much better in release mode.

soccer/planning/DynamicObstacle.hpp
```diff
@@ -1,54 +1,14 @@
...
-    // Radius = radius of obstacle
-    float getRadius() const { return radius; }
-
-    std::shared_ptr<Geometry2d::Circle> getStaticObstacle() const {
-        return staticObstacle;
-    }
+class Trajectory;
+struct DynamicObstacle {
+    DynamicObstacle(const std::shared_ptr<Geometry2d::Circle>& circ,
+                    const Trajectory* pAth)
```
JNeiger, line 9: `pAth` -> `path`

soccer/planning/DynamicObstacle.hpp
```diff
@@ -1,54 +1,14 @@
...
-
-    // Radius = radius of obstacle
-    float getRadius() const { return radius; }
-
-    std::shared_ptr<Geometry2d::Circle> getStaticObstacle() const {
-        return staticObstacle;
-    }
+class Trajectory;
+struct DynamicObstacle {
+    DynamicObstacle(const std::shared_ptr<Geometry2d::Circle>& circ,
```
JNeiger, line 8: If we're going to copy the shared_ptr, it might be better to show that in the args by just making it a `shared_ptr` instead of `const shared_ptr&`

soccer/planning/PlannerNode.cpp
oswinso: Actually I prefer guard conditions as I think it improves readability

soccer/planning/PlannerNode.cpp
```diff
@@ -0,0 +1,147 @@
...
+
+namespace Planning {
+
+PlannerNode::PlannerNode(Context* context)
+    : context_(context), plannerIdx(Num_Shells, -1) {
+    planners_.push_back(std::make_unique<PathTargetPlanner>());
+    planners_.push_back(std::make_unique<SettlePlanner>());
+    planners_.push_back(std::make_unique<CollectPlanner>());
+    planners_.push_back(std::make_unique<LineKickPlanner>());
+    planners_.push_back(std::make_unique<PivotPathPlanner>());
```
JNeiger, line 18: It feels weird to be treating these planners as state agnostic by having only one instantiation for each planner, then to be storing the state of each robot within the planners themselves. It adds a bunch of complexity in the code without adding anything? I'm only halfway through this PR so I might not have seen it yet.

JNeiger (changes requested):
Only about halfway through, I'll probably need to do a few passes anyways.

soccer/planning/planner/LineKickPlanner.cpp
JNeiger: lol, it barely even used to work on slow moving balls

soccer/CMakeLists.txt
oswinso: Im not sure why you would add O3 to DEBUG instead of just using RELWITHDEBINFO

soccer/planning/trajectory/Trajectory.cpp
```diff
@@ -127,6 +133,44 @@ std::optional<RobotInstant> Trajectory::EvaluateTime(RJ::Time time) const {
...
+
+    // This code disregards obstacles which the robot starts in. This allows the
+    // robot to move out a obstacle if it is already in one.
+    std::set<std::shared_ptr<Shape>> startHitSet = obstacles.hitSet(instants_[start].pose.position());
+
+    for (size_t i = start; i < instants_.size() - 1; i++) {
+        std::set<std::shared_ptr<Shape>> newHitSet = obstacles.hitSet(Segment(
+                instants_[i].pose.position(), instants_[i + 1].pose.position()));
+        if (!newHitSet.empty()) {
+            for (std::shared_ptr<Shape> hit : newHitSet) {
```
oswinso: I think this is an unintended copy?

oswinso:
is there documentation somewhere about what each planner is supposed to do?

JNeiger:
I don't think so, unless there is some paragraph at the top of the header. 

PathTargetPlanner should be the standard plan to this point with this final velocity. Pretty bog standard.

SettlePlanner gets in front of the ball and slows it down to a reasonable capture possibility. Think first touch in soccer.

CollectPlanner grabs a slow moving or stopped ball to control it for whatever we want to do next. This actually puts the ball on the dribbler and gives us control when it's done running.

LineKickPlanner moves through the ball and kicks on break beam. Think of it like foot ball where you have to take a few steps before you actually kick the ball to get more force behind it. It's good for free kicks because it helps stop against double touches and is more consistent in actually getting movement on the ball. 

PivotPathPlanner pivots around the ball. The ball should stay in the same position on the field, with a back spin towards the robot, as the robot circles around the ball. This is used for pivot kick

soccer/planning/PlannerNode.cpp
kylestach: Yeah, I think you're right. I'd be fine with having a `vector<vector<unique_ptr<Planner>>>`

kylestach:
Also, we should really consider using e.g. https://github.com/brendangregg/FlameGraph to try and diagnose where the slowdowns are. You're adding a somewhat more complex approach to some parts but it really should have similar performance, unless you're just sampling a ton more.

JNeiger:
I have had success profiling custom representative unit tests with callgrind to figure out where slowdowns are

oswinso:
I guess my point was that this information should be somewhere, either documented in the header files or like some other doc

kylestach:
ss
Flamegraph attached, zoomed into planning stack only (I had to turn all optimizations off to get it, so it's not 100% representative). Not much unexpected here, but it is worth looking at. Basically, settling is taking up most of the time.

oswinso:
I don't think it's really representative if optimizations are all turned off...

kylestach:
@oswinso you're right that it's not representative of absolute timing at all. It's obviously impossible to profile specific function calls when inlining is a thing, though, and the first thing the optimizer is going to do is heavily inline. I can mess around with settings to see if I can get useful data from it with optimizations on, I might be able to just slap `__attribute__((noinline))` on a few key methods if we want to look at particular targets.

oswinso:
it inlines that much? I wouldn't expect everything to get inlined though?

soccer/planning/planner/CollectPlanner.hpp
```diff
@@ -0,0 +1,75 @@
...
+    Trajectory course(PlanRequest&& request);
+    // approach the ball at constant velocity until contact
+    Trajectory fine(PlanRequest&& request);
+    // stop without losing the ball
+    Trajectory control(PlanRequest&& request);
+
+    void processStateTransitions(const PlanRequest& request);
+
+    // find the direction the robot will approach the ball (as a unit vector)
+    Geometry2d::Point findApproachDirection(const PlanRequest& request) const {
```
JNeiger, line 28: I would throw this in the cpp instead of the hpp

soccer/planning/planner/CollectPlanner.hpp
JNeiger: lmao, that was probably me

JNeiger: Wait, make pretty fucks with comments? https://github.com/RoboJackets/robocup-software/commit/faa1cc07bae00c9d562e4f59e2cd89b8cf94f376

soccer/planning/planner/InterceptPlanner.cpp
```diff
@@ -0,0 +1,57 @@
...
+    // a lower target velocity is safer because we are more likely to stay there
+    Trajectory path{{}};
+    constexpr int interpolations = 21;
+    for (int i = 0; i < interpolations; i++) {
+        double percent = static_cast<double>(i) / (interpolations - 1);
+        targetInstant.velocity.linear() =
+            percent * request.constraints.mot.maxSpeed * botToTarget.norm();
+        path = CreatePath::simple(startInstant, targetInstant,
+                                  request.constraints.mot);
+        PlanAngles(path, startInstant, AngleFns::facePoint(ball.pos),
```
JNeiger, line 48: What happens when you tell the path planner to plan to an invalid velocity target? AKA the target is .1 meters away and we want to be at a 5 m/s velocity when we get there.

soccer/planning/planner/InterceptPlanner.cpp
```diff
@@ -0,0 +1,57 @@
...
+        path.setDebugText("AtPoint");
+        return std::move(path);
+    }
+
+    // Find the minimum viable target velocity using Brute Force
+    // a lower target velocity is safer because we are more likely to stay there
+    Trajectory path{{}};
+    constexpr int interpolations = 21;
+    for (int i = 0; i < interpolations; i++) {
+        double percent = static_cast<double>(i) / (interpolations - 1);
```
JNeiger, line 43: You can just make the for loop `<= interpolations` so you don't have to do this subtraction. `interpolations` would be `20`. This way it's more obvious that this is every `0.05%`

soccer/planning/planner/SettlePlanner.cpp
```diff
@@ -0,0 +1,177 @@
...
+    // Shortcut Path: if we are almost done, move directly toward the ball line
+    // This prevents the planner from suddenly changing the target just before
+    // we get there because it thinks we won't make it in time.
+    Point closestPt = ballLine.nearestPoint(startInstant.pose.position());
+    bool inFrontOfBall = ball.vel.dot(closestPt - ball.pos) > 0;
+    bool withinShortcut =
+        startInstant.pose.position().distTo(closestPt) <
+        *_shortcutDist;  // && closestPt.distTo(targetPose.position()) <
+                         // *_shortcutDist;
+    if (inFrontOfBall && withinShortcut) {
```
JNeiger, line 118: Just to confirm, this also makes sure that if you are very close to the intercept line in terms of distance but far away from the ball, you will not shortcut but instead move towards the ball to intercept sooner? I can describe in pictures if that helps

JNeiger: The last two lines of this comment basically https://github.com/RoboJackets/robocup-software/blob/staging/soccer/planning/SettlePathPlanner.cpp#L350

JNeiger (commented):
Feel free to rip tons of comments from the source material. Especially for the planners themselves. 

As an aside, if you replan a path every frame, will the robot actually make progress along it? See #1184. A lot of hacks were built due to this and if its not solved, those hacks will have to be reinserted into the planners.

kylestach:
@JNeiger this keeps the old replanning behavior.

soccer/planning/planner/InterceptPlanner.cpp
ghost: the Velocity Profile handles that case. it basically just clamps the final velocity to values that will work.

soccer/planning/planner/SettlePlanner.cpp
ghost: Ok. I see what you mean.

JNeiger: So this is actually the "opposite" case. If I have a receiver at my goal, and the kicker at their goal. During the pass, assume the ball goes directly to the receiver. In this case, the receiver will be in front of the ball and within shortcut range, so they will sit there at the goal line waiting for the ball. But the ball must travel the entire field. In most every case, there is a better intercept position closer to the half line that will allow you to intercept the ball in less time. The optimal behavior in my head was to cheat forward and intercept at the half line instead of waiting at the goal line for the ball to arrive. The last two lines in the comment refer to this situation. We only want to shortcut if the target point is at or behind us. (Behind refers to intercepting at a later point in time)

JNeiger: Yeah, It's a weird case, but I saw it a lot with passes

soccer/planning/planner/InterceptPlanner.cpp
```diff
@@ -0,0 +1,57 @@
...
+        path = CreatePath::simple(startInstant, targetInstant,
+                                  request.constraints.mot);
+        PlanAngles(path, startInstant, AngleFns::facePoint(ball.pos),
+                   request.constraints.rot);
+        if (path.duration() <= ballToPointTime) {
+            break;
+        }
+    }
+    if (path.empty()) return reuse(std::move(request));
+    return std::move(path);
```
JNeiger, line 55: I would add the debug text back in about giving up. It makes it clearer when reviewing logs to see who's "fault" it is that we couldn't block a shot

soccer/planning/planner/SettlePlanner.cpp
```diff
@@ -0,0 +1,177 @@
...
+    _shortcutDist = new ConfigDouble(cfg, "Capture/Settle/shortcutDist");
+}
+using namespace Geometry2d;
+// todo(motion planning) find a better way to handle moving targets
+RJ::Time SettlePlanner::bruteForceIntercept(const PlanRequest& request) {
+    Ball& ball = request.context->state.ball;
+    std::vector<double> dists;
+    double endDist = *_searchEndDist;
+    const Rect& fieldRect = Field_Dimensions::Current_Dimensions.FieldRect();
+    Segment ballLine{ball.pos, ball.pos + ball.vel.normalized(endDist)};
```
JNeiger, line 36: This ball velocity direction is actually fairly noisy. You may need to filter this. I want to say it was causing a lot of replans which would ruin the smoothness of a normal path. You'll have to see how it looks on the real field. Depending on your grsim config, the friction is either perfect for dribbling or passing, not both.

JNeiger: The actual velocity is fairly noisy as well. #1233 has some graphs of velocity vs time as I roll the ball across the field. This will make the target point fairly noisy because of time estimate.

soccer/planning/planner/CollectPlanner.cpp
```diff
@@ -0,0 +1,139 @@
...
+Trajectory CollectPlanner::course(PlanRequest&& request) {
+    RobotInstant startInstant = request.start;
+    RotationConstraints rotationConstraints = request.constraints.rot;
+    const Ball& ball = request.context->state.ball;
+    Point approachDir = findApproachDirection(request);
+    Point noisyTargetPoint =
+        ball.pos - approachDir * (Ball_Radius + Robot_MouthRadius +
+                                  *_bufferDistBeforeContact);
+    Point noisyTargetVel =
+        approachDir * (ball.vel.dot(approachDir) + *_touchDeltaSpeed);
```
JNeiger, line 54: How is this with a slow moving ball perpendicular to the approach direction?

JNeiger, line 54: I was using a ball velocity average on the original code. I think if the touchDeltaSpeed was low enough, you could see the noise being the main determination of approach velocity. This will be something you have to test on a real field.

soccer/planning/planner/CollectPlanner.cpp
```diff
@@ -0,0 +1,139 @@
...
+    RobotInstant targetInstant{Pose{targetPoint, 0}, Twist{targetVel, 0},
+                               RJ::Time{0s}};
+    request.motionCommand = PathTargetCommand{targetInstant};
+    Trajectory path = _pathTargetPlanner.plan(std::move(request));
+    PlanAngles(path, startInstant, AngleFns::faceAngle(approachDir.angle()),
+               rotationConstraints);
+    return std::move(path);
+}
+Trajectory CollectPlanner::control(PlanRequest&& request) {
+    Point approachDir = findApproachDirection(request);
```
JNeiger, line 88: If there is any increase in ball velocity due to touching it, you may get runaway ball behavior because you try to match it's speed. I had to make it only plan once to get around this.

soccer/planning/planner/CollectPlanner.cpp
```diff
@@ -0,0 +1,139 @@
...
+    RotationConstraints rotationConstraints = request.constraints.rot;
+    RobotInstant startInstant = request.start;
+    // move with constant negative acceleration until stop
+    double accel =
+        request.constraints.mot.maxAcceleration * *_controlAccelScalePercent;
+    double stoppingDist =
+        std::pow(startInstant.velocity.linear().mag(), 2) / (2 * accel);
+    Point targetPoint =
+        startInstant.pose.position() + stoppingDist * approachDir;
```
JNeiger, line 97: You may need to add a config to extend this path slightly. If motion control is under-tuned (over-tuned?), it's possible this will not hit the ball due to stopping too soon.

JNeiger: It also allowed for a few initial bounces on the ball off the dribbler and a way to tune to that

soccer/planning/planner/CollectPlanner.cpp
```diff
@@ -0,0 +1,139 @@
...
+    RobotInstant targetInstant{Pose{targetPoint, 0}, {}, RJ::Time{0s}};
+    request.motionCommand = PathTargetCommand{targetInstant};
+    request.constraints.mot.maxSpeed = std::min(
+        startInstant.velocity.linear().mag(), request.constraints.mot.maxSpeed);
+    request.constraints.mot.maxAcceleration = accel;
+    Trajectory path = _pathTargetPlanner.plan(std::move(request));
+    PlanAngles(path, startInstant, AngleFns::faceAngle(approachDir.angle()),
```
JNeiger, line 104: If the robot covers the ball during capture, the ball position will not update and will end up inside the robot. At that point, doing the subtraction will result in a direction behind the robot, resulting in a rapid spin that will kick the ball out

soccer/planning/planner/CollectPlanner.cpp
```diff
@@ -0,0 +1,139 @@
...
+    if (collectState == CollectState::Fine) {
+        if (distToBall < *_distCutoffToControl + distToBallAtContact) {
+            collectState = CollectState::Control;
+        }
+    }
+    // backward transitions
+    bool fineToCourse =
+        collectState == CollectState::Fine &&
+        distToBall > distCutoffToFine + distToBallAtContact + 0.2;
+    bool controlToCourse =
```
JNeiger, line 131: If you have access to robot heading, a good check would be to see if the ball is not in the mouth or inside the robot. If the robot misses on the final approach, it's possible that the distance trigger will not happen. (Make sure to account for our robot blocking vision updates as well since that's a significant cause of problems close up

JNeiger (commented):
Settle/Collect/Intercept comments based on old ball behavior and hacks needed to get it working

soccer/planning/trajectory/RRTUtil.cpp
ghost: I mean, this is super hacky anyways, it should really be handling dynamic obstacles while planning so we don't need a loop here, but hey if it works it works...

soccer/Robot.hpp
```diff
@@ -439,6 +412,13 @@ class OurRobot : public Robot {
     const std::unique_ptr<Planning::MotionCommand>& motionCommand() const {
         return intent().motion_command;
     }
+    void setMotionCommand(std::unique_ptr<Planning::MotionCommand> newCmd) {
```
ashaw596, line 415: The unique_ptr should probably be removed. The variants are all tiny I think

ghost:
Closing this for now since its sorta dead. we can reopen later if needed

Issue #1431, opened 2020-04-06
Isolating Protobuf


HussainGynai:
## Description
Want to use protobuf as little as possible
## Requirements
 - Moving Processor configs to Context
 - Moving RadioRx and RadioTx to structs
 - Splitting out logging functionality (all protobuf-related code) to Logger

PR #1432, opened 2020-04-06
Move Processor Configs to Context

HussainGynai:
## Description
Created a new struct called gameSettings for context, that holds a bunch of basic data. Switched from using logFrame to context wherever necessary in code base. 

## Associated Issue
Issue #1431

### DIFF
--- a/soccer/GameSettings.hpp
+++ b/soccer/GameSettings.hpp
@@ -1,0 +1,40 @@
+struct GameSettings {
+    
+    bool is_simulation;
+
+    bool useOurHalf, useOpponentHalf;
+
+    // True if we are running with a simulator.
+    // This changes network communications.
+    bool simulation;
+
+    // True if we are blue.
+    // False if we are yellow.
+    bool blueTeam;
+
+    // _teamTrans is used for positions, not angles.
+    // _teamAngle is used for angles.
+    float teamAngle;
+
+    // Board ID of the robot to manually control or -1 if none
+    int manualID;
+    // Use multiple joysticks at once
+    bool multipleManual;
+
+    /// Measured framerate
+    float framerate;
+
+    // joystick damping
+    bool dampedRotation;
+    bool dampedTranslation;
+
+    bool kickOnBreakBeam;
+
+    // If true, rotates robot commands from the joystick based on its
+    // orientation on the field
+    bool useFieldOrientedManualDrive = false;
+
+    bool initialized;
+
+    bool paused;
+};
--- a/soccer/GameSettings.hpp
+++ b/soccer/GameSettings.hpp
@@ -36,4 +36,6 @@ struct GameSettings {
     bool initialized;
 
     bool paused;
+
+    int visionChannel;
 };
--- a/soccer/GameSettings.hpp
+++ b/soccer/GameSettings.hpp
@@ -1,6 +1,4 @@
 struct GameSettings {
-    
-    bool is_simulation;
 
     bool useOurHalf, useOpponentHalf;
 
@@ -10,14 +8,15 @@ struct GameSettings {
 
     // True if we are blue.
     // False if we are yellow.
-    bool blueTeam;
+    bool requestedBlueTeam;
 
-    // _teamTrans is used for positions, not angles.
     // _teamAngle is used for angles.
     float teamAngle;
 
     // Board ID of the robot to manually control or -1 if none
     int manualID;
+    //Board ID of the goalie robot
+    int goalieID;
     // Use multiple joysticks at once
     bool multipleManual;
 
--- a/soccer/Processor.cpp
+++ b/soccer/Processor.cpp
@@ -26,6 +26,7 @@
 #include "radio/PacketConvert.hpp"
 #include "radio/RadioNode.hpp"
 #include "vision/VisionFilter.hpp"
+#include "Network.hpp"
 
 REGISTER_CONFIGURABLE(Processor)
 
@@ -64,23 +65,24 @@ void Processor::createConfiguration(Configuration* cfg) {
 Processor::Processor(bool sim, bool defendPlus, VisionChannel visionChannel,
                      bool blueTeam, std::string readLogFile = "")
     : _loopMutex(), _readLogFile(readLogFile) {
-    _context.game_settings.blueTeam = blueTeam;
     _running = true;
-    _context.game_settings.manualID = -1;
-    _context.game_settings.framerate = 0;
     _context.game_settings.useOurHalf = true;
     _context.game_settings.useOpponentHalf = true;
-    _context.game_settings.initialized = false;
     _context.game_settings.simulation = sim;
-    _radio = nullptr;
-
+    _context.game_settings.requestedBlueTeam = blueTeam;
+    _context.game_settings.manualID = -1;
     _context.game_settings.multipleManual = false;
     setupJoysticks();
-
-    _context.game_settings.dampedTranslation = true;
+    _context.game_settings.framerate = 0;
     _context.game_settings.dampedRotation = true;
-
+    _context.game_settings.dampedTranslation = true;
     _context.game_settings.kickOnBreakBeam = false;
+    _context.game_settings.initialized = false;
+    _radio = nullptr;
+    _defendPlus = defendPlus;
+
+
+
 
     // Configuration-time variables.
     _context.robot_config = std::move(robot_config_init);
@@ -90,12 +92,7 @@ Processor::Processor(bool sim, bool defendPlus, VisionChannel visionChannel,
         robot_status_init.pop_back();
     }
 
-    // Initialize team-space transformation
-    defendPlusX(defendPlus);
-
     QMetaObject::connectSlotsByName(this);
-
-    _context.game_settings.is_simulation = _context.game_settings.simulation;
 
     _context.field_dimensions = *currentDimensions;
 
@@ -109,7 +106,7 @@ Processor::Processor(bool sim, bool defendPlus, VisionChannel visionChannel,
     _motionControl = std::make_unique<MotionControlNode>(&_context);
     _radio = std::make_unique<RadioNode>(&_context,
                                          _context.game_settings.simulation,
-                                         _context.game_settings.blueTeam);
+                                         _context.game_settings.requestedBlueTeam);
     _visionReceiver = std::make_unique<VisionReceiver>(
         &_context, sim, sim ? SimVisionPort : SharedVisionPortSinglePrimary);
     _grSimCom = std::make_unique<GrSimCommunicator>(&_context);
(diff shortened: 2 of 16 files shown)


soccer/GameSettings.hpp
```diff
@@ -0,0 +1,40 @@
...
+    // True if we are running with a simulator.
+    // This changes network communications.
+    bool simulation;
+
+    // True if we are blue.
+    // False if we are yellow.
+    bool blueTeam;
+
+    // _teamTrans is used for positions, not angles.
+    // _teamAngle is used for angles.
```
kylestach, line 13: Fix these names in the comments. Also, where did teamTrans go?

soccer/GameSettings.hpp
```diff
@@ -0,0 +1,40 @@
+struct GameSettings {
+    
+    bool is_simulation;
+
+    bool useOurHalf, useOpponentHalf;
+
+    // True if we are running with a simulator.
+    // This changes network communications.
+    bool simulation;
```
kylestach, line 7: We don't need the above `is_simulation` now...

soccer/GameSettings.hpp
```diff
@@ -0,0 +1,40 @@
...
+
+    // True if we are blue.
+    // False if we are yellow.
+    bool blueTeam;
```
kylestach: This is duplicated from `GameState`: https://github.com/RoboJackets/robocup-software/blob/staging/soccer/GameState.hpp#L56.

Actually, I suppose they aren't quite duplicated: this one is the _requested_ team (from the soccer program), and the one in `GameState` is from the referee. Probably should rename this accordingly, like `requestedBlueTeam` or something.

soccer/Processor.cpp
```diff
@@ -150,14 +153,16 @@ void Processor::stop() {
...
         joy->reset();
     }
 }
 
-void Processor::multipleManual(bool value) { _multipleManual = value; }
+void Processor::multipleManual(bool value) {
+    _context.game_settings.multipleManual = value;
+}
 
 void Processor::goalieID(int value) {
```
kylestach: Goalie ID should be a field in this struct as well, that gameplay is allowed to read. You'll need to add it to the Python bindings most likely:
 - Delete this function: https://github.com/RoboJackets/robocup-software/blob/staging/soccer/gameplay/GameplayModule.cpp#L278
 - In the Python root play's spin method (probably this one: https://github.com/RoboJackets/robocup-software/blob/staging/soccer/gameplay/root_play.py#L32), update the goalie ID accordingly.

soccer/Processor.cpp
```diff
@@ -171,17 +176,17 @@ int Processor::goalieID() {
...
     QMutexLocker locker(&_loopMutex);
-    _dampedTranslation = value;
+    _context.game_settings.dampedTranslation = value;
 }
 
 void Processor::joystickKickOnBreakBeam(bool value) {
     QMutexLocker locker(&_loopMutex);
-    _kickOnBreakBeam = value;
+    _context.game_settings.kickOnBreakBeam = value;
 }
```
kylestach: All of these functions (`joystickKickOnBreakBeam`, `dampedTranslation`, `dampedRotation`, `goalieID`, `manualID`) should be deleted in favor of directly writing to the `game_settings` struct in `Context`. However, you'll need to take care to lock the mutex when you access it. I'd suggest giving both `MainWindow` a pointer to `_loopMutex`, having it keep its own internal copy of the game settings, and then at the end of its update method (https://github.com/RoboJackets/robocup-software/blob/staging/soccer/ui/MainWindow.cpp#L307) locking the mutex and actually writing that copy into context (make sure you don't hold on to the mutex for very long, though). Alternatively, you could just lock the mutex where it's used.

soccer/Processor.cpp
```diff
@@ -207,9 +212,9 @@ void Processor::blueTeam(bool value) {
     // This is called from the GUI thread
     QMutexLocker locker(&_loopMutex);
 
-    if (_blueTeam != value) {
-        _blueTeam = value;
-        if (_radio) _radio->switchTeam(_blueTeam);
+    if (_context.game_settings.blueTeam != value) {
+        _context.game_settings.blueTeam = value;
```
kylestach: This will be incorrect when you are overridden by the referee, it should be checking `game_state.blueTeam`. (See my comment above about renaming `blueTeam`). This applies to basically everything in this section.

soccer/Processor.cpp
```diff
@@ -735,9 +747,9 @@ void Processor::defendPlusX(bool value) {
     _context.game_state.defendPlusX = value;
 
     if (value) {
-        _teamAngle = -M_PI_2;
+        _context.game_settings.teamAngle = -M_PI_2;
     } else {
-        _teamAngle = M_PI_2;
+        _context.game_settings.teamAngle = M_PI_2;
     }
```
kylestach: Same as the above functions. It should be okay to just always recalculate the world-to-team transform (in `Processor::run`'s main loop), and then `defendPlusX`, `changeVisionChannel`, etc. functions can be deleted (for vision channel you'll need to check against the previous vision port and then call `_visionReceiver->setPort(...)` if it's changed.)

soccer/Processor.cpp
```diff
@@ -758,7 +770,8 @@ void Processor::changeVisionChannel(int port) {
 void Processor::recalculateWorldToTeamTransform() {
     _worldToTeam = Geometry2d::TransformMatrix::translate(
         0, Field_Dimensions::Current_Dimensions.Length() / 2.0f);
-    _worldToTeam *= Geometry2d::TransformMatrix::rotate(_teamAngle);
+    _worldToTeam *=
+        Geometry2d::TransformMatrix::rotate(_context.game_settings.teamAngle);
 }
 
 void Processor::setFieldDimensions(const Field_Dimensions& dims) {
```
kylestach, line 734: Field dimensions should really be a part of `Context` as well, at which point my advice above about deleting this function and having `MainWindow` set it in the game state applies again.

soccer/Processor.hpp
```diff
@@ -126,7 +128,7 @@ class Processor : public QThread {
     std::vector<int> getJoystickRobotIds();
 
     void blueTeam(bool value);
-    bool blueTeam() const { return _blueTeam; }
+    bool blueTeam() const { return _context.game_settings.blueTeam; }
```
kylestach: Delete all of these setters, per the comment above (the getters are fine if you really need them, but I don't think we do).

soccer/Processor.hpp
```diff
@@ -155,9 +157,11 @@ class Processor : public QThread {
     void closeLog() { _logger.close(); }
 
     // Use all/part of the field
-    void useOurHalf(bool value) { _useOurHalf = value; }
+    void useOurHalf(bool value) { _context.game_settings.useOurHalf = value; }
 
-    void useOpponentHalf(bool value) { _useOpponentHalf = value; }
+    void useOpponentHalf(bool value) {
+        _context.game_settings.useOpponentHalf = value;
+    }
```
kylestach: More setters to delete.

soccer/Processor.hpp
```diff
@@ -175,7 +179,7 @@ class Processor : public QThread {
 
     bool isInitialized() const;
 
-    void setPaused(bool paused) { _paused = paused; }
+    void setPaused(bool paused) { _context.game_settings.paused = paused; }
```
kylestach: Delete.

soccer/VisionReceiver.cpp
```diff
@@ -158,9 +158,9 @@ void VisionReceiver::receivePacket(const boost::system::error_code& error,
 }
 
 bool VisionReceiver::shouldRemove(bool defendPlusX, double x) {
-    return (!_context->state.logFrame->use_our_half() &&
+    return (!_context->game_settings.useOurHalf &&
             ((defendPlusX && x < 0) || (!defendPlusX && x > 0))) ||
-           (!_context->state.logFrame->use_our_half() &&
+           (!_context->game_settings.useOurHalf &&
```
kylestach, line 163: :+1:

soccer/ui/FieldView.cpp
```diff
@@ -249,14 +249,14 @@ void FieldView::drawTeamSpace(QPainter& p) {
     }
 
     // Block off half the field
-    if (!frame->use_our_half()) {
```
kylestach: `FieldView` is the one place that should actually use the log frame instead of the context, because it is looking at the log (rather than current values). I think everything here can be reverted.

soccer/ui/SimFieldView.cpp
```diff
@@ -35,14 +35,14 @@ void SimFieldView::mousePressEvent(QMouseEvent* me) {
         for (const LogFrame::Robot& r : frame->self()) {
             if (pos.nearPoint(r.pos(), Robot_Radius)) {
                 _dragRobot = r.shell();
-                _dragRobotBlue = frame->blue_team();
+                _dragRobotBlue = _context->game_settings.blueTeam;
```
kylestach: Again, these should be based on the log frame not the context. Although, I'm not sure what the hell is supposed to happen when we drag a robot when we're at a previous point in the log.

kylestach (changes requested):
Great progress! I think the biggest thing that needs to be fixed here is that you're still just using `Processor` as a shim to access these structs from `MainWindow`, which kind of defeats the point - your `MainWindow` still needs a pointer to `Processor`; ideally we would be able to delete that (maybe we would still need it temporarily for play/testbooks, but ideally that would be all, and those would be fixed in a later PR). Basically, look through all of the usages of `_processor` in `MainWindow` and try to replace them with accesses to context.

soccer/GameSettings.hpp
JNeiger: If you could throw in a description of what it actually is, that'd be awesome too

soccer/GameSettings.hpp
```diff
@@ -0,0 +1,40 @@
...
+
+    // _teamTrans is used for positions, not angles.
+    // _teamAngle is used for angles.
+    float teamAngle;
+
+    // Board ID of the robot to manually control or -1 if none
+    int manualID;
```
JNeiger, line 17: With multiple manual enabled, how does this work?

Something like this also exists in context already

kylestach: As far as I remember multiple manual is broken. Even if it's not it should get an overhaul. Regardless, this is what it was like previously, so I think fixing multiple manual stuff is beyond the scope of this PR.

JNeiger: I was moreso making sure that this wasn't a straight duplicate of what's in context already

kylestach: Oh, right. https://github.com/RoboJackets/robocup-software/blob/staging/soccer/Context.hpp#L34.

I think probably the best thing to do would be to move the `is_joystick_controlled` array into `GameSettings` and then set it in `MainWindow` (or maybe make it a `std::array<int, 16>` where it's -1 for not joystick controlled and the joystick ID otherwise). That would mean we could get rid of all of the multiple manual code and deal with that later.

soccer/GameSettings.hpp
HussainGynai: I don't know what teamTrans is I just copied the comments over from Processor.hpp  and there was no _teamTrans there

kylestach: https://github.com/RoboJackets/robocup-software/blob/staging/soccer/Processor.hpp#L240

It's actually _worldToTeam

HussainGynai: So should I move that to GameSettings too?

kylestach: Yes.

soccer/GameSettings.hpp
kylestach: @HussainGynai ping on this. Again, I think the best thing to do would be a `std::array<int, 16> robot_joystick_ids` in `GameSettings`, with `-1` corresponding to "no joystick controller".

soccer/Processor.hpp
```diff
@@ -93,16 +93,18 @@ class Processor : public QThread {
     }
 
     void manualID(int value);
-    int manualID() const { return _manualID; }
+    int manualID() const { return _context.game_settings.manualID; }
```
kylestach, line 96: You should be able to delete these, right?

Either way, you should _definitely_ get rid of the setters' declarations in here.

soccer/gameplay/GameplayModule.hpp
```diff
@@ -72,7 +72,7 @@ class GameplayModule {
     bool checkPlaybookStatus();
 
     void goalieID(int value);
-    int goalieID() { return _goalieID; }
+    int goalieID() { return _context->game_settings.goalieID; }
```
kylestach, line 75: Do you actually need this (either one) anymore?

soccer/ui/MainWindow.hpp
```diff
@@ -251,6 +251,15 @@ private Q_SLOTS:
...
     QLabel* _logMemory;
 
+    // Locked when processing loop stuff is happening (not when blocked for
+    // timing or I/O). This is public so the GUI thread can lock it to access
+    // SystemState, etc.
+    QMutex _loopMutex;
+
+    Context* _context;
+
+    GameSettings _settings_copy{true, true, true, true, -1, -1, false, 0, true, true, false, true, false};
```
kylestach, line 261: You should probably just set defaults in `GameSettings` and default-initialize this.

kylestach (changes requested):

kylestach:
Closing in favor of #1441

Issue #1338, opened 2019-09-22
Multiple robots assigned to capture the ball at the same time


liquidmetal9015:
## Description
Submissive defender will enter the clearing state, while at the same time an midfielder will attempt to pivot kick the ball either for a coordinated pass or a shot on goal.

This issue was brought up in [the September 15th Virtual Test](https://docs.google.com/document/d/1YEqDBBlug9qEPL_0H8Gkff8k3KdWfy8u4cjPsxDtlOM/edit) and the subsequent [issue creation meeting](https://docs.google.com/document/d/1rmvEsIQpT6IzLGL4BVeG0XDHLTS-O3TTEezgHVygMy4/edit).

This has been tested in adaptive formation, but should work in any play that uses submissive defender. There was a bandied fix for this at some point in the past, possibly just for basic122.

The method used to solve this has been up for discussion but it comes down to there being no communication between behaviors at the same level of the tree. Having a skill "tag" or property system that indicated ball interaction was floated but devising a new feature to overcome this limitation may be part of the solution to this issue.

The issue can be seen below where robots 3 and 5 are both running capture at the same time.

![MultipleBotsTryingToCapture](https://user-images.githubusercontent.com/8135314/65393092-83cbca00-dd4a-11e9-8000-625333353b28.png)

## Steps to reproduce
1. Place ball near our goal
2. Watch multiple robots try to capture at the same time, it will only really be evident if the robots are having trouble and/or are equidistant.

Result: Robots run into each other trying to capture the ball. Two robots will have capture as their current skill in the tree.

- [ ] This bug has been reproduced by someone else

## Additional Information

DefensiveForward has also been spotted colliding with submissive defender

![DefensiveForwardCaptureCollision](https://user-images.githubusercontent.com/8135314/65393209-eb364980-dd4b-11e9-873a-a5e75759ecb7.png)

liquidmetal9015:
This issue does kind of need a bump as it continues to come up in testing, such as the November 10th testing session where default defense and defensive forward are both going for the ball at the same time.

![DefenceAndDefenceGoingForBall](https://user-images.githubusercontent.com/8135314/69198257-1a75f680-0b02-11ea-90fa-ef833807d0d4.png)


For the right team right team in the below image.

![MarkingInCloseProximityAndMutltipleGoingForTheBall](https://user-images.githubusercontent.com/8135314/69198562-18606780-0b03-11ea-8b5a-c060f719d6ce.png)

PR #1434, opened 2020-04-09
Neutralize default defense

liquidmetal9015:
## Description
The standard defense that we use is incompatible with our current systems on a number of levels. This PR removes the clearing functionality from defense.py and renames it coordinated_block.py to better reflect what it does, and removes the standard defense code from standard_play.py

There are some places that probably should be using coordinated block but aren't, but it's a simple matter of adding the sub behavior and setting the handles goalie function to true.

## Associated Issue
The critical issue #1338 should be pretty much fixed in it's current incarnation although not entirely.

### Test Case 1
1. Run the altered plays
2. See if it works

Expected result: Things generally work and coordinated_block is only activated when called


liquidmetal9015:
So my two computers disagree about style, and it appears that CI disagrees with both of them.

soccer/gameplay/standard_play.py
```diff
@@ -18,7 +18,7 @@ class StandardPlay(play.Play):
     # Not just on selection
     def __init__(self, continuous):
         super().__init__(continuous)
```
JNeiger, line 21: The only reason standard play exists is to setup the defense. It might be a good idea in the future or now to full remove standard play and inherit from play.Play.

soccer/gameplay/tactics/defense_old.py
```diff
@@ -1,535 +0,0 @@
-import composite_behavior
```
JNeiger, line 1: Good night sweet prince

JNeiger (approved):
Looks good! I don't have the code with me, but at one point the legacy basic 122 and legacy adaptive had code to coordinate who is grabbing the ball between defense and the higher level play. If those still exist in the new situation system, we should make sure those hooks do not exist anymore

HussainGynai (approved):
Looks good, I will create an issue for removing standard play, like Joe suggested

oswinso:
Make sure to merge in changes from staging

PR #1436, opened 2020-05-06
Replace Qt libs with std::/boost::.

kylestach:
Still need to do QString but that's a battle for another day.

## Description
This PR gets rid of all instances of `QUdpSocket`, `QThread` and `QMutex`.

## Associated Issue
N/A. This is a step towards #1432.

## Design Documents
N/A

### Basic soccer regression testing
1. Open soccer in sim mode. Make sure ball/robot placement still works.

Expected result: no regressions

### DIFF
--- a/soccer/Node.hpp
+++ b/soccer/Node.hpp
@@ -6,7 +6,7 @@
  */
 class Node {
 public:
-    virtual ~Node() {}
+    virtual ~Node() = default;
 
     virtual void start(){};
 
--- a/soccer/Node.hpp
+++ b/soccer/Node.hpp
@@ -6,6 +6,8 @@
  */
 class Node {
 public:
+    virtual ~Node() {}
+
     virtual void start(){};
 
     // The callback
--- a/soccer/Referee.hpp
+++ b/soccer/Referee.hpp
@@ -18,8 +18,6 @@
 #include "SystemState.hpp"
 #include "TeamInfo.hpp"
 
-class QUdpSocket;
-
 /**
  * @brief A packet we received over the network from ssl-refbox
  *
@@ -51,7 +49,7 @@ struct RefereePacket {
 class Referee : public Node {
 public:
     explicit Referee(Context* const context);
-    ~Referee();
+    ~Referee() override{};
 
     Referee(const Referee&) = delete;
     Referee& operator=(const Referee&) = delete;
@@ -60,7 +58,6 @@ public:
 
     void start() override;
     void run() override;
-    void stop() override;
 
     void getPackets(std::vector<RefereePacket>& packets);
 
--- a/soccer/Referee.hpp
+++ b/soccer/Referee.hpp
@@ -49,7 +49,7 @@ struct RefereePacket {
 class Referee : public Node {
 public:
     explicit Referee(Context* const context);
-    ~Referee() override{};
+    ~Referee() override = default;
 
     Referee(const Referee&) = delete;
     Referee& operator=(const Referee&) = delete;
--- a/soccer/joystick/GamepadController.cpp
+++ b/soccer/joystick/GamepadController.cpp
@@ -41,7 +41,7 @@ GamepadController::GamepadController()
 }
 
 GamepadController::~GamepadController() {
-    QMutexLocker(&mutex());
+    auto lock = lock_mutex();
     SDL_GameControllerClose(_controller);
     _controller = nullptr;
     SDL_Quit();
@@ -103,7 +103,7 @@ void GamepadController::closeJoystick() {
 bool GamepadController::valid() const { return connected; }
 
 void GamepadController::update() {
-    QMutexLocker(&mutex());
+    auto lock = lock_mutex();
     SDL_GameControllerUpdate();
 
     RJ::Time now = RJ::now();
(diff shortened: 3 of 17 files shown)


soccer/Referee.hpp
```diff
@@ -51,7 +49,7 @@ struct RefereePacket {
 class Referee : public Node {
 public:
     explicit Referee(Context* const context);
-    ~Referee();
+    ~Referee() override{};
```
oswinso: =default?

kylestach: Fixed.

JNeiger (approved):

soccer/radio/SimRadio.cpp
```diff
@@ -15,15 +15,22 @@
...
-    : _context(context), _blueTeam(blueTeam) {
-    switchTeam(blueTeam);
+    : _context(context),
+      _blueTeam(blueTeam),
+      _socket(_io_service, ip::udp::endpoint(ip::udp::v4(),
+                                             blueTeam ? SimBlueStatusPort
+                                                      : SimYellowStatusPort)) {
+    _grsim_endpoint = ip::udp::endpoint(ip::udp::v4(), SimCommandPort);
+
+    // TODO: Make sure the buffer is big enough.
```
EvanMorcom: Can we get an issue # for this?

EvanMorcom (commented):
LGTM :fire: :100:

soccer/radio/SimRadio.cpp
kylestach: Gah, left this in while I was working and forgot to delete it. The return packets from grSim are actually only a byte long in this version of the protocol (in later versions it's a proper protobuf, so a lot of this is about to be rewritten when we upgrade grSim versions).

soccer/Node.hpp
```diff
@@ -6,6 +6,8 @@
  */
 class Node {
 public:
+    virtual ~Node() {}
```
oswinso: default?

soccer/joystick/GamepadController.cpp
```diff
@@ -41,7 +41,7 @@ GamepadController::GamepadController()
 }
 
 GamepadController::~GamepadController() {
-    QMutexLocker(&mutex());
+    auto lock = lock_mutex();
```
oswinso, line 44: you should standardize how you're locking your mutexes...

soccer/joystick/Joystick.hpp
```diff
@@ -35,7 +34,7 @@ struct JoystickControlValues {
  */
 class Joystick : boost::noncopyable {
 public:
-    Joystick() : _mutex(QMutex::Recursive){};
+    Joystick(){};
```
oswinso: why not just remove the constructor then or have it as default?

soccer/radio/SimRadio.cpp
```diff
@@ -79,68 +85,80 @@ void SimRadio::send(Packet::RadioTx& radioTx) {
...
+    // 5: robot_on
+    const uint8_t robot_id_mask = 0x7;
+    const uint8_t touching_ball_mask = 0x1 << 3;
+    const uint8_t just_kicked_mask = 0x1 << 4;
+    const uint8_t robot_on_mask = 0x1 << 5;
+
+    int this_robot_id = data & robot_id_mask;
+    bool ball_sense = data & touching_ball_mask;
+    bool just_kicked = data & just_kicked_mask;
+    bool robot_on = data & robot_on_mask;
```
oswinso, line 127: for clarity I think you could separate the parsing logic out into a different function and return a struct that contains all the information that a packet could contain

soccer/radio/SimRadio.cpp
```diff
@@ -154,32 +172,29 @@ void SimRadio::stopRobots() {
...
+
     int status_port = blueTeam ? SimBlueStatusPort : SimYellowStatusPort;
-    if (!_rx_socket.bind(status_port)) {
-        throw runtime_error(
-            QString("Can't bind to the %1 team's radio status port.")
-                .arg(blueTeam ? "blue" : "yellow")
-                .toStdString());
-    }
+
+    std::cout << status_port << std::endl;
```
oswinso: Is a debug print?

soccer/radio/SimRadio.cpp
kylestach: This function is about to be rewritten (grSim update introduces a new protocol) so I don't really want to bother until it's actually being updated.

soccer/radio/SimRadio.cpp
kylestach: ...oof. ACK, sigh.

soccer/joystick/GamepadController.cpp
kylestach: What's non-standardized here? The mutex is actually a private member of the parent now, and I just gave it a protected getter for a lock instead of for the mutex itself.

soccer/joystick/Joystick.hpp
kylestach: ACK.

soccer/joystick/GamepadController.cpp
oswinso: as in like you're using a getter for a lock guard here, but in the other class you're creating the lock guard directly in the function. I mean standardize as in like maybe use either getters or construct the lock guard directly in both cases, so that it's standardized across the codebase

oswinso (approved):
lgtm

Issue #1325, opened 2019-09-01
Joystick Refactor


CollinAvidano:
## Description
Processor is a mess and really should not contain all the joystick code directly. I want to move the majority of joystick code out of processor and into a separate class (will probably be named manualmanager) that would hold all joystick and other manual input related state and functionality.

## Requirements
- No joystick-specific code in Processor
- All joystick logic is kept locally in something that is not Processor; that thing should be responsible for modifying packet commands
- Should support an unlimited number of joysticks being plugged in
- Should support any number of robots being manually controlled
- Should support extension via keyboards and other input methods

## Design
https://docs.google.com/document/d/1V0zQN19UZK7d2m-9Otd1NLVLHHrPHqdAuiDt_6dU6iw/edit?usp=sharing

kylestach:
Closed by #1461

PR #1441, opened 2020-05-10
Add GameSettings struct to Context

kylestach:
This allows several parts of MainWindow to stop relying on Processor
directly.

It also temporarily breaks multiple manual (if you can really consider that as "working" as it was). Joystck and GUI code is going to be rewritten anyway.

## Description
This moves most of the things that need to be written my `MainWindow` into a `GameSettings` struct that is read by `Processor`.

## Associated Issue
Issue #1329 

## Design Documents
[ROS Roadmap](https://docs.google.com/document/d/1wix-ZK9NPBY0zC8T9LyUihT1ONubXMKxHvb8KYmJkNQ/edit#heading=h.whcz8gaza9uh)

### Regression Testing
 - Basic start/halt/stop and restarts
 - Manual control (single joystick)
 - Orientation (defend +/- X)
 - Partial field (use opponent half/use our half)
 - Yellow and blue team
 - Log viewing

Expected result: No regressions


soccer/GameSettings.hpp
```diff
@@ -0,0 +1,57 @@
...
+    bool use_their_half = true;
+
+    bool paused = false;
+
+    // Joystick configuration
+    int manualID = -1;
+    bool dampedTranslation = true;
+    bool dampedRotation = true;
+    bool useKickOnBreakBeam = false;
+    bool useFieldOrientedDrive = false;
```
oswinso: break this out into another struct for clarity

soccer/ui/MainWindow.cpp
```diff
@@ -322,6 +306,15 @@ void MainWindow::updateViews() {
...
     }
 
+    GameState game_state;
+    {
+        std::lock_guard<std::mutex> lock(_context_mutex);
+        game_state = _context->game_state;
+    }
+    bool blueTeam = game_state.blueTeam;
+
+    // TODO(Kyle): Fix multiple manual
```
oswinso, line 316: issue number?

soccer/ui/qt/MainWindow.ui
```diff
@@ -2198,7 +2198,7 @@
           <attribute name="title">
            <string>Testing Tab</string>
           </attribute>
-          <layout class="QVBoxLayout" name="verticalLayout_12">
+          <layout class="QVBoxLayout" name="verticalLayout_14">
```
oswinso, line 2201: why these name changes?

oswinso (changes requested):
small changes

soccer/Referee.cpp
```diff
@@ -31,7 +31,6 @@ Referee::Referee(Context* const context)
       command_(HALT),
       sent_time{},
       received_time{},
-      stage_time_left{},
```
JNeiger, line 35: It defaults to 0, but other RJ:Times are in the constructor. We should probably be consistent

JNeiger (approved):

soccer/ui/qt/MainWindow.ui
kylestach: Oops? The warnings were bugging me (they're duplicated) so I fixed them and committed...

soccer/ui/MainWindow.cpp
kylestach: https://github.com/RoboJackets/robocup-software/issues/1325

soccer/GameSettings.hpp
```diff
@@ -0,0 +1,60 @@
...
+    // Whether or not we're in simulation.
+    bool simulation = true;
+
+    // Whether external referee is allowed.
+    // If this is set to true but no external referee is connected,
+    // none will be used and the control panel can be used as always.
+    bool use_external_referee = false;
+
+    // Requests. These can be overridden by the referee if it's enabled
+    bool requestBlueTeam = true;
```
garethellis0, line 21: not sure what the naming convention is for ya'll, but looks like `CamelCase` and `snake_case` are both being used for variable names here?

soccer/GameState.hpp
```diff
@@ -48,47 +51,52 @@ class GameState {
     int theirScore;
 
     // Time in seconds remaining in the current period
-    int secondsRemaining;
+    RJ::Seconds stage_time_left;
 
     TeamInfo OurInfo;
     TeamInfo TheirInfo;
```
garethellis0, line 57: Similar comment here, looks like we've got `snake_case`, `CamelCaseUpperFirst` and `camelCaseLowerFirst` all here?

There a few other places with similar issues, but just commenting once here to avoid comment spam.

soccer/Processor.cpp
```diff
@@ -599,43 +528,18 @@ void Processor::sendRadioData() {
         }
     }
 
+    // TODO(Kyle): Reimplement multiple manual code
```
garethellis0, line 531: Should we reference #1325 here?

soccer/Referee.cpp
```diff
@@ -278,9 +285,13 @@ GameState Referee::updateGameState(Command command) const {
...
 
-    const bool blue_team = _game_state.blueTeam;
+    bool blue_team = _game_state.blueTeam;
 
-    switch (command) {
+    if (!_isRefControlled) {
+        blue_team = _context->game_settings.requestBlueTeam;
+    }
+
+    switch (command_) {
```
garethellis0, line 294: bit of a nit, but looks like we're mixing prefixing/suffixing `_` on variables?

garethellis0, line 294: I'm not sure if this intentional or not, but seems like the `command` given to this function is never used. Instead we're using `command_`? Perhaps add `command_ = command` at the start of the function?

soccer/main.cpp
```diff
@@ -176,9 +158,14 @@ int main(int argc, char* argv[]) {
     std::shared_ptr<Configuration> config =
         Configuration::FromRegisteredConfigurables();
 
-    auto processor =
-        std::make_unique<Processor>(sim, defendPlus, visionChannel, blueTeam, readLogFile);
-    processor->refereeModule()->useExternalReferee(!noref);
+    auto processor = std::make_unique<Processor>(sim, blueTeam, readLogFile);
+
+    Context* context = processor->context();
```
garethellis0, line 163: Might want to consider making this a `shared_ptr` instead of a raw one?

soccer/Processor.cpp
```diff
@@ -314,18 +242,22 @@ void Processor::run() {
...
+        updateOrientation();
+
         // TODO(Kyle): Don't do this here.
         // Because not everything is on modules yet, but we still need things to
         // run in order, we can't just do everything via the for loop (yet).
         _visionReceiver->run();
 
         if (_context.field_dimensions != *currentDimensions) {
+            std::cout << "Updating field geometry based off of vision packet."
+                      << std::endl;
```
garethellis0, line 260: Not super sure what the logging structure is, but perhaps use `status` here?

soccer/ui/MainWindow.hpp
```diff
@@ -263,4 +249,7 @@ private Q_SLOTS:
     std::vector<QComboBox*> _robotConfigQComboBoxes{};
 
     std::vector<QComboBox*> _robotDebugResponseQComboBoxes{};
+
+    std::mutex _context_mutex;
+    Context* _context;
```
garethellis0, line 254: perhaps consider making this a `shared_ptr`?

soccer/ui/qt/MainWindow.ui
garethellis0: Maybe a good opportunity to give them slightly more descriptive names?

garethellis0 (commented):
Just dropping in from UBC, we chatted a bit in slack a while back about trading some reviews, so I figured I'd start it off! Obviously I'm not familiar with any conventions/standards/architecture that ya'll have/use, so just made some suggestions.

Feel free to slide into our PR comments! https://github.com/UBC-Thunderbots/Software/pulls

soccer/ui/MainWindow.hpp
kylestach: Raw pointers are still useful to represent non-owning pointers (depending on who you ask, you could also reasonably use a reference here). This is not an "a owns b" relationship, so it makes sense for it to be a regular raw pointer.

soccer/GameSettings.hpp
oswinso: yeah our variable naming scheme is overall a big mess right now, as different people have used different schemes over the years. We were planning on doing a full sweep of renaming sometime during our ROS refactor after we come to a decision on what naming scheme we're sticking to, but for now there's like basically no naming convention.

soccer/ui/MainWindow.hpp
garethellis0: Ah, makes sense. Have you considered `weak_ptr`? This seems like the textbook use-case for it (it definitely adds some overhead having to check that the pointer is valid, but in general it's good practice to check that a non-owning pointer is valid before using it).

oswinso: In this case though `Context` is something that basically is guaranteed to outlive `MainWindow` itself, so it doesn't _really_ matter that much I think.

kylestach: Yeah. Personally I consider using weak_ptr/shared_ptr when lifetimes can be statically determined to be a bit of an antipattern. In this case, Context is guaranteed to outlive all of the modules holding a pointer to it.

oswinso (approved):
lgtm

soccer/ui/MainWindow.hpp
garethellis0: Can it be statically (ie. compile-time) determined that `_context` outlives `MainWindow`? I suppose it can by looking at the code, but if that were ever to change, there'd certainly be no compile-time indication of it, and the only runtime indication would be random segfaults (if we're lucky) or the program somehow misbehaves (much worse).

Totally not here to tell ya'll how to handle this sort of stuff, just a suggestion. Especially for code for student teams I tend to be _very_ conservative in terms of making code as safe as possible.

kylestach: Context lives for the duration of the program (as does MainWindow, but it is destructed just before Context). IMO non-owning pointers only make sense to use in two situations:

 - The pointer is passed into a function, used in that function, and never stored anywhere.
 - The pointer points to an object that has whole-program lifetime (i.e. `'static` in Rust).

In these two situations, there's no ambiguity and so it's acceptable. In addition, IMO it is more clear than a reference (for example, it forces you to make it explicit that the function is allowed to mutate a given parameter). The reason I don't like `shared_`/`weak_ptr` is that they tend to act as a crutch to avoid having to think about ownership, which I've seen lead to messy messy spaghetti code (we've got a fair bit of that in this repository that we're currently trying to clean out). However, they are obviously very useful tools when you actually need any sort of dynamic lifetime.

PR #1442, opened 2020-05-10
Updated clang_format_6 to clang_format_10

oswinso:
## Description
`clang_format_6` is pretty retarded about c++17 attributes, ie. `[[nodiscard]]`:
```c++
    /**
     * Implicit conversion from Eigen::Vector3d
     */
    Pose(const Eigen::Vector3d& other)
        : _position(other(0), other(1)),
          _heading(other(2)){}

              /**
               * Compute the pose specified using this pose as coordinates in a
               * frame of reference specified by `other`, in the global space.
               *
               * i.e.
               *
               * x:
               *       |  x
               *       | /
               *       |/
               * ------|-------
               *       |
               *       |
               *       |
               *
               * y:
               *     |
               *     |
               * ----|----
               *     |\
               *     | y
               *
               * y.withOrigin(x):
               *       |  x
               *       | / \
               *       |/   y
               * ------|------
               *       |
               */
                  [[nodiscard]] Pose withOrigin(Pose other) const {
        Point rotated = position().rotated(other.heading());
        return other + Pose(rotated, heading());
    }
```

This PR:
1. Updates `make pretty-lines` and `make checkstyle-lines` to use `clang-format-9` instead of `clang-format` (which defaults to `clang-format-6` on 18.04)
2. Adds `clang-format-10` on `ubuntu-packages.txt`
3. Added llvm apt repository to the ubuntu setup script for `clang-format-10`

### `[[nodiscard]]` is formatted properly on CI
1. Let CI run

Expected result: The addition of `[[nodiscard]]` to `Pose.hpp` shouldn't screw up the comment formatting, and it shouldn't error out because it can't find `clang-format-9`.

### DIFF
--- a/util/ubuntu-packages.txt
+++ b/util/ubuntu-packages.txt
@@ -39,6 +39,7 @@ g++
 
 # C/C++ (and more) code-formatting tool
 clang-format
+clang-format-9
 
 # contains a ton of great C++ libraries
 libboost-all-dev
--- a/util/ubuntu-packages.txt
+++ b/util/ubuntu-packages.txt
@@ -38,7 +38,6 @@ clang
 g++
 
 # C/C++ (and more) code-formatting tool
-clang-format
 clang-format-10
 
 # contains a ton of great C++ libraries
--- a/util/ubuntu-packages.txt
+++ b/util/ubuntu-packages.txt
@@ -39,7 +39,7 @@ g++
 
 # C/C++ (and more) code-formatting tool
 clang-format
-clang-format-9
+clang-format-10
 
 # contains a ton of great C++ libraries
 libboost-all-dev
--- a/common/Geometry2d/Pose.hpp
+++ b/common/Geometry2d/Pose.hpp
@@ -69,7 +69,7 @@ public:
      * ------|------
      *       |
      */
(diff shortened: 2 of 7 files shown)


util/ubuntu-packages.txt
```diff
@@ -39,6 +39,7 @@ g++
 
 # C/C++ (and more) code-formatting tool
 clang-format
```
JNeiger: Do we need this version of clang-format if we're forcing v10?

JNeiger (approved):
Removing stylize removes the go dependency. I would just rip that out of `util/*-packages.txt` and `util/*-setup`. It'll probably be good to make a note of this dependency within the non-ubuntu files though so when someone gets around to using the mac/arch version, they can more easily fix the problems

util/ubuntu-packages.txt
oswinso: We don't, I can remove this

oswinso: alright removed.

kylestach (approved):
LGTM

kylestach (approved):
LGTM

PR #1443, opened 2020-05-10
Static analysis fixes

### LINKED ISSUE #1427: Fix cppcheck and clang-tidy warnings
## Description

The current codebase has way too many [cppcheck](http://cppcheck.sourceforge.net/) and [clang-tidy](https://clang.llvm.org/extra/clang-tidy/) warnings, which is bad in terms of undefined-behavior, readability and performance.

This number currently also includes warnings from 3rd party modules, but a rough number I'm getting:
`cpplint` - 1193
`clang-tidy` - 823

## Requirements
- Integrate `cppcheck` / `clang-tidy` into CI / the project more
- Creating a set of rules / lints that we want to use
- Reducing the number of warnings from these tools by at least an order of magnitude.

oswinso:
## Description
This PR reduced a large number of warnings that `clang-tidy` is giving.

On staging, with the current `.clang-tidy` file, `clang-tidy` gives 1033 warnings. After this PR, we now have only 454 warnings. If we ignore `new` related warnings such as `assigning newly created 'gsl::owner<>' to non-owner 'QTreeWidgetItem *' [cppcoreguidelines-owning-memory]`, then this number goes down to 364 (ie. all the QT and Config things).

In addition, this PR also creates a python tool in `util/analyze_clang_tidy` that gives a summary of how many of each warnings there are.

Note: none of the files in planning were touched, because of #1429.

Notable warnings:
- `bugprone-narrowing-conversions`, `cppcoreguidelines-narrowing-conversions`
    - Our entire codebase has implicit conversions between `double` and `float` everywhere. We really need to be consistent so we don't keep converting back and forth.
- `readability-implicit-bool-conversion`
    - Same with implicit bool conversions.
- `cppcoreguidelines-owning-memory`
    - This is basically everytime we do `new` into a raw pointer because memory leaks. I originally tried making the config system use `unique_ptr`, but I realized that the config system relies on memory leaking, and actually breaks if you clean up after yourself.
- `readability-braces-around-statements`
    - Personally I think this helps with bugs and readability.
- `misc-throw-by-value-catch-by-reference`
    - There are so many places where we are throwing heap allocated exceptions, or catching exceptions by value.

This was the breakdown before:
```
bugprone-narrowing-conversions·····················: 104
cppcoreguidelines-narrowing-conversions············: 104
readability-implicit-bool-conversion···············:  99
cppcoreguidelines-owning-memory····················:  98
readability-braces-around-statements···············:  79
cppcoreguidelines-pro-type-vararg··················:  59
readability-convert-member-functions-to-static·····:  43
performance-unnecessary-value-param················:  40
readability-make-member-function-const·············:  32
misc-throw-by-value-catch-by-reference·············:  29
cppcoreguidelines-pro-bounds-pointer-arithmetic····:  28
readability-container-size-empty···················:  26
modernize-use-auto·································:  25
misc-unused-parameters·····························:  24
cppcoreguidelines-init-variables···················:  22
cppcoreguidelines-pro-type-member-init·············:  16
readability-qualified-auto·························:  16
hicpp-exception-baseclass··························:  15
modernize-deprecated-headers·······················:  14
google-readability-todo····························:  14
readability-inconsistent-declaration-parameter-name:  13
readability-else-after-return······················:  12
bugprone-integer-division··························:   9
modernize-loop-convert·····························:   8
cppcoreguidelines-pro-type-union-access············:   8
cppcoreguidelines-explicit-virtual-functions·······:   8
modernize-use-override·····························:   8
cppcoreguidelines-pro-type-cstyle-cast·············:   7
bugprone-branch-clone······························:   7
cppcoreguidelines-pro-type-reinterpret-cast········:   6
readability-isolate-declaration····················:   6
modernize-use-nodiscard····························:   6
cppcoreguidelines-pro-bounds-array-to-pointer-decay:   5
cppcoreguidelines-avoid-c-arrays···················:   5
modernize-avoid-c-arrays···························:   5
clang-analyzer-deadcode.DeadStores·················:   5
performance-type-promotion-in-math-fn··············:   4
google-runtime-int·································:   3
performance-inefficient-vector-operation···········:   2
performance-unnecessary-copy-initialization········:   2
modernize-use-equals-default·······················:   2
readability-simplify-boolean-expr··················:   2
readability-static-accessed-through-instance·······:   2
clang-analyzer-core.CallAndMessage·················:   1
google-runtime-references··························:   1
modernize-use-nullptr······························:   1
modernize-avoid-bind·······························:   1
cppcoreguidelines-pro-type-static-cast-downcast····:   1
llvm-namespace-comment·····························:   1
readability-const-return-type······················:   1
modernize-use-emplace······························:   1
readability-named-parameter························:   1
modernize-use-bool-literals························:   1
cppcoreguidelines-interfaces-global-init···········:   1
--------------------------------
Total warnings: 1033
```

Here's a breakdown of the number of each currently:
```
cppcoreguidelines-owning-memory····················:  90
cppcoreguidelines-pro-type-vararg··················:  59
bugprone-narrowing-conversions·····················:  53
cppcoreguidelines-narrowing-conversions············:  53
readability-make-member-function-const·············:  31
cppcoreguidelines-pro-bounds-pointer-arithmetic····:  26
readability-braces-around-statements···············:  15
google-readability-todo····························:  14
readability-qualified-auto·························:  10
bugprone-integer-division··························:  10
readability-implicit-bool-conversion···············:  10
cppcoreguidelines-init-variables···················:   9
readability-inconsistent-declaration-parameter-name:   8
cppcoreguidelines-pro-type-union-access············:   8
modernize-use-auto·································:   8
cppcoreguidelines-pro-type-reinterpret-cast········:   6
cppcoreguidelines-avoid-c-arrays···················:   5
modernize-avoid-c-arrays···························:   5
clang-analyzer-deadcode.DeadStores·················:   5
cppcoreguidelines-pro-bounds-array-to-pointer-decay:   3
cppcoreguidelines-pro-type-member-init·············:   3
google-runtime-int·································:   3
readability-convert-member-functions-to-static·····:   2
readability-isolate-declaration····················:   2
performance-inefficient-vector-operation···········:   2
performance-unnecessary-value-param················:   2
modernize-use-equals-default·······················:   2
misc-unused-parameters·····························:   2
cppcoreguidelines-pro-type-cstyle-cast·············:   2
clang-analyzer-core.CallAndMessage·················:   1
performance-move-const-arg·························:   1
google-runtime-references··························:   1
cppcoreguidelines-pro-type-static-cast-downcast····:   1
modernize-avoid-bind·······························:   1
cppcoreguidelines-interfaces-global-init···········:   1
--------------------------------
Total warnings: 454
```

## Associated Issue
Issue #1427

### Run clang-tidy on the codebase
0. Install `clang-tidy-10` via here: https://apt.llvm.org/
1. Create a `compile_commands.json` by doing a cmake build with the `-DCMAKE_EXPORT_COMPILE_COMMANDS=ON`
2. Run `run-clang-tidy-10 -p build-debug $(cat files.txt) > clang_tidy.log`, where files.txt is just a file containing all the source files to run `clang-tidy` on. I ran this only on a [subset of files that are important](https://gist.github.com/oswinso/031b4f8738041e3e5c9fab7bee55428b)
3. Run `python3 util/analyze_clang_tidy.py clang_tidy.log`

Expected result: it should show that the number of warnings is 454, which is much lower than 1033.

### Everything still works
1. I guess run simulation or something?

Expected result: There shouldn't be any functional changes, so everything that worked before should still work.


soccer/Configuration.cpp
```diff
@@ -2,35 +2,35 @@
...
+#include <iostream>
+#include <utility>
 
 std::list<Configurable*>* Configurable::_configurables;
 
 // Role for tree column zero for storing ConfigItem pointers.
 static const int ConfigItemRole = Qt::UserRole;
 
-Q_DECLARE_METATYPE(ConfigItem*)  // FIXME: verify this
+Q_DECLARE_METATYPE(ConfigItem*)  // FIXME: verify this NOLINT
```
kylestach, line 15: Oh god I don't trust anything in the config system. It's slated for removal as soon as we have ROS.

soccer/Logger.cpp
```diff
@@ -1,9 +1,12 @@
 #include "Logger.hpp"
```
kylestach, line 1: Logger changes are going to cause major conflict with an upcoming PR. Just a heads up, I'll probably just redo these fixes on top of the new version.

soccer/Processor.cpp
```diff
@@ -1,25 +1,27 @@
-#include <gameplay/GameplayModule.hpp>
```
kylestach, line 1: Ditto with Processor changes.

soccer/DebugDrawer.hpp
```diff
@@ -28,56 +28,55 @@ class DebugDrawer {
 
     /** @ingroup drawing_functions */
     void drawPolygon(const Geometry2d::Point* pts, int n,
-                     const QColor& color = Qt::black,
+                     const QColor& qc = Qt::black,
```
JNeiger, line 31: What was the reasoning behind this change?

kylestach (approved):
I didn't actually read it all obviously, but this is a good step towards cleaning up the codebase. 

What should we do with regards to integrating this into CI? If it still has some unresolved lints obviously it shouldn't be required, but I'd like to enforce that new code shouldn't make _new_ warnings without reason. Thoughts?

soccer/KickEvaluator.hpp
```diff
@@ -80,11 +79,10 @@ class KickEvaluator {
      * @return F(X), F'(X)
      */
     static std::tuple<float, float> eval_calculation(
-        float x, const float kmean, const float kstdev,
```
JNeiger, line 83: The removed `const` qualifiers for the arguments for all functions in this file aren't reflected in the cpp

JNeiger: It actually looks like this is the case for almost every change to `const` and `&` qualifiers on arguments in all the files.

soccer/Robot.cpp
```diff
@@ -22,55 +23,57 @@ using Planning::MotionInstant;
...
 
+    self_avoid_radius =
+        new ConfigDouble(cfg, "PathPlanner/selfAvoidRadius", Robot_Radius);
+    opp_avoid_radius = new ConfigDouble(cfg, "PathPlanner/oppAvoidRadius",
+                                        Robot_Radius - kOppAvoidRadiusThresh);
+    opp_goalie_avoid_radius =
+        new ConfigDouble(cfg, "PathPlanner/oppGoalieAvoidRadius",
+                         Robot_Radius + kOppGoalieAvoidRadiusBuffer);
+    dribble_out_of_bounds_offset = new ConfigDouble(
+        cfg, "PathPlanner/dribbleOutOfBoundsOffset", kDribbleOutOfBoundsOffset);
```
JNeiger: I'm not sure I fully agree with this change. To me it doesn't add any extra information to these default config values

soccer/Robot.hpp
```diff
@@ -112,7 +114,7 @@ class Robot {
     Context* _context;
 
 private:
-    const unsigned int _shell;
+    const int _shell;
```
JNeiger, line 117: Why was this reverted to an int instead of unsigned?

soccer/SystemState.cpp
```diff
@@ -20,38 +21,41 @@ class BallPath : public Planning::Path {
 public:
     BallPath(const Ball& ball) : ball(ball){};
 
-    virtual bool hit(const Geometry2d::ShapeSet& obstacles,
-                     RJ::Seconds startTimeIntoPath,
-                     RJ::Seconds* hitTime) const {
-        throw new std::runtime_error("Unsupported Opperation");
+    bool hit(const Geometry2d::ShapeSet& /*obstacles*/,
+             RJ::Seconds /*startTimeIntoPath*/,
+             RJ::Seconds* /*hitTime*/) const override {
```
JNeiger, line 26: I assume this is to stop any unused variable warnings or something?

soccer/SystemState.hpp
```diff
@@ -67,7 +68,7 @@ class Context;
  */
 class SystemState {
 public:
-    SystemState(Context* const context);
+    SystemState(Context* context);
```
JNeiger: This to me removes context about the variable `context` and how it's to be used. If it's complaining that  `Robot`'s constructor isn't treating const correctly, I would rather fix that than remove context here.

soccer/WindowEvaluator.cpp
```diff
@@ -227,7 +235,7 @@ void WindowEvaluator::fill_shot_success(Window& window, Point origin) {
     auto shot_vector = window.segment.center() - origin;
     auto shot_distance = shot_vector.mag();
 
-    // get the angle between the shot vector and the target segment, then
+    // get the angle between the shot std::vector and the target segment, then
```
JNeiger: lol, this should actually be vector

soccer/gameplay/GameplayModule.cpp
```diff
@@ -36,11 +36,9 @@ void GameplayModule::createConfiguration(Configuration* cfg) {
 }
 
 bool GameplayModule::hasFieldEdgeInsetChanged() const {
-    if (abs(_fieldEdgeInset->value() - _oldFieldEdgeInset) >
-        numeric_limits<double>::epsilon()) {
-        return true;
-    }
-    return false;
+    return abs(_fieldEdgeInset->value() - _oldFieldEdgeInset) >
+
```
JNeiger: this spacing is weird?

soccer/gameplay/robocup-py.cpp
```diff
@@ -27,6 +29,7 @@ using namespace boost::python;
 #include <exception>
 #include <motion/MotionControl.hpp>
 #include <rc-fshare/pid.hpp>
+#include <utility>
```
JNeiger: This is included twice?

soccer/gameplay/robocup-py.cpp
```diff
@@ -356,7 +364,9 @@ boost::python::object Line_segment_intersection(Geometry2d::Line* self,
 boost::python::tuple Line_intersects_circle(Geometry2d::Line* self,
                                             Geometry2d::Circle* circle) {
     if (circle == nullptr) throw NullArgumentException("circle");
-    Geometry2d::Point a, b;
+    Geometry2d::Point a;
+
```
JNeiger: Weird space

soccer/motion/MotionControlNode.cpp
```diff
@@ -23,7 +23,7 @@ void MotionControlNode::runMotion(
         MotionControl& controller = _controllers[i];
         MotionSetpoint* setpoint = &(*setpoints)[i];
         if (force_stop) {
-            controller.stop(setpoint);
+            MotionControl::stop(setpoint);
```
JNeiger, line 26: I'm not a fan of this one, but it's hard to find reasons besides it doesn't look right in terms of the data flow within this function.

common/Geometry2d/Arc.cpp
```diff
@@ -10,24 +10,25 @@ namespace Geometry2d {
...
+    const auto cy = static_cast<float>(_center.y());
 
-    float x1 = line.pt[0].x() - cx;
-    float y1 = line.pt[0].y() - cy;
-    float x2 = line.pt[1].x() - cx;
-    float y2 = line.pt[1].y() - cy;
+    const auto x1 = static_cast<float>(line.pt[0].x() - cx);
+    const auto y1 = static_cast<float>(line.pt[0].y() - cy);
+    const auto x2 = static_cast<float>(line.pt[1].x() - cx);
+    const auto y2 = static_cast<float>(line.pt[1].y() - cy);
```
JNeiger, line 19: What's the current C++ thoughts behind using auto? I only use it for iterators or for each loops because I feel like it makes reading code somewhat difficult when types aren't obvious.

common/Geometry2d/Util.hpp
```diff
@@ -14,19 +14,23 @@
...
     if (f < T(0)) {
         return -1;
-    } else if (f > T(0)) {
+    }
+    if (f > T(0)) {
         return 1;
-    } else {
-        return 0;
     }
+    return 0;
```
JNeiger, line 33: Is this something it complained about or is this more a style thing?

soccer/DebugDrawer.hpp
oswinso: This is `readability-inconsistent-declaration-parameter-name` [link](https://clang.llvm.org/extra/clang-tidy/checks/readability-inconsistent-declaration-parameter-name.html), ie. the parameter name in the header being different than the parameter name in the definition. `color` is a function defined in `LogUtils.hpp`, and so the definition can't use `color` as the name for the parameter. There definitely are better solutions to this, but I'm leaving this as future work when we refactor the GUI.

soccer/Robot.cpp
oswinso: oh I think I can revert this change. Originally I tried to use `unique_ptr`, but that caused some bugs so I was debugging

soccer/SystemState.cpp
oswinso: Yeah. `misc-unused-parameters` [link](https://clang.llvm.org/extra/clang-tidy/checks/misc-unused-parameters.html)

common/Geometry2d/Arc.cpp
oswinso: The warning for here is that the variable is defined with a cast, and so having auto means you don't need to write the type twice. [link](https://clang.llvm.org/extra/clang-tidy/checks/modernize-use-auto.html#cast-expressions)

JNeiger: Whoops, this was more in general, not this specific cast

common/Geometry2d/Util.hpp
oswinso: It's a readability / style thing it complained about: `readability-else-after-return` [link](https://clang.llvm.org/extra/clang-tidy/checks/readability-else-after-return.html)

soccer/gameplay/robocup-py.cpp
oswinso: Woops thanks for catching this

soccer/Robot.hpp
oswinso: I guess the main reason is that even though it's used as an `int` in so many places that I thought it would be better to just have it as an `int` instead of an `unigned int`. I can revert this if you like

JNeiger: That's fair, it's not that important

soccer/SystemState.hpp
JNeiger: If it's more than one or two lines, it's not a big deal and can be fixed another time

soccer/DebugDrawer.hpp
JNeiger: RIP, that's unfortunate

soccer/KickEvaluator.hpp
oswinso: ah yeah you're right. In `readability-avoid-const-params-in-decls` [link](https://clang.llvm.org/extra/clang-tidy/checks/readability-avoid-const-params-in-decls.html), it only checks for `const` in declarations because they are superfluous, but they're not superfluous in definitions. I think for consistency we should also remove them in definitions though

soccer/WindowEvaluator.cpp
oswinso: oh woops my find and replace didn't work out

soccer/gameplay/GameplayModule.cpp
oswinso: yup fixed

soccer/gameplay/robocup-py.cpp
oswinso: fixed

soccer/motion/MotionControlNode.cpp
oswinso: Yeah... currently I don't think the `MotionControl::stop` function makes sense. Maybe it makes more sense to have a `setpoint.zero()` function or something?

common/Geometry2d/Arc.cpp
oswinso: oh. I think the community is still divided between always use auto and like just using it for iterators and stuff.

soccer/SystemState.hpp
oswinso: This is a "ptr to `const Context`" vs "`const` ptr to `Context`" issue I think. It's currently a `const` ptr to `Context`", so the `const` is superfluous. Should this be a "ptr to `const Context`?

JNeiger: It's supposed to be const ptr to Context. If it's superfluous, nevermind then. I guess I'm still stuck on the const everything that shouldn't change train lol

oswinso: yeah... it's superfluous the same way as like how a `const float` parameter is superfluous....

JNeiger (approved):

common/Geometry2d/Arc.cpp
kylestach: Personally I'm more on the side of only for long/opaque types. Regards it's pretty much a judgment call.

PR #1444, opened 2020-05-11
Added PID resetting to stop(), moved fshare to use master branch

oswinso:
## Description
`MotionControl::stop()` didn't actually reset the PID controllers. This PR adds that in. Also, it changes the branch for `fshare` from `comp2019` to `master`.

### IDK how to test this


kylestach (approved):
Looks good to me. I don't think there's really an easy test case for this, just make sure it doesn't explode with large derivative when you stop.

oswinso:
how do we test that

kylestach:
Honestly a unit test would be pretty simple...reset it and make sure effort remains relatively low the cycle after.

JNeiger (approved):

Issue #1445, opened 2020-05-11
Support Ubuntu 20.04


kylestach:
## Description
This project currently supports at most Ubuntu 19.10. 20.04 adds in a few new changes that we'll have to deal with:
 - No Qt4 support (requires a grSim upgrade)

## Requirements

 - Project works out-of-the-box on 20.04 after running `ubuntu-setup`.
 - No regressions on older operating systems including 18.04 and 19.* 

## Design

N/A

PR #1446, opened 2020-05-11
Support for 20.04

kylestach:
## Description
Upgrade to the latest version of grSim. The corresponding PR in our fork of grSim is [here](https://github.com/RoboJackets/grSim/pull/7).

This also gets rid of some old packages that are no longer necessary. It should work out-of-the-box on Ubuntu 20.04 LTS.

## Associated Issue
Issue #1445

## Design Documents
N/A

### Standard regression suite
 - Run basic tests of soccer in simulation on multiple operating systems:
    - Ubuntu 18.04
    - Ubuntu 20.04

Expected result: no regressions


kylestach:
Okay this PR is now slightly more in-depth, but basically it should:
 - Update submodules (RRT at https://github.com/RoboJackets/rrt/pull/95, grSim at https://github.com/RoboJackets/grSim/pull/7)
 - Remove all dependency on python 2
 - Clean up the format-diff script (for python 2->3)
 - Move around some configs and get rid of unnecessary packages

util/requirements3.txt
```diff
@@ -8,3 +8,7 @@ pylint # static checker for python
 mypy # static arugment checker
 
 yapf # python checker
+
+# Computing libraries
+numpy
+matplotlib
```
JNeiger: Are we actually using these?

util/ubuntu-setup
```diff
@@ -179,9 +179,8 @@ PACKAGES="$(sed 's/#.*//;/^$/d' $BASE/util/ubuntu-packages.txt)"
 apt-get install $ARGS $PACKAGES
 
 # install python3 requirements
+pip3 install --upgrade pip
 pip3 install -r $BASE/util/requirements3.txt
-# install python2 requirements
-pip2 install -r $BASE/util/requirements2.txt
```
JNeiger, line 184: We can remove the requirements2.txt as well if we're taking out python2 overall

soccer/radio/SimRadio.cpp
```diff
@@ -101,57 +107,51 @@ void SimRadio::startReceive() {
...
 
-    rx.set_fpga_status(FpgaGood);
-    rx.set_timestamp(RJ::timestamp());
+    for (size_t pkt_idx = 0; pkt_idx < packet.robots_status_size(); pkt_idx++) {
+        const Robot_Status& status = packet.robots_status(pkt_idx);
+        int this_robot_id = status.robot_id();
+        bool ball_sense = status.infrared();
+        bool just_kicked = status.has_flat_kick() || status.has_chip_kick();
+        // TODO: do we even get results if the robot isn't on?
+        bool robot_on = true;
```
JNeiger: Since they changed their data structure, I don't think we need this flag anymore or even to get this data. It doesn't look like it's being used anyway

kylestach: ACK

util/requirements3.txt
kylestach: Looks like only in an old process-latency script that hasn't been touched since 2011. I removed it, I'm pretty confident that it no longer works.

JNeiger (approved):

JNeiger:
It's good that we're finally getting around to do some cleanup. Thanks @kylestach and @oswinso!

liquidmetal9015:
I'm currently getting the wrong config loading from a fresh install as far as I can tell, so 8 robots on the larger field and possibly having our usual multi-cast problem although sim2-play works out of the box. I deleted my .grsim.xml from my home directory but it doesn't seem to be generating a new one and the config is still the same. It's not hard to set 6 robots and division B field but if it can be avoided for new people that would be good. I've also had some very strange performance issues in grSim where robots are dropping out of vision and radio on an off for a period before returning to normal but that might just be me. Other than that things seem to be working well on 20.04. When I run headmore grsim often spits out a constant stream of "Sending UDP datagram failed (maybe too large?). Size was: 228 byte(s)." for a period before stopping. I think the flickering in and out might correspond with "Sending UDP datagram failed (maybe too large?). Size was: 1216 byte(s)." Edit: with 6 robots you get 1182 bytes it would appear.

JNeiger:
Wait, is this using the grsim master or the robojackets_mods branch?

liquidmetal9015:
Ball placement by clicking in the soccer UI is also reversed for one of the sim 2 play teams.

JNeiger:
Long story short, I think we've been keeping config changes and a few features (like headless) in a different branch than master for some reason. I don't remember why exactly. https://github.com/RoboCup-SSL/grSim/compare/master...RoboJackets:robojackets_mods_v2

liquidmetal9015:
I might be dumb but when I merge this into staging locally I get pages and pages of build errors, even running submodule update and ubuntu-setup just to be safe.

kylestach:
Did you `make clean` @liquidmetal9015?

Also, @JNeiger TL;DR submodules just use a hash not a branch so it is fine as is, but I was going to merge it back into our grsim main. If we'd rather not I'm fine with cancelling that PR though.

JNeiger:
It's moreso all the config defaults we're on that branch (8 -> 6 robots, vision broadcast ip, friction, and robot info). If we're just moving to grsim master, that's fine, but we definitely should make sure some of the changes get reflected into the new version. IP and robot numbers being the big one

kylestach:
Really we should just provide a `.grsim.xml` file that is installed standard in `*-setup`. This should solve our problem right?

kylestach:
The flipped ball placement issue has been an issue for a while, and is unrelated to this PR. It actually looks like an issue with propagating `defendPlusX` through the code, so it should be fixable after #1441 gets through.

kylestach:
@liquidmetal9015 could you check and see if you're still running into issues? You'll need to re-run `./util/ubuntu-setup` in order to get the new grsim configuration.

liquidmetal9015 (approved):
The grsim config appears to be installing so I see no other pertinent issues.

JNeiger (approved):
Check if kick on breakbeam works. If that works, it should be good to go

kylestach:
Yeah it seems to work fine.

PR #1447, opened 2020-05-11
Hotfix/fsm types

oswinso:
## Description
`pylint` is now broken on 16.04 after updating the docker image from the error, ie. see [here](https://app.circleci.com/pipelines/github/RoboJackets/robocup-software/136/workflows/d32174df-40c2-47aa-871f-5b6636c6c407/jobs/13040)
```
soccer/gameplay/fsm.py:115:44: E1101: Instance of 'list' has no 'name' member (no-member)
```

~This PR adds type information to `fsm.py` to fix this error~
nvm, now `mypy` is angry. This PR just adds a comment to disable that `pylint` from checking that line because `pylint` stills infers wrong types even after adding all the type information

### Check CI

Expected result: CI doesn't break after running `pylint`

### DIFF
--- a/soccer/gameplay/fsm.py
+++ b/soccer/gameplay/fsm.py
@@ -18,7 +18,6 @@ StateMethod = Callable[[], None]  # Takes nothing, returns nothing
 OnEnterMethod = Callable[[], None]  # Takes nothing, returns nothing
 OnExitMethod = Callable[[], None]  # Takes nothing, returns nothing
 
-
 ## @brief generic hierarchial state machine class.
 #
 # states can have substates.  If the machine is in a state, then it is also implicitly in that state's parent state
@@ -77,7 +76,8 @@ class StateMachine:
             # transition if an 'event' fires
             next_states = []
             if self.state in self._transitions:
-                for next_state, transition in self._transitions[self.state].items():
+                for next_state, transition in self._transitions[
+                        self.state].items():
                     if transition['condition']():
                         next_states += [next_state]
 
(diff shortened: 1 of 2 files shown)


soccer/gameplay/fsm.py
```diff
@@ -81,25 +96,34 @@ def spin(self):
...
-                       event_name: str):
+    def add_transition(self, from_state: State, to_state: State,
+                       condition: Union[bool, TransitionFunction],
+                       event_name: str) -> None:
         if isinstance(condition, bool):
-            condition = lambda: condition
+            condition_bool: bool = cast(bool, condition)
+
+            def condition_fn() -> bool:
+                return condition_bool
```
kylestach: For clarity could you make this a lambda? I feel like declaration is less clear than assignment. Also, you already know condition is a bool from the isinstance check so there should be no need for the cast.

kylestach (approved):
Small changes, but not going to block

soccer/gameplay/fsm.py
oswinso: `mypy` disagrees : (

kylestach: What the hell? Why does mypy say no lambdas here? We aren't getting rid of lambdas everywhere in our code just to make mypy happy...

kylestach: If it's just typing concerns this answer: https://stackoverflow.com/questions/33833881/is-it-possible-to-type-hint-a-lambda-function seems to give a direction.

You would just have to let `condition: Callable[[], bool] = lambda: condition_bool` and then you should be good, right? Or am I missing something?

oswinso: ok wtf it works now

JNeiger (approved):
Looks good 👍

PR #1451, opened 2020-05-12
Origin/hdemusg/pileup

hdemusg:
## Description
Implemented a basic defensive pileup using ball location to allocate collecting robots upon the resolution of the pileup

## Associated Issue
Defensive Pileup

### 9 test cases
TC 1: center field (-1 <= x <= 1, 1.5 <= y <= 7.5)
TC 2: left flank (x < -1, 1.5 <= y <= 7.5)
TC 3: right flank (1 < x, 1.5 <= y <= 7.5)
TC 4: our box (-1 <= x <= 1, y < 1.5)
TC 5: our left corner (x < -1,  y < 1.5)
TC 6: our right corner (1 < x,  y < 1.5)
TC 7: their box (-1 <= x <= 1, y > 7.5)
TC 8: their left corner (x < -1,  y > 7.5)
TC 9: their right corner (1 < x,  y > 7.5)

Expected results:
TC 1: two robots 180 degrees apart arrive at randomly determined locations (one of two possible configurations) one meter away from the ball. 
TC 2: two robots 90 degrees apart arrive at fixed positions one meter away from the ball
TC 3:  two robots 90 degrees apart arrive at fixed positions one meter away from the ball
TC 4: AVOID THE REGION (there should not be a pileup in this region)
TC 5: one robot arrives at a fixed point 1 meter away from the ball
TC 6: one robot arrives at a fixed point 1 meter away from the ball
TC 7: AVOID THE REGION (there should not be a pileup in this region)
TC 8: one robot arrives at a fixed point 1 meter away from the ball
TC 9: one robot arrives at a fixed point 1 meter away from the ball


soccer/gameplay/plays/defense/pileup.py
```diff
@@ -0,0 +1,115 @@
...
+            #print("Angle: ", t.angle)
+        #avg = sum(directions) / len(directions)
+        #print("Average angle: ", avg)
+        #print(str(x) + "," + str(y))
+        print(len(combatants), " robots fighting for the ball")
+        self.standbyBot1 = None
+        self.standbyBot2 = None
+        '''
+
+        if (y >= 7.5):
```
HussainGynai: Don't use literal numbers for field positions. Instead use something proportional to constants.Field.Length and constants.Field.Width; this is because the field is not always set up like it is in sim

soccer/gameplay/plays/defense/pileup.py
```diff
@@ -0,0 +1,115 @@
...
+            if (x <= -1): # left corner of attacking half
```
HussainGynai: Same as above comment

soccer/gameplay/plays/defense/pileup.py
```diff
@@ -0,0 +1,115 @@
...
+                print("pileup in attacking left corner")
+                pos1 = robocup.Point(x, y - 1)
```
HussainGynai: same here with the -1

soccer/gameplay/plays/defense/pileup.py
```diff
@@ -0,0 +1,115 @@
...
+                self.add_subbehavior(skills.move.Move(pos1), 'move', required=True)
+                self.standbyBot1 = self.subbehavior_with_name('move').robot
+            elif (x >= 1): #right corner of attacking half
```
HussainGynai: Literal here too

soccer/gameplay/plays/defense/pileup.py
```diff
@@ -0,0 +1,115 @@
...
+                print("pileup in attacking right corner")
+                pos1 = robocup.Point(x, y - 1)
```
HussainGynai: Isn't this just the same position as before? If so, you don't need an if else statement since it seems they are the same

soccer/gameplay/plays/defense/pileup.py
```diff
@@ -0,0 +1,115 @@
...
+                self.standbyBot1 = self.subbehavior_with_name('move').robot
+            else: #attacking box
+                print("pileup in illegal zone. let go immediately.")
+        elif (y <= 1.5):
+            if (x <= -1): #left corner of defending half
+                print("pileup in defending left corner")
+                pos1 = robocup.Point(x, y + 1)
+                self.add_subbehavior(skills.move.Move(pos1), 'move', required=True)
+                self.standbyBot1 = self.subbehavior_with_name('move').robot
+            elif (x >= 1): #right corner of defending half
```
HussainGynai: Same thing here, the if and elif are the same except for the string you print

soccer/gameplay/plays/defense/pileup.py
```diff
@@ -0,0 +1,115 @@
...
+                print("pileup in defending right corner")
+                pos1 = robocup.Point(x, y + 1)
+                self.add_subbehavior(skills.move.Move(pos1), 'move', required=True)
+                self.standbyBot1 = self.subbehavior_with_name('move').robot
+            else: #defending box
+                print("pileup in illegal zone. this should never occur.")
+        else:
+            if (x <= -1): #left flank
+                print("pileup on left flank")
+                pos1 = robocup.Point(x + math.sqrt(.5), y + math.sqrt(.5))
```
HussainGynai: Again make sure to use field constants instead of actual numbers

soccer/gameplay/plays/offense/two_side_attack.py
```diff
@@ -1,3 +1,4 @@
+
```
HussainGynai: Are you trying to change anything in two_sided_attack? If not please revert changes made to this file.

soccer/gameplay/plays/defense/pileup.py
```diff
@@ -0,0 +1,115 @@
...
+
+    def __init__(self):
+        super().__init__(continuous=False)
+        
+        x = main.ball().pos.x #x-coordinate of ball's position
+        y = main.ball().pos.y #y-coordinate of ball's position
+        us = main.our_robots()
+        them = main.their_robots()
+
+        ''' THIS IS TO BE IMPLEMENTED NEXT. GOAL IS TO CALCULATE AVERAGE POSITION OF OUR BOTS AND THEIR BOTS TO DETERMINE WHERE TO PLACE COLLECTING ROBOTS IN THE EVENT OF A GIVEN PILEUP
```
HussainGynai: Is this something you're still planning on doing?  If it is, I think  there might already be a function for this somewhere, I would ask on the robocup slack to see if anyone knows

HussainGynai (changes requested):
The stuff you have so far seems pretty good, just get rid of all the literal numbers you used for positions. I commented on some of them, but there were a lot so I didn't do all of them. Also ask about the average position function in slack and try to implement it.

soccer/gameplay/plays/defense/pileup.py
hdemusg: For the time being, I'll keep the basic version, but I'll ask around sometime next week to try and add this to this play.

soccer/gameplay/plays/defense/pileup.py
hdemusg: they are not the same. -1 was the left side, 1 was the right side, since the range of ball's x is -3 to 3

soccer/gameplay/plays/defense/pileup.py
```diff
@@ -0,0 +1,125 @@
...
+       
+        their_end = (length * 7)/8
+        our_end = length/8
+        left = base_width + (width/3)
+        right = base_width + (width*2)/3
+
+        if (y >= their_end): 
+            if (x <= left): # left corner of attacking half
+                print("pileup in attacking left corner")
+                pos1 = robocup.Point(x, y - 1)
```
HussainGynai: You still wanna change the y-1, to something like y - (c * length) , where c is a constant, because depending on the field a -1 to the coordinates can mean different things

soccer/gameplay/plays/defense/pileup.py
```diff
@@ -0,0 +1,125 @@
...
+                self.add_subbehavior(skills.move.Move(pos1), 'move', required=True)
+                self.standbyBot1 = self.subbehavior_with_name('move').robot
+            elif (x >= right): #right corner of attacking half
+                print("pileup in attacking right corner")
+                pos1 = robocup.Point(x, y - 1)
```
HussainGynai: Same as before with y-1

soccer/gameplay/plays/defense/pileup.py
```diff
@@ -0,0 +1,125 @@
...
+                self.add_subbehavior(skills.move.Move(pos1), 'move', required=True)
+                self.standbyBot1 = self.subbehavior_with_name('move').robot
+            else: #attacking box
+                print("pileup in illegal zone. let go immediately.")
+        elif (y <= our_end):
+            if (x <= left): #left corner of defending half
+                print("pileup in defending left corner")
+                pos1 = robocup.Point(x, y + 1)
```
HussainGynai: Same with y+1

soccer/gameplay/plays/defense/pileup.py
```diff
@@ -0,0 +1,125 @@
...
+                print("pileup in defending right corner")
+                pos1 = robocup.Point(x, y + 1)
+                self.add_subbehavior(skills.move.Move(pos1), 'move', required=True)
+                self.standbyBot1 = self.subbehavior_with_name('move').robot
+            else: #defending box
+                print("pileup in illegal zone. this should never occur.")
+        else:
+            if (x <= left): #left flank
+                print("pileup on left flank")
+                pos1 = robocup.Point(x + math.sqrt(.5), y + math.sqrt(.5))
```
HussainGynai: same with the +sqrt(.5)

HussainGynai (changes requested):
You got rid of most of the magic numbers, but there are still some for when you are determining pos1. You add and subtract literal numbers to the coordinates, which on different fields can mean different things.

soccer/gameplay/plays/defense/pileup.py
```diff
@@ -0,0 +1,125 @@
...
+        self.standbyBot1 = None
+        self.standbyBot2 = None
+        '''
+       
+        their_end = (length * 7)/8
+        our_end = length/8
+        left = base_width + (width/3)
+        right = base_width + (width*2)/3
+
+        if (y >= their_end):
```
JNeiger: I don't know how reasonable it is, but it might be cleaner to do something like

```
pos1 = None
pos2 = None

if ...
   if ...
      pos1 = rc.Point(x, y)
   elif ...
      pos1 = ...
...

if pos1 is not None:
    self.add_subbehavior(...)
    self.standbyBot1 = ...
if pos2 is not None:
    ...
```

JNeiger: This makes the if statements much shorter and cleans up all the add subbehaviors

soccer/gameplay/plays/defense/pileup.py
hdemusg: done, added constants dist and half_dist (for pythagorean calculations)

soccer/gameplay/evaluation/position.py
```diff
@@ -0,0 +1,43 @@
+import main
+import robocup
+import constants
+import math
+
+'''calculates the distance of a robot to the ball'''
+def distance(robot):
```
JNeiger: You can just do `(main.ball().pos - robot.pos).mag()` to get distance

soccer/gameplay/evaluation/position.py
```diff
@@ -0,0 +1,48 @@
...
+	y = abs(main.ball().pos.y - robot.pos.y)
+	dist = math.sqrt((x*x) + (y*y))
+	return dist
+'''
+	
+'''returns all robots, as well as their average positions, that are within a given radius of the ball'''
+def ball_vicinity(us, them, radius):
+    contending = []
+    #contending_us = []
+    #contending_them = []
```
HussainGynai: Get rid of comments

soccer/gameplay/evaluation/position.py
```diff
@@ -0,0 +1,48 @@
+import main
+import robocup
+import constants
+import math
+
+'''calculates the distance of a robot to the ball
+def distance(robot):
```
HussainGynai: Since you don't use this function anywhere, you might just wanna delete it.

soccer/gameplay/evaluation/position.py
```diff
@@ -0,0 +1,48 @@
...
+	x = abs(main.ball().pos.x - robot.pos.x)
+	y = abs(main.ball().pos.y - robot.pos.y)
+	dist = math.sqrt((x*x) + (y*y))
+	return dist
+'''
+	
+'''returns all robots, as well as their average positions, that are within a given radius of the ball'''
+def ball_vicinity(us, them, radius):
```
HussainGynai: Since us and them are always going to be in main.our_robots() and main.their_robots() you don't need to have them be passed in and can just directly access them in the eval function itself.

soccer/gameplay/plays/defense/advanced_pileup.py
```diff
@@ -0,0 +1,91 @@
...
+    def __init__(self):
+        super().__init__(continuous=False)
+        ball = main.ball().pos
+        us = main.our_robots()
+        them = main.their_robots()
+        radius = constants.Ball.Radius
+        range = 20 * radius
+
+        c, pos = evaluation.position.ball_vicinity(us, them, range)
+        print(c, pos)
```
HussainGynai: Get rid of debugging prints

HussainGynai (changes requested):
Looks pretty good overall, just make the small changes I wrote in the comments and run make pretty-lines

soccer/gameplay/evaluation/position.py
```diff
@@ -0,0 +1,48 @@
...
+    		contending.append(t)
+    		#contending_them.append(t)
+
+    cpos = 0 #position sum to calculate average position of robots (-pi to pi)
+   
+    for w in contending:
+        x = w.pos.x - main.ball().pos.x
+        y = w.pos.y - main.ball().pos.y
+
+        cpos += math.atan2(y,x)
```
JNeiger: You can do `cpos += (w.pos - main.ball().pos).angle()`

soccer/gameplay/evaluation/position.py
```diff
@@ -0,0 +1,48 @@
+import main
```
JNeiger, line 1: tabs/spaces aren't consistent. It's 4 spaces sometimes, sometimes 2

soccer/gameplay/evaluation/position.py
```diff
@@ -0,0 +1,48 @@
...
+
+'''calculates the distance of a robot to the ball
+def distance(robot):
+	x = abs(main.ball().pos.x - robot.pos.x)
+	y = abs(main.ball().pos.y - robot.pos.y)
+	dist = math.sqrt((x*x) + (y*y))
+	return dist
+'''
+	
+'''returns all robots, as well as their average positions, that are within a given radius of the ball'''
```
JNeiger, line 5: What does average position mean here, you're returning the average angle around the ball?

Average angle also isn't very good of metric in this regard if you just add the angles and divide by the number of robots. Something at -.99PI and .99PI shouldn't result in an average of 0 PI, it should be PI. I've converted to Cartesian coord, averaged them in the past. This comes out a little better in terms of averaging angle

soccer/gameplay/plays/defense/advanced_pileup.py
```diff
@@ -0,0 +1,91 @@
...
+        pos1 = None
+        pos2 = None
+        their_end = (length * 7)/8
+        our_end = length/8
+        left = base_width + (width/3)
+        right = base_width + (width*2)/3
+        dist = (length / 9)
+
+        #initial positioning
+        if (pos == -math.pi or pos == 0):
```
JNeiger: The white space on this nested if statement is super weird. The `else` on line 49 isn't highlighting correctly on github

soccer/gameplay/plays/defense/advanced_pileup.py
```diff
@@ -0,0 +1,91 @@
...
+import constants
+import play
+import skills
+import tactics
+import main
+import evaluation.position
+import math
+import random
+
+class AdvancedPileup(play.Play):
```
JNeiger: This will only execute when this play starts, after that first frame this starts, the positions don't get updated. Is that the goal?

soccer/gameplay/plays/defense/advanced_pileup.py
JNeiger: Since these are `==`, it's very unlikely that this, or the first else if will be hit. It might be better to specify a range if you want these to be hit

soccer/gameplay/plays/defense/advanced_pileup.py
```diff
@@ -0,0 +1,91 @@
...
+            else: 
+            	pos1 = robocup.Point(ball.x, ball.y + dist)
+        else: 
+            tangent = math.tan(pos)
+            if tangent > 0: # tangent = y over x
+                x_sq = 1 #x^2 + y^2 = 1
+                y_sq = tangent * tangent
+                factor = x_sq + y_sq #to get the sum to 1
+                x = math.sqrt(x_sq / factor)
+                y = math.sqrt(y_sq / factor)
```
JNeiger: If this is trying to get XY from an angle, you should be able to do `delta1 = robocup.Point.direction(angle)`, `pos1 = main.ball().pos - delta1`

soccer/gameplay/plays/defense/advanced_pileup.py
JNeiger: Wait, I just looked at this hard, see the code suggestion below and that should be enough to simplify a lot of this

soccer/gameplay/plays/defense/advanced_pileup.py
```diff
@@ -0,0 +1,91 @@
...
+                pos1 = robocup.Point(ball.x - x, ball.y + y)
+                pos2 = robocup.Point(ball.x + x, ball.y - y)
+         
+
+        left_end = - (width / 2)
+        right end = width / 2
+        our_edge = 0
+        their_edge = length
+
+        if pos1 != None:
```
JNeiger: The field should be a rectangle, so you should be able to check if a point is contained in the rectangle. It'll be `if constants.Field.FieldRect.contains_point(pt):`

soccer/gameplay/plays/defense/advanced_pileup.py
hdemusg: Since this is only run when the pileup begins, I am assuming that the pileup does not move in any way during its duration, and just getting the original positions to prepare for collection when the ball is out of the pileup.

soccer/gameplay/evaluation/position.py
hdemusg: I am returning both pi and 0, and whichever of the two corresponding points (both or neither is also acceptable) is within the field of play are assigned robots to wait at those positions.

soccer/gameplay/plays/defense/advanced_pileup.py
```diff
@@ -0,0 +1,103 @@
...
+        range = 20 * radius
+
+        c, pos = evaluation.position.ball_vicinity(range)
+
+        pos1 = None
+        pos2 = None
+        their_end = (length * 7) / 8
+        our_end = length / 8
+        left = base_width + (width / 3)
+        right = base_width + (width * 2) / 3
```
HussainGynai: Does this work? Where is base_width coming from?

soccer/gameplay/plays/defense/pileup.py
```diff
@@ -0,0 +1,129 @@
...
+        length = constants.Field.Length
+        #print((length * 5)/6)
+        #print(base_width, base_width + ((width * 2)/3))
+
+        # TO BE IMPLEMENTED IN A NEW PLAY
+
+        # for u in us:
+        #    ux = u.pos.x
+        #    uy = u.pos.y
+        #    print(ux, uy)
```
HussainGynai: You can delete these comments

HussainGynai (commented):
Check my comments, and fix the issue and I think you should be good to go.

soccer/gameplay/plays/defense/advanced_pileup.py
hdemusg: These were the same stipulations I used in the basic pileup, and they worked fine. I removed base_width (which was just a variable representing the left-most x value in basic pileup) and just used the left and right ends of the field as benchmarks.

HussainGynai (approved):
Looks good, but you're failing pylint and mypy, have you tried pulling and then pushing?

kylestach:
Closing, outdated

PR #1461, opened 2020-05-20
Joystick refactor

oswinso:
## Description
This PR refactors the joystick logic out of `Processor`, by creating a few classes (and scrapping the rest):
- `SDLGamepad`
    - This is responsible for all the libsdl logic for getting gamepad inputs and stuff
- `SDLJoystickNode`
    - This is responsible for initializing libsdl, and calling any callbacks that have been registered whenever it gets new joystick commands
- `ManualControlNode`
    - This registers callbacks with `SDLJoystickNode`, and contains the logic for how controls on the joystick change the translation / rotation and what not.

Note: this PR does not handle #1424.

### Joystick still works
1. Open up sim and connect a joystick
2. Move robots around

Expected Result: Robots still move around normally


soccer/CMakeLists.txt
```diff
@@ -12,11 +12,10 @@ set(ROBOCUP_LIB_SRC
...
     "gameplay/robocup-py.cpp"
-    "joystick/Joystick.cpp"
-    "joystick/GamepadController.cpp"
-    "joystick/GamepadJoystick.cpp"
-    "joystick/SpaceNavJoystick.cpp"
-    "KickEvaluator.cpp"
+        "joystick/SDLJoystickNode.cpp"
+    "joystick/SDLGamepad.cpp"
+    "joystick/ManualControlNode.cpp"
+        "KickEvaluator.cpp"
```
JNeiger: Tabbing is weird

soccer/Processor.hpp
```diff
@@ -190,8 +181,8 @@ class Processor {
 
     std::vector<Node*> _nodes;
 
-    // joystick control
-    std::vector<Joystick*> _joysticks;
+    std::unique_ptr<joystick::SDLJoystickNode> _sdl_joystick_node;
+    std::unique_ptr<joystick::ManualControlNode> _manual_control_node;
```
JNeiger, line 192: Minor nit, these should probably be with the list of modules in the section just above

soccer/Processor.cpp
```diff
@@ -85,6 +81,16 @@ Processor::Processor(bool sim, bool blueTeam, const std::string& readLogFile)
...
 
+    // Joystick
+    _sdl_joystick_node = std::make_unique<joystick::SDLJoystickNode>();
+    _manual_control_node =
+        std::make_unique<joystick::ManualControlNode>(&_context);
+
+    // Register callbacks
+    _sdl_joystick_node->addCallbacks(_manual_control_node->getCallback(),
+                                     _manual_control_node->getOnConnect(),
+                                     _manual_control_node->getOnDisconnect());
```
JNeiger: This is fairly tight coupling between the manual control node and the joystick driver compared to what we have done so far with context. Is this the direction we want to go? An alternate form would be the joystick node pushing a vector of joystick ID's and messages into context that the manual control node consumes every frame

soccer/Processor.cpp
```diff
@@ -483,6 +449,7 @@ void Processor::run() {
 
         // Send motion commands to the robots
         sendRadioData();
+        _manual_control_node->applyControlsToRobots(&_context.state.self);
```
JNeiger: Should this be in the manual control node `run` function?

JNeiger (commented):
Just a few things here and there

soccer/Processor.cpp
oswinso: The basic idea was to model this as a ROS pub/sub architecture, ie. joystick node publishing joystick messages, manual control node subscribing to those messages. I just came up with this as like a "placeholder" for that when we move over to ROS. What you suggested works as well tho.

soccer/Processor.cpp
oswinso: yeah, I think so

soccer/CMakeLists.txt
oswinso: fixed

soccer/Processor.hpp
oswinso: done

soccer/Processor.cpp
oswinso: one thing I'm confused about though is like how this is tightly coupled? The `ManualControlNode` doesn't need to know anything about how the joystick driver works, only that it will call the callbacks when new information is added.

JNeiger: They're not like super super tightly coupled, but compared to the context pub/sub model, it's much more coupled. The institutional knowledge of how the context is setup allows much safer editing of one module without the other. Information about how often the callbacks are being called, whether you need to do any multi-thread safety checks, what guarantees there are about which callbacks are called in what order etc, are very important when editing these modules. If this gets added as another model for pub/sub and the institutional knowledge builds assumptions about how data is shared, it becomes much less coupled.

TLDR: If this becomes a common occurrence and pub/sub model alongside the shared memory context model, it's not super coupled. If it doesn't become commonly used and the contract isn't fully defined, it's very coupled

soccer/Processor.cpp
```diff
@@ -483,6 +449,7 @@ void Processor::run() {
 
         // Send motion commands to the robots
         sendRadioData();
+        _manual_control_node->run();
```
JNeiger, line 450: Could you add a comment describing that the order of these two functions is very specific? If sendRadioData is called after, it will override all manual control if the robot is not visible since it sets the `intent.is_active`

soccer/Processor.cpp
```diff
@@ -532,99 +499,14 @@ void Processor::sendRadioData() {
     // Add RadioTx commands for visible robots and apply joystick input
```
JNeiger: Don't forget to update this comment

oswinso: Changed the name of the function instead

soccer/Processor.cpp
oswinso: I've changed it so that `intent.is_active` is not touched if it's being controlled by a joystick

soccer/Processor.cpp
oswinso: Ok, sure. I've changed it to use `context`

soccer/joystick/ManualControlNode.hpp
```diff
@@ -0,0 +1,96 @@
...
+#include <Robot.hpp>
+#include <SystemState.hpp>
+#include <functional>
+#include <joystick/GamepadMessage.hpp>
+#include <unordered_map>
+
+namespace joystick {
+using GamepadCallbackFn = std::function<void(const GamepadMessage&)>;
+using GamepadConnectedFn = std::function<void(int unique_id)>;
+using GamepadDisconnectedFn = std::function<void(int unique_id)>;
```
JNeiger: Leftover from callbacks

JNeiger (approved):
Minus the one thing, it looks good, nice job!

soccer/joystick/ManualControlNode.hpp
oswinso: Removed, woopsies

oswinso:
oh lmao woops accidentally re-requested a review...

soccer/Context.hpp
```diff
@@ -33,13 +34,19 @@ struct Context {
...
+    bool joystick_valid;
 
     std::unique_ptr<RobotConfig> robot_config;
 
     SystemState state;
     GameState game_state;
     DebugDrawer debug_drawer;
 
+    /** \brief Stack of unique IDs of gamepads. First is oldest to connect. */
+    std::vector<int> gamepad_stack;
```
kylestach: Unless I'm missing something, why is this a stack? You can disconnect a joystick at any time, it doesn't have to be FILO.

soccer/joystick/SDLGamepad.hpp
```diff
@@ -0,0 +1,100 @@
...
+class SDLGamepad {
+public:
+    SDLGamepad(int device_index);
+    ~SDLGamepad();
+
+    // Noncopyable
+    SDLGamepad(const SDLGamepad&) = delete;
+    SDLGamepad& operator=(const SDLGamepad&) = delete;
+    SDLGamepad(SDLGamepad&&) = delete;
+    SDLGamepad& operator=(SDLGamepad&&) = delete;
```
kylestach: Nitpick, should probably be default-movable.

soccer/Processor.cpp
```diff
@@ -510,119 +475,27 @@ void Processor::run() {
...
+    // Intent is active if it's being joystick controlled, or
+    // if it's visible.
+    for (OurRobot* r : _context.state.self) {
+        RobotIntent& intent = _context.robot_intents[r->shell()];
+        intent.is_active = r->isJoystickControlled() || r->visible();
     }
-    return vals;
 }
 
 void Processor::recalculateWorldToTeamTransform() {
```
kylestach, line 501: I just feel like I need to do a :tada: every time we cut code from Processor.

That's all.

soccer/joystick/GamepadMessage.hpp
```diff
@@ -0,0 +1,60 @@
+#pragma once
+#include <time.hpp>
+
+namespace joystick {
```
kylestach, line 4: Alright, question time: walk me through the process of adding a new joystick type. We would have to add a new type similar to GamepadMessage, right? Why not have a `ControllerMessage` instead of this that is a common interface between all controller types? (For reference I'd kind of like to add a WASD controller at some point in time so you don't have to grab a joystick for super basic tests - might be a good new member project actually).

soccer/joystick/ManualControlNode.cpp
```diff
@@ -0,0 +1,246 @@
...
+
+    // Field oriented control
+    if (robot->visible() &&
+        context_->game_settings.joystick_config.useFieldOrientedDrive) {
+        Geometry2d::Point translation{x_vel, y_vel};
+        translation.rotate(-M_PI / 2 - robot->angle());
+
+        x_vel = static_cast<float>(translation.x());
+        y_vel = static_cast<float>(translation.y());
+    }
```
kylestach, line 80: If FOC is on and the bot is not visible I would just stop it.

soccer/joystick/SDLJoystickNode.hpp
```diff
@@ -0,0 +1,87 @@
...
+    using GamepadDisconnectedFn = std::function<void(int unique_id)>;
+
+    /**
+     * Add a callback function
+     * @param callback Function that will get called on every connected joystick
+     * each update
+     */
+    void addCallbacks(const GamepadCallbackFn& callback,
+                      const GamepadConnectedFn& on_connected,
+                      const GamepadDisconnectedFn& on_disconnected);
```
kylestach: Is this actually used anywhere?

soccer/joystick/SDLJoystickNode.hpp
```diff
@@ -0,0 +1,87 @@
...
+#include <Context.hpp>
+#include <Node.hpp>
+#include <functional>
+#include <joystick/GamepadMessage.hpp>
+#include <joystick/SDLGamepad.hpp>
+#include <vector>
+
+namespace joystick {
+constexpr auto GameControllerDBPath =
+    "../external/sdlcontrollerdb/gamecontrollerdb.txt";
```
kylestach: We should probably just make a symlink from this to the `run` directory.

kylestach (changes requested):
Mostly this looks like a good change. A few changes, and I'd like a clarification on the thought behind the `GamepadMessage` struct.

soccer/joystick/GamepadMessage.hpp
oswinso: The reason would be that it's impossible to come up with any "common interface" for controller types. When we're actually using ROS, it would be like
```
           (GamepadMsg)                           (IntentMessage)
[ Gamepad ]------------> [ GamepadControlScheme ] --------------> [Something ]

            (KeyboardMsg)                            (IntentMessage)
[ Keyboard ]-------------> [ KeyboardControlScheme ] --------------> [ Something ]
```

Right now our `IntentMessage` would be basically `context_->robot_intents` and `context_->motion_setpoints`. We're not using ROS yet though, and I think it's not worth setting some sort of the above setup right now.

soccer/joystick/SDLJoystickNode.hpp
oswinso: removed

soccer/joystick/ManualControlNode.cpp
oswinso: done

soccer/joystick/SDLGamepad.hpp
oswinso: done

soccer/Context.hpp
oswinso: changed

soccer/joystick/GamepadMessage.hpp
kylestach: Yeah, that's what I was figuring you'd say.

Thanks for the clarification! I think this is fine for now.

run/gamecontrollerdb.txt
```diff
@@ -0,0 +1 @@
+../external/sdlcontrollerdb/gamecontrollerdb.txt
```
kylestach, line 1: Nice. I didn't realize symlinks showed up like this in github lmao.

kylestach (approved):
LGTM

PR #1462, opened 2020-05-21
Pass Tactic

HussainGynai:
## Description
Implementation of allowing use of heuristics to determine pass points into coordinated pass. Created a pass to point heuristic that takes in coordinates and just returns a point. Also created a TestHeuristic play that just uses the pass to point heuristic to pass to a location on the field. 

## Associated Issue
Issue #1458 

## Design Documents
[Link](https://docs.google.com/document/d/1vE7y_K5OppqjfXBsLNgMPDJiGpCbL5azZzDsPcYiDrw/edit?usp=sharing)

### Test Case 1
1. Run Soccer
2. Run the TestHeursitic play
3. Observe

Expected result: Should pass to (1,2)

### Test Case 2
1. Try creating your own heuristic with the function calc_pass_point()!

### DIFF
--- a/soccer/gameplay/plays/testing/testHeurPass.py
+++ b/soccer/gameplay/plays/testing/testHeurPass.py
@@ -9,14 +9,16 @@ import situational_play_selection
 import tactics.coordinated_pass
 from tactics.pass_to_point_heuristic import pass_to_point_heuristic
 
+
 ##
 # A test play that just uses the pass to point heuristic that takes in coodrinates and passes there
 #
 class TestHeuristic(standard_play.StandardPlay):
-    def __init__(self):
+    def __init__(self) -> None:
         super().__init__(continuous=False)
 
-    def on_enter_start(self):
+    def on_enter_start(self) -> None:
         self.add_subbehavior(
-            tactics.coordinated_pass.CoordinatedPass(pass_to_point_heuristic(constants.Field.Width/2, constants.Field.Length/2)),
-            'pass')
+            tactics.coordinated_pass.CoordinatedPass(
+                pass_to_point_heuristic(constants.Field.Width / 2,
+                                        constants.Field.Length / 2)), 'pass')
--- a/soccer/gameplay/plays/testing/testHeurPass.py
+++ b/soccer/gameplay/plays/testing/testHeurPass.py
@@ -2,36 +2,21 @@ import main
 import robocup
 import behavior
 import constants
-import enum
 
 import standard_play
 import evaluation
 import situational_play_selection
 import tactics.coordinated_pass
-import skills.move
-import skills.capture
-import random
-import tactics.pass_to_point_heuristic
-
-#Based on clear sitation
-#    Have ball in defense
-#   Want to Kick ball up field and get other robots in position
-#
-
+from tactics.pass_to_point_heuristic import pass_to_point_heuristic
 
 ##
-# A basic clearing play written in responce to the fact that we didn't have a clearing play
-#
-# It needs to be renamed as "clear" is the name of the situation, and each play that fufills it
-# needs to have a unique name beyond that
+# A test play that just uses the pass to point heuristic that takes in coodrinates and passes there
 #
 class TestHeuristic(standard_play.StandardPlay):
     def __init__(self):
         super().__init__(continuous=False)
 
     def on_enter_start(self):
-        self.heuristic = tactics.pass_to_point_heuristic.PassToPointHeuristic(
-            1, 2)
         self.add_subbehavior(
-            tactics.coordinated_pass.CoordinatedPass(heuristic=self.heuristic),
+            tactics.coordinated_pass.CoordinatedPass(pass_to_point_heuristic(constants.Field.Width/2, constants.Field.Length/2)),
             'pass')
--- a/soccer/gameplay/plays/testing/testHeurPass.py
+++ b/soccer/gameplay/plays/testing/testHeurPass.py
@@ -1,0 +1,37 @@
+import main
+import robocup
+import behavior
+import constants
+import enum
+
+import standard_play
+import evaluation
+import situational_play_selection
+import tactics.coordinated_pass
+import skills.move
+import skills.capture
+import random
+import tactics.pass_to_point_heuristic
+
+#Based on clear sitation
+#    Have ball in defense
+#   Want to Kick ball up field and get other robots in position
+#
+
+
+##
+# A basic clearing play written in responce to the fact that we didn't have a clearing play
+#
+# It needs to be renamed as "clear" is the name of the situation, and each play that fufills it
+# needs to have a unique name beyond that
+#
+class TestHeuristic(standard_play.StandardPlay):
+    def __init__(self):
+        super().__init__(continuous=False)
+
+    def on_enter_start(self):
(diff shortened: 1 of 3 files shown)


soccer/gameplay/tactics/pass_to_point_heuristic.py
```diff
@@ -0,0 +1,12 @@
+import constants
```
JNeiger, line 1: If the thought is to have tons of these heuristics, somewhere else besides tactics might be a good idea. Evaluation is getting big as a folder, but that's always a possible place to put it

soccer/gameplay/tactics/pass_to_point_heuristic.py
```diff
@@ -0,0 +1,12 @@
...
+import robocup
+import time
+
+
+class PassToPointHeuristic():
+    def __init__(self, x, y):
+        self.x = x
+        self.y = y
+
+    def calc_pass_point(self):
```
JNeiger: Is there a case where this will be anything more than just a function that will be called? If not, should this just be a function that you pass into the coordinated pass?

A fun idea that I saw around is a callable. So you if change `calc_pass_point` to `__call__`, you can call the class instance like a function. Then, a function pointer, lambda function, or this heuristic class can be passed in and executed without changing any code in coordinated pass using just `self.heuristic()`. We should def pick a single way to do it, but it's a fun concept
https://en.wikipedia.org/wiki/Callable_object#In_Python

soccer/gameplay/plays/testing/testHeurPass.py
```diff
@@ -0,0 +1,37 @@
...
+#   Want to Kick ball up field and get other robots in position
+#
+
+
+##
+# A basic clearing play written in responce to the fact that we didn't have a clearing play
+#
+# It needs to be renamed as "clear" is the name of the situation, and each play that fufills it
+# needs to have a unique name beyond that
+#
```
oswinso: Could you change this to a RST style doc?

soccer/gameplay/tactics/coordinated_pass.py
```diff
@@ -55,7 +55,8 @@ def __init__(self,
                  prekick_timeout=None,
                  receiver_required=True,
                  kicker_required=True,
-                 use_chipper=False):
+                 use_chipper=False,
+                 heuristic=None):
```
oswinso: Can you add types for these parameters?

soccer/gameplay/tactics/coordinated_pass.py
```diff
@@ -165,6 +165,8 @@ def receive_point(self, value):
                 'receiver').receive_point = self.receive_point
 
     def on_enter_running(self):
+        if self.heuristic != None:
```
oswinso: use `is not None` instead

soccer/gameplay/tactics/coordinated_pass.py
```diff
@@ -201,24 +203,6 @@ def use_chipper(self, value):
...
-                                    ball_velocity_unit.y * path_unit.x)
-            distance = path_vector.mag()
-            if distance < self.TOO_CLOSE_TO_FAIL_DISTANCE:
-                return True  # If the ball is already close to the receive point but not heading towards it, dont fail
-            if (distance > self.FAR_DISTANCE and cross_product_mag > self.STRICT_CROSS_THRESHOLD)\
-               or (distance < self.FAR_DISTANCE and cross_product_mag > self.RELAXED_CROSS_THRESHOLD)\
-               or ball_velocity.mag() < self.BALL_STOPPED_SPEED:
-                return False  # If ball is too off course or too slow, fail the pass
-        return True
-
```
oswinso, line 221: Why was this removed?

oswinso (changes requested):
Some initial changes

soccer/gameplay/tactics/coordinated_pass.py
HussainGynai: Situational Analysis already does this

soccer/gameplay/tactics/pass_to_point_heuristic.py
kylestach: These heuristics don't really mean anything outside of tactics, though. We could always make an entire `tactics.pass` directory and put them in there (and the actual tactic itself).

oswinso: What about a putting it under `heurustics/pass`?

soccer/gameplay/plays/testing/testHeurPass.py
```diff
@@ -0,0 +1,24 @@
...
+class TestHeuristic(standard_play.StandardPlay):
+    def __init__(self) -> None:
+        super().__init__(continuous=False)
+
+    def on_enter_start(self) -> None:
+        self.add_subbehavior(
+            tactics.coordinated_pass.CoordinatedPass(
+                pass_to_point_heuristic(constants.Field.Width / 2,
+                                        constants.Field.Length / 2)), 'pass')
```
JNeiger, line 24: What's the difference between this and what we had previously with the passing positioning? The pass point is being decided at this time in code instead of inside the coordinated pass

soccer/gameplay/tactics/pass_to_point_heuristic.py
```diff
@@ -0,0 +1,7 @@
+import constants
+import robocup
+import time
+
+
+def pass_to_point_heuristic(x: int, y: int) -> robocup.Point:
```
JNeiger, line 6: What's the thought behind doing (x, y) instead of point?

oswinso, line 6: I don't understand this.... what are `x` and `y` supposed to be?

HussainGynai: It's just a way to test stuff and as a template for new members for heuristics, like since we aren't making any changes to coordinated pass, we would just pass a point straight into it like we already do if we wanted to pass to a point.

HussainGynai: They are x and y coordinates on the field. See the comment I left for Joe above for reasoning.

oswinso: So...... what's the point of passing in the `x` and `y` coordinates on the field?

oswinso: What is this supposed to represent?

HussainGynai: Its where you want to pass to, like this isn't really meant to be a real heuristic we use just like a thing for new members to look at so they what one should look like. Like if I just passed in a point and immediately returned it wouldn't be helpful for new members to look at.

oswinso: So..... how are you supposed to know where you want to pass to if the heurustic is telling you where to pass to?

HussainGynai: The heuristic isn't. The heuristic in this case is just a function that takes in an x and y coordinate that you feed it when writing the play and returns a point. In an actual useful heuristic that a freshman would write they would pass in like other stuff, like maybe acceptable risk of pass or something,  and create a function that returns a point of where to pass to, which is then passed into coordinated pass.

Issue #1463, opened 2020-05-22
Wrong usage of eval_best_receive_point in many places


oswinso:
## Description
We use `eval_best_receive_points` incorrectly in a lot of different places:

https://github.com/RoboJackets/robocup-software/blob/0c0f1448968d65b6429e48ec0e380d0d465aa064/soccer/gameplay/plays/Clear/adaptive_clear.py#L222-L237
https://github.com/RoboJackets/robocup-software/blob/757b8d3b90c290e2da23fd829c3efb279a7969a0/soccer/gameplay/plays/restarts/our_free_kick.py#L54-L57
https://github.com/RoboJackets/robocup-software/blob/757b8d3b90c290e2da23fd829c3efb279a7969a0/soccer/gameplay/plays/Legacy/adaptive_formation.py#L327-L331

However, the `eval_best_receive_point` has a signature of
https://github.com/RoboJackets/robocup-software/blob/0c0f1448968d65b6429e48ec0e380d0d465aa064/soccer/gameplay/evaluation/passing_positioning.py#L106-L115

There are parameters that are missing in these calls to `eval_best_receive_point`

- [ ] This bug has been reproduced by someone else

oswinso:
Related to #1382

PR #1465, opened 2020-05-24
Bug/wall wrong type

### LINKED ISSUE #1448: TypeError in wall tactic
## Description
There appears to be a crash in the wall tactic creating the following stack trace. I think it should be fairly self explanatory it appears to happen semi-frequently when running the full stack. I can try to add more details if the stack trace does not cover it.

```
ERROR:root:Exception occurred when spinning subbehavior named 'form wall': <class 'TypeError'>
  File "/home/clindbeck9/robocup/robocup-software/soccer/gameplay/composite_behavior.py", line 94, in spin
    bhvr.spin()
  File "/home/clindbeck9/robocup/robocup-software/soccer/gameplay/composite_behavior.py", line 77, in spin
    super().spin()
  File "/home/clindbeck9/robocup/robocup-software/soccer/gameplay/fsm.py", line 56, in spin
    state_method()
  File "/home/clindbeck9/robocup/robocup-software/soccer/gameplay/tactics/wall.py", line 66, in execute_defense_wall
    self._remove_wall_defenders()
  File "/home/clindbeck9/robocup/robocup-software/soccer/gameplay/tactics/wall.py", line 87, in _remove_wall_defenders
    for i in range(len(self.active_defenders)):
TypeError: object of type 'int' has no len()
ERROR:root:Play 'WingerWall' encountered an exception: <class 'TypeError'>.  Dropping and temp. blacklisting current play...
  File "/home/clindbeck9/robocup/robocup-software/soccer/gameplay/composite_behavior.py", line 94, in spin
    bhvr.spin()
  File "/home/clindbeck9/robocup/robocup-software/soccer/gameplay/composite_behavior.py", line 97, in spin
    self.handle_subbehavior_exception(name, exc)
  File "/home/clindbeck9/robocup/robocup-software/soccer/gameplay/composite_behavior.py", line 94, in spin
    bhvr.spin()
  File "/home/clindbeck9/robocup/robocup-software/soccer/gameplay/composite_behavior.py", line 77, in spin
    super().spin()
  File "/home/clindbeck9/robocup/robocup-software/soccer/gameplay/fsm.py", line 56, in spin
    state_method()
  File "/home/clindbeck9/robocup/robocup-software/soccer/gameplay/tactics/wall.py", line 66, in execute_defense_wall
    self._remove_wall_defenders()
  File "/home/clindbeck9/robocup/robocup-software/soccer/gameplay/tactics/wall.py", line 87, in _remove_wall_defenders
    for i in range(len(self.active_defenders)):
TypeError: object of type 'int' has no len()
```

oswinso:
## Description
This PR fixes #1448 by adding type annotations to `wall.py`. After fixing the type bug, it turned out that there was another bug in `wall.py` so I fixed that as well =.=

### Run `sim2play`

Expected result: `wall.py` doesn't throw exceptions (though other things do, such as #1463)

### DIFF
--- a/soccer/gameplay/tactics/wall.py
+++ b/soccer/gameplay/tactics/wall.py
@@ -40,7 +40,8 @@ class Wall(composite_behavior.CompositeBehavior):
         self.active_defenders = num_defenders
         self.number_of_defenders = num_defenders
         self.curvature = 1 * curvature
-        self._mark_point = main.ball().pos if mark_point is None else mark_point
+        self._mark_point = main.ball(
+        ).pos if mark_point is None else mark_point
         self._defense_point = defender_point
         self.dist_from_mark = dist_from_mark
         self.defender_spacing = defender_spacing
@@ -83,8 +84,8 @@ class Wall(composite_behavior.CompositeBehavior):
                 self.remove_subbehavior(subbhvr_name)
             self.add_subbehavior(self.WallMove(
                 pt, pt if i != 0 and i != len(pts) - 1 else None),
-                name=subbhvr_name,
-                required=False)
+                                 name=subbhvr_name,
+                                 required=False)
 
     ## Remove wall behaviors
     def _remove_wall_defenders(self) -> None:
@@ -107,16 +108,16 @@ class Wall(composite_behavior.CompositeBehavior):
     ## Finds the point on the arc the defender should move to
     def calculate_destination(self, robot_number: int) -> robocup.Point:
         defender_number = robot_number - self.number_of_defenders / 2 + .5
-        direct: robocup.Point = (
-            self.mark_point - self.defense_point).normalized()
+        direct: robocup.Point = (self.mark_point -
+                                 self.defense_point).normalized()
         arc_angle = defender_number * self.curvature + math.pi / 2
         direct.rotate_origin(arc_angle)
         return self.midpoint - direct * constants.Robot.Radius * self.defender_spacing * defender_number
 
     ## Update wall's midpoint based on the mark and defense points
     def update_midpoint(self) -> None:
-        self.midpoint = self.mark_point + (
-            self.defense_point - self.mark_point).normalized() * self.dist_from_mark
+        self.midpoint = self.mark_point + (self.defense_point - self.mark_point
+                                           ).normalized() * self.dist_from_mark
 
     @property
     def defense_point(self) -> robocup.Point:
@@ -158,7 +159,8 @@ class Wall(composite_behavior.CompositeBehavior):
 
     ## Move behavior that prioritizes the center point of the wall
     class WallMove(skills.move.Move):
-        def __init__(self, pos: Optional[robocup.Point] = None,
+        def __init__(self,
+                     pos: Optional[robocup.Point] = None,
                      center_pt: Optional[robocup.Point] = None):
             super().__init__(pos)
             self.center_pt = center_pt
--- a/soccer/gameplay/tactics/wall.py
+++ b/soccer/gameplay/tactics/wall.py
@@ -82,7 +82,7 @@ class Wall(composite_behavior.CompositeBehavior):
             subbhvr_name = f"robot{i}"
 
             # This logic is so that the center robots are prioritized for role
-            # selection, and we don't leave a whole in the middle
+            # selection, and we don't leave a hole in the middle
             is_center = i != 0 and i != len(pts) - 1
             center_pt = pt if is_center else None
 
--- a/soccer/gameplay/tactics/wall.py
+++ b/soccer/gameplay/tactics/wall.py
@@ -78,9 +78,12 @@ class Wall(composite_behavior.CompositeBehavior):
         # prioritize the middle elements of the wall
         for i in range(len(pts)):
             pt = pts[i]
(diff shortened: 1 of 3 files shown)


soccer/gameplay/tactics/wall.py
```diff
@@ -76,84 +79,93 @@ def _add_wall_defenders(self):
         # prioritize the middle elements of the wall
         for i in range(len(pts)):
             pt = pts[i]
+            subbhvr_name = f"robot{i}"
+            if self.has_subbehavior_with_name(subbhvr_name):
+                self.remove_subbehavior(subbhvr_name)
```
JNeiger: This may have some unintended consequences. Removing and adding a subbehavior removes the `robot_change_cost` in the subbehaviors role requirements which may cause swapping of roles when they get readded. I'll have to look over the code more to see which is better.

If the wall builds it in such a way where the numbers ascend from one side of the wall to the other, it's better to just add the new one to the end of the list. This way the rest of the wall stays somewhat intact and the new robot moves to the edge of the wall.

If the wall builds it such that robot0 is the center, we should remove all and add all the subbehaviors again since the middle one is the most important.

soccer/gameplay/tactics/wall.py
```diff
@@ -76,84 +79,93 @@ def _add_wall_defenders(self):
...
+    def _remove_wall_defenders(self) -> None:
         self.update_midpoint()
-        for i in range(len(self.active_defenders)):
+        for i in range(self.active_defenders):
             name = "robot" + str(i)
             pt = self.calculate_destination(i)
             if i < self.number_of_defenders:
-                self.subbehavior_with_name(name).pos = pt
+                move_subbhvr = self.subbehavior_with_name(name)
+                assert isinstance(move_subbhvr, self.WallMove)
```
JNeiger, line 112: I'll comment here to bring up a conversation about asserts in our python code. @kylestach @chachmu @liquidmetal9015 @HussainGynai 

When things break, it's more obvious where, but if something fucks up in comp and it may have still worked, it'll cause a blacklist loop and we'll have to timeout

soccer/gameplay/tactics/wall.py
```diff
@@ -1,4 +1,6 @@
+from typing import Optional
 import composite_behavior
+from role_assignment import RoleRequirements
```
JNeiger: `from ... import ..` should probably be `import ...`

soccer/gameplay/tactics/wall.py
JNeiger: Honestly, based on the `_remove_wall_defenders`, this should be an if statement based on if i is > self.active_defenders, and add the subbehavior, otherwise, just update the pt

soccer/gameplay/tactics/wall.py
oswinso: What about a util that just changes between real assert and no op depending on whether we're in an actual game or not?

soccer/gameplay/tactics/wall.py
oswinso: Ah, I see. I forgot you could update the point directly.

soccer/gameplay/tactics/wall.py
kylestach: I usually tend towards assert for states that should be unreachable (aka statically-guaranteed). This should be one of those cases - basically, if this is failing from a particular call site, it will _always_ fail from that call site, so it is okay.

soccer/gameplay/tactics/wall.py
oswinso: done

soccer/gameplay/tactics/wall.py
```diff
@@ -76,84 +79,106 @@ def _add_wall_defenders(self):
...
+            # so we need to check that we're not adding a duplicate
+            if self.has_subbehavior_with_name(subbhvr_name):
+                move_subbhvr = self.subbehavior_with_name(subbhvr_name)
+                assert isinstance(move_subbhvr, self.WallMove)
+
+                # Update pos and center_pt instead of removing and adding
+                # the subbehavior so it doesn't cause swapping of roles
+                move_subbhvr.pos = pt
+                move_subbhvr.center_pt = center_pt
+                continue
```
JNeiger: Functionally this works, but a continue feels weird instead of just doing an else, up to you if you want to change it

JNeiger (approved):
Looks good!

soccer/gameplay/tactics/wall.py
oswinso: ok yeah I'll change it to an else

soccer/gameplay/tactics/wall.py
```diff
@@ -76,84 +79,108 @@ def _add_wall_defenders(self):
...
         for i in range(len(pts)):
             pt = pts[i]
-            self.add_subbehavior(self.WallMove(
-                pt, pt if i != 0 and i != len(pts) - 1 else None),
-                                 name="robot" + str(i),
-                                 required=False)
+            subbhvr_name = f"robot{i}"
+
+            # This logic is so that the center robots are prioritized for role
+            # selection, and we don't leave a whole in the middle
```
kylestach: *hole

kylestach (approved):
LGTM

PR #1467, opened 2020-05-25
Processor logging refactor

### LINKED ISSUE #1431: Isolating Protobuf

kylestach:
## Description
This PR accomplishes several changes:
 - Makes Logger a node
 - Removes logging code from the remainder of the codebase (now Logger is the only thing generating the protobuf tree)
 - Makes separate RadioStatus struct to replace RadioRx (protobuf struct) as the standard in-memory representation of a returned packet

It also contains several fixes to logging things that we broke along the way:
 - Strip charts and the protobuf tree
 - Pausing now properly stops Processor

## Associated Issue
Issue #1431 

## Design Documents
[ROS Roadmap](https://docs.google.com/document/d/1wix-ZK9NPBY0zC8T9LyUihT1ONubXMKxHvb8KYmJkNQ/edit)

### Regression testing
There should be no regressions on:
 - Basic soccer functionality
 - Field view and robot status list
 - Behavior tree
 - Protobuf tree*
 - Strip chart*
 - Saving logs to file*
 - Viewing logs*
 - Pausing/resuming

*these items may have been broken previously


soccer/ui/MainWindow.cpp
```diff
@@ -306,7 +314,7 @@ void MainWindow::updateViews() {
 
     GameState game_state;
     {
-        std::lock_guard<std::mutex> lock(_context_mutex);
+        std::lock_guard<std::mutex> lock(*_context_mutex);
         game_state = _context->game_state;
     }
```
oswinso, line 319: Can we not have to wait on this mutex? We sometimes spend a lot of time waiting on this mutex here. Maybe use a SPSC queue or something?

soccer/ui/MainWindow.cpp
```diff
@@ -430,32 +457,19 @@ void MainWindow::updateViews() {
...
+    if (live_frame) {
+        updateDebugLayers(*live_frame);
     }
 
     // Get the frame at the log playback time
-    const std::shared_ptr<LogFrame> currentFrame = _history[0];
+    const std::shared_ptr<LogFrame> currentFrame = _history.back();
 
     if (currentFrame) {
+        _ui.logTree->message(*currentFrame);
```
oswinso: This call is taking a long time. What exactly does the protobuf tree do? Can we do this in a separate thread or something?

soccer/ui/MainWindow.cpp
```diff
@@ -544,234 +536,290 @@ void MainWindow::updateViews() {
...
-            statusWidget->setShowstopper(showstopper);
+                // if there is an error bad enough that we should get this robot
+                // off the field, alert the user through the UI that there is a
+                // "showstopper"
+                bool showstopper = !hasVision || !hasRadio || hasMotorFault ||
+                                   kickerFault || ballSenseFault ||
+                                   (batteryLevel < 0.25) || !fpgaWorking;
+                statusWidget->setShowstopper(showstopper);
 
 #endif
```
oswinso: This entire section is a monstrosity. Can you like break this out into multiple functions so it's readable?

oswinso (commented):
Just some comments but oh god UI fps is so low

soccer/ui/MainWindow.cpp
kylestach: We spend a lot of time waiting on this mutex because we have to wait for processor to finish its cycle. However, this shouldn't bring frame rate down substantially below processor's frame rates on its own.

soccer/ui/MainWindow.cpp
kylestach: It's using a lot of reflection and updating GUI, which is probably why it's so slow. I could make this only update every few frames - wouldn't really _fix_ the problem, but would at least make it better. Is it taking a substantial portion of frame times?

soccer/ui/MainWindow.cpp
kylestach: I'd been hoping to not have to do much in the way of trying to clean this up, because it's all about to be rewritten (@chachmu). Yeah it's a mess, but it's a mess that's about to be deleted.

kylestach:
How bad are the frame rates you're getting? I'm getting 20-40 which is pretty much on par with before this patch. Yes, our GUI code is horribly unoptimized, and yes this probably makes it slightly worse (given that apparently we previously had no synchronization whatsoever on log frames?), but it's not something I'm worried about unless you're seeing a more serious regression. Doubly so given that a lot of the GUI code will soon go away.

common/protobuf/RadioRx.proto
```diff
@@ -30,6 +30,8 @@ enum HardwareVersion
 	RJ2008 = 1;
 	RJ2011 = 2;
 	RJ2015 = 3;
+    RJ2018 = 4;
+    Simulation = 5;
```
garethellis0, line 30: bit of a nit, but indenting here?

soccer/Logger.cpp
```diff
@@ -1,141 +1,275 @@
...
 
-void Logger::close() {
-    QWriteLocker locker(&_lock);
-    if (_fd >= 0) {
-        ::close(_fd);
-        _fd = -1;
-        _filename = QString();
+void Logger::start() { _context->logs.start_time = RJ::now(); }
+
+void Logger::run() {
```
garethellis0, line 101: Seems like there are several fairly distinct sections to this (fairly long) function. Perhaps consider breaking it up into a few calls to smaller functions?

soccer/Logger.cpp
```diff
@@ -1,141 +1,275 @@
...
+    } else {
+        log_frame->set_team_name_yellow(_context->game_state.OurInfo.name);
+        log_frame->set_team_name_blue(_context->game_state.TheirInfo.name);
+    }
+
+    log_frame->set_timestamp(RJ::timestamp());
+
+    if (_context->logs.state == Logs::State::kWriting) {
+        _context->logs.size_bytes += log_frame->ByteSize();
+        google::protobuf::io::OstreamOutputStream output(&_log_file.value());
```
garethellis0: Might want to check whether the optional is valid before getting it's value like this, though I do see that it should be if you're writing.

oswinso:
I'm seeing like 15 fps with heavy spikes to like 5 fps sometimes

kylestach:
Interesting. You mentioned some timing specifics, do you have profiling results from your machine?

oswinso:
actually nvm, disregard what I said. I think I forgot to charge my laptop when I ran it and my cpu was throttling because of low battery lmao

soccer/CMakeLists.txt
```diff
@@ -85,7 +85,7 @@ set(ROBOCUP_LIB_SRC
     "vision/robot/WorldRobot.cpp"
     "vision/util/VisionFilterConfig.cpp"
     "vision/VisionFilter.cpp"
-    "WindowEvaluator.cpp")
+    "WindowEvaluator.cpp" radio/RobotStatus.hpp)
```
oswinso: I don't think this should be here

kylestach:
Okay, will do. Hopefully most of this will get cleaned up soon anyway, and once we're based off of ROS/rqt it'll be decoupled from the processor updates which would be nice.

garethellis0 (commented):
Not sure what your folks procedure is for validating changes like this, but it might be worth considering some unit tests (at least for the stuff that's easily unit-testable, like serialization/de-serialization).

soccer/CMakeLists.txt
JNeiger: Formatting

JNeiger (approved):
There's a lot going on and changing, so I haven't been able to follow it as well as I usually do. Nothing stands out as wrong though.

The regression testing list looks like it covers everything so if it passes, we should be good to go

soccer/Processor.cpp
```diff
@@ -87,13 +85,15 @@ Processor::Processor(bool sim, bool blueTeam, const std::string& readLogFile)
...
-        firstLogTime = _logger.startTime();
+        _logger.read(readLogFile);
     }
 
+    _logger.start();
+
     _nodes.push_back(_visionReceiver.get());
     _nodes.push_back(_motionControl.get());
     _nodes.push_back(_grSimCom.get());
+    _nodes.push_back(&_logger);
```
oswinso: Why not use a `std::unique_ptr` like all the other nodes?

soccer/ui/FieldView.cpp
```diff
@@ -401,14 +404,15 @@ void FieldView::drawTeamSpace(QPainter& p) {
 
     /// populate @cometTrails with the past locations of each robot
     int pastLocationCount = 40;  // number of past locations to show
-    for (int i = 0; i < pastLocationCount + 1 && i < _history->size(); i++) {
+    int start = std::max(0, (int)_history->size() - pastLocationCount);
```
oswinso: `static_cast` ?

soccer/ui/MainWindow.cpp
```diff
@@ -544,234 +536,290 @@ void MainWindow::updateViews() {
...
+                _ui.robotStatusList->setItemWidget(item, statusWidget);
+
+                // set shell ID
+                statusWidget->setShellID(shell);
+
+                // set team
+                statusWidget->setBlueTeam(
+                    _context->game_settings.requestBlueTeam);
+
+                // TODO(Kyle): set board ID
```
oswinso: Make an issue for it and include the issue number?

oswinso (changes requested):
Just the one thing about the issue number

soccer/CMakeLists.txt
kylestach: Sigh. CLion keeps adding those automatically.

soccer/ui/MainWindow.cpp
kylestach: We actually don't have board IDs anymore. I'm going to do one more pass through this and clean up the variables that don't apply anymore from the log frame, considering this is already sort of breaking backwards compatibility anyway I'm okay with that.

oswinso (approved):
lgtm

Issue #1455, opened 2020-05-15
Remove Standard Play


HussainGynai:
## Description
Now that we have removed default defense, @JNeiger suggested that we remove standard play since it was only used to setup default defense. This will mean switching everything to inherit from play.Play instead. 
## Requirements

 - Remove standard play
 - switch to inheriting from play.Play

PR #1477, opened 2020-05-28
Add coordinated block to many plays, remove goalie management from root_play, reorganize plays folder

liquidmetal9015:
## Description
This PR does a couple of things that I came across when I was re-adding defense to many of our plays.
(I probably should have tried to break it into multiple PRs but too late now)

I've gone through a ton of plays and added coordinated block as a sub-behavior in the init so they behave approximately like they did with standard defense. This was a bit more invasive than it should be in some cases due to the use of remove_all_subbehaviors() in many places.

I've gone into root play and removed the goalie management and removed the last of the provides_goalie and use_defense flags from the play inheritance tree. Keeping in mind that this wan't actually used almost ever because every standard play provided it's own goalie anyway through defense.py and it assigns what I understand to be an outdated goalie anyway even if it did ever trigger.

~~I've added a dict in root play that keeps track of each time a play is dropped through error handling and plays will not be considered for selection if they have been dropped more than x times. The counts all reset if there is a frame where no play is selected due to the filtered and enabled play list being empty.~~

As a result of difficulties adding coordinated block two_sided_attack and distraction play are both broken and I've made them raise exceptions when you try to run them. ~~Which coincidentally works as a good test for the blacklist functionality as if you include them along with valid plays they will get blacklisted within a few frames for being broken.~~

I have also reorganized the plays folder, the restarts folder is gone, the legacy folder no longer has an __init__.py in it (along with skel and training) and some plays are renamed to fall in line with the naming convention of the others. I think this generally makes sense and cleans up the plays window a bit. I'm still a bit apprehensive about testing being fully expanded by default since it has a lot of junk in it (some of which results in an instant crash or is totally out of date).

So after this, when you write a play you are responsible for ALL 6 ROBOTS, INCLUDING THE GOALIE. I do think this is a very positive change overall, as now there are no spooky forces outside of your control moving robots for you (like literally 4 layers of checks every frame if there was defense and/or a goalie that you had to explicitly opt out of) and I think it will give us more creative flexibility to swap out defense and goalie behaviors as well as making the code cleaner and more understandable. Now when you write a play the first step will probably be "add defense and a goalie" which hopefully we will soon have a number of defensive and goalie tactics to choose from for each of those spots.

This means that you really shouldn't be using remove_all_subbehaviors() at all anymore unless you really want to take out all the subbehaviors which I think is a positive change as manually managing the subbehaviors across the states and transitions makes it much more clear what the code is actually doing.

One thing I'm considering in general is making some warnings about common mistakes that plays could have, like checking to see if there is a goalie assigned via some method. If we could get something that would check if there is a goalie assigned and that there are exactly 6 robots assigned sub-behaviors that would be good. I think that if we are absolutely adamant about having safeguards against not having defense/goalie assigned we should just blacklist plays that fail those checks.

I would also like to potentially add a like one line "Dead goalie" that just moves the goalie to the goal box for use in like the stopped play for example if we wanted to have a goalie check.

Feel free to chime in if any of this doesn't sit right with you I've just kind of YOLO'd and done it and am seeing if it passes review. 

I'll probably take another look at everything myself but I want to go ahead and get this PR out so people can start looking at it.

## Associated Issue
Issue #1455 is related

## Steps to test
You kind of need to test everything relating to game play. Testing most of the individual plays for behavior should suffice.
Especially the plays that have remove_all_behaviors removed and replaced with individual removes. 

Expected result: Everything works like we think it should.


liquidmetal9015:
Oh yeah and this is going to be failing a bunch of checks for awhile so I'm going to have to fix that over a period of time.

I'll make some comments on some things later when I have a little bit of time.

soccer/gameplay/plays/Attack_Goal/adaptive_attack.py
```diff
@@ -195,14 +201,16 @@ def dribbler_has_ball(self):
         return any(r.has_ball() for r in main.our_robots())
 
     def on_enter_collecting(self):
-        self.remove_all_subbehaviors()
+        #We shouldn't need this here, I suspect this was a quick bug fix
+        #self.remove_all_subbehaviors()
```
JNeiger: It was probably from the legacy version, it's good to remove

soccer/gameplay/plays/Attack_Goal/adaptive_attack.py
```diff
@@ -195,14 +201,16 @@ def dribbler_has_ball(self):
...
 
         # 2 man to man defenders and 1 zone defender
         defensive_forward = tactics.defensive_forward.DefensiveForward()
         self.add_subbehavior(defensive_forward, 'defend', required=True)
 
     def on_exit_collecting(self):
-        self.remove_all_subbehaviors()
+        self.remove_subbehavior('defend')
+        #self.remove_all_subbehaviors()
```
JNeiger: Good to remove

soccer/gameplay/plays/Attack_Goal/distraction_attack.py
```diff
@@ -160,11 +160,13 @@ def __init__(self):
             subbehavior_with_name('shooting').is_done_running(), 'repeat')
 
     def on_enter_setup(self):
-        self.remove_all_subbehaviors()
         #capture ball and get striker and distractor in position
         self.add_subbehavior(
             skills.capture.Capture(), 'capture', required=True)
 
+        #I've put this here instead of in init in part to test the new blacklist system
+        raise NotImplementedError("This play has subbehavior removal issues, has no defense, and needs some work in general")
```
JNeiger: We should probably add issues for these

soccer/gameplay/plays/Attack_Goal/two_sided_attack.py
```diff
@@ -88,6 +89,10 @@ def score(cls):
             return 9 if main.game_state().is_playing() else float("inf")
 
     def on_enter_setup(self):
+   
+        #I've put this here partially to test the blacklist system
+        raise NotImplementedError("This play does not have a defense and must be fixed before it can be used")
```
JNeiger: Issue as well

soccer/gameplay/plays/Clear/basic_clear.py
```diff
@@ -100,3 +109,13 @@ def on_enter_clear_ball(self):
...
+
+    def on_exit_clear_ball(self):
+        self.remove_subbehavior('clear')
+        self.remove_subbehavior('keep moving')
+
+
+
+
+
+
```
JNeiger: Tons of extra spaces

soccer/gameplay/plays/legacy/our_free_kick.py
```diff
@@ -13,14 +13,14 @@
...
+class OurFreeKick(play.Play):
     Running = False
     BumpKickPower = 0.01
     FullKickPower = 1
     MaxShootingAngle = 80
     # Untested as of now
-    MaxChipRange = 3
-    MinChipRange = 0.3
+    MaxChipRange = constants.Robot.Chipper.MaxChipRange
+    MinChipRange = constants.Robot.Chipper.MinChipRange
```
JNeiger, line 23: This should just get fully refactored out, this was before we had the chipper constants

soccer/gameplay/root_play.py
```diff
@@ -2,13 +2,13 @@
...
 from behavior import Behavior
-import plays.stopped
+import plays.Stopped.basic_stopped
 import plays.testing.test_coach
 import logging
 from PyQt5 import QtCore
 import main
 import evaluation.double_touch
-import tactics.positions.goalie
+#import tactics.positions.goalie
```
JNeiger: Remove

soccer/gameplay/root_play.py
```diff
@@ -2,13 +2,13 @@
 from play import Play
 import robocup
 from behavior import Behavior
-import plays.stopped
+import plays.Stopped.basic_stopped
```
JNeiger: The caps don't match the rest of the codebase

soccer/gameplay/play.py
```diff
@@ -1,5 +1,8 @@
 import composite_behavior
-
+from typing import List
+from situations import Situation
```
JNeiger, line 3: Should switch to `import situations as situations` if you really want the shortened name

soccer/gameplay/constants.py
```diff
@@ -29,6 +29,10 @@ class Robot:
     ChipClearance = (
         .1, .6)  # min and max distance a chip will go over another robot
 
+    class Chipper:
+        MinChipRange = 0.3
+        MaxChipRange = 3
```
JNeiger, line 34: So there's 3 different versions of chip stuff now, `Robot.Chipper.*`, `Robot.ChipClearance` and `OurChpiping.*`. These should be combined

JNeiger (changes requested):

soccer/gameplay/plays/Attack_Goal/distraction_attack.py
oswinso: Yeah. I agree. Any new todos need issue numbers

soccer/gameplay/root_play.py
```diff
@@ -30,6 +30,11 @@ def __init__(self):
         self._currently_restarting = False
         self._robots: List[robocup.Robot] = []
 
+        ##Counts the number of times a given play has been dropped
+        self.play_drop_count = dict()
+        ##Once a play has been dropped more than this number of times, it won't be considered to be run in the future
+        self.play_drop_threshold = 3
```
HussainGynai: Do we want this to be a set number for the entire play base. I could also see this being a thing that each play has its own threshold for.

HussainGynai: Also how long does blacklisting a play in this manner last, like will it never be picked for the rest of the match?

HussainGynai (changes requested):
Looks pretty good, just had some questions about how we want to handle the actual blacklisting of plays once they've been dropped a certain number of times. Also run style.

soccer/gameplay/root_play.py
liquidmetal9015: So the way I've rearranged this is that Stopped is basically a situation folder, and all the situation folders are capitalized. See the link below for the file structure under this PR. Every folder that is capitalized has an __init__.py in it and the folders that are lower-case do not.

It might not be super obvious that these changes have been made due to how github displays "name changes" but I would like people to look at this reorganization in general.

[Everyone make sure to check it out to see if we are ok with these changes](https://github.com/RoboJackets/robocup-software/tree/reAddDefense/soccer/gameplay/plays)

(legacy plays have been moved to lowercase legacy, no __init__.py, stopped.py which was in no directory has been moved to Stopped, the restarts folder has been emptied and it's contents distributed amongst other folders. Amongst other things)

soccer/gameplay/constants.py
liquidmetal9015: Yeah I was just trying to do this quickly to remove a dependency where a play imported these parameters from another play (like they were in the body of another play, arbitrarily assigned there) that got renamed without looking too much. I'll try to clean it up if we are having chipper parameter set fragmentation.

soccer/gameplay/root_play.py
liquidmetal9015: So I think we have decided to remove this from this PR and move it to another one to evaluate if we want something more advanced. But this simple implementation just keeps track of all the plays in the play registry and only resets the count when there is a frame where there are no valid plays. If we port this to another PR it would also work that way by default. We could have it decay or something but realistically I think that since our matches aren't that long anyway and we can call time outs to fix issues, if a play is crashing more than a few times we should just kill it and risk a less optimal play running rather than give more chances to a play that is crashing over and over. We should ideally have quite a few plays overall so the loss of a play or two shouldn't really affect us that much.

soccer/gameplay/plays/Attack_Goal/distraction_attack.py
liquidmetal9015: Will do. These plays might need some more in-depth work in general to be in the running to actually be used so we might want to include that in those issues if there are obvious fixes.

soccer/gameplay/root_play.py
kylestach: I'm not a fan of this. IMO the best solution is to stop making directory names correspond 1:1 with display names, i.e. by exporting a string from `__init__.py` for the situation folder name is.

liquidmetal9015:
I'm going to change directory names to lower case to make imports fall in with general python convention.

HussainGynai (approved):
looks good

PR #1480, opened 2020-05-31
Created viz for heading

oswinso:
## Description
This PR creates a visualization for the robot heading, so we can easily tell which direction the robots are facing.

Screenshot from 2020-05-31 15-34-51

### Run sim

Expected result: There's a line indicating which direction the robots are facing.


soccer/gameplay/visualization/robot_viz.py
```diff
@@ -0,0 +1,38 @@
...
+    """ Draws a line representing the heading of the robot
+    :param robot: Robots to draw heading for
+    :param debug_drawer: DebugDrawer
+    :param r: How long to draw the heading line
+    :param color: What color to draw the heading line
+    """
+    theta = robot.angle
+    endpoint = robot.pos + robocup.Point(r * cos(theta), r * sin(theta))
+    heading_line = robocup.Line(robot.pos, endpoint)
+    debug_drawer.draw_line(heading_line, color, "Heading")
```
JNeiger: This can be main.debug_drawer()

soccer/gameplay/visualization/robot_viz.py
```diff
@@ -0,0 +1,38 @@
...
+                   their_robots: Optional[List[robocup.OpponentRobot]],
+                   debug_drawer: robocup.DebugDrawer):
+    """ Performs robot visualization agnostic to any play
+    :param our_robots: List of our robots
+    :param their_robots: List of their robots
+    :param debug_drawer: Instance of robocup.DebugDrawer
+    :return:
+    """
+    assert our_robots is not None
+    assert their_robots is not None
```
JNeiger: If you don't want nones, is it possible to just change the param list to remove the optionals?

soccer/gameplay/main.py
```diff
@@ -227,6 +228,7 @@ def run():
         if root_play() is not None:
             situationAnalysis.updateAnalysis()
             root_play().spin()
+        robot_viz.draw_robot_viz(our_robots(), their_robots(), debug_drawer())
```
JNeiger: Personally, if this is just being drawn every frame, I think this fits better in the GUI. 

https://github.com/RoboJackets/robocup-software/blob/fa29951e7c3a3de608252b77511093a15ce0bbd8/soccer/ui/FieldView.cpp#L654

soccer/gameplay/visualization/robot_viz.py
```diff
@@ -0,0 +1,38 @@
...
+    :param r: How long to draw the heading line
+    :param color: What color to draw the heading line
+    """
+    theta = robot.angle
+    endpoint = robot.pos + robocup.Point(r * cos(theta), r * sin(theta))
+    heading_line = robocup.Line(robot.pos, endpoint)
+    debug_drawer.draw_line(heading_line, color, "Heading")
+
+
+def draw_robot_viz(our_robots: Optional[List[robocup.OurRobot]],
```
JNeiger: If it's just to draw the headings, I would probably change the name to reflect that. If there's going to be more, then it's probably fine

soccer/gameplay/visualization/robot_viz.py
```diff
@@ -0,0 +1,38 @@
...
+                 r: float = 0.5,
+                 color: Tuple[int, int, int] = (0xff, 0xc1, 0xda)):
+    """ Draws a line representing the heading of the robot
+    :param robot: Robots to draw heading for
+    :param debug_drawer: DebugDrawer
+    :param r: How long to draw the heading line
+    :param color: What color to draw the heading line
+    """
+    theta = robot.angle
+    endpoint = robot.pos + robocup.Point(r * cos(theta), r * sin(theta))
```
JNeiger: You should be able to do r*robocup.Point

JNeiger: There's a robocup.Point.Direction(theta) that does the cos and sin automatically

soccer/gameplay/visualization/robot_viz.py
oswinso: I'm not too sure about the circular imports though... it feels kind of wrong imo

soccer/gameplay/main.py
oswinso: ah ok, I guess that makes more sense.

soccer/gameplay/visualization/robot_viz.py
oswinso: it is possible, but that means that I would have to change the getters to not return `None`

soccer/gameplay/main.py
kylestach: I would agree with Joe.

kylestach:
I would put this in FieldView.cpp. Also, I would prefer a visualization like we had for the viewer on the RL project (visualize the kicker bar) - I think that is going to get a lot less cluttered when you have a lot of robots. You could just as well make it turn green/red depending on whether or not the robot has the ball (in which case we could get rid of the other visualization that does that, which is a red circle but gets confusing because there are multiple things that draw red circles of different sizes).

oswinso:
@kylestach I prefer the line though in this case, just because the robots are much smaller. Even if we do add a kicker bar, you would probably still need to strain your eyes to see exactly which direction it was. Also, I think you can distinguish between smaller changes in angles when there's a line sticking out vs if there is just a small bar.

JNeiger (approved):

kylestach:
That's fair. Maybe we could have a kicker bar always displayed in the GUI and then a debug layer with lines like this (like you have it now), so it can be turned off?

oswinso:
I'm too lazy to do the kicker bar stuff rn, can we make that a separate PR?

oswinso:
@kylestach do we want to make it a debug layer? Or should we have it displayed always?

kylestach:
@OswinSo I'd rather it be an optional debug layer.

oswinso:
@kylestach `FieldView` doesn't have access to `DebugDrawer` though. Do we want `FieldView` to have a pointer to `Context`? Or how are you thinking of having this work?

kylestach:
Yeah, I think reverting to the python version if it is an optional debug layer would be better. @JNeiger thoughts?

JNeiger:
Let's back up a second, what is this solving? Do we have trouble figuring out which directions robots are facing and need a better display? Or do we need this feature for debugging purposes within plays/planning/motioncontrol?

For the former, that should definitely be in fieldview imo. If we don't like the line, we should just choose a better way to show it.

For the latter, we can put it into python, but I don't think it should be in main.py. I think it should be called in whatever play/tactic/skill that cares about displaying that as debug.

I think I'm mostly against putting something into main.py that doesn't fit there. It's easy to get another processor type situation where anything that doesn't really have an obvious place gets throw into there.

oswinso:
My intention with this was simply as a way to see what direction the robots were pointing. I think it would be something that would be useful at all times, so I don't think it should be something that's under debug drawer. Sure, the robots look a bit more retarded, but I don't see why you wouldn't want to know what direction each robot was facing.

kylestach:
I mean, my personal opinion is that the line is a bit cluttered, but we can reevaluate that if we actually decide that it's annoying. For now I guess I'm fine with this.

kylestach (approved):

Issue #1435, opened 2020-05-05
Pass Tactic


HussainGynai:
## Description
A tactic that will take in some sort of pass point heuristic and pass to the point determined by it. 
See [design doc](https://docs.google.com/document/d/1vE7y_K5OppqjfXBsLNgMPDJiGpCbL5azZzDsPcYiDrw/edit?usp=sharing) for more details.

PR #1484, opened 2020-06-12
Pass heuristic optimizer

HussainGynai:
## Description
Created a function called findPassPoint that takes in a heuristic and finds the minima and returns it as a point to be used by coordinated pass.

## Associated Issue
Issue #1435 

## Design Documents
[Link](https://docs.google.com/document/d/1vE7y_K5OppqjfXBsLNgMPDJiGpCbL5azZzDsPcYiDrw/edit?usp=sharing)

### Test Case 1
1. Run the test heuristic
2. Create your own heuristics to see if they work

Expected result: the test heuristic play should pass to (0.915239, 1.38493)

### DIFF
--- a/soccer/gameplay/plays/testing/TestHeursitc.py
+++ b/soccer/gameplay/plays/testing/TestHeursitc.py
@@ -1,0 +1,21 @@
+import main
+import robocup
+import behavior
+import constants
+import standard_play
+import tactics.coordinated_pass
+from tactics.find_pass_point import findPassPoint
+from tactics.test_cost_function import testCostFunction
+
+## A test of the pass heuristic system using cost function optimization
+# should pass to around (0.915239, 1.38493)
+class TestHeuristic(standard_play.StandardPlay):
+
+    def __init__(self) -> None:
+        super().__init__(continuous=False)
+
+    def on_enter_start(self) -> None:
+        pass_point = findPassPoint(testCostFunction)
+
+        self.add_subbehavior(
+            tactics.coordinated_pass.CoordinatedPass(pass_point), 'pass')
--- a/soccer/gameplay/plays/testing/TestHeursitc.py
+++ b/soccer/gameplay/plays/testing/TestHeursitc.py
@@ -4,18 +4,19 @@ import behavior
 import constants
 import standard_play
 import tactics.coordinated_pass
-from tactics.find_pass_point import findPassPoint
-from tactics.test_cost_function import testCostFunction
+from tactics.find_pass_point import find_pass_point
+from tactics.test_cost_function import create_test_cost_function
+
 
 ## A test of the pass heuristic system using cost function optimization
 # should pass to around (0.915239, 1.38493)
 class TestHeuristic(standard_play.StandardPlay):
-
     def __init__(self) -> None:
         super().__init__(continuous=False)
 
     def on_enter_start(self) -> None:
-        pass_point = findPassPoint(testCostFunction)
+        pass_point = find_pass_point(create_test_cost_function(0.4, 0.6),
+                                     (1, 1), 100)
 
         self.add_subbehavior(
             tactics.coordinated_pass.CoordinatedPass(pass_point), 'pass')
--- a/soccer/gameplay/tactics/find_pass_point.py
+++ b/soccer/gameplay/tactics/find_pass_point.py
@@ -1,0 +1,24 @@
+import robocup
+import constants
+from scipy.optimize import minimize
+
+## Function that optimizes cost function for passing
+# uses scipy minimize using the BFGS method
+def findPassPoint(heuristic) -> robocup.Point:
+    width = constants.Field.Width
+    length = constants.Field.Length
+    initial_guess = [0.00, 0.00]
+    result = minimize(heuristic, initial_guess, bounds=((-width/2.00, width/2.00), (0.00, length)), tol=1e-3)
+    if (not result.success):
+        raise OptimizationError()
+    point = robocup.Point(result.x[0], result.x[1])
+    print(point)
+    return point
+
+## Exception for when optimization fails
+# without throwing an exception the bot will just pass to 0,0
+class OptimizationError(Exception):
+
+    def __init__(self, message='Error during optimization of cost function, please check expression being optimized for mathematical errors'):
+        self.message = message
+        super().__init__(self.message)
--- a/soccer/gameplay/tactics/find_pass_point.py
+++ b/soccer/gameplay/tactics/find_pass_point.py
@@ -1,24 +1,37 @@
 import robocup
 import constants
 from scipy.optimize import minimize
+from typing import Callable, Tuple, Optional, Any, Sequence
+
 
 ## Function that optimizes cost function for passing
 # uses scipy minimize using the BFGS method
-def findPassPoint(heuristic) -> robocup.Point:
+def find_pass_point(heuristic: Callable[[Tuple[float, float]], float],
+                    initial_guess: Optional[Tuple[float, float]] = (0, 0),
+                    max_iter: Optional[int] = None) -> robocup.Point:
     width = constants.Field.Width
     length = constants.Field.Length
-    initial_guess = [0.00, 0.00]
-    result = minimize(heuristic, initial_guess, bounds=((-width/2.00, width/2.00), (0.00, length)), tol=1e-3)
-    if (not result.success):
-        raise OptimizationError()
+    if (max_iter is not None):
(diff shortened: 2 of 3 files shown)


soccer/gameplay/tactics/find_pass_point.py
```diff
@@ -0,0 +1,24 @@
+import robocup
+import constants
+from scipy.optimize import minimize
+
+## Function that optimizes cost function for passing
+# uses scipy minimize using the BFGS method
+def findPassPoint(heuristic) -> robocup.Point:
```
oswinso: `heuristic` needs a type too

soccer/gameplay/tactics/find_pass_point.py
```diff
@@ -0,0 +1,24 @@
...
+    width = constants.Field.Width
+    length = constants.Field.Length
+    initial_guess = [0.00, 0.00]
```
oswinso: Why is this the initial guess? Let the user put their own initial guess

soccer/gameplay/tactics/find_pass_point.py
```diff
@@ -0,0 +1,24 @@
...
+    result = minimize(heuristic, initial_guess, bounds=((-width/2.00, width/2.00), (0.00, length)), tol=1e-3)
+    if (not result.success):
+        raise OptimizationError()
+    point = robocup.Point(result.x[0], result.x[1])
+    print(point)
```
oswinso: Remove this debugging print

soccer/gameplay/plays/testing/TestHeursitc.py
```diff
@@ -0,0 +1,21 @@
...
+import behavior
+import constants
+import standard_play
+import tactics.coordinated_pass
+from tactics.find_pass_point import findPassPoint
+from tactics.test_cost_function import testCostFunction
+
+## A test of the pass heuristic system using cost function optimization
+# should pass to around (0.915239, 1.38493)
+class TestHeuristic(standard_play.StandardPlay):
```
oswinso, line 13: I don't think this should be a play...

oswinso (changes requested):
Some comments

soccer/gameplay/plays/testing/TestHeursitc.py
kylestach: This is just a test for the functionality introduced here. It's fine for now.

soccer/gameplay/tactics/find_pass_point.py
kylestach: Agreed.

soccer/gameplay/tactics/find_pass_point.py
kylestach: It should be `find_pass_point(heuristic)`, and should probably be a higher-order function. Weren't we going to have `CoordinatedPass` accept a heuristic function rather than a single target point, so we could later extend it to optimize continuously? Am I mistaken there?

soccer/gameplay/tactics/find_pass_point.py
```diff
@@ -0,0 +1,24 @@
...
+
+## Function that optimizes cost function for passing
+# uses scipy minimize using the BFGS method
+def findPassPoint(heuristic) -> robocup.Point:
+    width = constants.Field.Width
+    length = constants.Field.Length
+    initial_guess = [0.00, 0.00]
+    result = minimize(heuristic, initial_guess, bounds=((-width/2.00, width/2.00), (0.00, length)), tol=1e-3)
+    if (not result.success):
+        raise OptimizationError()
```
kylestach: I'm okay with this raising an exception as long as the only thing that calls it directly is the pass tactic, but that's not how it's currently set up. See my above comment.

Otherwise, people are probably not going to stick the exception handling boilerplate everywhere they use this, because it's not a regular program case and people are lazy.

soccer/gameplay/plays/testing/TestHeursitc.py
HussainGynai: Yup what Kyle said

soccer/gameplay/tactics/find_pass_point.py
HussainGynai: So we've gone back and forth if we want this to be in coordinated pass or not, I could put it in coordinated pass if we wanted

soccer/gameplay/tactics/find_pass_point.py
```diff
@@ -0,0 +1,24 @@
...
+import constants
+from scipy.optimize import minimize
+
+## Function that optimizes cost function for passing
+# uses scipy minimize using the BFGS method
+def findPassPoint(heuristic) -> robocup.Point:
+    width = constants.Field.Width
+    length = constants.Field.Length
+    initial_guess = [0.00, 0.00]
+    result = minimize(heuristic, initial_guess, bounds=((-width/2.00, width/2.00), (0.00, length)), tol=1e-3)
```
JNeiger: I think we should set the max num iterations arg. I'm not sure whether it should just be hard coded or passed in as a param though. The "correct" value can change significantly depending on the speed of execution of the heuristic.

soccer/gameplay/tactics/test_cost_function.py
```diff
@@ -0,0 +1,9 @@
+import robocup
+import constants
+import math
+
+## A test cost function for passes that should have a minima at (0.915239, 1.38493)
+def testCostFunction(coordinates) -> float:
```
JNeiger: If the heuristic wants extra args, is the plan to bind them similar to std::bind leaving the coordinates as the only arg? The other option would be a tuple of args that get forwarded into the minimize function, and is passed directly into this every call

JNeiger (approved):
Looks good! Just a few discussion points about possible enhancements. 

Lgtm pending other comments

soccer/gameplay/tactics/find_pass_point.py
HussainGynai: Also the function isn't really specific to passing, like you could have a cost function for positioning or shooting that could use the same function, so we might not wanna constrain it to passing.

soccer/gameplay/tactics/find_pass_point.py
HussainGynai: Okay so the minimizer fails if it doesn't find the exact minima within the number of allotted iterations, but that's sort of the point of limiting iterations, so I'll get rid of the exception, but should I like print something that says that the minimizer failed?

soccer/gameplay/tactics/find_pass_point.py
```diff
@@ -0,0 +1,37 @@
...
+    else:
+        options = None
+    result = minimize(heuristic,
+                      initial_guess,
+                      bounds=((-width / 2.00, width / 2.00), (0.00, length)),
+                      tol=1e-3,
+                      options=options)
+    # if (not result.success):
+    #     raise OptimizationError()
+    point = robocup.Point(result.x[0], result.x[1])
```
JNeiger, line 25: If results.success is false, does this still return "good" values?

soccer/gameplay/tactics/test_cost_function.py
```diff
@@ -0,0 +1,15 @@
+import robocup
+import constants
+import math
+from typing import Tuple, Callable
+
+
+## A test cost function for passes that should have a minima at (0.915239, 1.38493)
+# Takes in an array of 2 cooridnates
+def create_test_cost_function(
```
JNeiger, line 9: You can also use `partial` to bind args to a function.

I do it [here](https://github.com/RoboJackets/robocup-software/blob/staging/soccer/gameplay/evaluation/passing_positioning.py#L116). Documentation is [here](https://docs.python.org/3/library/functools.html#functools.partial)

soccer/gameplay/tactics/find_pass_point.py
HussainGynai: Yeah, I removed the error because even if it does return false, it'll return the closest value it got to within the number of allotted iterations, which is the point of limiting iterations.

soccer/gameplay/tactics/find_pass_point.py
```diff
@@ -0,0 +1,37 @@
+import robocup
+import constants
+from scipy.optimize import minimize
+from typing import Callable, Tuple, Optional, Any, Sequence
+
+
+## Function that optimizes cost function for passing
+# uses scipy minimize using the BFGS method
+def find_pass_point(heuristic: Callable[[Tuple[float, float]], float],
+                    initial_guess: Optional[Tuple[float, float]] = (0, 0),
```
oswinso, line 10: Why `Tuple[float, float]` and not `robocup.Point`?

Issue #1485, opened 2020-06-12
Segment.interects fails in RELEASE


oswinso:
## Steps to reproduce
1. Compile tests in RELEASE
2. Run tests

Result: `Segment.intersects` fails.

- [ ] This bug has been reproduced by someone else

## Additional information
Float bullshit. Probably because of all the casting between float and double or something.

PR #1486, opened 2020-06-12
Increases FLOAT_EPSILON to 1e-4 to fix bug with floating point in Release

oswinso:
## Description
Temporarily increases `FLOAT_EPSILON` to `1e-4` so that tests don't fail when compiled in Release.

## Associated Issue
#1485 

## Design Documents
[Link](link-to-design-doc)

### Build in Release
1. Run test

Expected result: Test doesn't fail.


JNeiger (approved):

kylestach (approved):
Stupid, but okay.

PR #1482, opened 2020-06-09
[DONT MERGE] Initial ROS2 refactor + Migrate vision receiver

oswinso:
## Description
This PR:
- Rearranges `common` a bit to be easier to refactor into ROS2 packages later
    - Moves `protobuf` into its own folder
    - Changes `common` to have headers in a separate `include` folder
    - Renames `Geometry2d` to `geometry2d` to conform to Google C++ standards
- Creates a `config` system
    - `config_server` which acts as a global parameter server and serves `FieldDimensions` and `GameSettings`
    - `config_client` which pulls the most recent `FieldDimensions` and `GameSettings` and has methods to update them
- Creates the ROS2 messages for communicating the `FieldDimensions` and `GameSettings` parameter updates, as well as the stuff that `vision_receiver` does.
- Refactors `vision_receiver` to be a separate `rclcpp::Node`
- Creates launch files to launch sim, ie. `sim.launch.py`
- (Temporarily) Refactors `Processor` to have a `_vision_receiver_sub`, `_config_client` and `_executor` just so that it can interface with the `vision_receiver` for now.
- (Temporarily) Comments out `simple_logger` for now since we don't really use it.

Note: CI still needs to be updated to correctly build the new folder structure.

### Sim still works
1. Get ROS2
2. Clone this to like `~/robocup_ros_ws/src/robocup-software`
3. `source /opt/ros/XXXX/setup.bash`. I'm on 18.04 so I tested this on eloquent
4. Run `colcon build`
5. `ros2 launch rj_robocup sim.launch.py`

Expected result: Things still run. Vision still works.

PR #1487, opened 2020-06-12
ROS2 build system

oswinso:
## Description
This PR
- Changes our build system to use the ROS2 build system `colcon`.
- Fixes various paths so that they work correctly now
- Deletes the 16.04 `Dockerfile` because only 18.04 and 20.04 have ROS2 binaries
- Updates CircleCI to build correctly
- Updates some Makefile tests to work properly
- Updates the python tests to reference `soccer.so` in the correct directory
- Changes tests to use a build flag so that it can be installed properly

Broken out of #1482.

## Design Documents
[Link](link-to-design-doc)

### Sim works
1. Install ros2
2. Put this in `robocup_ws/src/robocup-software`
3. `source /opt/ros/ROS_VERSION/setup.sh`
4. Run `colcon build`
5. `source install/setup.sh`
6. Launch sim via `ros2 launch rj_robocup sim.launch.py`

Expected result: sim still runs and works

### Tests work
1. Do the same stuff above
2. Run `./install/rj_robocup/lib/rj_robocup/test-soccer `

Expected result: Tests still run and pass


JNeiger (approved):
Nothing looks too out there, looks good to me

kylestach:
Let's merge this into a `ros2` base branch to avoid breaking the main workflow for everybody.

package.xml
```diff
@@ -0,0 +1,36 @@
+<?xml version="1.0"?>
+<?xml-model href="http://download.ros.org/schema/package_format3.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
+<package format="3">
+    <name>rj_robocup</name>
+    <version>0.0.0</version>
+    <description>TODO: Package description</description>
+    <maintainer email="kylestach99@gmail.com">Kyle Stachowicz</maintainer>
+    <maintainer email="oswinso@gmail.com">Oswin So</maintainer>
+    <license>TODO: License declaration</license>
```
kylestach, line 9: We should probably agree on a license. It looks like we're on Apache right now: https://github.com/RoboJackets/robocup-software/blob/staging/LICENSE. I don't have any major concerns with carrying it over, but we should make sure everyone's still okay with that.

run/2bot-simulator.cfg
```diff
@@ -1,28 +0,0 @@
-<simulation>
```
kylestach, line 1: LMFAO this shouldn't even exist. I actually didn't even know it did.

util/ubuntu-setup
```diff
@@ -27,35 +27,16 @@ if cat /etc/os-release | grep -iq '^NAME=.*Debian'; then # dont add repositories
...
-    elif cat /etc/os-release | grep -iq '^VERSION=.*19.04'; then # Using Ubuntu 19.04
-        echo "Ubuntu 19.04 Detected..."
-        SYSTEM="ubuntu-19.04"
-        ADD_REPOS=false
     elif cat /etc/os-release | grep -iq '^VERSION=.*20.04'; then # Using Ubuntu 20.04
         echo "Ubuntu 20.04 Detected..."
         SYSTEM="ubuntu-20.04"
         ADD_REPOS=false
     else
+        echo "Sorry, we only support 18.04 and 20.04 due to ROS2." >&2
```
kylestach: Foxy isn't backwards-compatible with 18.04, is it? And IIRC there were breaking API changes from Eloquent to Foxy; Foxy is the first time it's (supposed to) have a "stable" API. IMO we should only support Foxy on 20.04 here.

kylestach (approved):
LGTM. As mentioned, I think this should be PR'd against a `ros2` base branch. This is probably not ready for daily-driver use yet (cc @HussainGynai @liquidmetal9015) and I expect we'll be breaking a lot of things in the near future.

package.xml
oswinso: Yeah, but we should just do that sometime later.

.circleci/config.yml
```diff
@@ -6,57 +6,64 @@ defaults_env: &setup_env
...
 
 save_workspace: &save_workspace
   persist_to_workspace:
-    root: ~/robocup-software
+    root: ~/robocup_ws
     paths:
+      - install/*
       - build/*
-      - run/*
+      - log/*
```
oswinso:
```suggestion
```

util/ubuntu-setup
oswinso: Ok removed eloquent

run/2bot-simulator.cfg
oswinso: I think it would be nice to have different grsim config for different tests though, ie. Maybe a passing test should only have 2 robots

kylestach: Agree. I wonder if there's a way to change grSim settings programmatically...probably not but it would be nice.

soccer/gameplay/run_tests.sh
```diff
@@ -2,4 +2,6 @@
 
 # runs all the tests in the tests directory
 # we prepend the run directory so that `import robocup` finds the `robocup.so` file located there
-PYTHONPATH="../../run/lib" python3 -m unittest discover -s tests
+export PYTHONPATH=$(readlink -f "../../install/lib")
+echo "PYTHONPATH: $PYTHONPATH"
```
kylestach: Can you remove these?

kylestach (approved):
LGTM, but please clean up extra prints from debugging.

soccer/gameplay/run_tests.sh
oswinso: done

PR #1488, opened 2020-06-12
removed soccer/gameplay/__init__.py to avoid ambiguous name and fix mypy error

oswinso:
## Description
`mypy` 0.780 errors out from error:
```
soccer/gameplay/timeout_behavior.py: error: Source file found twice under different module names: 'gameplay.main' and 'main'
Found 1 error in 1 file (checked 204 source files)
```

This is [this issue on mypy](https://github.com/python/mypy/issues/8944). On there, it is recommended that `mypy_path` should not point to a folder that has a `__init__.py` file to avoid ambiguous names.

This PR removes `soccer/gameplay/__init__.py` to fix that `mypy` error.

### `mypy` doesn't error out
1. Run `make mypy`
Expected result: No errors


oswinso:
Just kidding `pylint` complains about this. I guess I'll have to look at this a bit more carefully.

oswinso:
I'll just pin mypy to `0.770` instead

PR #1490, opened 2020-06-14
Motion planning rewrite

kylestach:
## Description
This PR refactors the motion planning interface and makes some major upgrades along the way. Anecdotally, motion planning and specifically capture after this PR is much, much snappier. This improvement comes in part due to removing curvature-based velocity limiting, which had some serious issues with the Bezier paths.

A brief breakdown of the interface changes:
 - Make a `PlannerNode`
 - Trim down and clean up `PlanRequest` entries; make all information flowing into the planner come from this struct.
 - Make `Trajectory` the only internally-used path representation
    - This is a piecewise cubic curve interpolating between waypoints. It should give us piecewise-smooth acceleration that is consistent with the velocity and position profiles.
    - Timestamps are now enforced throughout the whole interface.
 - Clean up the `Ball` interface and local obstacle generation.
 - Make angle functions return a Jacobian matrix for calculation of angular velocity.
    - This allows angle planning to be smooth

Minor quality-of-life fixes/improvements:
 - `nearly_equal` methods for `Pose`, `RobotInstant`, etc.
 - Separating out `RobotInstant` and `LinearMotionInstant`.

Things that still need to happen:
 - [ ] Clean up pivot kicks. They are still a little sloppy. (This can happen in a later PR).
    - [ ] Robot shots are sometimes slightly at the wrong angles (at least in grSim); this is annoying.
 - [ ] Find a cleaner way to do angle planning. Currently, this does accomplish the goal of providing angular targets that are self-consistent (i.e. velocities are valid) but does _not_ ensure that they are consistent with the current position and velocity. (This can happen in a later PR).
 - [ ] Re-enable the line kick planner (has not yet been ported over).
 - [ ] **[BLOCKER]** Fix the `EscapeObstaclesPlanner`
 - [ ] **[BLOCKER]** Draw local obstacles
 - [ ] **[BLOCKER]** Create local obstacles for goal zones (for non-goalie robots) and goals in the gameplay module.

Other improvements that would be nice:
 - `LinearMotionTrajectory` for obstacles and pre-angle-profiling trajectories
    - We could even make `Trajectory` templated over a type, and then treat angle planning as splicing together an angular trajectory and a linear trajectory.
    - Right now these are just regular `Trajectory` objects, so they contain useless semantic information (angles)

## Associated Issues
 - #1323 
 - #1184 

## Design Documents
N/A

### Regular playbook
1. Run the regular playback
2. General performance improvement expected in capture times, etc.

### Repeated line up
 - No regressions expected; this was already pretty good.

### Capture test
 - Should experience great improvement in performance.

### Obstacle avoidance
1. Should be able to avoid robots (ours and enemy).
2. Should be able to avoid static field obstacles (goals) and the edge of the field
3. Should avoid "tactical" obstacles (goal zones, shot cones, etc.)


soccer/gameplay/plays/defense/defend_goal.py
```diff
@@ -0,0 +1,68 @@
+import main
```
kylestach, line 1: What's the deal with these? No idea why they're showing up here.

soccer/gameplay/robocup-py.cpp
```diff
@@ -8,6 +8,8 @@
 #include <string>
 #include <utility>
 
+#include "WorldState.hpp"
```
kylestach, line 11: Changes in `robocup-py.cpp` are essentially trivial and just relate to the ball interface changes.

soccer/gameplay/skills/angle_receive.py
```diff
@@ -16,24 +16,12 @@
 # Kick is a single_robot_behavior, so no need to import both
 class AngleReceive(skills.pass_receive.PassReceive):
     def __init__(self):
-        super().__init__(
-            captureFunction=(lambda: skills.settle.Settle(self._target_point)))
-        self._target_point = None
+        super().__init__(captureFunction=(lambda: skills.settle.Settle()))
```
kylestach, line 19: TODO: Add this back in.

soccer/gameplay/skills/settle.py
```diff
@@ -22,8 +21,4 @@ def execute_running(self):
         if (self.robot is not None):
             self.robot.disable_avoid_ball()
             self.robot.set_dribble_speed(constants.Robot.Dribbler.MaxPower)
-
-            if (self.target is None):
-                self.robot.settle()
-            else:
-                self.robot.settle_w_bounce(self.target)
+            self.robot.settle()
```
kylestach, line 24: TODO reset this.

soccer/motion/MotionControl.cpp
```diff
@@ -1,12 +1,12 @@
 #include "MotionControl.hpp"
 
-#include <Context.hpp>
-#include <Geometry2d/Util.hpp>
-#include <Robot.hpp>
-#include <RobotConfig.hpp>
-#include <Utils.hpp>
 #include <optional>
-#include <planning/MotionInstant.hpp>
+
```
kylestach, line 4: Trivial changes based on the change in the Trajectory API.

Only exception is that angle targets should be fully resolved by control time, so they are no longer represented by optionals.

soccer/planning/MotionConstraints.hpp
```diff
@@ -23,7 +23,6 @@ struct MotionConstraints {
      */
     double maxSpeed;
```
kylestach, line 24: TODO clean this up and make it part of `RobotIntent`.

soccer/planning/low_level/TrapezoidalMotion.hpp
```diff
@@ -0,0 +1,24 @@
+#pragma once
```
kylestach, line 1: TODO dedup with `motion/TrapezoidalMotion.hpp`, which is broken for some of the cases this needs to handle.

Regardless, this should live in planning rather than control.

soccer/planning/low_level/Replanner.hpp
```diff
@@ -0,0 +1,54 @@
+#pragma once
```
kylestach, line 1: TODO make these free functions (maybe - that might not interact nicely with the config system?)

soccer/planning/planner/CollectPlanner.cpp
```diff
@@ -0,0 +1,470 @@
+#include "CollectPlanner.hpp"
+
```
kylestach, line 2: This is almost line-for-line the same as the original collect planner. There should be no major functional changes.

For some reason it's not showing up as a rename, I think I may have renamed a bunch of locals :(

soccer/planning/planner/CollectPlanner.hpp
```diff
@@ -1,8 +1,8 @@
 #pragma once
 
-#include "DirectTargetPathPlanner.hpp"
-#include "RRTPlanner.hpp"
-#include "SingleRobotPathPlanner.hpp"
+#include "planning/Instant.hpp"
```
kylestach, line 3: This is almost line-for-line the same as the original collect planner. There should be no major functional changes.

soccer/planning/planner/EscapeObstaclesPathPlanner.cpp
```diff
@@ -2,14 +2,13 @@
 
 #include <Configuration.hpp>
 #include <optional>
+#include <vector>
```
kylestach, line 5: TODO re-enable this file

soccer/planning/planner/InterceptPlanner.cpp
```diff
@@ -0,0 +1,88 @@
+#include "InterceptPlanner.hpp"
+
```
kylestach, line 2: This is almost a direct port from the original intercept planner. There should be no major functional changes.

oswinso:
So... you're gonna hate what I'm gonna say, but.... could you break up this PR?

soccer/planning/planner/InterceptPlanner.hpp
```diff
@@ -0,0 +1,24 @@
+#pragma once
+
+#include "Planner.hpp"
```
kylestach, line 3: This is almost a direct port from the original intercept planner. There should be no major functional changes.

soccer/planning/planner/LineKickPlanner.cpp
```diff
@@ -0,0 +1,247 @@
+#include "LineKickPlanner.hpp"
+
+#include "planning/Instant.hpp"
```
kylestach, line 3: TODO re-enable this file.

oswinso:
Maybe something like:
1. Changes to the `Geometry2d` lib
2. Separating out RobotInstant and LinearMotionInstant
3. Clean up the Ball interface and local obstacle generation.
4. The rest?

kylestach:
@oswinso I can split it into a set of stacked PRs but none of them will pass CI on their own. It'll just be a massive interactive rebase, lmao.

I'm doing my best to comment things accordingly but it's unfortunately all really tied together. In theory it should be split into:
 - Ball state changes
 - Trajectory class
 - Low-level planning refactor (this is really big, basically a rewrite of path smoothing and RRT drivers)
 - Planner interface and PlanRequest refactor
 - Planners (mostly no-op changes, just updating them to fit the interfaces and not own instances of other high-level planners, i.e. `CollectPlanner` no longer owns an `RRTPlanner/PathTargetPlanner`). Also, moving out replanning functionality.
 - A few other minor things.

kylestach:
Closing in favor of #1492.

PR #1492, opened 2020-06-14
[MP Rewrite Part 0] Base for motion planning rewrite

kylestach:
This is the base PR for the motion planning rewrite. Other motion planning PRs will be stacked onto this one.

## Description
This PR refactors the motion planning interface and makes some major upgrades along the way. Anecdotally, motion planning and specifically capture after this PR is much, much snappier. This improvement comes in part due to removing curvature-based velocity limiting, which had some serious issues with the Bezier paths.

A brief breakdown of the interface changes:
 - Make a `PlannerNode`
 - Trim down and clean up `PlanRequest` entries; make all information flowing into the planner come from this struct.
 - Make `Trajectory` the only internally-used path representation
    - This is a piecewise cubic curve interpolating between waypoints. It should give us piecewise-smooth acceleration that is consistent with the velocity and position profiles.
    - Timestamps are now enforced throughout the whole interface.
 - Clean up the `Ball` interface and local obstacle generation.
 - Make angle functions return a Jacobian matrix for calculation of angular velocity.
    - This allows angle planning to be smooth

Minor quality-of-life fixes/improvements:
 - `nearly_equal` methods for `Pose`, `RobotInstant`, etc.
 - Separating out `RobotInstant` and `LinearMotionInstant`.

Things that still need to happen:
 - [x] Clean up pivot kicks. They are still a little sloppy. (This can happen in a later PR).
    - [x] Robot shots are sometimes slightly at the wrong angles (at least in grSim); this is annoying.
 - [ ] Find a cleaner way to do angle planning. Currently, this does accomplish the goal of providing angular targets that are self-consistent (i.e. velocities are valid) but does _not_ ensure that they are consistent with the current position and velocity. (This can happen in a later PR).
 - [x] Re-enable the line kick planner (#1541).
 - [x] **[BLOCKER]** Fix the `EscapeObstaclesPlanner`
 - [x] **[BLOCKER]** Draw local obstacles (#1526)
 - [x] **[BLOCKER]** Create local obstacles for goal zones (for non-goalie robots) and goals in the gameplay module. (#1526)

Other improvements that would be nice:
 - `LinearMotionTrajectory` for obstacles and pre-angle-profiling trajectories
    - We could even make `Trajectory` templated over a type, and then treat angle planning as splicing together an angular trajectory and a linear trajectory.
    - Right now these are just regular `Trajectory` objects, so they contain useless semantic information (angles)

## Associated Issues
 - #1323 
 - #1184 

## Design Documents
N/A

### Regular playbook
1. Run the regular playback
2. General performance improvement expected in capture times, etc.

### Repeated line up
 - No regressions expected; this was already pretty good.

### Capture test
 - Should experience great improvement in performance.

### Obstacle avoidance
1. Should be able to avoid robots (ours and enemy).
2. Should be able to avoid static field obstacles (goals) and the edge of the field
3. Should avoid "tactical" obstacles (goal zones, shot cones, etc.)

### Unit testing
 - [X] Ball state changes
 - [X] Trajectory
 - [x] Plan request
    - [x] ~~Obstacle hit checkers~~
 - [x] Primitives
    - [X] Angle planning
    - [X] RRT
    - [x] ~~`CreatePath` utils~~
    - [x] ~~Replanner~~
    - [X] Trapezoidal motion
    - [X] Velocity profiling
 - Planners: not much we can do here other than integration testing, unless anyone else has good ideas for how to test these.


JNeiger (approved):
🎉 🥇 
(Assuming nothing changed from the other 12 PR's lmao)

kylestach:
*13

But no. Nothing changed.

HussainGynai (approved):

oswinso (approved):
There are some issues with dribble and aiming for passes, but we can solve those later.

kylestach:
@oswinso those have always been there, I'm pretty sure they're issues with the gameplay side of things. I decided not to address them in this rewrite.

PR #1494, opened 2020-06-14
 [MP Rewrite Part 2] Introduce the Trajectory abstraction

kylestach:
A Trajectory is a piecewise-cubic function interpolating between a
vector of time-indexed RobotInstant waypoints.

This is not expected to pass CI.


common/Utils.hpp
```diff
@@ -61,10 +61,7 @@ inline void debugThrowIf(const std::string& string, bool condition) {
  */
 template <typename T>
 static inline T fixAngleRadians(T a) {
-    a = remainder(a, 2 * M_PI);
-    while (a < -M_PI) a += 2.0 * M_PI;
-    while (a > M_PI) a -= 2.0 * M_PI;
-    return a;
+    return std::atan2(std::sin(a), std::cos(a));
```
oswinso: Why this change?

soccer/motion/MotionControl.cpp
```diff
@@ -102,7 +98,7 @@ void MotionControl::run(const RobotState& state,
         result_world.angular());
 
     // Use default constraints. Planning should be in charge of enforcing
-    // constraints on the path, here we just follow it.
+    // constraints on the trajectory, here we just follow it.
     // TODO(Kyle): Use this robot's constraints here.
```
oswinso: I know this isn't part of this PR but issue number plz

soccer/planning/Instant.hpp
```diff
@@ -0,0 +1,94 @@
...
+#include <time.hpp>
+
+#include "DebugDrawer.hpp"
+#include "Utils.hpp"
+#include "planning/DynamicObstacle.hpp"
+
+namespace Planning {
+
+/**
+ * This class represents a robot's motion "state" at a given time, including
```
oswinso: brief

soccer/planning/Instant.hpp
```diff
@@ -0,0 +1,94 @@
...
+
+    friend std::ostream& operator<<(std::ostream& stream,
+                                    const LinearMotionInstant& instant) {
+        return stream << "LinearMotionInstant(position=" << instant.position
+                      << ", velocity=" << instant.velocity << ")";
+    }
+};
+
+/**
+ * Represents the current state of a robot in a planned trajectory.
```
oswinso: brief

soccer/planning/TrajectoryUtils.hpp
```diff
@@ -0,0 +1,42 @@
+#pragma once
+
+#include "Trajectory.hpp"
+
+namespace Planning {
+
+// TODO(Kyle): both of these functions should be able to be abstracted under a
+//  single interface templated on either a shape set or a collection of dynamic
+//  obstacles.
```
oswinso: Issue number pls

soccer/planning/Trajectory.cpp
```diff
@@ -0,0 +1,344 @@
...
+    RJ::Time actual_end = std::min(endTime, end_time());
+
+    // If the start and end times are identical, we are trying to grab an
+    // infinitesimal trajectory. This is weird but technically not wrong.
+    if (startTime == actual_end) {
+        return Trajectory({cursor.value()});
+    }
+
+    RobotInstantSequence result_instants;
+    // TODO(Kyle): Reserve space in result_instants.
```
oswinso: Issue num

oswinso (changes requested):
Preliminary changes

common/Utils.hpp
kylestach: O(1) instead of O(n), lmao. This should really just be an `std::remainder` though.

soccer/motion/MotionControl.cpp
kylestach: #1500

soccer/planning/Trajectory.hpp
```diff
@@ -0,0 +1,420 @@
...
+     * @brief Get a reference to the ith instant.
+     *
+     * @param i the instant index to retrieve. Must be in range
+     *  [0, num_instants())
+     * @return a reference to the ith instant.
+     *
+     * @details Intended for use when editing a trajectory in-place; for other
+     *  cases it may be simpler to use the iterator or Cursor interfaces.
+     */
+    [[nodiscard]] auto& instant_at(int i) { return instants_.at(i); }
```
oswinso: Why the `auto` in these functions?

soccer/planning/TrajectoryUtils.cpp
```diff
@@ -0,0 +1,141 @@
...
+
+    const auto& start_hits = obstacles.hitSet(cursor.value().position());
+    while (cursor.has_value()) {
+        RobotInstant instant = cursor.value();
+
+        // Calculate the list of hits.
+        // TODO(Kyle): We could save a lot of allocation and smart pointer
+        //  reference counting overhead by making hitSet return a custom
+        //  iterator type. Then, we could use that to construct a set at the
+        //  beginning but just iterate over it in standard fashion here.
```
oswinso: issue number

soccer/planning/TrajectoryUtils.hpp
kylestach: #1501

oswinso (changes requested):
more things

common/Utils.hpp
oswinso: I'm pretty sure with the old one the `while` acts more like an `if` though... but I think it's fine. Just a bit worried about the performance of calling `sin`, `cos` and `atan2`

kylestach: Remainder is preferable to any of the above. I was meaning to swap that out but this was the first change I thought of (I just really don't like the idea that we could take an unbounded amount of time in this stupid little function, especially if something hits infinity or a really really large number).

kylestach: Alright `std::remainder` has the correct behavior, I'll swap out to that.

kylestach: I didn't catch that we were doing a remainder beforehand, my bad. That should already be correct (always between -pi and pi), and we shouldn't even have needed the loops.

ghost (approved):

soccer/planning/Trajectory.cpp
```diff
@@ -0,0 +1,344 @@
...
+}
+
+void Trajectory::ScaleDuration(RJ::Seconds final_duration) {
+    // Note: even though this modifies the trajectory, it does not break
+    // angle planning and so we do not need to mark the angle profile as
+    // invalid.
+    ScaleDuration(final_duration, begin_time());
+}
+
+void Trajectory::ScaleDuration(RJ::Seconds final_duration,
```
ghost, line 69: I forgot where this is used, but this will be a bit slower than the old version. It's probably fine though

soccer/planning/Trajectory.hpp
```diff
@@ -0,0 +1,422 @@
...
+class Trajectory {
+public:
+    Trajectory() = default;
+    ~Trajectory() = default;
+
+    /**
+     * @brief Create a trajectory from several "instants", each with a pose,
+     * velocity, and timestamp. This
+     */
+    explicit Trajectory(RobotInstantSequence instants)
```
ghost, line 41: should use an r-value reference or a smart pointer in the function parameter. I think this makes a copy

soccer/planning/Trajectory.hpp
```diff
@@ -0,0 +1,422 @@
...
+        : instants_(std::move(instants)) {}
+
+    /**
+     * @brief Create a trajectory from two other trajectories.
+     *  Assumes a.last() == b.first() so b.first() is skipped.
+     *
+     * @details Is able to move out of the first trajectory, but not the second.
+     */
+    Trajectory(Trajectory a, const Trajectory& b);
```
ghost, line 50: Maybe have an r-value reference / smart pointer version of this function, so we can move out of the first Trajectory. Here we are copying both.

soccer/planning/Trajectory.hpp
kylestach: To pass into this you can do `Trajectory(std::move(instants_arg))` and it will move `instants_arg` into the `instants` argument (call the move constructor for the parameter. This is generally preferred because you can pass either a const reference or an rvalue.

soccer/planning/Trajectory.cpp
kylestach: It's not actually used anywhere, I think. Yeah it'll be O(n) instead of O(1) but it's super infrequently used.

soccer/planning/Trajectory.hpp
kylestach: See above

soccer/planning/Trajectory.cpp
ghost: Yeah that's fine. Trajectories are smaller now. I didn't realize that when making this comment so this is not really an issue.

soccer/planning/Trajectory.cpp
```diff
@@ -0,0 +1,344 @@
...
+    : Cursor{trajectory, trajectory.instants_.begin()} {}
+
+Trajectory::Cursor::Cursor(const Trajectory& trajectory, RJ::Time start_time)
+    : Cursor{trajectory} {
+    seek(start_time);
+}
+
+RobotInstant Trajectory::Cursor::value() const {
+    RobotInstant instant_0 = *iterator_;
+    if (instant_0.stamp == time_) {
```
JNeiger, line 259: `*iterator_.stamp` and `time_` are only equal if it's the last instant in the vector right?

soccer/planning/Trajectory.hpp
```diff
@@ -0,0 +1,422 @@
...
+    /**
+     * @brief Stamp this trajectory for completion at the specified time.
+     */
+    void stamp(RJ::Time time) { creation_stamp_ = time; }
+
+    /*
+     * @brief Allows seeking to arbitrary positions in a Trajectory and
+     *  incrementing by arbitrary steps.
+     */
+    class Cursor {
```
JNeiger, line 301: This header and source file are pretty large, would it be simple to split them up?

soccer/planning/TrajectoryUtils.cpp
```diff
@@ -0,0 +1,141 @@
...
+    Trajectory::Cursor cursor = trajectory.cursor(start_time);
+
+    // If the trajectory has already ended, we don't need to check it.
+    if (!cursor.has_value()) {
+        return false;
+    }
+
+    // Limit iterations to 400. This will continue to operate at dt = 0.05 until
+    // we hit a 20 second trajectory. If our trajectory is longer than that,
+    // something is probably wrong.
```
JNeiger: You should specify that the dt increases after 20 seconds as well

common/Geometry2d/Pose.hpp
```diff
@@ -128,12 +128,38 @@ class Pose {
...
+     */
+    bool operator==(const Pose& other) const {
+        return _position == other._position && _heading == other._heading;
+    }
+
+    /**
+     * Inequality comparison operation.
+     */
+    bool operator!=(const Pose& other) const {
+        return _position != other._position || _heading != other._heading;
```
JNeiger: Since you have the equality operator, you can just do `!(*this == other)`. This reduces copy paste errors

JNeiger (commented):
Just a few things here and there

soccer/planning/Trajectory.hpp
kylestach: It would be a bit annoying, `Cursor` and `Trajectory::Cursor` are pretty much interdependent.

soccer/planning/Trajectory.cpp
kylestach: Yes. This check is necessary to avoid dereferencing the `end()` iterator in the case that we're pointing to the end.

common/Geometry2d/Pose.hpp
```diff
@@ -289,12 +313,38 @@ class Twist {
...
+     * Equality comparison operation.
+     */
+    bool operator==(const Twist& other) const {
+        return _linear == other._linear && _angular == other._angular;
+    }
+
+    /**
+     * Inequality comparison operation.
+     */
+    bool operator!=(const Twist& other) const {
```
JNeiger: `!=` oeprator that you can swap to `!( == )`

soccer/planning/Instant.hpp
```diff
@@ -0,0 +1,94 @@
...
+     */
+    bool operator==(const RobotInstant& other) const {
+        return pose == other.pose && velocity == other.velocity &&
+               stamp == other.stamp;
+    }
+
+    /**
+     * Inequality comparison operator.
+     */
+    bool operator!=(const RobotInstant& other) const {
```
JNeiger, line 88: `!=` operator that you can swap to `!( == )`

soccer/planning/Trajectory.hpp
JNeiger: Nevermind then

JNeiger (commented):
Missed two `!=` operators. LGTM after that

JNeiger (approved):

oswinso (commented):
lgtm

oswinso (approved):
lgtm

PR #1497, opened 2020-06-14
[MP Rewrite Part 5] Port previous planners to new framework

kylestach:
Most of these are not functional changes from the old planners, but
rather simple ports to the new primitives. The biggest changes are that
we now use `Replanner` static functions and manually manage our own
"previous trajectory" objects instead of having "behavior-level"
planners (i.e. settle) own instances of an RRT planner and interacting
with it through plan requests.

This is the final PR in the rewrite, and it should pass CI. At the point
where it passes CI and all PRs are ready for merge, each PR should be
merged into its parent in sequence.


soccer/planning/planner/CollectPlanner.hpp
```diff
@@ -0,0 +1,156 @@
...
+
+class Configuration;
+class ConfigDouble;
+
+namespace Planning {
+
+/**
+ * @brief Planner that tries to move onto and gain control of a slow moving ball
+ *
+ * TODO: Clean up description
```
oswinso: issue num

soccer/planning/planner/LineKickPlanner.hpp
```diff
@@ -0,0 +1,39 @@
+#pragma once
+
+#include "Planner.hpp"
+#include "planning/Instant.hpp"
+#include "planning/low_level/Replanner.hpp"
+
+namespace Planning {
+class LineKickPlanner : public PlannerForCommandType<LineKickCommand> {
```
oswinso, line 14: docstring

oswinso (changes requested):
issue num & docstring

oswinso (approved):
lgtm

soccer/planning/planner/CollectPlanner.cpp
```diff
@@ -0,0 +1,470 @@
...
+
+    // If we are in range to the slow dist
+    if (dist < *_approachDistTarget + Robot_MouthRadius &&
+        currentState == CourseApproach) {
+        currentState = FineApproach;
+    }
+
+    // If we are close enough to the target point near the ball
+    // and almost the same speed we want, start slowing down
+    // TODO(Kyle): Check for ball sense?
```
JNeiger: Unrelated to this PR, but do we have angle information in these planners now? The old iteration only had XY information

kylestach: Yeah, that's a big change to how these planners work. They now plan angles on the trajectory directly rather than returning an angle function.

soccer/planning/planner/CollectPlanner.cpp
```diff
@@ -0,0 +1,470 @@
...
+    // Moving at us
+    if (averageBallVel.angleBetween((ball.position - start.position())) >
+        3.14 / 2) {
+        currentSpeed = *_touchDeltaSpeed;
+        velocityScale = 0;
+    }
+
+    motionConstraints.maxAcceleration *= *_controlAccelScalePercent;
+    motionConstraints.maxSpeed =
+        std::min(currentSpeed, motionConstraints.maxSpeed);
```
JNeiger, line 368: This is the max speed thing I mentioned in #1513

soccer/planning/planner/CollectPlanner.cpp
JNeiger: That was a big problem I had when trying to "match" the state machines inside the planner and in python. It was possible to have the ball behind us but close enough such that the planner thought we had the ball, but the python side knew we didn't. I don't think I ever found a good hack to make it completely consistent

soccer/planning/planner/LineKickPlanner.cpp
```diff
@@ -0,0 +1,247 @@
...
+REGISTER_CONFIGURABLE(LineKickPlanner);
+
+void LineKickPlanner::createConfiguration(Configuration* cfg) {
+    // NOLINTNEXTLINE
+    _approachSpeed = new ConfigDouble(cfg, "LineKickPlanner/approachSpeed");
+}
+
+using namespace Geometry2d;
+Trajectory LineKickPlanner::plan(const PlanRequest& /* request */) {
+#if 0
```
JNeiger, line 18: LineKick is disabled?

soccer/planning/planner/Planner.cpp
```diff
@@ -0,0 +1,28 @@
+#include "Planner.hpp"
+
+#include "planning/Instant.hpp"
+#include "planning/Trajectory.hpp"
+
+namespace Planning {
+
+#if 0
```
JNeiger: Disabled code?

soccer/planning/planner/Planner.hpp
```diff
@@ -0,0 +1,78 @@
...
+    [[nodiscard]] bool isApplicable(
+        const MotionCommand& command) const override {
+        return std::holds_alternative<CommandType>(command);
+    }
+};
+
+template <typename T>
+inline T applyLowPassFilter(const T& oldValue, const T& newValue, double gain) {
+    return gain * newValue + (1 - gain) * oldValue;
+}
```
JNeiger: Copy paste error?

soccer/planning/planner/SettlePlanner.cpp
```diff
@@ -0,0 +1,633 @@
...
+            previous.subTrajectory(previous.begin_time(), startInstant->stamp);
+        double botDistToBallMovementLine =
+            ballMovementLine.distTo(pathSoFar.last().position() - deltaPos);
+
+        // Intercept -> Dampen
+        //  Almost intersecting the ball path and
+        //  Almost at end of the target path or
+        //  Already in line with the ball
+        //
+        // TODO(Kyle): Check ball sense?
```
JNeiger: Based on how I imagined dampen, it should already be in the dampen state before the ball hits. It should have some negative velocity when the ball hits and slow to a stop, minimizing the delta velocity between the robot and ball.

It was disabled through the config gains fairly early on during testing so no promises how well this actually works.

soccer/planning/planner/SettlePlanner.cpp
```diff
@@ -0,0 +1,633 @@
...
+                               nullptr, nullptr)) {
+        RJ::Seconds timeRemaining = previous.end_time() - startInstant.stamp;
+        Point pathAtEnd = previous.last().position();
+
+        Point nearPoint;
+        RJ::Seconds timeToIntersect =
+            ball.query_seconds_near(pathAtEnd, &nearPoint);
+        if (timeRemaining <
+                timeToIntersect + RJ::Seconds(*_interceptBufferTime) &&
+            nearPoint.distTo(pathAtEnd) < Replanner::goalPosChangeThreshold()) {
```
JNeiger: This will need to be tested, but you may run into problems with the `goalPosChangeThreshold` being larger than the `mouthWidth/2` so the reused path won't actually get you close enough to the ball path to actually catch it

JNeiger: This is especially problematic since the "shortcut" stuff is cut out with this early return. The ball velocity is surprisingly noisy

soccer/planning/planner/CollectPlanner.cpp
kylestach: I'm going to add a TODO issue for this. It should be much more reasonable to incorporate once we move to the "behavior planner" approach to these planners, where (in this case) it will actually have control and visibility over the dribbler.

kylestach: #1518

soccer/planning/planner/SettlePlanner.cpp
kylestach: Yeah, I'll remove this. It's no longer really an issue now that replanning is working a lot better.

soccer/planning/planner/Planner.cpp
kylestach: Removed.

soccer/planning/planner/Planner.hpp
kylestach: Moving to utils in #1512

soccer/planning/planner/LineKickPlanner.cpp
kylestach: I was going to leave it disabled in this PR, and make a separate PR into `mp-rewrite/base` where I actually port it over.

JNeiger (approved):

kylestach:
This now contains everything except for line kick, which I am planning on doing in a separate PR to mp-rewrite/base (logic being, it needs to share a lot of logic with settle and I'd like to clean that up a lot).

JNeiger (approved):

PR #1519, opened 2020-06-17
ROS2 config client / server

oswinso:
## Description
This PR:
1. Creates `rj_geometry_msgs`, `rj_msgs` folders for ROS2 messages
2. Creates some conversion code between `Geometry2d` and `rj_geometry_msgs`
3. Creates `config_server` and `config_client`
4. Creates `rj_utils/logging.hpp` as a `clang_tidy` compatible way to do logging (by default, `RCLCPP_INFO` raises `clang_tidy` warnings due to vararg stuff). Also, this creates `EZ_` prefixed macro calls that automatically use `get_logger()`, so that logging calls are simpler.

### `config_server` works
1. `ros2 launch rj_robocup config_server.launch.py`
2. `ros2 topic echo --qos-profile services_default --qos-durability transient_local /config/game_settings`
3.  `ros2 service call /config/set_game_settings rj_msgs/srv/SetGameSettings "{game_settings:{paused: true}}"`

Expected result: It should show
```
simulation: false
use_external_referee: false
request_blue_team: false
request_goalie_id: 0
defend_plus_x: false
use_our_half: false
use_their_half: false
paused: false
---
simulation: false
use_external_referee: false
request_blue_team: false
request_goalie_id: 0
defend_plus_x: false
use_our_half: false
use_their_half: false
paused: true
---
```


Geometry2d/include/Geometry2d/Line.hpp
```diff
@@ -67,6 +71,16 @@ class Line {
...
 
+    [[nodiscard]] LineMsg toROS() const {
+        LineMsg msg{};
+        msg.pt = {pt[0].toROS(), pt[1].toROS()};
+        return msg;
+    }
+
+    static Line fromROS(const LineMsg& msg) {
+        return Line{Point::fromROS(msg.pt[0]), Point::fromROS(msg.pt[1])};
+    }
```
kylestach: Do we want these to be implicit conversions? It's a trivial conversion, so I feel like it is going to reduce a lot of friction if it's implicit.

Any reason you can think of to not do that?

install/setup.bash
```diff
@@ -24,9 +24,16 @@ _ament_prefix_path_add() {
...
+_pythonpath_add() {
+    if [ -d "$1" ] && [[ ":$PYTHONPATH:" != *":$1:"* ]]; then
+        export PYTHONPATH="$1${PYTHONPATH:+":$PYTHONPATH"}"
+    fi
+}
+
 _path_add "${_INSTALL_PATH}/bin"
 _ld_library_path_add "${_INSTALL_PATH}/lib"
 _ament_prefix_path_add "${_INSTALL_PATH}"
+_pythonpath_add "${_INSTALL_PATH}/lib/python3.6/site-packages"
```
kylestach: Is this hardcoding python versions? Looks pretty brittle. Also, isn't 20.04 on 3.8 by default?

install/setup.zsh
```diff
@@ -24,9 +24,16 @@ _ament_prefix_path_add() {
...
+_pythonpath_add() {
+    if [ -d "$1" ] && [[ ":$PYTHONPATH:" != *":$1:"* ]]; then
+        export PYTHONPATH="$1${PYTHONPATH:+":$PYTHONPATH"}"
+    fi
+}
+
 _path_add "${_INSTALL_PATH}/bin"
 _ld_library_path_add "${_INSTALL_PATH}/lib"
 _ament_prefix_path_add "${_INSTALL_PATH}"
+_pythonpath_add "${_INSTALL_PATH}/lib/python3.6/site-packages"
```
kylestach: Same here.

rj_utils/include/rj_utils/logging.hpp
```diff
@@ -0,0 +1,189 @@
+#pragma once
+
+/**
+ * \file logging.hpp
+ * \brief This file defines macros that wrap the ones in rclcpp/logging.hpp
```
kylestach, line 5: IIRC we usually use `@brief`

rj_config/src/config_server/config_server.cpp
```diff
@@ -0,0 +1,60 @@
...
+    : Node{"config_server", node_options} {
+    const auto latching_qos = rclcpp::QoS(1).transient_local();
+
+    // Game Settings
+    game_settings_publisher_ = create_publisher<GameSettingsMsg>(
+        topics::kGameSettingsPub, latching_qos);
+
+    const auto game_settings_cb =
+        [this](const SetGameSettingsSrvReqPtr request,   // NOLINT
+               SetGameSettingsSrvRespPtr /*response*/) {  // NOLINT
```
kylestach, line 19: Why do you need NOLINT here?

install/setup.bash
JNeiger: Are we pinned to 3.6 in the ubuntu_setup script?

rj_msgs/msg/VisionPacket.msg
```diff
@@ -0,0 +1,2 @@
+builtin_interfaces/Time time
+RawProtobuf wrapper
```
kylestach: We should _really_ properly decode this at the vision receiver node. I don't think that this belongs in this PR, the `VisionPacket.msg` is unused at this point anyway.

launch/config_server.launch.py
```diff
@@ -0,0 +1,15 @@
+import launch
+from launch import LaunchDescription
+
+import launch_ros.actions
+
+
+def generate_launch_description():
+    return LaunchDescription([
```
kylestach, line 8: This node should also be added to `sim.launch.py`, right? I guess it's not integrated yet so we can ignore it for now, but personally I don't see a reason not to add it now.

rj_msgs/srv/SetGameSettings.srv
```diff
@@ -0,0 +1,2 @@
+GameSettings game_settings
```
kylestach, line 1: This means that we can't set game settings individually, correct? We should really consider whether this approach makes sense. This is more important for the subscription end, though - I don't think you usually want to subscribe to the entire GameSettings object.

rj_config/include/config_client/config_client.h
```diff
@@ -0,0 +1,50 @@
...
+public:
+    ConfigClient(rclcpp::Node* node);
+
+    [[nodiscard]] const GameSettingsMsg& gameSettings() const {
+        return game_settings_.value();
+    }
+
+    [[nodiscard]] const FieldDimensionsMsg& fieldDimensions() const {
+        return field_dimensions_.value();
+    }
```
kylestach, line 34: I would like a way of making a callback for when a parameter changes. This should be specifiable on a per-member basis of the `GameSettings` object.

From a design perspective, I'm still honestly not sure if this approach is the right one, rather than building off of the parameter server.

rj_utils/include/rj_utils/logging.hpp
```diff
@@ -0,0 +1,189 @@
...
+ * but with a NOLINT at the end, so that we don't trip up clang_tidy every time
+ * we make a logging call. This also defines EZ prefixed log calls that
+ * automatically use get_logger() to get the current logger, so that logging
+ * calls can be shorter.
+ */
```
JNeiger, line 12: I would add the link to the log stuff since there's tons of extra functions. Something like...

```suggestion
 * See link for log descriptions  http://docs.ros2.org/foxy/api/rclcpp/logging_8hpp.html
 */
```

rj_geometry_msgs/package.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0"?>
```
kylestach, line 1: Weren't we going to delete `package.xml` files, considering we don't use them?

rj_msgs/package.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0"?>
```
kylestach, line 1: Delete `package.xml`

rj_config/src/config_client/config_client_node.cpp
```diff
@@ -0,0 +1,6 @@
+#include <config_client/config_client_node.h>
+
+namespace config_client {
+ConfigClientNode::ConfigClientNode(const std::string& name)
```
kylestach, line 4: What is the purpose of this node? I'm confused as to why it's necessary - is it just a test?

rj_config/include/config_server/config_server.h
```diff
@@ -0,0 +1,41 @@
...
+using SetGameSettingsSrv = rj_msgs::srv::SetGameSettings;
+using SetGameSettingsSrvReqPtr = SetGameSettingsSrv::Request::SharedPtr;
+using SetGameSettingsSrvRespPtr = SetGameSettingsSrv::Response::SharedPtr;
+
+using FieldDimensionsMsg = rj_msgs::msg::FieldDimensions;
+using SetFieldDimensionsSrv = rj_msgs::srv::SetFieldDimensions;
+using SetFieldDimensionsSrvReqPtr = SetFieldDimensionsSrv::Request::SharedPtr;
+using SetFieldDimensionsSrvRespPtr = SetFieldDimensionsSrv::Response::SharedPtr;
+
+class ConfigServer : public rclcpp::Node {
```
kylestach, line 26: Please add documentation to `ConfigServer`, `ConfigClient`, and associated functions.

rj_utils/include/rj_utils/logging.hpp
```diff
@@ -0,0 +1,189 @@
...
+
+#define RJ_DEBUG_STREAM(logger, stream_arg) RCLCPP_DEBUG_STREAM(logger, stream_arg)  // NOLINT
+#define RJ_DEBUG_STREAM_ONCE(logger, stream_arg) RCLCPP_DEBUG_STREAM_ONCE(logger, stream_arg)  // NOLINT
+#define RJ_DEBUG_STREAM_EXPRESSION(logger, expression, stream_arg) RCLCPP_DEBUG_STREAM_EXPRESSION(logger, expression, stream_arg)  // NOLINT
+#define RJ_DEBUG_STREAM_FUNCTION(logger, function, stream_arg) RCLCPP_DEBUG_STREAM_FUNCTION(logger, function, stream_arg)  // NOLINT
+#define RJ_DEBUG_STREAM_SKIPFIRST(logger, stream_arg) RCLCPP_DEBUG_STREAM_SKIPFIRST(logger, stream_arg)  // NOLINT
+#define RJ_DEBUG_STREAM_THROTTLE(logger, clock, duration, stream_arg) RCLCPP_DEBUG_STREAM_THROTTLE(logger, clock, duration, stream_arg)  // NOLINT
+#define RJ_DEBUG_STREAM_SKIPFIRST_THROTTLE(logger, clock, duration, stream_arg) RCLCPP_DEBUG_STREAM_SKIPFIRST_THROTTLE(logger, clock, duration, stream_arg)  // NOLINT
+
+#define EZ_DEBUG(...) RCLCPP_DEBUG(get_logger(), __VA_ARGS__)  // NOLINT
```
kylestach, line 37: Do we ever anticipate using the `RJ_*` macros from above? I think we'll almost always be using the default logger. Why not make the default ones `RJ_*` and name the custom logger ones something else?

kylestach (changes requested):
Dumb question: after the previous PR removing `colcon`, what is the recommended method of building and running?

Also, I'm a little bit confused by some of the design decisions you've made with the config system. Some of these are mentioned in other comments, but for visibility:
1. What is the point of the config client node?
2. There doesn't appear to be a way to set and get (or callback on change) on individual parameters. This is a very important and necessary feature for most of the nodes we use, which care about changes in individual nodes but not in the entire `GameSettings` struct.
(Also, aren't you supposed to be at work?)

install/setup.bash
oswinso: I thought ament was hardcoding versions, turns out I was wrong.

rj_utils/include/rj_utils/logging.hpp
oswinso: I think it varies file to file

rj_config/src/config_server/config_server.cpp
oswinso: `clang-tidy` complains about an unnecessary copy of a `std::shared_ptr`, but because ros2 uses SFINAE I can't change this to be a const ref

rj_msgs/srv/SetGameSettings.srv
oswinso: Yeah, it means that when you want to set one thing, you need to send an updated version of the entire struct. I don't think that's really that bad, considering how small `GameSettings` is. Also, it means that keeping track of state is easy, as you don't have to manually apply each "update",

rj_config/include/config_client/config_client.h
oswinso: Using the parameter server doesn't give you a callback. This does. I'll add in a way to specify a callback.

rj_geometry_msgs/package.xml
oswinso: Yes, except the `ros2` cli uses them, ie. if you don't have this for messages specifically then `ros2 topic` doesn't work.

rj_config/src/config_client/config_client_node.cpp
oswinso: It's so that something like `Processor` which isn't itself a `rclcpp::Node` can use this, since `config_client::ConfigClient` basically tacks on subscribers to an existing node.

rj_utils/include/rj_utils/logging.hpp
oswinso: Yeah, I think we'll usually be using the default logger. There might be use cases for using different loggers for like when a node is multithreaded and you want to filter out logs by thread or something.

Geometry2d/include/Geometry2d/Line.hpp
oswinso: I guess implicit conversions would make it nice to use.

oswinso:
@kylestach I honestly don't really see a use case for a node to know particularly _which_ field changed. I would imagine that for a normal node, if a param it uses changes, then it would just run something that updates a bunch of stuff based on the entirety of the new struct, even though it could do it just on the one updated field, though honestly it feels like the performance difference is probably something that is negligible. Maybe for more specialized things we could do something that's more tailored towards that node's use case, but in general I don't really see it.

install/setup.bash
oswinso: Ok changed.

oswinso: Ok I changed this to use the same thing ament is using

kylestach:
It's a semantics issue, not really a performance issue. When `blueTeam` or `defendPlusX` changes you want to reset all of the robots and planners, `blueTeam` changes = changing ports, joystick change => resetting the relevant robots, etc.

rj_config/src/config_client/config_client_node.cpp
kylestach: ACK, so it's a temporary measure. Can you document it as such?

rj_config/src/config_server/config_server.cpp
kylestach: Weird, but okay.

rj_config/include/config_client/config_client.h
kylestach: ACK

rj_geometry_msgs/package.xml
kylestach: Sigh, okay.

kylestach:
Obviously those cases can be handled by just caching the previous value though, so not a big deal. Please make sure we get callbacks in, though.

Other than that, nice work! This is looking pretty good.

install/setup.zsh
oswinso: Done.

rj_config/src/config_client/config_client_node.cpp
oswinso: Sure.

rj_config/include/config_server/config_server.h
oswinso: Done.

rj_msgs/msg/VisionPacket.msg
oswinso: Deleted.

launch/config_server.launch.py
oswinso: I'll add it in later - it'll make more sense.

oswinso:
LLVM apt repo is down or something, all of our CI runs are failing.

rj_config/include/config_client/config_client.h
```diff
@@ -0,0 +1,67 @@
...
+using SetGameSettingsReq = SetGameSettingsSrv::Request;
+
+using FieldDimensionsMsg = rj_msgs::msg::FieldDimensions;
+using SetFieldDimensionsSrv = rj_msgs::srv::SetFieldDimensions;
+using SetFieldDimensionsReq = SetFieldDimensionsSrv::Request;
+
+// TODO(1520): Add callback functionality to ConfigClient.
+
+/**
+ * \brief Helper util that sets up the proper subscribers for
```
JNeiger:
```suggestion
 * @brief Helper util that sets up the proper subscribers for
```

JNeiger (approved):
One typo, but lgtm after that

rj_utils/include/rj_utils/logging.hpp
kylestach: Okay, we'll cross that bridge when we get to it I guess.

kylestach (approved):
LGTM

kylestach (approved):
LGTM

PR #1526, opened 2020-06-21
[MP Rewrite Part 7] Add global obstacles; fix dynamic obstacle properties

kylestach:
This PR adds global and goal-zone obstacles to the planning stack. It also fixes how we deal with dynamic obstacles by essentially treating them as static after they reach the end of their trajectory.


soccer/gameplay/GameplayModule.cpp
```diff
@@ -314,6 +314,9 @@ void Gameplay::GameplayModule::run() {
     _ballMatrix = Geometry2d::TransformMatrix::translate(
         _context->world_state.ball.position);
 
+    _context->globalObstacles = globalObstacles();
+    _context->goalZoneObstacles = goalZoneObstacles();
```
JNeiger, line 281: Does it make sense to have the global obstacles set here? It doesn't really feel like it fits. For the sake of this PR though, this is fine

JNeiger (approved):

soccer/planning/low_level/CreatePath.cpp
```diff
@@ -52,7 +52,7 @@ Trajectory rrt(const LinearMotionInstant& start,
 
     ShapeSet obstacles = static_obstacles;
     Trajectory path{{}};
-    constexpr int attemptsToAvoidDynamics = 10;
+    constexpr int attemptsToAvoidDynamics = 3;
```
oswinso: I guess probably make this a param as well?

oswinso (approved):
lgtm

soccer/gameplay/GameplayModule.cpp
kylestach: Yeah this has always been where they're created but I think it really belongs with the field geometry calculations.

PR #1527, opened 2020-06-21
Vision Receiver refactor to ROS2

oswinso:
## Description
This PR pulls our vision receiver out of `soccer` and into a separate node:
1. Created conversions for `RJ::Time`, `Field_Dimensions` and `GameSettings`
2. Created `VisionReceiver` node that publishes vision packets and the raw protobuf packets.
3. Changed `VisionFilter` to have another thread that listens for the vision packets, bypassing `Processor`

### Test 1
1. `source /opt/ros/foxy/setup.sh && mkdir build && cd build && cmake -GNinja -DCMAKE_INSTALL_PREFIX=../install .. && ninja install`
2. `cd .. && source install/setup.sh`
3. `ros2 launch sim.launch.py`

Expected result: Everything still works


rj_vision_receiver/include/rj_vision_receiver/vision_receiver.h
```diff
@@ -0,0 +1,138 @@
...
+
+    void setPort(int port);
+
+private:
+    void startReceive();
+    void receivePacket(const boost::system::error_code& error,
+                       std::size_t num_bytes);
+
+    // Helper function to decide whether or not to remove a robot at a given
+    // x position
```
JNeiger: This should just get merged with the doc comment if it's actually needed

rj_vision_receiver/include/rj_vision_receiver/vision_receiver.h
```diff
@@ -0,0 +1,138 @@
...
+    /**
+     * @brief Given the current config on which half we are enabling vision on,
+     * return whether the x coordinate lies in an enabled half.
+     * @param defend_plus_x Whether we are defending x>0.
+     * @param x The x coordinate.
+     * @return Whether the x coordinate lies in an enabled half.
+     */
+    [[nodiscard]] bool InUsedHalf(bool defend_plus_x, double x) const;
```
JNeiger, line 67: Whats the style for private class functions? It's inconsistent in this class

rj_vision_receiver/src/vision_receiver.cpp
```diff
@@ -0,0 +1,361 @@
+#include <rj_vision_receiver/vision_receiver.h>
```
JNeiger, line 1:
```suggestion
#include "rj_vision_receiver/vision_receiver.h"
```

rj_vision_receiver/src/vision_receiver.cpp
```diff
@@ -0,0 +1,361 @@
...
+#include <stdexcept>
+
+namespace vision_receiver {
+using boost::asio::ip::udp;
+
+VisionReceiver::VisionReceiver()
+    : Node{"vision_receiver"}, config_{this}, port_{-1}, _socket{_io_context} {
+    _recv_buffer.resize(65536);
+
+    // There should be at most four packets: one for each camera on each of two
```
JNeiger: The number of cameras shouldn't be hard coded if possible. Id rather just have a larger buffer

soccer/src/Processor.cpp
```diff
@@ -200,30 +146,24 @@ void Processor::run() {
...
         _manual_control_node->run();
 
-        updateOrientation();
+        _config_client->run();
 
-        // TODO(Kyle): Don't do this here.
-        // Because not everything is on modules yet, but we still need things to
-        // run in order, we can't just do everything via the for loop (yet).
-        _visionReceiver->run();
+        updateOrientation();
```
JNeiger: `updateOrientation` can be removed. This includes `_teamAngle` and `_worldToTeam`

soccer/src/Processor.cpp
```diff
@@ -200,30 +146,24 @@ void Processor::run() {
         _sdl_joystick_node->run();
         _manual_control_node->run();
 
-        updateOrientation();
+        _config_client->run();
```
JNeiger, line 152: This is just temporary?

soccer/src/ros2_temp/detection_frame_sub.h
```diff
@@ -0,0 +1,59 @@
...
+
+    /**
+     * @brief Returns whether this node is ok, ie. config_client_ is connected.
+     * @return
+     */
+    [[nodiscard]] bool ok() const { return config_client_->connected(); }
+
+    /**
+     * @brief Returns the team angle (It's horrible, but it's only temporary
+     * until VisionFilter gets refactored).
```
JNeiger, line 34: Does it make sense to do the rotation in the vision filter or the receiver? To me, if we're going to ingest data into the system, we should convert correctly at that boundary and rather than deal with it later.

Before, the ros conversions were
`SSL Protobuf {->} Camera* -> Vision filter`
Now its
`SSL Protobuf -> ROS Msg {->} Camera* -> Vision Filter`
where `{}` is the rotation.

I'd rather the `ROS Msg` just be the `Camera*`
`SSL Proto {->} POD Camera* ROS Msg -> Camera* -> Vision Filter`

soccer/src/ros2_temp/message_queue.h
```diff
@@ -0,0 +1,100 @@
...
+     * @param qos
+     */
+    MessageQueueNode(const std::string& name, const std::string& topic,
+                     const rclcpp::QoS& qos = rclcpp::QoS{5});
+
+    /**
+     * @brief Inserts all of the messages in the queue into the passed in
+     * vector.
+     * @param vector
+     * @return
```
JNeiger: @return should be filled in more in this whole file

soccer/src/ros2_temp/message_queue.h
```diff
@@ -0,0 +1,100 @@
+#include <queue>
+#include <rclcpp/rclcpp.hpp>
+#include <rj_utils/logging.hpp>
+
+namespace ros2_temp {
+/**
+ * @brief A temporary node that acts as a message queue for messages.
```
JNeiger, line 9: So is this getting removed once we fully convert?

soccer/src/vision/VisionFilter.cpp
```diff
@@ -22,9 +25,28 @@ VisionFilter::~VisionFilter() {
...
+void VisionFilter::run() {
+    // Fill the list of our robots/balls based on whether we are the blue team
+    // or not
+    fillBallState(context_->state);
+    fillRobotState(context_->state, context_->game_state.blueTeam);
+}
+
+void VisionFilter::GetFrames() {
+    std::vector<DetectionFrameMsg::UniquePtr> raw_frames =
+        detection_frame_sub_->GetFrames();
```
JNeiger, line 37: Will we have multi-thread problems with the sub msg receive callback?

soccer/src/vision/VisionFilter.cpp
```diff
@@ -91,7 +115,7 @@ void VisionFilter::updateLoop() {
 
         {
             // Do update with whatever is in frame buffer
-            std::lock_guard<std::mutex> lock1(frameLock);
+            GetFrames();
             {
```
JNeiger: I would remove this set of `{ }`

soccer/src/vision/VisionFilter.cpp
```diff
@@ -22,9 +25,28 @@ VisionFilter::~VisionFilter() {
...
 }
 
-void VisionFilter::addFrames(const std::vector<CameraFrame>& frames) {
-    std::lock_guard<std::mutex> lock(frameLock);
-    frameBuffer.insert(frameBuffer.end(), frames.begin(), frames.end());
+void VisionFilter::run() {
+    // Fill the list of our robots/balls based on whether we are the blue team
+    // or not
+    fillBallState(context_->state);
+    fillRobotState(context_->state, context_->game_state.blueTeam);
```
JNeiger, line 32: Is this just to make it more "roslike" for the future?

soccer/src/vision/VisionFilter.hpp
```diff
@@ -51,16 +46,33 @@ class VisionFilter {
...
     std::mutex worldLock;
     World world;
 
     std::atomic_bool threadEnd{};
 
-    std::mutex frameLock;
     std::vector<CameraFrame> frameBuffer{};
+    Context* context_;
+    std::unique_ptr<ros2_temp::DetectionFrameSub> detection_frame_sub_;
+    rclcpp::Time last_update_time_;
```
JNeiger: I would prefer getting everything into RJ time if possible. I think people were on board to limit the ROS types as much as possible

launch/sim.launch.py
```diff
@@ -1,17 +1,37 @@
...
+                  on_exit=Shutdown())
+
+    config_server = Node(package='rj_robocup',
+                         executable='config_server',
+                         output='screen',
+                         on_exit=Shutdown())
+
+    vision_receiver = IncludeLaunchDescription(
+        PythonLaunchDescriptionSource(
+            os.path.join(launch_dir, "vision_receiver.launch.py")))
```
JNeiger, line 34: Are long term nodes going to all be built this way?

rj_vision_receiver/include/rj_vision_receiver/vision_receiver.h
oswinso: I'm just basically using whatever style for now until we do the big style refactor.

rj_vision_receiver/src/vision_receiver.cpp
oswinso: I'm gonna save this for when we enforce includes style using `clang-format`

rj_vision_receiver/src/vision_receiver.cpp
oswinso: This was what it was like before. I'll make a TODO.

rj_vision_receiver/src/vision_receiver.cpp
JNeiger: Cool cool. I just did that recently for another project. I'm on clang-format-8 in that project and it wasn't able to convert `<>` to `""` automatically when doing the main include stuff. I had to manually go back to every file and change to `""` for the include sorting to work.

soccer/src/Processor.cpp
oswinso: Yeah, we still need a way for all the old stuff in Processor to get access to the updated `FieldDimensions` and `GameSettings` stuff, so this is here until we can move everything off of Processor.

soccer/src/ros2_temp/detection_frame_sub.h
oswinso: That makes sense. I'll move the logic to the receiver.

soccer/src/ros2_temp/message_queue.h
oswinso: Yeah.

soccer/src/vision/VisionFilter.cpp
oswinso: Yeah, the next PR will have VisionFilter directly publish this to the nodes that need it and so hopefully `context_->state` should not be a thing anymore.

soccer/src/vision/VisionFilter.hpp
oswinso: @kylestach thoughts on `RJ::Time` vs `rclcpp::Time`?

launch/sim.launch.py
oswinso: What do you mean by that?

JNeiger: Basically, for the receiver, you have a second file, but for the other ones, you just build it in this file

oswinso: Oh. I think long term we're going to want to split it up into separate files and then include them, so that we can compose them in different configurations and whatnot

rj_vision_receiver/include/rj_vision_receiver/vision_receiver.h
oswinso: Done.

soccer/src/Processor.cpp
oswinso: Done.

soccer/src/vision/VisionFilter.cpp
oswinso: There shouldn't be. `frameBuffer` is only accessed by the `worker` thread. The internal queue has two threads accessing it, but it is protected by a mutex:
https://github.com/RoboJackets/robocup-software/blob/ca006d54675c1e2ef99d77fa43bd64ce97d4087e/soccer/src/ros2_temp/message_queue.h#L77-L89
and
https://github.com/RoboJackets/robocup-software/blob/ca006d54675c1e2ef99d77fa43bd64ce97d4087e/soccer/src/ros2_temp/message_queue.h#L53-L63

soccer/src/ros2_temp/message_queue.h
oswinso: Done.

soccer/src/vision/VisionFilter.cpp
oswinso: Done.

launch/vision_receiver.launch.py
```diff
@@ -0,0 +1,21 @@
+from launch import LaunchDescription
+
+from launch_ros.actions import Node
+
+
+def generate_launch_description():
+    sim_vision_port = 10020
```
kylestach, line 7: The idea is that we eventually specify one of these ports based on different command-line options or a different top-level sim launch, right?

rj_common/include/rj_common/time.hpp
```diff
@@ -33,6 +35,19 @@ constexpr Timestamp timestamp(Time time) {
...
+                        time.time_since_epoch())
+                        .count();
+
+    return rclcpp::Time{nanos};
+}
+
+inline RJ::Time FromROSTime(const rclcpp::Time& time) {
+    const std::chrono::nanoseconds dur(time.nanoseconds());
+    return RJ::Time{dur};
+}
```
kylestach, line 49: Eventually we should really unify these types.

rj_config/include/config_client/config_client_node.h
```diff
@@ -15,23 +15,29 @@ class ConfigClientNode : public rclcpp::Node {
...
+    [[nodiscard]] const GameSettingsMsg& gameSettingsThreaded() const {
+        return config_client_.gameSettingsThreaded();
     }
 
-    inline void updateGameSettings(const GameSettingsMsg& msg) {
+    [[nodiscard]] const FieldDimensionsMsg& fieldDimensionsThreaded() const {
+        return config_client_.fieldDimensionsThreaded();
+    }
+
+    [[nodiscard]] bool connected() const { return config_client_.connected(); }
```
kylestach, line 34: _technically_ should we not have a `connectedThreaded`?

rj_config/src/config_client/config_client.cpp
```diff
@@ -1,26 +1,34 @@
...
+#include <rj_constants/topic_names.hpp>
 #include <rj_utils/logging.hpp>
 
 namespace config_client {
-ConfigClient::ConfigClient(rclcpp::Node* node) : node_{node} {
+ConfigClient::ConfigClient(rclcpp::Node* node)
+    : node_{node},
+      game_settings_{std::nullopt},
+      field_dimensions_{std::nullopt} {
     const auto latching_qos = rclcpp::QoS(1).transient_local();
```
kylestach, line 11: We should abstract the common QoS builder functions out into a package like `rj_ros_utils` or something.

rj_vision_receiver/include/rj_vision_receiver/stamped_wrapper_packet.h
```diff
@@ -0,0 +1,17 @@
...
+
+#include <rj_protos/messages_robocup_ssl_wrapper.pb.h>
+
+#include <rclcpp/time.hpp>
+
+namespace vision_receiver {
+/**
+ * @brief A struct that adds receive time information to a SSL_WrapperPacket.
+ */
+struct StampedSSLWrapperPacket {
```
kylestach, line 11: We do `StampedX` often enough (`StampedRobotInstant`, `StampedTrajectory`, etc.) that we should really just introduce a templated struct for it at some point. Not something we should address here though.

rj_vision_receiver/include/rj_vision_receiver/vision_receiver.h
kylestach: This is fine with me but also, lmao.

rj_vision_receiver/src/vision_receiver.cpp
```diff
@@ -0,0 +1,362 @@
...
+    msg.height = robot.height();
+    return msg;
+}
+
+void VisionReceiver::SyncDetectionTimestamp(DetectionFrameMsg* frame,
+                                            const rclcpp::Time& receive_time) {
+    // Assume that the frame was sent when we received it, so offset the capture
+    // time by the difference (ie. estimated drift in clock between this
+    // computer and the vision computer).
+    frame->t_capture = receive_time - frame->t_sent + frame->t_capture;
```
kylestach, line 179: I think this is going to give us the opportunity to see a big improvement vision filter side, considering we don't currently try to keep track of timestamps at all. :smile:

rj_vision_receiver/include/rj_vision_receiver/vision_receiver.h
```diff
@@ -0,0 +1,138 @@
...
+
+    /**
+     * @brief Converts from a SSL_WrapperPacket to a DetectionFrameMsg,
+     * respecting the "exclude half" config as well as performing timestamp
+     * synchronization.
+     * @param frame
+     * @param received_time Time that this packet was received.
+     * @return
+     */
+    [[nodiscard]] DetectionFrameMsg ToROSMsg(
```
kylestach: I would prefer not to call this `ToROSMsg` given that it does some non-trivial computation (rather than just "conversion"). Could you change the name?

rj_vision_receiver/include/rj_vision_receiver/vision_receiver.h
```diff
@@ -0,0 +1,138 @@
...
+        const SSL_DetectionFrame& frame,
+        const rclcpp::Time& received_time) const;
+
+    /**
+     * @brief Converts from SSL_WrapperPacket to a RawProtobufMsg.
+     * @param packet
+     * @return
+     */
+    [[nodiscard]] static RawProtobufMsg::UniquePtr ToROSMsg(
```
kylestach: I think this probably belongs in a free function rather than associated with the vision receiver (all of the `static` variants do)

rj_vision_receiver/src/vision_receiver.cpp
```diff
@@ -0,0 +1,362 @@
...
+    // TODO(#1528): Don't hardcode the number of cameras.
+    _packets.reserve(4);
+
+    // Parameters
+    declare_parameter<int>("port", SharedVisionPortSinglePrimary);
+    get_parameter<int>("port", port_);
+
+    double hz{};
+    declare_parameter<double>("hz", 60.0);
+    get_parameter<double>("hz", hz);
```
kylestach: I think at some point we should have a nicer interface for setting and getting parameters that keeps you from messing things up at compile time. Something like:

```cpp
// vision_parameters.hpp
struct SharedVisionPortSinglePrimary
{
    constexpr static std::string Name = "port";
    using Type = int;
};

// vision_receiver.cpp
double hz = declare_and_get_parameter<SharedVisionPortSinglePrimary>();
```
(Probably with a macro to make the declaration side simpler)

rj_vision_receiver/src/vision_receiver.cpp
```diff
@@ -0,0 +1,362 @@
...
+
+    auto period = std::chrono::duration<double>(1 / hz);
+
+    raw_packet_pub_ =
+        create_publisher<RawProtobufMsg>(topics::kRawProtobufPub, 10);
+    detection_frame_pub_ =
+        create_publisher<DetectionFrameMsg>(topics::kDetectionFramePub, 10);
+
+    // Create timer for callback
+    timer_ = this->create_wall_timer(period, [this]() { run(); });
```
kylestach: Is there a simple-ish way to change this into a blocking wait in another thread? I think that would be preferable to polling at some (relatively low) polling rate. If that's not straightforward, we should just bump the poll rate up decently high (~180Hz) - right now the receive time could be off by as much as 16ms, which is not ideal.

rj_vision_receiver/src/vision_receiver.cpp
```diff
@@ -0,0 +1,362 @@
...
+    if (!connected) {
+        connected = true;
+        EZ_INFO("Connected!");
+    }
+
+    // Let boost::asio check for new packets
+    _io_context.poll();
+
+    // Process and clear new packets.
+    processNewPackets();
```
kylestach: Why do these in two separate steps? It seems like "publishing the packet" should just be part of the receive handler.

rj_vision_receiver/src/vision_receiver.cpp
```diff
@@ -0,0 +1,362 @@
...
+ * Updates the geometry packet in `Context` based on data from the vision packet
+ */
+void VisionReceiver::UpdateGeometryPacket(
+    const SSL_GeometryFieldSize& fieldSize) {
+    if (fieldSize.field_lines_size() == 0) {
+        return;
+    }
+
+    const SSL_FieldCicularArc* center = nullptr;
+    float penaltyShortDist = 0;  // default value
```
kylestach, line 240: Could we just standardize to doubles for all of these?

soccer/src/ros2_temp/detection_frame_sub.h
kylestach: I'm not sure if I agree. Actually, I think it probably makes the most sense for the vision filter to operate in _world_ coordinates (not team coordinates) and then do the conversion at publish time.

For that matter, I think the vision filter should also operate on "blue/yellow" robots, and convert to "ours/theirs" at the boundary. This would avoid the (admittedly trivial, but still mildly annoying) "reset" that happens when you switch color/team directions.

That obviously wouldn't be a change that belongs in this PR though, so for now I don't really care where the conversion goes.

soccer/src/vision/VisionFilter.hpp
kylestach: So, I have mixed thoughts on this.

I _really_ think we should be careful about how we use `Time` types. You should pretty much _never_ call `now()` (or equivalent) for anything meaningful except at very defined boundaries (i.e. at the node level) - this makes it a lot easier to mock things out later on. I think ROS does a really good job of enforcing this by forcing you to have a `Node` in order to actually get the current time, but on the other hand limiting ROS types except at the boundaries is a Good Thing™.

I think a good compromise-ish would be to never call `RJ::now()` (use the ROS equivalent instead) but to use `RJ::Time` internally.

Tangentially, `RJ::Seconds` is evil; change my mind.

soccer/src/vision/ball/CameraBall.hpp
```diff
@@ -16,6 +19,13 @@ class CameraBall {
     CameraBall(RJ::Time timeCaptured, Geometry2d::Point pos)
         : timeCaptured(timeCaptured), pos(pos) {}
 
+    /**
+     * @brief Implicit conversion from DetectionBallMsg.
```
kylestach: This isn't an implicit conversion, there are multiple parameters...

That said, this is fine.

soccer/src/vision/VisionFilter.cpp
```diff
@@ -91,27 +115,27 @@ void VisionFilter::updateLoop() {
...
         }
 
         // Wait for the correct loop timings
         RJ::Seconds diff = RJ::now() - start;
-        RJ::Seconds sleepLeft = RJ::Seconds(*VisionFilterConfig::vision_loop_dt) - diff;
+        RJ::Seconds sleepLeft =
+            RJ::Seconds(*VisionFilterConfig::vision_loop_dt) - diff;
 
         if (diff > RJ::Seconds(0)) {
             std::this_thread::sleep_for(sleepLeft);
```
kylestach, line 135: `std::this_thread::sleep_until` is what you want here.

soccer/src/vision/camera/CameraFrame.hpp
```diff
@@ -19,15 +23,41 @@ class CameraFrame {
...
-                  cameraRobotsYellow(cameraRobotsYellow),
-                  cameraRobotsBlue(cameraRobotsBlue) {}
+        : tCapture(tCapture),
+          cameraID(cameraID),
+          cameraBalls(std::move(cameraBalls)),
+          cameraRobotsYellow(std::move(cameraRobotsYellow)),
+          cameraRobotsBlue(std::move(cameraRobotsBlue)) {}
+
+    /**
+     * @brief Implicit conversion from DetectionFrameMsg.
```
kylestach: Still not implicit.

soccer/src/ros2_temp/raw_vision_packet_sub.h
```diff
@@ -0,0 +1,38 @@
...
+private:
+    Context* context_;
+    rclcpp::executors::SingleThreadedExecutor executor_;
+    std::shared_ptr<MessageQueueNode<RawProtobufMsg>> queue_;
+    std::thread worker_;
+
+    /**
+     * @brief Calls executor_.spin().
+     */
+    void spin();
```
kylestach: This should probably be `spin_forever()` considering we have other places in the codebase (all of gameplay) where `spin()` = `spin_once()`.

kylestach (changes requested):
Ideally the vision receiver node should spin up a worker thread and queue up messages using blocking `io_context::run()`, rather than polling, in order to shave off some milliseconds of incorrectness on receive time estimate.

Otherwise, most of these comments are notes for things we could do in the future to make code nicer. I think for the most part this is okay, I really hate the raw protobuf hack but it is fine until we take a hatchet to the whole logging system.

rj_vision_receiver/src/vision_receiver.cpp
oswinso: I'm just copy-pasting whatever logic we were previously using, nothing new here.

rj_vision_receiver/src/vision_receiver.cpp
oswinso: or actually the other thread can just publish directly.

rj_vision_receiver/src/vision_receiver.cpp
kylestach: Oops. Well, I guess we just never used it, cool.

rj_vision_receiver/src/vision_receiver.cpp
oswinso: I'm just copy-pasting. Standardizing `float` to `double` is out of the scope of this PR imo

rj_vision_receiver/src/vision_receiver.cpp
kylestach: I mean, yeah. Point is that this should be moved into another thread.

soccer/src/vision/VisionFilter.cpp
oswinso: I'd rather keep this PR related `vision_receiver` as much as possible and avoid changing `VisionFilter` related functionality even if its bad, and leave that for a different PR just to keep this one focused and small.

launch/vision_receiver.launch.py
oswinso: yeah

soccer/src/vision/VisionFilter.cpp
kylestach: Oh sure, that's fine. Just add a TODO.

Edit: on second thought, honestly the update cycle of the vision filter should be reworked anyway so this is probably fine for now.

rj_vision_receiver/src/vision_receiver.cpp
kylestach: I would disagree in that we want to make the `.msg` itself a `double`. But if you'd rather do it later I'm not going to block over it.

soccer/src/vision/ball/CameraBall.hpp
oswinso: Used to be, until I realized that it needed more information to be converted lmao

soccer/src/ros2_temp/detection_frame_sub.h
JNeiger: That's probably the best option in the end. The filter should be built in a way to support this change without anything internally changing. If it's on the docket to change during this ros update, I'm ok with it where ever, but if it's a thing we're throwing on the todo list for far into the future, I'd rather do the conversion in the receiver as it's a better "temporary" solution imo

soccer/src/vision/VisionFilter.hpp
JNeiger: I'm curious now, what's the issue with `RJ::Seconds`?

soccer/src/ros2_temp/detection_frame_sub.h
oswinso: I'm refactoring `VisionReceiver` next, so it'll probably happen after this PR gets merged

rj_vision_receiver/src/vision_receiver.cpp
oswinso: Made #1529 and added a todo

rj_vision_receiver/src/vision_receiver.cpp
oswinso: Sure.

soccer/src/vision/VisionFilter.hpp
kylestach: I mean in general it's not like it shouldn't be used at all, we just use it far too often to represent "some time in the future" when we really should be using an `RJ::Time`. I think at least half of it's usage in our codebase (at least pre-motion planning rewrite) should have been replaced with a `Time`.

oswinso: Any examples of this misuse?

kylestach: https://github.com/RoboJackets/robocup-software/blob/8ed909cf54317f176b4ba71369532897228af4c9/soccer/planning/trajectory/Trajectory.hpp#L65
Literally everything in the old `Path` interface (and everything that used it).
https://github.com/RoboJackets/robocup-software/blob/b0cad9cac38505a84617d72ed14fd32410745fa5/soccer/planning/paths/TrapezoidalPath.hpp#L43
https://github.com/RoboJackets/robocup-software/blob/7b68a9ebef777944fdee07d4bb2b824b5471f571/soccer/GameState.hpp#L54

oswinso: I think we should create our own `RJ::Time` and `RJ::Duration` that isn't just a type alias. I think if we call it `RJ::Duration` then it'll be easier to not misuse, ie. `RJ::Duration initialTime` just looks and sounds so much more wrong than `RJ::Seconds initialTime` (Who knows, it might be seconds since epoch or something).

rj_config/src/config_client/config_client.cpp
oswinso: wdym? whats wrong with their builder?

kylestach: Eh, scratch that. I think the correct way to address the "get the latest value" use case looks a bit different (i.e. make a `LatestValue<T>` struct), but this is fine for now.

rj_vision_receiver/src/vision_receiver.cpp
oswinso: I'm keeping these in two separate steps so that the we get the receive time as accurate as possible if we want to try and estimate clock skew things later on.

soccer/src/ros2_temp/raw_vision_packet_sub.h
oswinso: Done.

soccer/src/vision/camera/CameraFrame.hpp
oswinso: Changed

rj_vision_receiver/include/rj_vision_receiver/vision_receiver.h
oswinso: Done.

rj_config/include/config_client/config_client_node.h
oswinso: done.

rj_vision_receiver/include/rj_vision_receiver/stamped_wrapper_packet.h
oswinso: #1532

rj_vision_receiver/include/rj_vision_receiver/vision_receiver.h
oswinso: done.

rj_vision_receiver/src/vision_receiver.cpp
oswinso: done.

rj_utils/include/rj_utils/concurrent_queue.hpp
```diff
@@ -0,0 +1,74 @@
...
+     */
+    T Get() {
+        std::unique_lock<std::mutex> lock{mutex_};
+        cv_.wait(lock, [this]() { return !queue_.empty(); });
+        auto item = std::move(queue_.front());
+        queue_.pop();
+        return item;
+    }
+
+    bool TryGet(T& item, const std::chrono::milliseconds& timeout) {
```
kylestach, line 28: Really this should just return an `std::optional`, right? Whatever, fine for now.

kylestach (approved):

JNeiger (approved):

Issue #1533, opened 2020-06-27
Optimize Geometry2D


kylestach:
## Description
Many components in our Geometry2d are horifically unoptimized. As part of the motion planning rewrite I've had to fix a few of them. Some examples:
 - `ShapeSet` collision detection: first called `hitSet` to construct a set of all obstacles, then returned whether that set was empty (huge allocation overhead, could be fixed just by checking if each shape hits in sequence)
 - `Rect`collision: constructs a vector of four `Segment`s and checks collision for each one independently (can be calculated very easily).

In general, there are likely many other cases of similar garbage in the Geometry2d library. Considering that it's basically used everywhere in our code, we should really put in the effort to vet it with a fine-toothed comb to make sure we're not throwing away CPU cycles and allocating for no reason.

## Requirements
 - It should be possible to construct a non-owning shape set, to avoid `shared_ptr` overhead
    - Idea: template it on the contained type
 - Each algorithm in `Geometry2d` should be checked to make sure that it is at least of reasonable complexity. Any known naïve algorithms in Geometry2d should be marked as such.

## Design
N/A

PR #1534, opened 2020-06-27
[MP Rewrite Part 8] Optimize geometry checking

kylestach:
## Description
This optimizes the field geometry checking to use sane rectangle collision algorithms.

## Associated Issue
Issue #1533


common/Geometry2d/Rect.cpp
```diff
@@ -237,20 +237,17 @@ bool Rect::nearPoint(Point other, float threshold) const {
...
-            return true;
-        }
-    }
+    double dx = std::max<double>({0.0, max_x - other.x(), other.x() - min_x});
+    double dy = std::max<double>({0.0, max_y - other.y(), other.y() - min_y});
 
-    return false;
+    double distance_to_point = std::sqrt(dx * dx + dy * dy);
+
+    return distance_to_point < threshold;
```
oswinso: I don't understand how this logic works, can you explain it?

kylestach: I'll add comments, but it's pretty simple. Because the box is axis-aligned, it's easy to find the minimum distance in the `x` and `y` directions separately (at most one of `x - x_max (above), 0 (inside), x_min - x (below)` will be positive, where "positive" represents "outside").

oswinso:
I'm a bit concerned about that typo making it through tests. Can you add a test that checks this?

common/Geometry2d/Rect.cpp
oswinso: ok yeah after you fix the typo it makes sense.

kylestach:
That typo didn't make it through tests, lol. I just never bothered to run them.

JNeiger (approved):

oswinso (approved):
Once you fix the tests, lgtm

kylestach:
@oswinso the tests are fine. The failures are for unrelated reasons.

PR #1535, opened 2020-06-28
build + install ROSIDL stuff at configure time so that cross-package dependencies work

oswinso:
## Description
This PR creates `cmake/AddROSIDLSubdirectory.cmake` which builds + installs things at configure time, so that `find_package()` of things like ROSIDL messages work.

### Everything still works
1. Follow the same steps as in #1527


kylestach (approved):
Dumb workaround but LGTM. :+1:

JNeiger (approved):

PR #1537, opened 2020-06-28
[MP Rewrite Part 10] Overall improvements to planners

kylestach:
## Description
Improve planners, especially pivot planner. Make angle overrides work correctly.


soccer/planning/planner/PivotPathPlanner.cpp
```diff
@@ -71,24 +75,33 @@ Trajectory PivotPathPlanner::plan(const PlanRequest& request) {
         points.push_back(point);
     }
 
-    BezierPath pathBezier(points, start_instant.linear_velocity(), Point(0, 0),
+    BezierPath pathBezier(points, Point(0, 0), Point(0, 0),
```
JNeiger: This feels a little weird. What's the reasoning behind this?

soccer/planning/planner/PivotPathPlanner.cpp
```diff
@@ -71,24 +75,33 @@ Trajectory PivotPathPlanner::plan(const PlanRequest& request) {
...
 
     Trajectory path =
         ProfileVelocity(pathBezier, start_instant.linear_velocity().mag(), 0,
                         linear_constraints, start_instant.stamp);
+    {
+        // Stay in place
+        RobotInstant instant = path.last();
+        instant.stamp = instant.stamp + RJ::Seconds(5.0);
+        path.AppendInstant(instant);
+    }
```
JNeiger, line 87: This might be a visual glitch because I'm on mobile. These braces seem useless?

soccer/planning/planner/SettlePlanner.cpp
```diff
@@ -282,10 +281,18 @@ Trajectory SettlePlanner::intercept(
                 path.duration() + RJ::Seconds(*_interceptBufferTime) <=
                     ballTime ||
             !fieldRect.containsPoint(ballVelIntercept)) {
+            ball_intercept_maybe = ballVelIntercept;
             break;
         }
     }
 
+    Geometry2d::Point ballVelIntercept;
```
JNeiger, line 297: Can you throw in a comment about the edge case this will be triggered? I assume it's to deal with the case the ball stops before we would reach it

soccer/Processor.cpp
```diff
@@ -268,14 +266,16 @@ void Processor::run() {
         // Processor Initialization Completed
         _initialized = true;
 
-        // Log this entire frame
-        _logger->run();
+        {
+            loopMutex()->lock();
```
kylestach, line 270: This just keeps the UI from locking up quite as aggressively. It should probably be reverted.

soccer/planning/planner/PivotPathPlanner.cpp
kylestach: For pivot, you actually want to explicitly create a circular(ish) path rather than respecting initial velocity.

Without this, when you're constantly replanning the robot just sort of pushes the ball forwards forever.

soccer/planning/planner/PivotPathPlanner.cpp
kylestach: They're useless, I like to do this to make a temporary that shouldn't be used anywhere else. It's just personal preference, I can get rid of it if you'd rather not have it.

soccer/planning/planner/PivotPathPlanner.cpp
JNeiger: That makes sense. How well does it work with a fast moving robot coming into a pivot or a robot far away from the pivot?

soccer/planning/planner/PivotPathPlanner.cpp
JNeiger: I gotcha. Personally I'm not a fan of it because it messes with the tabbing whitespace, but I'm also not doing much coding anymore sooooo it's up to you guys

soccer/planning/planner/PivotPathPlanner.cpp
kylestach: It's a little jank when the robot comes in fast, but it actually gives decent results. It definitely needs to be fixed eventually, if we want to tune motion control really aggressively (which we do). No idea how it works if the robot is very far away from the pivot - with the transition to the action/behavior controller (hey, that's ABC!) we should honestly just fail if we attempt to do a pivot in that context.

JNeiger: Sounds good! So the idea would be to move close to a ball/have it captured, before calling this?

JNeiger (approved):
Nothing blocking, LGTM from my pov

soccer/planning/planner/PivotPathPlanner.cpp
```diff
@@ -39,12 +39,15 @@ Trajectory PivotPathPlanner::plan(const PlanRequest& request) {
 
     const auto& command = std::get<PivotCommand>(request.motionCommand);
 
-    if (!shouldReplan(command)) {
+    double radius = _pivotRadiusMultiplier->value() * Robot_Radius;
+    auto pivot_point = command.pivotPoint;
+
+    if (cached_pivot_point.has_value() && cached_pivot_point.value().distTo(pivot_point) < Robot_MouthWidth / 2) {
```
oswinso: What's the logic here? Why mouthwidth / 2?

soccer/planning/planner/PivotPathPlanner.cpp
oswinso: I think its a bit of visual clutter. I'm not a big fan of it either

soccer/planning/planner/PivotPathPlanner.hpp
```diff
@@ -17,11 +17,16 @@ class PivotPathPlanner : public PlannerForCommandType<PivotCommand> {
...
+    void reset() override {
+        previous = Trajectory{};
+        cached_pivot_point = std::nullopt;
+    }
 
 private:
-    [[nodiscard]] bool shouldReplan(const PivotCommand& command) const;
     Trajectory previous;
+
+    std::optional<Geometry2d::Point> cached_pivot_point;
```
oswinso, line 29: A comment about when this is cached / invalidated, and why we need to cache maybe?

oswinso (approved):
lgtm

soccer/planning/planner/PivotPathPlanner.cpp
JNeiger: The usual thought is that if you did the calculations perfectly, you have about this much wiggle room before you won't hit the ball with the mouth. Replanning for micrometers isn't worth it since replanning caused more performance problems than than being a little off historically.

This measurement falls apart a little when you have to worry about forwards backwards motion, but the thought process was always kinda "meh, there's worse things causing problems"

soccer/planning/planner/PivotPathPlanner.cpp
kylestach: Fixed in #1542 by deleting this code anyway, lmao

PR #1541, opened 2020-06-29
[MP Rewrite Part 12] Port the line kick planner

kylestach:
## Description
This is a straight port of the line kick planner that's currently on staging. I didn't put too much time into it - it honestly needs an overhaul but I think that's going to be separate from the rest of motion planning.


JNeiger (approved):
I ended up manually diffing against the master linekick and it looks good

soccer/planning/planner/LineKickPlanner.hpp
```diff
@@ -1,46 +1,43 @@
...
- * @brief Planner to kick a (possibly moving) ball without stopping to pivot
- * around it. The robot will move "through" the ball, and kick just as it comes
- * into contact with it.
+ * Planner which plans a path to line kick a ball.
+ * Uses the System State object to get the position of the ball
+ * and predict its Path. It chooses the closest intersection point
+ * with the ball Path it can reach in time and plans a Path so the
+ * ball and robot intersect at the same time.
+ *
+ * TODO(Kyle): Overhaul this entire planner. It's sketchy right now.
```
JNeiger, line 20: Whoops forgot this todo

oswinso (approved):
lgtm

soccer/planning/planner/LineKickPlanner.hpp
kylestach: Honestly we don't even really have a good idea of what we want to do with it, so I'm not comfortable making an issue for it yet.

oswinso: I think it'd be nice to make an issue to put it somewhere on the todo list so that we don't forget that we have a sketchy `LineKickPlanner`

PR #1542, opened 2020-06-30
[MP Rewrite Part 13] Stop play fixes and holding trajectory

kylestach:
## Description
Hold position at the end of the trajectory. This prevents "jumpy" behavior (also change the default move behavior to respect the planner's new requirement for the command to be newly set each frame).

Make a local obstacle for the ball in planning for each of the robots.


oswinso (approved):

JNeiger (approved):
Code looks good! I'm having trouble thinking how adding this behavior changes the overall behavior effectiveness, but if it looks good in sim its good enough for me

PR #1543, opened 2020-07-01
Merge MP rewrite into ROS 2

kylestach:
## Description
In theory there are no functional changes here. However, I'm experiencing dramatically reduced planner performance (both execution time and capability - collect is _really_ bad, it looks like it's hitting an old failure case), so I'm leaving this as a draft until I can figure out what is wrong.


kylestach:
This is ready for merge. Ideally reviewers could give it a quick test for sanity's sake.

JNeiger:
I don't have the right setup to test this, rip. You'll have to get someone else to do the test

JNeiger (approved):
Code looks good though

oswinso (approved):
Tried it out and I don't see anything broken

PR #1546, opened 2020-07-07
Set default build type to relwithdebinfo

oswinso:
## Description
Fixes another performance issue that was noticed in #1543 by ensuring that we default to `RelWithDebInfo` if no build type is specified.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -19,10 +19,7 @@ if(NOT NO_WALL AND CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Cl
 endif()
 
 # Set a default build type if none was specified
-set(default_build_type "Release")
-if(EXISTS "${CMAKE_SOURCE_DIR}/.git")
-    set(default_build_type "RelWithDebInfo")
-endif()
+set(default_build_type "RelWithDebInfo")
 
 if(NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
     message(STATUS "Setting build type to '${default_build_type}' as none was specified.")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -18,6 +18,21 @@ if(NOT NO_WALL AND CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Cl
     add_compile_options(-Wall -Wextra -Wpedantic)
 endif()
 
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -18,6 +18,21 @@ if(NOT NO_WALL AND CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Cl
     add_compile_options(-Wall -Wextra -Wpedantic)
 endif()
 
+# Set a default build type if none was specified
+set(default_build_type "Release")
+if(EXISTS "${CMAKE_SOURCE_DIR}/.git")
+    set(default_build_type "RelWithDebInfo")
```
JNeiger: What's the thought behind this?

JNeiger (approved):

CMakeLists.txt
oswinso: I think the only valid choices for debug cycles are `debug` and `relwithdebinfo`. Of the two, I prefer `relwithdebinfo` as it gives a closer approximation to the actual performance during comp with release mode than `debug`. Even though `relwithdebinfo` might inline stuff, I don't think needing to drop into gdb happens _that_ often.

I took this snippet from [here](https://blog.kitware.com/cmake-and-the-default-build-type/)

JNeiger: I was more talking about the `.git` check

oswinso: yeah I guess we don't really need that check, since its not like we're creating this as a library for others to use or something.

CMakeLists.txt
```diff
@@ -18,6 +18,18 @@ if(NOT NO_WALL AND CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Cl
...
+# Set a default build type if none was specified
+set(default_build_type "RelWithDebInfo")
+
+if(NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
+    message(STATUS "Setting build type to '${default_build_type}' as none was specified.")
+    set(CMAKE_BUILD_TYPE "${default_build_type}" CACHE
+        STRING "Choose the type of build." FORCE)
+    # Set the possible values of build type for cmake-gui
+    set_property(CACHE CMAKE_BUILD_TYPE PROPERTY STRINGS
+        "Debug" "Release" "MinSizeRel" "RelWithDebInfo")
```
kylestach, line 30: Imagine using CMake GUI...

kylestach (approved):
LGTM

PR #1547, opened 2020-07-08
ROS2 parameter library

### LINKED ISSUE #1529: [TODO] Parameter "library" with type safety for compile time checking
Pull request: #1527 

Link to TODO, or file number/line: `vision_receiver.cpp:29`

> I think at some point we should have a nicer interface for setting and getting parameters that keeps you from messing things up at compile time. Something like:
> 
> ```c++
> // vision_parameters.hpp
> struct SharedVisionPortSinglePrimary
> {
>     constexpr static std::string Name = "port";
>     using Type = int;
> };
> 
> // vision_receiver.cpp
> double hz = declare_and_get_parameter<SharedVisionPortSinglePrimary>();
> ```
> 
> (Probably with a macro to make the declaration side simpler)

_Originally posted by @kylestach in https://github.com/_render_node/MDE3OlB1bGxSZXF1ZXN0UmV2aWV3NDM0NTI3NjIw/pull_request_reviews/more_threads_

oswinso:
## Description
This PR creates a parameter library `rj_param_utils` for easy parameter declarations.

Closes #1529.

### Sim still works
1. Launch sim


rj_param_utils/include/rj_param_utils/param.h
```diff
@@ -0,0 +1,167 @@
...
+        : name_{name},
+          help_{help},
+          filename_{filename},
+          default_value_{param},
+          param_{param} {}
+
+    /**
+     * @brief Updates the parameter with the new value.
+     * @param new_value The new value of the parameter.
+     * @throws std::runtime_error if the type is set.
```
JNeiger: Maybe it's too early in the morning, but I don't think I get this comment?

rj_param_utils/include/rj_param_utils/param.h
```diff
@@ -0,0 +1,167 @@
...
+
+    template <typename ParamType>
+    ParamMap<ParamType>& GetParamMap();
+};
+
+}  // namespace params
+
+/**
+ * @brief Defines a namespaced parameter. The defined parameter can be used as
+ * PARAM_#, where * denotes # denotes the name passed in.
```
JNeiger: Typo? 
` * denotes # denotes`

rj_param_utils/include/rj_param_utils/param.h
```diff
@@ -0,0 +1,167 @@
...
+ * @param type The type of the parameter.
+ * @param prefix The prefix / namespace of the parameter, using . as a
+ * separator.
+ * @param name The variable name of the parameter.
+ * @param val The default value of the parameter.
+ * @param description A description of the parameter.
+ */
+#define DEFINE_NAMESPACED_VARIABLE(type, prefix, name, val, description)   \
+    namespace params::variables {                                          \
+    static type PARAM_##name##_storage = val;                              \
```
JNeiger: Since only the name is being used here, does it conflict if you have two config values "config1/number" and "config2/number" as the prefix/name?

rj_param_utils/include/rj_param_utils/param.h
```diff
@@ -0,0 +1,167 @@
...
+    const std::string filename_;
+    const T& default_value_;
+    T& param_;
+};
+
+/**
+ * @brief A ParamProvider represents a "provider" for parameters. In particular,
+ * it should provide a way of updating the parameters.
+ */
+class ParamProvider {
```
JNeiger, line 87: Can put comments on the functions here since they're the main interface?

rj_param_utils/include/rj_param_utils/param.h
```diff
@@ -0,0 +1,167 @@
...
+#define DEFINE_VARIABLE(type, name, val, description)          \
+    namespace params::variables {                              \
+    static type PARAM_##name##_storage = val;                  \
+    const type& PARAM_##name = PARAM_##name##_storage;         \
+    static ::params::internal::ParamRegisterer o_##name(       \
+        #name, description, __FILE__, PARAM_##name##_storage); \
+    }                                                          \
+    using params::variables::PARAM_##name;
+
+// Define the DEFINE_* macro for all supported types.
```
JNeiger, line 191: I'll let Kyle jump in, but I really like forcing the idea of namespaced parameters, it really keeps things from getting out of control in terms of keeping things clean. I would almost completely remove the root versions on just make the namespaced the default

rj_vision_receiver/src/vision_receiver.cpp
```diff
@@ -9,17 +9,21 @@
...
+      port_{-1},
+      _socket{_io_context},
+      param_provider_(this) {
     _recv_buffer.resize(65536);
 
-    // Parameters. TODO(#1529): Better parameter "library".
-    declare_parameter<int>("port", SharedVisionPortSinglePrimary);
-    get_parameter<int>("port", port_);
-    setPort(port_);
+    setPort(PARAM_port);
```
JNeiger, line 28: So this is set up in the "launch" files right? What would be the process to have two different configs ala sim and IRL and being able to swap between them in a simple way?

JNeiger (commented):
Can you document more how the system works from a high level? The amount of indirection through templates/defines makes it hard to follow easily

rj_param_utils/src/ros2_param_provider.cpp
```diff
@@ -0,0 +1,77 @@
...
+    for (const auto& [param_name, param] : GetParamMap<type>()) { \
+        rcl_interfaces::msg::ParameterDescriptor descriptor;      \
+        descriptor.description = param->help();                   \
+        const type& val = node->declare_parameter(                \
+            param_name, param->default_value(), descriptor);      \
+        Update(param_name, val);                                  \
+    }
+
+    // Declare the parameters
+    DECLARE_AND_UPDATE_PARAMS(bool)
```
JNeiger, line 43: These should be untabbed one

oswinso:
Yeah, sure. I'll probably dump it in a google doc somewhere so that we can easily copy paste it when we do the documentation-a-thon week.

rj_vision_receiver/src/vision_receiver.cpp
oswinso: Yeah. When the `ROS2ParamProvider` is instantiated, that connects the `ROS2` parameter system with the parameters defined here.

If you wanted to have 2 different configs, you would need to create 2 different `.yaml` files, and then have some logic in the `.launch.py` file that chooses between the two.

rj_param_utils/include/rj_param_utils/param.h
oswinso: Yeah, in that case it will conflict. I don't think it will conflict if they're in different translation units though.

Do you think this use case should be considered?

JNeiger: PID constants for different controllers is a good example. XY and Heading controllers both use a PID controller but with different gains

JNeiger: https://github.com/RoboJackets/robocup-software/blob/staging/soccer/config/sim.xml#L88
https://github.com/RoboJackets/robocup-software/blob/staging/soccer/config/sim.xml#L97

oswinso: I guess the easiest solution I can think of right now would be to put the `DEFINE_XXX` inside different namespaces on the client side.

JNeiger: Is it not possible to attach the prefix to all the stuff here? So like `PARAM_##prefix##__##name##_storage` or something like that. The prefix + name should be unique imo

oswinso: The problem with prefix is that ROS2 uses `.` as the separator for namespaces, ie. `a.b.c.example_param`, and `.` is an invalid character for C++ identifiers.

One way I could see this working would be to use `::` as the separator for the prefixes, ie.
`DEFINE_NAMESPACED_BOOL(a::b::c, thing, true, "description.")`, and then access it as `a::b::c::PARAM_thing` or something, and then do the conversion between `::` and `.` in `ROS2ParamProvider`.

rj_param_utils/include/rj_param_utils/param.h
oswinso: yeah this comment is pretty retarded and doesn't make sense.

rj_param_utils/include/rj_param_utils/param.h
oswinso: I think for simple nodes it does make sense to not need to have a namespace. ie. For the `port` parameter of `vision_receiver`

rj_param_utils/src/ros2_param_provider.cpp
oswinso: `clang-format` put these here so I'll leave them here.

rj_param_utils/include/rj_param_utils/param.h
JNeiger: I would argue no namespaces is the exception to the rule. Unless a bunch is going to change, there aren't any non-namespace configs currently.

Unless I'm missing something, port as a parameter wouldn't be descriptive enough by itself in the context of the whole system. We would need to say vision_reciever_port or something, which is just a namespace tied into the name. I feel like I'm missing something though if that's the example.

oswinso: In ROS2, parameters are tied to the node, ie. the hierarchy is like
```
/vision_receiver
  - port
```

and so `port` would be in the "root namespace" under the `/vision_reciever` node.

JNeiger: That makes a lot more sense

rj_param_utils/include/rj_param_utils/param.h
oswinso: Fixed.

rj_param_utils/include/rj_param_utils/param.h
oswinso: Done.

JNeiger (approved):
👍

rj_param_utils/.clang-tidy
```diff
@@ -0,0 +1,30 @@
...
+                  -fuchsia-default-arguments,
+                  -readability-redundant-member-init,
+                  -readability-magic-numbers,
+                  -readability-uppercase-literal-suffix,
+                  -misc-non-private-member-variables-in-classes,
+                  -cppcoreguidelines-non-private-member-variables-in-classes,
+                  -clang-analyzer-cplusplus.Move,
+                  -cppcoreguidelines-macro-usage,
+                  -bugprone-macro-parentheses
+                  '
```
kylestach, line 28: What is left out of this list?

rj_param_utils/include/rj_param_utils/param.h
```diff
@@ -0,0 +1,233 @@
+/** @file */
+#pragma once
+
+#include <rclcpp/node.hpp>
+
+namespace params {
+namespace internal {
```
kylestach, line 7: nit: standard is `detail`, no?

kylestach (approved):
LGTM, at least for the time being.

rj_param_utils/.clang-tidy
oswinso: More like `-cppcoreguidelines-macro-usage` and `-bugprone-macro-parentheses` were added.

rj_param_utils/include/rj_param_utils/param.h
oswinso: what do you mean by standard is `detail`?

PR #1563, opened 2020-08-03
Use ROS converters everywhere

### LINKED ISSUE #1470: Line equality operator is incorrect
## Description
Line equality compares the arrays like `arr1 == arr2`. This decays to comparing pointer which results in false in almost all cases unless you are comparing the exact same line to itself. You must compare the values of the two arrays or use std::array

## Steps to reproduce
1. Uncomment the line in the `Line` unit test in the `==` overload test.
2. Run `make test-cpp`

Result: Fails

- [x] This bug has been reproduced by someone else

## Additional information

### LINKED ISSUE #1560: Create uniform ROS conversion functions
## Description
We should have a uniform way of converting to/from ROS types that is consistent across all datatypes with 1:1 conversions.

## Requirements

 - Should provide a simple and consistent way of defining conversions between C++ and ROS datatypes

## Design

We will create a `RosConverter` struct, templated on `CppType` and `RosType`, with static `to_ros` and `from_ros` methods. Using a struct rather than a templated function lets us do partial specialization, which lets us do useful things like defining a converter for `std::vector<CppType>` to and from `std::vector<RosType>` for arbitrary types.

kylestach:
## Description
Implement ROS converters for most of our regular message types, and remove the implicit conversions.

In the process, accidentally fixes #1470.

## Associated Issue
Closes #1560.

## Design Documents
See #1560

## Steps to test
No functional changes, covered by unit tests.


JNeiger (approved):

rj_common/include/rj_common/Field_Dimensions.hpp
```diff
@@ -276,3 +232,38 @@ struct Field_Dimensions {
...
+        to.goal_depth = from.GoalDepth();
+        to.goal_height = from.GoalHeight();
+        to.penalty_short_dist = from.PenaltyShortDist();
+        to.penalty_long_dist = from.PenaltyLongDist();
+        to.center_radius = from.CenterRadius();
+        to.center_diameter = from.CenterDiameter();
+        to.goal_flat = from.GoalFlat();
+        to.floor_length = from.FloorLength();
+        to.floor_width = from.FloorWidth();
+        return to;
```
oswinso: I'd prefer if you used the builder, just so that if we change message definitions there's a compiler error and we don't run into bugs where we forget to convert a field

rj_msgs/msg/TeamInfo.msg
```diff
@@ -1,10 +1,10 @@
 string name
 int32 score
-int32 num_red_cards
-int32 num_yellow_cards
+uint64 num_red_cards
+uint64 num_yellow_cards
 builtin_interfaces/Duration[] yellow_card_remaining_times
 
-int32 timeouts_left
+uint64 timeouts_left
```
oswinso, line 7: Why these changes?

rj_vision_receiver/src/vision_receiver.cpp
```diff
@@ -314,7 +317,9 @@ void VisionReceiver::UpdateGeometryPacket(
             (fieldSize.field_length() / 1000.0f + (fieldBorder)*2),
             (fieldSize.field_width() / 1000.0f + (fieldBorder)*2)};
 
-        config_.updateFieldDimensions(new_field_dim);
+        config_.updateFieldDimensions(
+            rj_convert::convert_to_ros<Field_Dimensions, Field_Dimensions::Msg>(
+                new_field_dim));
```
oswinso: Looking at it now, this is pretty verbose... would it be possible to infer the second argument?

Geometry2d/include/Geometry2d/Line.hpp
```diff
@@ -77,23 +72,13 @@ class Line {
...
-    }
-
     /**
      * Returns the point on the line closest to p.
      */
     Point nearestPoint(Point p) const;
 
     /** the line consists of two points */
-    Point pt[2];
+    std::array<Point, 2> pt;
```
jonathanlew, line 81: consider `std::pair` to avoid the complexity of arrays, e.g. out of bounds errors

soccer/src/soccer/TeamInfo.hpp
```diff
@@ -4,33 +4,36 @@
...
 class TeamInfo {
 public:
+    using Msg = rj_msgs::msg::TeamInfo;
+
     // The team's name (empty string if operator has not typed anything).
-    std::string name = "";
+    std::string name;
     // The number of goals scored by the team during normal play and overtime.
-    uint score = 0;
+    int score = 0;
```
jonathanlew, line 19: hm, can score be negative?

jonathanlew (commented):
Hey, Jonathan from Thunderbots dropping by with a review

rj_msgs/msg/TeamInfo.msg
kylestach: Consistency with the C++ structure

Geometry2d/include/Geometry2d/Line.hpp
kylestach: This is not intended to do any significant refactoring on `Line`; we're just bringing it into this decade (real reason is I didn't want to do a converter templatized over regular fixed-length arrays).

soccer/src/soccer/TeamInfo.hpp
kylestach: Not In Real Life (:tm:), but I've found that unsigned integers are a pretty poor way to express the constraint of "nonnegative integer" (at least, not without overflow/underflow checking). I usually use them for memory/size-related variables and otherwise avoid using them when there's a choice.

The reason is that `uint` _implicitly_ captures the constraint while giving the illusion that it's capturing it _explicitly_. If there's any chance I'd subtract one thing from another - something like:
```cpp
const auto goal_differential = our_info.score - their_info.score;
if (goal_differential > SOME_POSITIVE_NUMBER) { /* be conservative */ }
if (goal_differential < SOME_NEGATIVE_NUMBER) { /* be aggressive */ }
```
then it's just easier to save the headache down the line and make it a regular int, assuming we don't need the range (and I'm pretty sure mercy rule kicks in before one team scores 2 billion goals :wink: )

---

TL;DR: I am a fool; I've fallen prey to one of the classic blunders!
```cpp
for (size_t i = N; i >= 0; i--) { ... }
```

rj_vision_receiver/src/vision_receiver.cpp
kylestach: I could probably make an association struct that defaults to `::Msg` pretty easily, let me see. (You'd still have to provide one template argument for the `from_ros` case, though).

Geometry2d/include/Geometry2d/Line.hpp
kylestach: This is probably a good suggestion for a future cleanup effort, though.

oswinso (approved):
lgtm

Issue #1483, opened 2020-06-11
Sort and group includes with clang-format


JNeiger:
## Description
Clang-format can automatically sort includes and group them based on regex and priority. 

```
File.cpp

#include "file.hpp"

#include <stdlib>
...

#include <external libraries> // eg Eigen or boost
...

#include <project headers from other modules .hpp>
...

#include "project headers from this modules .hpp"
...
```

Can be done using the following in the clang-format config file

```
SortIncludes: true
IncludeBlocks: Regroup
IncludeCategories: 
  # Librarys
  - Regex:           '<(Eigen|behaviortree_cpp_v3|rapidxml)\/([A-Za-z0-9.\/-_])+>'
    Priority:        2
  # STD LIB
  - Regex:           '<([A-Za-z0-9\/-_])+>'
    Priority:        1
  # <*.hpp>
  - Regex:           '<([A-Za-z0-9.\/-_])+>'
    Priority:        3
  # "*.hpp"
  - Regex:           '"([A-Za-z0-9.\/-_])+"'
    Priority:        4
```
Where the external libraries are specified in the clang regex. Anything not hit by that regex and doesn't have a .hpp gets pushed into the stdlibs. Any <*.hpp> that haven't been associated with those two are caught by the third group and are assumed to be internal project configs. The last one catches everything else in quotes.

## Requirements

 - Sorts includes correctly

## Design

Feel free to close this is the current Google includes settings are good enough

PR #1567, opened 2020-08-05
All style things. Part 1 - Config files

oswinso:
## Description
This PR sets up:
1. Naming conventions in `.clang-tidy`
2. Include order stuff in `.clang-format`
3. cmake styling stuff using `cmake-format` tool in `.cmake-format.py`

Note: I changed the symlink for `.clang-format` to [style-configs](https://github.com/RoboJackets/style-configs) to just the actual file.

Changes based on these config files can be seen [here](https://github.com/RoboJackets/robocup-software/pull/1568/files)

## Associated Issue
Issue #1483

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -30,23 +30,26 @@ Checks:          '-*,
 AnalyzeTemporaryDtors: false
 # Use Google style guide.
 CheckOptions:
-# Classes / Structs / Enums are CamelCase.
+  # Classes / Structs / Enums are CamelCase.
   - { key: readability-identifier-naming.ClassCase,               value: CamelCase  }
   - { key: readability-identifier-naming.StructCase,              value: CamelCase  }
   - { key: readability-identifier-naming.EnumCase,                value: CamelCase  }
-# Variables / Parameters are lower_case.
+  # Variables / Parameters are lower_case.
   - { key: readability-identifier-naming.VariableCase,            value: lower_case }
   - { key: readability-identifier-naming.ParameterCase,           value: lower_case }
   - { key: readability-identifier-naming.PublicMemberCase,        value: lower_case }
-# Constexpr / const things are kCamelCase.
+  # Private members are lower_case_
+  - { key: readability-identifier-naming.PrivateMemberCase,       value: lower_case }
+  - { key: readability-identifier-naming.PrivateMemberSuffix,     value: "_" }
+  # Constexpr / const things are kCamelCase.
   - { key: readability-identifier-naming.ConstexprVariablePrefix, value: "k" }
   - { key: readability-identifier-naming.ConstexprVariableCase,   value: CamelCase }
-  - { key: readability-identifier-naming.ClassConstantPrefix, value: "k" }
-  - { key: readability-identifier-naming.ClassConstantCase,   value: CamelCase }
-  - { key: readability-identifier-naming.GlobalConstantPrefix, value: "k" }
-  - { key: readability-identifier-naming.GlobalConstantCase,   value: CamelCase }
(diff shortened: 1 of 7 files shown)


JNeiger (approved):
No idea about the cmake stuff, everything else looks good though

.clang-tidy
```diff
@@ -28,4 +28,25 @@ Checks:          '-*,
...
+  - { key: readability-identifier-naming.PublicMemberCase,        value: lower_case }
+# Constexpr / const things are kCamelCase.
+  - { key: readability-identifier-naming.ConstexprVariablePrefix, value: "k" }
+  - { key: readability-identifier-naming.ConstexprVariableCase,   value: CamelCase }
+  - { key: readability-identifier-naming.ClassConstantPrefix, value: "k" }
+  - { key: readability-identifier-naming.ClassConstantCase,   value: CamelCase }
+  - { key: readability-identifier-naming.GlobalConstantPrefix, value: "k" }
+  - { key: readability-identifier-naming.GlobalConstantCase,   value: CamelCase }
+# Functions are CamelCase.
+  - { key: readability-identifier-naming.FunctionCase,            value: CamelBack  }
```
kylestach: Ideally either `CamelCase` or `lower_case` rather than `CamelBack`, right? I think I personally have a _slight_ preference towards `lower_case` but we should probably come to consensus.

.cmake-format.py
```diff
@@ -0,0 +1,221 @@
+# ----------------------------------
+# Options affecting listfile parsing
+# ----------------------------------
+with section("parse"):
+    # Specify structure for custom cmake functions
+    additional_commands = {'foo': {'flags': ['BAR', 'BAZ'],
```
kylestach: lol?

kylestach (approved):
Approving pending the `camelBack` vs. `CamelCase` vs. `lower_case` for function names.

.clang-tidy
```diff
@@ -28,4 +28,25 @@ Checks:          '-*,
...
+# Use Google style guide.
+CheckOptions:
+# Classes / Structs / Enums are CamelCase.
+  - { key: readability-identifier-naming.ClassCase,               value: CamelCase  }
+  - { key: readability-identifier-naming.StructCase,              value: CamelCase  }
+  - { key: readability-identifier-naming.EnumCase,                value: CamelCase  }
+# Variables / Parameters are lower_case.
+  - { key: readability-identifier-naming.VariableCase,            value: lower_case }
+  - { key: readability-identifier-naming.ParameterCase,           value: lower_case }
+  - { key: readability-identifier-naming.PublicMemberCase,        value: lower_case }
```
kylestach, line 40: Forgot to mention, but we need a `PrivateMemberCase` and `PrivateMemberPrefix` (or suffix if that's what we decide to use).

.clang-tidy
oswinso: lol the comment says CamelCase

.cmake-format.py
oswinso: I literally did `dump-config` and this is what it output.

.clang-tidy
oswinso: ah that's true.

.clang-tidy
oswinso: If we're following google then I think it makes sense to go google all the way and do `CamelCase`.

.cmake-format.py
oswinso: Fixed.

.clang-tidy
oswinso: Fixed.

PR #1568, opened 2020-08-05
All style things. Part 2 - Config files 

oswinso:
## Description
This PR:
- Creates scripts to run `clang-tidy`, `clang-format` and `cmake-format` on the entire codebase.
- Runs `./util/reformat_all.sh`


soccer/src/soccer/gameplay/robocup-py.cpp
```diff
@@ -43,6 +44,8 @@ using namespace boost::python;
 #include "planning/MotionConstraints.hpp"
 #include "referee/ExternalReferee.hpp"
 
+#include <rc-fshare/pid.hpp>
```
JNeiger, line 47: The includes here look a little weird. It might have problems with the using. Also `rj-fshare` doesn't fit the regex, it needs a `-` also in the `([A-Za-z0-9.\/-_])+`, eg `([A-Za-z0-9.\/-_-])+`

JNeiger (approved):
Minus the one thing, everything else looks good. (I only checked a file here and there, not every one)

oswinso:
Lol this was just supposed to show what the config files look like and not an actual PR

PR #1569, opened 2020-08-07
Feat/format everything

### LINKED ISSUE #1483: Sort and group includes with clang-format

oswinso:
## Description
This PR fixes a few last config issues, ie. adding `builtin_interfaces` to the 3rd party libraries section of `.clang-format` and specifying a suffix `_` for protected members as well. `clang-tidy` was not run as it resulted in a lot of broken code that was nontrivial to fix.

The changes in the last commit can be reproduced by commenting out the `run-clang-tidy` part in `reformat_all.sh` and running the file.

## Associated Issue
Fixes #1483.

### Everything still works.
1. run sim

Expected result: Sim works.


JNeiger (approved):
Looks reasonable, lgtm

PR #1570, opened 2020-08-08
Fix naming on everything

kylestach:
## Description
Standardize naming across the C++ codebase.

This makes all variables `snake_case` (`snake_case_trailing_` for members), classes `CamelCase`, etc. I think I got everything except for ui-related files, because those are kind of a mess (Qt stuff is backCamelCase and it's hard to keep straight what's what.

## Associated Issue
N/A

## Design Documents
N/A

### CI Testing

Expected result: Everything passes except for UI

### Sim Testing
Nothing breaks when running in sim, especially with respect to python code (tested locally).


oswinso (approved):
lgtm

JNeiger (approved):
Glanced through, looks reasonable

Issue #1566, opened 2020-08-03
ROS2 independent logging system.


oswinso:
# Description
We need a ROS2 independent logging system so that we can unify our logging between things that don't use any ROS2 (currently prints using `std::cout`) and things that do use ROS2 (`EZ_*` and `RJ_*`).

## Requirements
- Allow for swappable backends (ie. using ROS)

## Design

todo

PR #1573, opened 2020-08-17
spdlog ros2 logging sink + set default logger

oswinso:
## Description
This PR creates a `spdlog` sink that dumps stuff into the `rclcpp` logging functions and replaces all logging calls with the `spdlog` variants.

Note: I also changed the `DIFFBASE` to `ros2` so that when we merge into `ros2` it doesn't take forever to lint all the changes.

## Associated Issue
Related to #1566.

## Design Documents
[Link](link-to-design-doc)

### Everything still works.
1. Test sim.


soccer/src/soccer/radio/network_radio.cpp
```diff
@@ -1,9 +1,12 @@
 #include "network_radio.hpp"
 
+#include <fmt/ostream.h>
```
JNeiger, line 3: I'm not sure fmt is being used in this file

soccer/src/soccer/optimization/python_function_wrapper.cpp
```diff
@@ -1,20 +1,21 @@
 #include "python_function_wrapper.hpp"
 
-#include <iostream>
+#include <fmt/ostream.h>
```
JNeiger, line 3: Is fmt needed as an include?

soccer/src/soccer/ui/strip_chart.cpp
```diff
@@ -4,16 +4,17 @@
 #include <cmath>
 #include <cstdio>
 #include <fstream>
-#include <iostream>
 
 #include <QDateTime>
 #include <QFileDialog>
 #include <QPainter>
+#include <fmt/ostream.h>
```
JNeiger, line 11: fmt here as well

soccer/CMakeLists.txt
```diff
@@ -86,6 +86,9 @@ if(NOT FOUND_A_BOOST_PYTHON)
   message(FATAL_ERROR "Unable to find a suitable version of boost python")
 endif()
 
+find_package(fmt REQUIRED)
+find_package(spdlog REQUIRED)
```
JNeiger, line 90: sdplog is never actually pulled in using submodules/static libs

soccer/src/soccer/radio/network_radio.cpp
oswinso: So `spdlog` uses `fmt` internally (which is where the `{}` syntax comes from). To allow `{}` to print the same representation as `operator<<`, we need to include `fmt/ostream.h`.

soccer/src/soccer/optimization/python_function_wrapper.cpp
oswinso: Same as above.

soccer/CMakeLists.txt
oswinso: `ros2` uses it internally so we're just freeloading off of that.

JNeiger (approved):

.circleci/config.yml
```diff
@@ -102,7 +102,7 @@ jobs:
       - run: |
           PR_NUM=$(echo "$CIRCLE_PULL_REQUEST" | grep -o "[0-9]*")
           if [ -z "$PR_NUM" ]; then
-            DIFFBRANCH="origin/staging"
+            DIFFBRANCH="ros2"
```
kylestach, line 105: I'm not sure that this is the right default behavior here. Should we even be running tidy/pretty-diff on regular branch builds?

rj_utils/src/logging.cpp
```diff
@@ -0,0 +1,61 @@
...
+            break;
+    }
+}
+
+void set_spdlog_default_ros2(const std::string& logger_name) {
+    auto ros2_sink = std::make_shared<Ros2Sink<spdlog::details::null_mutex>>();
+    auto ros2_logger = std::make_shared<spdlog::logger>(logger_name, std::move(ros2_sink));
+
+    // Also include the source filename and line number, but remove all the other level / timestamp
+    // information as that is provided by ros2.
```
kylestach, line 57: Why?

soccer/src/rj_vision_filter/ball/world_ball.cpp
```diff
@@ -22,11 +23,12 @@ WorldBall::WorldBall(RJ::Time calc_time, const std::list<KalmanBall>& kalman_bal
...
     // Above 2 would just be super noisy
     if (PARAM_ball_merger_power < 1 || PARAM_ball_merger_power > 2) {
-        std::cout << "WARN: ball_merger_power should be between 1 and 2" << std::endl;
+        SPDLOG_WARN("ball_merger_power should be between 1 and 2");
     }
 
     if (kalman_balls.empty()) {
-        throw std::runtime_error("ERROR: Zero balls are given to the WorldBall constructor");
+        spdlog::critical("Zero balls are given to the WorldBall constructor");
+        throw std::runtime_error("Zero balls are given to the WorldBall constructor");
```
kylestach: Can we replace this with the debug throw from utils?

kylestach: Also, why use `spdlog::critical` here but the macro versions elsewhere?

soccer/include/rj_vision_filter/vision_filter.hpp
```diff
@@ -139,14 +138,13 @@ class VisionFilter : public rclcpp::Node {
...
-        detection_frame_queue_;
+    rj_utils::ConcurrentQueue<DetectionFrameMsg::UniquePtr> detection_frame_queue_;
 
     /**
      * @brief Publisher for WorldStateMsg.
      */
     rclcpp::Publisher<WorldStateMsg>::SharedPtr world_state_pub_;
 
-    params::ROS2ParamProvider param_provider_;
+    ::params::ROS2ParamProvider param_provider_;
```
kylestach, line 148: Should this just be renamed to `rj_params`?

soccer/src/soccer/joystick/sdl_gamepad.cpp
```diff
@@ -1,15 +1,16 @@
...
 #include <joystick/sdl_gamepad.hpp>
 
 namespace joystick {
 SDLGamepad::SDLGamepad(int device_index) {
     controller_ = SDL_GameControllerOpen(device_index);
 
     if (controller_ == nullptr) {
-        std::cerr << "ERROR: Could not open controller! SDL Error: " << SDL_GetError() << std::endl;
+        spdlog::critical("ERROR: Could not open controller! SDL Error: {}", SDL_GetError());
         throw std::runtime_error("Failed to open controller!");
```
kylestach: Debug throw function.

soccer/src/soccer/joystick/sdl_joystick_node.cpp
```diff
@@ -1,16 +1,14 @@
...
 namespace joystick {
 SDLJoystickNode::SDLJoystickNode(Context* context) : context_{context} {
     // initialize using the SDL joystick
     if (SDL_Init(SDL_INIT_GAMECONTROLLER | SDL_INIT_EVENTS) != 0) {
-        std::cerr << "ERROR: SDL could not initialize game controller system! SDL "
-                     "Error: "
-                  << SDL_GetError() << std::endl;
+        SPDLOG_ERROR("SDL could not initialize game controller system! SDL Error: {}",
+                     SDL_GetError());
         throw std::runtime_error("");
```
kylestach: Debug throw? Also, lmao empty exception string.

soccer/src/soccer/logger.cpp
```diff
@@ -122,8 +123,7 @@ void Logger::read(const std::string& filename) {
     context_->logs.frames.clear();
 
     if (!log_file_->good()) {
-        std::cerr << "Log file " << filename << " does not exist." << std::endl;
-        std::exit(-1);
+        spdlog::critical("Log file {} does not exist.", filename);
```
kylestach: Should still exit, or use debug throw

rj_common/include/rj_common/utils.hpp
```diff
@@ -24,18 +23,15 @@ inline void debug_log_if(const std::string& e, bool condition) {
...
     debug_log(e);
     if (kThrowDebugExceptions) {
         throw e;
     }
 }
 
-inline void debug_throw(const std::string& string) {
-    debug_throw(std::runtime_error(string));
-}
+inline void debug_throw(const std::string& string) { debug_throw(std::runtime_error(string)); }
```
kylestach: Nit: Shouldn't these be in `rj_utils/logging`?

kylestach (approved):
I think there are a few cases where the pattern:
```cpp
spdlog::critical(...);
throw some_exception(...);
```
could be replaced with
```cpp
debug_throw(some_exception(...));
```
Other than that, LGTM.

rj_utils/src/logging.cpp
oswinso: What do you mean by "why"? When you log using ROs, it already shows the log level + timestamp, unless you think it would be better for those pieces of information should be repeated? ie.
> [INFO] [1598296733.836129931] [processor]: [2020-08-24 15:18:53.836] [processor] [info] [processor.cpp:157] Updating field geometry based off of vision packet.

kylestach: Oh god yeah nvm lmfao

oswinso:
I don't think the semantics of `debug_throw` match up with what we want to happen here, since `debug_throw` conveys that we only want to throw if we are debugging. I've created `fatal_throw` and `FATAL_THROW` just so that it's more clear that it's a fatal error that needs to be looked at.

rj_common/include/rj_common/utils.hpp
oswinso: Done.

soccer/src/soccer/logger.cpp
oswinso: Changed to `fatal_throw`.

soccer/src/soccer/joystick/sdl_joystick_node.cpp
oswinso: Changed to `fatal_throw`.

soccer/src/soccer/joystick/sdl_gamepad.cpp
oswinso: Changed to `fatal_throw`.

soccer/include/rj_vision_filter/vision_filter.hpp
oswinso: lol rip I guess this is for another PR then.

soccer/src/rj_vision_filter/ball/world_ball.cpp
oswinso: Changed to `fatal_throw`.

kylestach:
Fine by me, although IMO debug_throw should really unconditionally throw. You very rarely _actually_ want to avoid throwing in release mode - that usually just leads to ignoring bugs because "they don't show up in release".

PR #1577, opened 2020-09-01
STP - rewrite of gameplay library

oswinso:
## Description
This PR contains the beginnings of the rewrite of the current gameplay library, based off of the design documents written by @HussainGynai [here](https://docs.google.com/document/d/1pf-aaHnvonTPC6ccI7RdHudpuev84fZslMVDq8lw59s/edit).

This PR contains two things:
- `stp` - A library for the STP(A) system.
- `rj_gameplay` - The actual application level code that uses the `stp` library.

In addition, this PR also adds a few things:
- Tests for `stp` using pytest
- Custom `pylint` linter to check that we don't do `from ... import ...` in any of our code as it breaks hot module reloading.
- Coverage using codecov


rj_gameplay/stp/play/__init__.py
```diff
@@ -0,0 +1,140 @@
...
+
+    def __str__(self) -> str:
+        return "{:3}: {} - {}".format(self._idx, self.concrete_cls.__name__, self.skill)
+
+    def __repr__(self) -> str:
+        return self.__str__()
+
+
+class TacticsEnumMeta(type):
+    """ Metaclass for
```
kylestach: ???

rj_gameplay/stp/play/__init__.py
```diff
@@ -0,0 +1,140 @@
...
+
+
+TacticT = TypeVar("TacticT", bound=tactic.ITactic)
+
+
+class TacticEntry(edict.EKey[TacticT]):
+    """ An entry in the TacticEnum for a play.
+    """
+
+    INVALID_IDX = -1
```
kylestach: I'm not a fan of sentinels, is there any engineering reason this can't explicitly be `None`?

rj_gameplay/stp/game_state.py
```diff
@@ -0,0 +1,61 @@
...
+
+    def __init__(self, pos: np.ndarray, vel: np.ndarray):
+        self.pos = pos
+        self.vel = vel
+
+    def __repr__(self) -> str:
+        return "Ball(pos:{}, vel:{})".format(self.pos, self.vel)
+
+
+class GameState:
```
kylestach: Nit: this is called `WorldState` C++-side.

rj_gameplay/stp/play/offensive_triangle.py
```diff
@@ -0,0 +1,20 @@
...
+    DEF_SUPPORT = play.TacticEntry(def_support.DefSupport)
+
+
+class OffensiveTriangle(pure_play.PurePlay):
+    """ A basic offensive play. The three offensive robots form an attack triangle while
+    the other two robots support the ball carrier.
+    """
+
+    def __init__(self, tactics_factory: tactic.Factory):
+        super().__init__(Tactics(tactics_factory))
```
kylestach: Clean, damn.

rj_gameplay/stp/play/pure_play.py
```diff
@@ -0,0 +1,43 @@
...
+    """
+
+    __slots__ = ["tactics"]
+
+    def __init__(self, tactics: play.TacticsEnum):
+        self.tactics = tactics
+
+    def tick(self) -> List[skill.ISkill]:
+        role_requests = self.collect_role_requests()
+
```
kylestach, line 37: `TODO` here for actually ticking the tactics and returning their actions?

rj_gameplay/stp/role/__init__.py
```diff
@@ -0,0 +1,125 @@
...
+
+class Priority(IntEnum):
+    LOW = 0
+    MEDIUM = 1
+    HIGH = 2
+    NUM_PRIORITIES = 3
+
+
+# (robot: Robot, prev_robot: Robot) -> cost.
+CostFn = Callable[[game_state.Robot, Optional[game_state.Robot]], float]
```
kylestach: `CostFn` should return a tuple of `(float, bool)` so we can express hard constraints.

kylestach: Also, should probably accept a `WorldState`.

rj_gameplay/stp/role/cost.py
```diff
@@ -0,0 +1,72 @@
...
+    """
+
+    def cost_fn(robot: game_state.Robot, prev_robot: Optional[game_state.Robot]):
+        dist: float = np.linalg.norm(robot.pose[:2] - pt)
+        cost: float = dist / saturate_dist
+
+        # Add the cost for switching robots.
+        cost += flat_switch_cost(robot, prev_robot, switch_cost)
+
+        return max(min(cost, 1.0), 0.0)
```
kylestach: Are we intending to limit all of our costs to the (0, 1) range? If so, this should happen in a "soft" way (i.e. by tweaking constants so it happens like that in normal operating conditions) rather than by hard-limiting. This just loses signal and makes it more confusing to combine things later.

rj_gameplay/stp/role/assignment/naive.py
```diff
@@ -0,0 +1,164 @@
...
+        return sorted_requests
+
+    @staticmethod
+    def compute_costs_matrix(
+        free_robots: np.ndarray,
+        flat_requests: assignment.FlatRoleRequests,
+        game_state: stp.game_state.GameState,
+        prev_results: assignment.FlatRoleResults,
+    ) -> np.ndarray:
+        """Computes the m x n cost matrix corresponding to the passed in free robots and
```
kylestach: What are `m` and `n`? (I know from reading the code, but put it in the docstring)

rj_gameplay/stp/role/assignment/naive.py
```diff
@@ -0,0 +1,164 @@
...
+        # Iterate over each role request.
+        for request_idx, (role_id, request) in enumerate(flat_requests.items()):
+            role_id: assignment.RoleId
+            request: role.RoleRequest
+
+            # For each role request, iterate over robots.
+            for robot_idx, robot in enumerate(free_robots):
+                # Otherwise, record the cost.
+                cost: float = request.cost_fn(robot, prev_results.get(role_id, None))
+                if not isfinite(cost):
```
kylestach, line 81: I think we were planning to express hard constraints as a separate `bool` return value. Infinities are annoying because inevitably someone will try to combine two costs and forget to deal with that case, making it explicit makes it impossible to express.

If there is a `NaN` in the cost function, role assignment should be noisy and throw an exception.

That said, is a `0` the right thing to do in this case? Because that's what you're currently putting into the cost matrix, which seems _really_ wrong.

rj_gameplay/stp/skill/__init__.py
```diff
@@ -0,0 +1,77 @@
...
+
+    @abstractmethod
+    def create_request(self) -> role.RoleRequest:
+        ...
+
+
+SkillT = TypeVar("SkillT", bound=ISkill)
+
+
+class Registry(MutableMapping):
```
kylestach: `SkillRegistry`?

rj_gameplay/stp/skill/__init__.py
```diff
@@ -0,0 +1,77 @@
...
+        return self._dict.__len__()
+
+    def __iter__(self):
+        return self._dict.__iter__()
+
+    def __contains__(self, item) -> bool:
+        return self._dict.__contains__(item)
+
+
+class Factory:
```
kylestach, line 69: `SkillFactory`?

rj_gameplay/stp/skill/ball_carrier.py
```diff
@@ -0,0 +1,37 @@
...
+
+    def __init__(self, passer: int, receiver: int, pt: np.ndarray):
+        self.passer = passer
+        self.receiver = receiver
+        self.pt = pt
+
+
+class IBallCarrier(skill.ISkill, ABC):
+    @abstractmethod
+    def get_pass(self) -> Optional[Pass]:
```
kylestach: I'm pretty sure this belongs at the tactic level. The functionality exposed should be `do_pass(self, pass: Pass)` and `do_shoot(self, shot: Shot)`.

rj_gameplay/stp/skill/ball_carrier.py
```diff
@@ -0,0 +1,37 @@
...
+    __slots__ = ["passer", "receiver", "pt"]
+
+    passer: int
+    receiver: int
+    pt: np.ndarray
+
+    def __init__(self, passer: int, receiver: int, pt: np.ndarray):
+        self.passer = passer
+        self.receiver = receiver
+        self.pt = pt
```
kylestach: This should be somewhere else. Also, please specify `pt`'s dimensions.

Also, pass should contain `velocity` and `pass_start_time`.

rj_gameplay/stp/skill/ball_carrier.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+class Pass:
+    __slots__ = ["passer", "receiver", "pt"]
+
+    passer: int
+    receiver: int
+    pt: np.ndarray
+
+    def __init__(self, passer: int, receiver: int, pt: np.ndarray):
```
kylestach: Can we make a `RobotId` alias for `int`? Been meaning to do that C++-side.

rj_gameplay/stp/utils/edict.py
```diff
@@ -0,0 +1,65 @@
...
+    __slots__ = ["concrete_cls"]
+
+    def __init__(self, cls: Type[ValueConcreteT]):
+        self.concrete_cls = cls
+
+    def value_t(self) -> Type[ValueConcreteT]:
+        return self.concrete_cls
+
+
+class EDict(Generic[ValueInterfaceT], MutableMapping):
```
kylestach: Honestly I have no idea what the `E` in the name here is intending to convey.

rj_gameplay/stp/tactic/__init__.py
```diff
@@ -0,0 +1,182 @@
...
+
+
+class SkillsDict(edict.EDict[List[skill.ISkill]]):
+    ...
+
+
+class ITactic(ABC):
+    @abstractmethod
+    def get_requests(self, prev_skills: SkillsDict) -> RoleRequests:
+        ...
```
kylestach, line 142: `TODO` for a `tick` method?

rj_gameplay/stp/tactic/def_support.py
```diff
@@ -0,0 +1,25 @@
...
+    """
+
+    __slots__ = ["skills", "STUB"]
+
+    def __init__(self, ctx: tactic.Ctx):
+        self.skills = Skills(ctx.skill_factory)
+
+        self.STUB = self.skills.STUB
+
+    def define(self, prev_skills: tactic.SkillsDict) -> tactic.RoleRequests:
```
kylestach: Wrong name?

rj_gameplay/stp/tactic/pass_or_shoot.py
```diff
@@ -0,0 +1,48 @@
...
+
+    def get_requests(self, prev_skills: tactic.SkillsDict) -> tactic.RoleRequests:
+        role_requests: tactic.RoleRequests = tactic.RoleRequests()
+
+        maybe_pass = self.get_pass(prev_skills)
+        if maybe_pass:
+            role_requests[self.RECEIVER] = self.RECEIVER.skill.create_request()
+
+        role_requests[self.BALL_CARRIER] = self.BALL_CARRIER.skill.create_request()
+        role_requests[self.SEEKERS] = self.SEEKERS.skill.create_request()
```
kylestach: How do we support multiple (an arbitrary number of) seekers in this paradigm?

kylestach (changes requested):
First pass. Looks mostly pretty clean, although I'm not sure how the dict-of-skills paradigm you've chosen works out:
 - What if you want arbitrarily many (i.e. seeker) skills?
 - How will the play keep track of which tactic gets which `prev_skills` (and, when `tick()` is implemented, which tactic gets which `assigned_skills`)?

rj_gameplay/stp/utils/edict.py
oswinso: I kind of needed a name for this but I couldn't think of anything good.

rj_gameplay/stp/role/cost.py
oswinso: I think it would be nice to limit all of our costs to [0, 1], just so it's easier to compare costs or something. I'll need to think about it more about what the best way of doing this is.

rj_gameplay/stp/role/__init__.py
oswinso: Yeah true.

rj_gameplay/stp/play/__init__.py
oswinso: Sure.

rj_gameplay/stp/play/__init__.py
oswinso: lmao nice.

rj_gameplay/stp/skill/__init__.py
oswinso: I was thinking about this, and because we're only doing `import .. as ...` calls, it's basically the difference between doing `skill.Registry` vs `skill.SkillRegistry`. I thought that the former was nice, but idk.

rj_gameplay/stp/role/assignment/naive.py
oswinso: The `isfinite` here was mainly to handle bad cost functions, ie trying to divide by 0 or something, rather than hard constraints. This still needs to be worked on, but I just thought that it would be nice if role assignment still worked, but it gets blacklisted + a log message or something.

rj_gameplay/stp/tactic/pass_or_shoot.py
oswinso: This is still WIP but I was thinking that the `RoleRequest` has a way of specifying the min / max number of robots it is accepting or something. So the `RoleRequest` that `self.SEEKERS.skill.create_request` returns will allow for accepting multiple robots.

rj_gameplay/stp/utils/edict.py
kylestach: Naming things is hard...maybe `TypedKeyDict`? Still a little bit weird but I think it gets the point across a little better.

rj_gameplay/stp/role/cost.py
kylestach: I personally really hate limiting costs - it tends to just mask over the weird edge cases. I would prefer to express all costs in (approximate) seconds or something, that tends to make it easier to reason about.

rj_gameplay/stp/role/assignment/naive.py
kylestach: Ok. Yeah honestly I think we should just throw for now.

rj_gameplay/stp/skill/__init__.py
kylestach: True...I am 90% sure someone will `from skill import Registry` at some point, though, and I'd rather not constantly police for that.

rj_gameplay/stp/tactic/pass_or_shoot.py
kylestach: Why not just create multiple role requests?

rj_gameplay/stp/role/__init__.py
HussainGynai: I thought cost functions were going to be different from the hard constraints

rj_gameplay/stp/role/cost.py
oswinso: What do you mean by (approximate) seconds?

rj_gameplay/stp/skill/__init__.py
oswinso: That breaks hot reload though... I think maybe I'll make a custom pylint checker that errors on any `from stp.* import .*` if it's not a test.

rj_gameplay/stp/role/cost.py
kylestach: I mean like we should scale them to a physically intuitive unit. Seconds was my first thought, because it's pretty easy to define and reason about.
My preferred definition of cost would be, roughly, "time until the robot is able to start actively participating in this role". For example, for a capture skill, this would equate to "time to capture the ball". In this context, I used "approximate" because obviously the cost function won't know _exactly_ how long it is.

rj_gameplay/stp/tactic/pass_or_shoot.py
oswinso: That could work as well. We would need to know how many robots there are though, so I guess we can pass world_state in here.

oswinso: That could work as well. We would need to know how many robots there are though, so I guess we can pass world_state in here.

rj_gameplay/stp/skill/__init__.py
kylestach: Oh shit, you right. Yeah that sounds like a better path.

rj_gameplay/stp/tactic/pass_or_shoot.py
kylestach: I think the original plan was that you just request as many roles as you can possibly use (but at lower priority), and then check if they're filled in the `tick` function, and then the roles are just stored as member variables. Is there a reason you moved away from this? (I don't love it given that it introduces additional statefullness, so if you have better ideas I'd love to hear them)

rj_gameplay/stp/skill/ball_carrier.py
oswinso: This isn't really meant to be an actual working play, but more as like something to think about how the library should look like in terms of ergonomics and stuff.

rj_gameplay/stp/skill/ball_carrier.py
oswinso: This isn't really meant to be an actual working play, but more as like something to think about how the library should look like in terms of ergonomics and stuff.

rj_gameplay/stp/utils/edict.py
oswinso: Sure.

rj_gameplay/stp/tactic/pass_or_shoot.py
oswinso: That's what I meant by the "we would need to know how many robots there are" part, as in like you need to know how many roles to make.

rj_gameplay/stp/play/pure_play.py
oswinso: Implemented.

rj_gameplay/stp/skill/ball_carrier.py
oswinso: Done

rj_gameplay/stp/tactic/pass_or_shoot.py
kylestach: Sure, the idea was just that we request more roles than will be fillable and you only fill some of them. But sure, the role request function can accept a world state.

oswinso: Yeah, but like how many roles do you actually request? 5? 10? 15? That's dependent on how many robots you actually have right?

rj_gameplay/stp/role/assignment/naive.py
oswinso: Done.

rj_gameplay/stp/skill/ball_carrier.py
oswinso: How would `do_pass` and `do_shoot` work? Why does it belong at the tactic level?

rj_gameplay/rj_gameplay/tactic/pass_or_shoot.py
```diff
@@ -0,0 +1,78 @@
...
+class Skills(tactic.SkillsEnum):
+    """Skills for PassOrShoot."""
+
+    BALL_CARRIER = tactic.SkillEntry(ball_carrier.IBallCarrier)
+    SEEKERS = tactic.SkillEntry(seeker.ISeeker)
+    RECEIVER = tactic.SkillEntry(capture.ICapture)
+
+
+@dataclass
+class Props:
```
kylestach: Move this inside `PassOrShoot`

rj_gameplay/rj_gameplay/tactic/pass_or_shoot.py
```diff
@@ -0,0 +1,78 @@
...
+    """Props (state) for PassOrShoot."""
+
+    maybe_pass: Optional[eval.Pass] = None
+
+
+class PassOrShoot(tactic.ITactic[Props]):
+    """Tactic that controls one ball carrier and multiple seekers."""
+
+    __slots__ = ["skills", "BALL_CARRIER", "RECEIVER", "SEEKERS"]
```
kylestach, line 35: Do we really need to have this everywhere? Kinda a premature optimization.

rj_gameplay/tests/stp/utils/test_edict.py
```diff
@@ -0,0 +1,169 @@
+from typing import Type
```
kylestach: This file should be renamed

rj_gameplay/stp/situation/analyzer.py
```diff
@@ -0,0 +1,67 @@
...
+        highest_score: float = max(situation_scores)
+        highest_score_idx: int = situation_scores.index(highest_score)
+
+        # Get the situation and play.
+        highest_score_situation = applicable_situations[highest_score_idx]
+        situation_play: play.IPlay = self._registry[highest_score_situation]
+
+        return highest_score_situation, situation_play
+
+    def _applicable_situations(
```
kylestach: We usually consider situation analysis as a function from WorldState -> Situation, not to [Situation]. I'm not opposed to changing it - I think this could be a more flexible model - but that's something that needs to be brought up to the team.

rj_gameplay/stp/role/cost.py
kylestach: Bump on this?

rj_gameplay/stp/rc.py
```diff
@@ -0,0 +1,64 @@
...
+WorldState"""
+
+from typing import List
+
+import numpy as np
+
+RobotId = int
+
+
+class Robot:
```
kylestach, line 12: We should make it explicit that this is the state of the robot and not a place to apply actions to the robot.

kylestach (changes requested):
Nits. Some more major things from the first review are still unresolved.

HussainGynai (approved):

rj_gameplay/rj_gameplay/tactic/pass_or_shoot.py
oswinso: I'm mainly using it as "documentation, in that it's useful to be able to tell all the instance attributes that a class has and have the IDE / interpreter enforce that for me.

rj_gameplay/stp/rc.py
oswinso: Done.

rj_gameplay/tests/stp/utils/test_edict.py
oswinso: Done.

rj_gameplay/stp/role/cost.py
oswinso: Done.

JNeiger:
Imma be honest, I don't have access to the meeting notes/most of the discussion around this; add in all the python framework, shorthand, and magic happening, there's no way for me to actually be able to do more than "yep this is python". And tbf, I'm not sure I'm sold enough on the rewrite to spend a bunch of time trying to reverse engineer dozens of meetings of discussion to figure it out enough to correctly do an in-depth review

I don't mind reviewing code that I've touched or some actual strategy stuff, but I think I'll have to skip most of the new internal framework things.

rj_gameplay/rj_gameplay/situation/decision_tree/analyzer.py
```diff
@@ -0,0 +1,197 @@
...
+    def __calc_ball_pos(
+        world_state: stp.rc.WorldState, game_info: stp.rc.GameInfo
+    ) -> BallPos:
+        """Computes the current ball possession.
+        :param world_state:
+        :param game_info:
+        :return: The current BallPos.
+        """
+
+        # TODO: Re-implement all the logic here.
```
kylestach: Fine for now, could you make an issue?

rj_gameplay/rj_gameplay/situation/decision_tree/analyzer.py
```diff
@@ -0,0 +1,197 @@
...
+
+        heuristics = HeuristicInformation(world_state, game_info)
+
+        if game_info.is_restart():
+            return self.__analyze_restart(world_state, game_info, heuristics)
+        else:
+            return self.__analyze_normal(world_state, game_info, heuristics)
+
+    @staticmethod
+    def __analyze_restart(
```
kylestach, line 119: Nit: Isn't convention usually one leading underscore, not two?

rj_gameplay/rj_gameplay/situation/decision_tree/plays.py
```diff
@@ -0,0 +1,168 @@
+"""This module contains all the plays used by the decision tree situation analyzer.
+Ported from the old situation alaysis."""
```
kylestach, line 2: Are these each supposed to contain a list of plays? That seems somewhat backwards. _All_ plays should provide an `is_applicable` method and then we should determine which plays we can use at runtime. Unless I'm misinterpreting this, it prevents us from dynamically selecting playbooks, which is a _very_ important feature.

rj_gameplay/rj_gameplay/tactic/pass_or_shoot.py
kylestach: Sure, fine with me.

rj_gameplay/stp/coordinator.py
```diff
@@ -0,0 +1,56 @@
...
+    def tick(self, world_state: rc.WorldState) -> None:
+        """Performs 1 ticks of the STP system:
+            1. Selects the best play to run given the passed in world state.
+            2. Ticks the best play, collecting the list of actions to run.
+            3. Ticks the list of actions.
+        :param world_state: The current state of the world.
+        """
+
+        # Call situational analysis to see which play should be running.
+        cur_situation, cur_play = self._play_selector.select(world_state)
```
kylestach, line 48: This will create a new play instance every tick, which kind of breaks everything. The current logic is, if the situation changes and `is_applicable` no longer applies, you can call `play.try_preempt()` which will either end the play or refuse to give up control (most plays just give up).

https://github.com/RoboJackets/robocup-software/blob/73c92878baf960844b5a4b34c72804093f1ea459/soccer/gameplay/situational_play_selection.py#L340-L348

rj_gameplay/stp/role/cost.py
```diff
@@ -0,0 +1,83 @@
...
+convenience."""
+
+from typing import Optional
+
+import numpy as np
+import stp.rc as rc
+import stp.role as role
+
+
+def flat_switch_cost(
```
kylestach, line 11: Previous comments on this still apply.

kylestach (changes requested):

rj_gameplay/rj_gameplay/situation/decision_tree/analyzer.py
oswinso: I originally intended this to be a "private" method, but then yeah could be single underscore.

rj_gameplay/rj_gameplay/situation/decision_tree/plays.py
oswinso: Nah, this is just a replacement for the enum we've been using before, but like it allows us to extend `ISituation` to in the future to require things for all situations, ie. "name" or "description" or whatever. This could be a big enum, but we'd need to do some metaclass magic to do the same thing. Both are identical approaches, I just thought that this was simpler.

rj_gameplay/stp/role/cost.py
oswinso: wdym?

rj_gameplay/stp/coordinator.py
kylestach: Resolved offline; we're going to try this stateless version to see how it works and if it is bad we'll remove it. For now, no changes needed.

rj_gameplay/rj_gameplay/situation/decision_tree/plays.py
kylestach: Sure, this is fine I guess. Could you leave a comment about the intentions (basically just an open enum)?

rj_gameplay/rj_gameplay/situation/decision_tree/analyzer.py
kylestach: Actually, this is fine.

rj_gameplay/stp/role/cost.py
kylestach: You marked this comment chain: https://github.com/RoboJackets/robocup-software/pull/1577/files/b8e6f7053e87e223dd7a2a5005eeda4799a7708d#r491723289 as resolved. My concern mentioned is still not addressed (i.e. I don't like trying to saturate costs).

oswinso: You're looking at old code. Check the Files Changed tab.

rj_gameplay/rj_gameplay/situation/decision_tree/analyzer.py
oswinso: Done.

rj_gameplay/stp/role/cost.py
```diff
@@ -0,0 +1,83 @@
...
+    :return: Cost function
+    """
+
+    def cost_fn(
+        robot: rc.Robot,
+        prev_result: Optional[role.RoleResult],
+        world_state: rc.WorldState,
+    ):
+        dist: float = np.linalg.norm(robot.pose[:2] - eval_pt)
+        cost: float = min(dist, saturate_dist)
```
kylestach, line 44: (From comment chain above) this still saturates, and we still haven't really defined units for cost which I think would address the original problem.

oswinso: We discussed and said that this was fine.

kylestach (approved):

Issue #1558, opened 2020-08-01
Backport `spin_all` and use it for temporary executors.


kylestach:
Currently we are using spin_some or `spin()` in a separate thread for our temporary executors. This should technically really be replaced with a `spin_all()`, which has been added to rclcpp master but has not yet been added to the foxy release.

Pull request: https://github.com/RoboJackets/robocup-software/pull/1554

Link to TODO, or file number/line: https://github.com/RoboJackets/robocup-software/blob/1aa4cee694700dd91c4fbb1eb343c8dc772a50fe/soccer/src/soccer/processor.cpp#L146

kylestach:
The following code from `rclcpp` will be helpful:

https://github.com/ros2/rclcpp/blob/aa159a5e8f6d3bf50c4c0c40dbfb656f7e8c6f97/rclcpp/src/rclcpp/executor.cpp#L406-L450

PR #1582, opened 2020-10-05
ROS 2 Debug Draw Interface

kylestach:
## Description
Debug drawer for ROS 2 nodes. This allows us to draw debug information to layers on the field by publishing to the `/viz/debug_draw` topic.

## Associated Issue
N/A

## Design
There is a one-to-one correspondence between the `DebugDraw` message and a layer. This means that when you publish a `DebugDraw` message, the previous contents of the layer will be cleared - but otherwise the most recent copy of each layer will be latched. This allows us to publish debug drawing information from nodes like planning that may not be exactly coupled to the GUI update frequency.

image
This will become more filled in as more ROS nodes require debug drawing.

### Manual test
1. Run soccer: `ros2 launch rj_robocup sim.launch.py`
2. Run the following command to publish a debug layer:
```sh
ros2 topic pub /viz/debug_draw rj_drawing_msgs/msg/DebugDraw "{layer: 'test', debug_text: [{text: 'TEST'}]}"
```
3. Run the following command to clear the debug layer:
```sh
ros2 topic pub /viz/debug_draw rj_drawing_msgs/msg/DebugDraw "{layer: 'test'}"
```

Expected result: After step 2, the text `TEST` should appear at our goal. It should be toggleable by checking/unchecking the `test` layer in the Layers tab. After step 3, the text should be cleared.


soccer/src/soccer/processor.cpp
```diff
@@ -143,7 +145,9 @@ void Processor::run() {
         ////////////////
         // Inputs
         // TODO(#1558): Backport spin_all and use it for our main executor.
-        ros_executor_->spin_some();
+        for (int i = 0; i < 10; i++) {
+            ros_executor_->spin_some();
+        }
```
kylestach, line 150: This is a hack so we can process multiple debug drawing messages in the same frame, which will be necessary. The queue length is 10, so that's the most messages we could process at once anyway.

Pending issue #1558 this will be removed.

soccer/src/soccer/ros2_temp/debug_draw_interface.cpp
```diff
@@ -0,0 +1,60 @@
...
+                                                  color_to_qt(shapes.color),
+                                                  QString::fromStdString(layer));
+        }
+        for (const auto& segment : debug_draw->segments) {
+            context_->debug_drawer.draw_segment(rj_convert::convert_from_ros(segment.segment),
+                                                color_to_qt(segment.color),
+                                                QString::fromStdString(layer));
+        }
+        for (const auto& pose : debug_draw->poses) {
+            // TODO: Handle poses
```
kylestach: We don't currently have a way of explicitly drawing poses, but eventually we will probably draw them as coordinate transforms (i.e. display the axes).

soccer/src/soccer/processor.hpp
```diff
@@ -4,33 +4,36 @@
...
-#include "node.hpp"
 #include "joystick/manual_control_node.hpp"
 #include "joystick/sdl_joystick_node.hpp"
 #include "motion/motion_control_node.hpp"
+#include "node.hpp"
 #include "planning/planner_node.hpp"
 #include "radio/radio.hpp"
 #include "radio/radio_node.hpp"
+
 #include "rc-fshare/rtp.hpp"
```
kylestach, line 37: The include reordering was done by the formatter, not sure why it decided to do everything.

rj_drawing_msgs/msg/DrawColor.msg
```diff
@@ -0,0 +1,3 @@
+int8 r
+int8 g
+int8 b
```
oswinso: why not RGBA? also, EOF newline.

rj_drawing_msgs/msg/DrawPose.msg
```diff
@@ -0,0 +1 @@
+rj_geometry_msgs/Pose pose
```
oswinso: `DrawColor color`?

rj_drawing_msgs/msg/DebugDraw.msg
```diff
@@ -0,0 +1,7 @@
+string layer
+
+DrawShapes[] shapes
+DrawSegment[] segments
+DrawPose[] poses
+DrawPath[] paths
+DrawText[] debug_text
```
oswinso, line 9: Do we not need a header, ie. timestamp?

soccer/src/soccer/ros2_temp/debug_draw_interface.cpp
```diff
@@ -0,0 +1,60 @@
...
+
+namespace ros2_temp {
+
+DebugDrawInterface::DebugDrawInterface(Context* context, rclcpp::Executor* executor)
+    : context_(context) {
+    node_ = std::make_shared<rclcpp::Node>("_debug_draw_interface");
+    debug_draw_sub_ = node_->create_subscription<rj_drawing_msgs::msg::DebugDraw>(
+        viz::topics::kDebugDrawPub, rclcpp::QoS(10),
+        [this](rj_drawing_msgs::msg::DebugDraw::SharedPtr debug_draw) {  // NOLINT
+            spdlog::info("Got message");
```
oswinso: Debug log?

soccer/src/soccer/ros2_temp/debug_draw_interface.cpp
```diff
@@ -0,0 +1,60 @@
...
+
+#include <rj_constants/topic_names.hpp>
+
+namespace ros2_temp {
+
+DebugDrawInterface::DebugDrawInterface(Context* context, rclcpp::Executor* executor)
+    : context_(context) {
+    node_ = std::make_shared<rclcpp::Node>("_debug_draw_interface");
+    debug_draw_sub_ = node_->create_subscription<rj_drawing_msgs::msg::DebugDraw>(
+        viz::topics::kDebugDrawPub, rclcpp::QoS(10),
```
oswinso: Change the `10` to some `constexpr` constant?

soccer/src/soccer/ros2_temp/debug_draw_interface.cpp
```diff
@@ -0,0 +1,60 @@
...
+            latest_[debug_draw->layer] = debug_draw;
+        });
+
+    executor->add_node(node_);
+}
+
+void DebugDrawInterface::run() {
+    const auto& color_to_qt = [](const rj_drawing_msgs::msg::DrawColor& color) {
+        return QColor::fromRgb(color.r, color.g, color.b);
+    };
```
oswinso: why a lambda and not some free function?

soccer/src/soccer/ros2_temp/debug_draw_interface.cpp
```diff
@@ -0,0 +1,60 @@
...
+
+    for (const auto& [layer, debug_draw] : latest_) {
+        spdlog::info("Got layer {}", layer);
```
oswinso: Debug log?

soccer/src/soccer/ros2_temp/debug_draw_interface.cpp
```diff
@@ -0,0 +1,60 @@
...
+            // TODO(#1584): Handle poses
+        }
+        for (const auto& path : debug_draw->paths) {
+            auto* debug_path = context_->debug_drawer.add_debug_path();
+            for (const auto& point : path.points) {
+                auto* new_point = debug_path->add_points();
+                new_point->mutable_pos()->set_x(point.x);
+                new_point->mutable_pos()->set_y(point.y);
+
+                // TODO(Kyle): Use color in trajectory
```
oswinso: Change from TODO(Kyle) to an issue number?

oswinso (changes requested):
Just a few changes and questions

rj_drawing_msgs/msg/DrawPose.msg
kylestach: We don't have a way of rendering poses yet but I was planning on drawing them with just red/blue axes in the standard format. Any thoughts on a better way that could represent colors?

soccer/src/soccer/ros2_temp/debug_draw_interface.cpp
kylestach: Yeah left this in, oops.

soccer/src/soccer/ros2_temp/debug_draw_interface.cpp
kylestach: There wasn't really anywhere else this was likely to be used? I'll just make it a free function and leave it in a utils/ or something.

soccer/src/soccer/ros2_temp/debug_draw_interface.cpp
oswinso: what about arrows?

kylestach: I guess that would work, or in general any kind of line segment+marker. I'll add in color so it's usable.

oswinso (approved):
lgtm

HussainGynai (approved):
looks good

PR #1593, opened 2020-10-27
Ball Classifier Skeleton

HussainGynai:
## Description
A classifier which will be used for testing to categorize certain events that happen within a test, like a pass, a shot, etc.


kylestach (commented):
Do we really want to bring in the old state machine abstraction? If we do it should probably be somewhere more general than `classifier`.

liquidmetal9015 (approved):
Looks good to me

adithya-gv (commented):
looks good!

adithya-gv (approved):

PR #1602, opened 2020-11-08
Add configuration variables for trajectory checking utilities

gauravgarre:
## Description
Add configuration variables for trajectory checking utilities

## Associated Issue
Issue #1525 

## Design Documents
[Link](link-to-design-doc)

### Test Case 1
1. Step 1
2. Step 2
3. Step 3

Expected result: ???

### DIFF
--- a/soccer/src/soccer/planning/params.cpp
+++ b/soccer/src/soccer/planning/params.cpp
@@ -1,0 +1,6 @@
+#include <rj_param_utils/param.hpp>
+
+namespace trajectory_utils {
+    DEFINE_INT64(kMaxIterations, max_iterations, 100, "Limit on iterations")
+    DEFINE_NS_FLOAT64(kExpectedDt, RJ::Seconds, expected_dt, 0.05, "Delta time between program is started and first message received")
+}
--- a/soccer/src/soccer/planning/params.hpp
+++ b/soccer/src/soccer/planning/params.hpp
@@ -1,0 +1,11 @@
+#pragma once
+
+#include <rj_param_utils/param.hpp>
+
+namespace trajectory_utils {
+    constexpr auto kMaxIterations = "trajectory_utils";
+    constexpr auto kExpectedDt = "trajectory_utils";
+
+    DECLARE_INT64(kMaxIterations, max_iterations)
+    DECLARE_NS_FLOAT64(kExpectedDt, RJ::Seconds, expected_dt)
+}
--- a/soccer/src/soccer/planning/trajectory_utils.cpp
+++ b/soccer/src/soccer/planning/trajectory_utils.cpp
@@ -27,12 +27,9 @@ bool trajectory_hits_static(const Trajectory& trajectory, const rj_geometry::Sha
     // we hit a 5 second trajectory. If our trajectory is longer than that,
     // something is probably wrong, but we'll still handle it (just scale dt
     // accordingly).
-    // TODO(#1525): Make these config variables.
-    DECLARE_NS_INT64(kMaxIterations, int, 100);
-    DECLARE_NS_INT64(kExpectedDt, RJ::Seconds, 0.05);
 
     RJ::Seconds time_left{trajectory.end_time() - start_time};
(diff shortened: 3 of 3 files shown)


gauravgarre:
Pretty positive I did not do this right

soccer/src/soccer/planning/trajectory_utils.cpp
```diff
@@ -28,8 +28,8 @@ bool trajectory_hits_static(const Trajectory& trajectory, const rj_geometry::Sha
     // something is probably wrong, but we'll still handle it (just scale dt
     // accordingly).
     // TODO(#1525): Make these config variables.
-    constexpr int kMaxIterations = 100;
-    constexpr RJ::Seconds kExpectedDt{0.05};
+    DECLARE_NS_INT64(kMaxIterations, int, 100);
```
kylestach: You need to `DECLARE_` in the `.hpp` file and `DEFINE_` in the `.cpp` file.

Also, the syntax for this isn't quite right. The arguments of the macro are:
```cpp
DECLARE_NS_<INT64|FLOAT64|...>(kParameterModuleName, some::namespace::for::the::param, param_name);
```

soccer/src/soccer/planning/params.cpp
```diff
@@ -0,0 +1,6 @@
+#include <rj_param_utils/param.hpp>
+
+namespace trajectory_utils {
+    DEFINE_INT64(kMaxIterations, max_iterations, 100, "Limit on iterations")
```
kylestach, line 4: I think it should be okay to put these directly in the `trajectory_utils.hpp/cpp` files.

soccer/src/soccer/planning/params.hpp
```diff
@@ -0,0 +1,11 @@
+#pragma once
+
+#include <rj_param_utils/param.hpp>
+
+namespace trajectory_utils {
+    constexpr auto kMaxIterations = "trajectory_utils";
+    constexpr auto kExpectedDt = "trajectory_utils";
```
kylestach, line 7: These variables should be deleted. There should be a constant for `kPlanningParams` (the string constant is the module name).

soccer/src/soccer/planning/params.hpp
```diff
@@ -0,0 +1,11 @@
...
+
+    DECLARE_INT64(kMaxIterations, max_iterations)
+    DECLARE_NS_FLOAT64(kExpectedDt, RJ::Seconds, expected_dt)
```
kylestach, line 10: Don't make this a `_NS_` variable. You're already in an outer-level namespace.

kylestach (changes requested):
Sorry for the late review! Here are some suggestions.

soccer/src/soccer/planning/trajectory_utils.cpp
kylestach: For reference the format for the non-namespaced version is just:
```cpp
DECLARE_<INT64|FLOAT64|...>(kParameterModuleName, param_name);
```

soccer/src/soccer/planning/params.hpp
```diff
@@ -0,0 +1,11 @@
+#pragma once
+
+#include <rj_param_utils/param.hpp>
+
+namespace trajectory_utils {
+    constexpr auto kMaxIterations = "trajectory_utils";
+    constexpr auto kExpectedDt = "trajectory_utils";
+
+    DECLARE_INT64(kMaxIterations, max_iterations)
```
mamarcus64, line 9: not sure if I understand this correctly, but if kMaxIterations represents the same value as max_iterations why not just use max_iterations and cut out the middle-man variable

adithya-gv: i think kMaxIterations is later modified to be a different value? but maybe im just understanding it differently as well.

kylestach: This should be `DECLARE_INT(kPlanningParams, max_iterations)`.

The first argument to this macro is a string representing the "module name", which is just how we group together a bunch of parameters. The second one is the variable name of the parameter, which can be given an initial value but is also able to be changed through ROS.

PR #1608, opened 2020-11-15
Action Stubs

HussainGynai:
## Description
Created actions tubs for some actions, and created a dummy Ctx for actions.


liquidmetal9015:
Should done just go in the action interface? Are there going to be actions that are never done?

Yudai-8-om (approved):
LGTM!
You may want an __init__ function under Capture class in capture.py as you did in dribble.py?

liquidmetal9015 (approved):
Looks good to me.

PR #1610, opened 2020-11-16
Expand stp types in rc.py (Early PR)

liquidmetal9015:
## Description
I'm going to go ahead and put this out so people can look at it. I think that the stp types in rc.py will need to be expanded. I still need to look over this and test the code so I expect there to be some changes.

1. Remove has_ball from Robot, add is_ours, visible, ball_sense_triggered, has_ball_sense, kicker_charged, kicker_healthy, lethal_fault. I think that there may be things to add or remove from this list or we may want to package them differently. is_ours is an additional mechanism to manage which robots are ours, so that robots can be lumped together.

I've used warnings for when you attempt to access information about an opposing robot that is unknown as obviously this should be avoided

2. Expand Field and move it into WorldState rather than being in GameInfo.

I need to create and run some tests, do some more thinking about what we need here, style etc. I'd like to hear any input.


rj_gameplay/stp/rc.py
```diff
@@ -133,24 +233,90 @@ class GameRestart(Enum):
...
+    """
+    A list of the information avalible in the FieldDimensions.msg
+    we probably need many of these included in WorldState
+    float32 length
+    float32 width
+    float32 border - What is border?
+    float32 line_width - Does gameplay need this? 
+    float32 goal_width
+    float32 goal_depth - Does gameplay need this?
+    float32 goal_height - Does gameplay need this?
```
HussainGynai: I don't think that these things will be useful for now, if we need them it should be simple to add later.

HussainGynai (approved):
LGTM

rj_gameplay/stp/rc.py
```diff
@@ -5,38 +5,66 @@
...
         :param robot_id: Shell id of the robot.
+        :param is_ours: Whether the robot is one of our robots
         :param pose: Pose of the Robot. [x, y, theta].
         :param twist: Twist of the robot. [dx, dy, dtheta].
-        :param has_ball: Whether the robot has the ball or not.
+        :param ball_sense_triggered: Whether the ball sensor is triggered
+        :param visible: Whether the robot is being seen by the global vision system
+        :param has_ball_sense: Whether the robots ball sensor is functional
+        :param kicker_charged: Whether the robots kicker capacitors are charged
+        :param kicker_healthy: Whether the robots kicker is "healthy" (check on what that means)
```
kylestach: "Healthy" just means the board is connected and responding.

kylestach (approved):

rj_gameplay/tests/stp/test_rc.py
```diff
@@ -0,0 +1,100 @@
...
+
+def test_generate_test_GameInfo() -> None:
+    info = GameInfo.generate_test_playing_gameinfo()
+    assert info.period == GamePeriod.FIRST_HALF
+    assert info.state == GameState.PLAYING
+    assert info.restart == GameRestart.NONE
+    assert info.our_restart == False
+
+
+def test_generate_test_WorldState() -> None:
```
kylestach:
```suggestion
def test_generate_test_world_state() -> None:
```
And et cetera with other spots. Just because you're "quoting" a type name doesn't mean you should break naming conventions (`snake_case` functions).

rj_gameplay/stp/rc.py
```diff
@@ -5,38 +5,91 @@
...
-        self.__has_ball = has_ball
+        self.__ball_sense_triggered = ball_sense_triggered
+        self.__visible = visible
+        self.__has_ball_sense = has_ball_sense
+        self.__kicker_charged = kicker_charged
+        self.__kicker_healthy = kicker_healthy
+        self.__lethal_fault = lethal_fault
+
+    @classmethod
+    def generate_test_robot(cls,
```
kylestach: This should honestly be in a separate module (i.e. `stp.testing.generate_test_robot()`), not a class function.

kylestach (approved):
More minor change requests.

rj_gameplay/stp/rc.py
liquidmetal9015: Yeah, I can agree with that.

mamarcus64 (commented):
looks good to me. might want to make the test files more dynamic for more robust tests, or move all of the magic number variables to the top for easy changes

liquidmetal9015:
@HussainGynai I think this should generally be ready, but we will of course have to be careful with some of the other PRs because this will change/break their functionality. Which is a reason that we should get this through sooner rather than later before more things outrun this overdue PR.

Or we could merge the game play entry point PR first and I could add modifications to that code into this PR, but either way we need to get these merged imo.

PR #1617, opened 2020-12-24
Gameplay Entry Point

HussainGynai:
## Description
An entry point to get the world state to the gameplay side of the codebase. Will need to be updated when a gamestate topic exists.

### Test Case 1
1. Run sim
2. In another terminal run the entry point test file

Expected result: 
You should see printed the ball's position and if you move the ball into the right half it will stop.


kylestach:
Could you add this to the launch file?

kylestach:
Also, it looks like tests are failing because our (python-side) CI script doesn't source ros2 setup scripts.

rj_gameplay/stp/gameplay_sub.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+class GameplaySub(Node):
+    """A node which subscribes to the world_state and game state topics and creates the converts the messages to python types."""
+
+    def __init__(self, world_state: Optional[rc.WorldState] = None) -> None:
+        rclpy.init()
+        super().__init__('minimal_subscriber')
+        self.world_state_sub = self.create_subscription(msg.WorldState, '/vision_filter/world_state', self.update_world_state, 10)
+        self.game_state_sub = self.create_subscription(msg.GameState, '/game_state', self.update_game_state, 10)
```
kylestach: Use relative paths here instead of absolute, so we can use node namespacing for sim2play.

rj_gameplay/stp/gameplay_sub.py
```diff
@@ -0,0 +1,41 @@
...
+
+from rj_msgs import msg
+import stp.rc as rc
+from stp.utils.world_state_converter import worldstate_message_converter
+import numpy as np
+
+from typing import List, Optional
+
+
+class GameplaySub(Node):
```
kylestach: I would do this as a `GameplayNode` that runs the gameplay spin function off of a ROS timer.

kylestach (changes requested):

.circleci/config.yml
```diff
@@ -59,9 +59,9 @@ jobs:
       - run: |
           source /opt/ros/foxy/setup.sh && source install/setup.bash && cd soccer/src/soccer/gameplay && ./run_tests.sh
       # Run tests for rj_gameplay
-      - run: pip3 install -r rj_gameplay/requirements.txt
-      - run: PYTHONPATH="$(readlink -f rj_gameplay):$PYTHONPATH" python3 -m pytest --cov rj_gameplay --cov stp rj_gameplay --cov-report xml
-      - run: bash <(curl -s https://codecov.io/bash)
+      - run: source /opt/ros/foxy/setup.sh && source install/setup.bash && pip3 install -r rj_gameplay/requirements.txt
```
kylestach: You should only have to do the source step once. You should be able to do it as a separate `run` step.

rj_gameplay/stp/gameplay_node.py
```diff
@@ -0,0 +1,66 @@
...
+from rj_msgs import msg
+import stp.rc as rc
+from stp.utils.world_state_converter import worldstate_message_converter
+import stp.situation as situation
+import stp.coordinator as coordinator
+import numpy as np
+
+from typing import List, Optional
+
+class EmptyPlaySelector(situation.IPlaySelector):
```
kylestach: Isn't `stp` the library and `rj_gameplay` is the application code? This should be there instead.

rj_gameplay/stp/gameplay_node.py
```diff
@@ -0,0 +1,66 @@
...
+
+    def gameplay_tick(self) -> None:
+        """
+        ticks the gameplay coordinator using recent world_state
+        """
+
+        if self.world_state is not None:
+            pass
+            # self.gameplay.tick(self.world_state)
+            # Uncomment when
```
kylestach: Yes, uncomment when?

rj_gameplay/stp/gameplay_sub.py
```diff
@@ -0,0 +1,41 @@
+import rclpy
```
kylestach: This file should be deleted, right?

rj_gameplay/tests/stp/entry_point/entry_point_test.py
```diff
@@ -0,0 +1,21 @@
...
+        rclpy.spin_once(world_state)
+        if world_state.get_world_state() is not None:
+            ball = world_state.get_world_state().ball
+            print(ball.pos)
+            if ball.pos[0] < 0:
+                world_state.shutdown()
+                print('shutdown')
+                break
+
+test_entry_point()
```
kylestach: This isn't an automated test, so it doesn't really belong here. You could put it somewhere else, but I'd just suggest deleting it for now.

kylestach (approved):
Looks good, a few minor suggestions.

rj_gameplay/stp/utils/world_state_converter.py
```diff
@@ -0,0 +1,67 @@
+from rj_msgs import msg
+import stp.rc as rc
+import numpy as np
+
+from typing import List
+
+def robotstate_to_robot(robot_msg: msg._robot_state.RobotState, index: int) -> rc.Robot:
```
liquidmetal9015: This will have to be changed with the rc.py rewrite but I can do that after this is merged and before the rc.py changes are merged in.

rj_gameplay/stp/utils/world_state_converter.py
```diff
@@ -0,0 +1,67 @@
...
+    twist = np.array([dx,dy,dtheta])
+
+    has_ball: bool
+    has_ball = None
+
+    robot = rc.Robot(robot_id, pose, twist, has_ball)
+
+    return robot
+
+def ballstate_to_ball(ball_msg: msg._ball_state.BallState) -> rc.Ball:
```
liquidmetal9015: This stuff as well, I'll have to add more information to most of the converters.

liquidmetal9015 (approved):

rj_gameplay/stp/utils/world_state_converter.py
kylestach: With the expanded rc.py this constructor will also take in a RobotStatus message. Not sure if I'm forgetting anything else.

(That's relevant to @liquidmetal9015, but @HussainGynai doesn't have to care about it for this PR.)

kylestach: You should still definitely be able to make automated tests for these converters, FWIW.

kylestach (approved):
Add automated tests for converters and then we should be good.

liquidmetal9015:
I will have to modify this code to match with the new rc.py stuff so it might be fine to wait until then and have me write the tests along with the rc.py testing functions in my PR.

PR #1620, opened 2021-01-17
Use clang and LLD to speed up compile times.

kylestach:
## Description
Use clang and LLD to speed up compile times. It turns out that these alternatives (especially LLD) cut out the bulk of our compile time.

Unfortunately clang/CMake don't interact nicely with precompiled headers, so this commit also disables the precompiled geometry header. That said, this was originally just a compilation speedup technique, so this is acceptable given that it results in a net compile time decrease.

On my machine this diff reduces compile times from ~2 minutes to under 30 seconds.

## Associated Issue
N/A (General improvement)

## Design Documents
N/A

### Test Case 1
1. `make clean`
2. `make`

Expected result: compile times should be improved.


kasohrab:
resolved by latest merge

Issue #1586, opened 2020-10-15
Global Parameter Server


kylestach:
## Description
There are some parameters, like world/robot parameters, that need to be stored globally. Some examples are listed:
 - Robot motion parameters (max linear/angular velocity and acceleration)
 - Kick and chip speed
 - Ball friction

## Requirements
 - Add a global parameter server node
 - Create a parameter provider that pulls parameters from another node
    - Use the `AsyncParameterClient` to listen to events on the global parameter server and update parameters accordingly
 - Add a way of creating a callback on a parameter so we can call some function when it updates

## Design
We will want a new `global_parameters` node responsible for storing global parameters. Then, we will want two parameter providers that should be used with the `params::kGlobalParamModule` tag: one for the actual global parameters node (which takes care of actually registering parameters) and one for the client nodes that internally uses `AsyncParameterClient` to dispatch callbacks to the correct parameter objects and apply changes.

PR #1621, opened 2021-01-21
Global Parameter Server

Tohwx:
## Description
A parameter provider for a global parameter server node. Uses a parameter client to listen for new/changed parameters on the global parameter server and updates the local parameter values accordingly.

## Associated Issue
Issue #1586 

### Automated test cases
 - Covers basic functionality of setting from within the same program
Expected result: pass

This node is currently not used by soccer.


rj_param_utils/include/rj_param_utils/global_param.hpp
```diff
@@ -0,0 +1,61 @@
...
+
+#include <rclcpp/rclcpp.hpp>
+#include <rclcpp/node.hpp>
+#include <rj_param_utils/param.hpp>
+#include <rj_param_utils/ros2_param_provider.hpp>
+
+/**
+ * @brief Create node to store global parameters.
+ */
+auto global_param_node = rclcpp::Node::make_shared("global_parameter_node");
```
kylestach: This should be in a `main()` function for the global parameter node. Put it in `global_param_server.cpp` or something (for now, each ROS node is a separate binary).

rj_param_utils/include/rj_param_utils/global_param.hpp
```diff
@@ -0,0 +1,61 @@
...
+     * @param new_value The new value of the parameter.
+     */
+    template <typename ParamType>
+    void SetGlobalParameter(const std::string& full_name, const ParamType& new_value);
+
+    /**
+     * @brief Propagates the values of the global parameters to a specified node.
+     * @param node Node which has parameters to update. 
+     */
+    void PropagateParameters(rclcpp::Node* node);
```
kylestach: You shouldn't need to do this explicitly. You should set it up so that parameters are automatically set when we get a new value on the parameter callback.

rj_param_utils/include/rj_param_utils/global_param.hpp
```diff
@@ -0,0 +1,61 @@
...
+
+private:
+    rclcpp::Node* node_;
+    auto parameters_client = std::make_shared<rclcpp::AsyncParametersClient>(node_);
```
kylestach: The initialization for this needs to be in the constructor.

rj_param_utils/include/rj_param_utils/global_param.hpp
```diff
@@ -0,0 +1,61 @@
...
+
+    /**
+     * @brief Sets a new global parameter with the passed in full_name and 
+     * with new_value.
+     * @tparam ParamType Type of the paramater.
+     * @param full_name The full_name of the parameter to set.
+     * @param new_value The new value of the parameter.
+     */
+    template <typename ParamType>
+    void SetGlobalParameter(const std::string& full_name, const ParamType& new_value);
```
kylestach: I don't think we need this for now.

build-debug/CMakeCache.txt
```diff
@@ -0,0 +1,375 @@
+# This is the CMakeCache file.
```
kylestach: This is my bad, I forgot to add `build-*` to the `.gitignore`.

In general, you should avoid doing `git add .` or related commands, though - instead prefer to add individual files you want to change.

rj_param_utils/include/rj_param_utils/global_param.hpp
```diff
@@ -0,0 +1,31 @@
...
+
+
+namespace params {
+/**
+ * Implementation of a global parameter provider used by every node.
+ */
+
+class ROS2GlobalParamProvider : public ::params::ROS2ParamProvider {
+public:
+    explicit ROS2GlobalParamProvider(rclcpp::Node* node, const rclcpp::Node* global_node);
```
kylestach: Last change on the skeleton: you don't need (and for some situations can't have) a pointer to the global parameter node, as it can be running in a separate binary. I would suggest just accepting a `const std::string&` for the node name (that's all the `AsyncParametersClient` needs for the remote node, anyway).

rj_param_utils/src/global_param.cpp
```diff
@@ -0,0 +1,23 @@
...
+        const std::string& global_node) : node_{node}, global_node_{global_node} {
+    params_client_ = std::make_shared<rclcpp::AsyncParametersClient>(node);
+
+    auto on_param_event_callback = 
+        [this](const rcl_interfaces::msg::ParameterEvent::SharedPtr event) -> void
+        {
+            std::vector<std::string> names;
+            for (auto& changed_parameter : event->changed_parameters) {
+                names.emplace_back(changed_parameter.name);
+                Update(changed_parameter.name, params_client_->get_parameters(names));
```
kylestach: I think the parameter values should be contained in the `changed_parameter` field, as `changed_parameter.value`. It is of type `ParameterValue`:

https://github.com/ros2/rcl_interfaces/blob/836c8653825f0cf6e6ee46d6a3837b370678004b/rcl_interfaces/msg/ParameterValue.msg#L1-L38

However, IIRC this type is not accepted by `Update` directly. I would break out this lambda:
https://github.com/RoboJackets/robocup-software/blob/1e65413efbcef57f50f516279854f22f1c2334f0/rj_param_utils/src/ros2_param_provider.cpp#L69-L115
into a separate `protected` method in `ROS2ParamProvider`, and then call it here.

This has the added benefit that you won't have to keep track of `names`, so that vector can be removed.

rj_param_utils/src/ros2_param_provider.cpp
```diff
@@ -8,6 +8,59 @@ ROS2ParamProvider::ROS2ParamProvider(rclcpp::Node* node, const std::string& modu
...
+                break;
+            default:
+                break;
+        }
+
+        // Failure could happen in two main cases:
+        //  - The parameter does not exist
+        //  - The parameter's type as declared does not match the parameter's current type
+        if (!success) {
+            // NOLINTNEXTLINE(bugprone-lambda-function-name)
```
kylestach, line 51: You can remove this `NOLINTNEXTLINE`, as it's no longer in a lambda.

rj_param_utils/src/ros2_param_provider.cpp
```diff
@@ -68,55 +121,7 @@ void ROS2ParamProvider::InitUpdateParamCallbacks(rclcpp::Node* node) {
...
-                // NOLINTNEXTLINE(bugprone-lambda-function-name)
-                SPDLOG_WARN("Failed to set parameter {}", param.get_name());
-                all_succeeded = false;
-            }
-        }
-
-        SetParametersResult set_parameters_result;
-        set_parameters_result.successful = all_succeeded;
-        return set_parameters_result;
+        UpdateParameters(params);
```
kylestach: This lambda needs to return something.
```suggestion
        return UpdateParameters(params);
```
Also, then you shouldn't have to explicitly specify that it returns `SetParametersResult` above.

rj_param_utils/include/rj_param_utils/global_param.hpp
```diff
@@ -0,0 +1,26 @@
...
+public:
+    explicit ROS2GlobalParamProvider(rclcpp::Node* node, const std::string& global_node); 
+
+private:
+    rclcpp::Node* node_;
+    const std::string& global_node_;
+    rclcpp::AsyncParameterClient::SharedPtr params_client_;
+    rclcpp::Subscription<rcl_interfaces::msg::ParameterEvent>::SharedPtr params_client_sub_;
+    
+}
```
kylestach: Compile nit: need a semicolon here.

kylestach (changes requested):
You also need a return value for the lambda here:
https://github.com/RoboJackets/robocup-software/blob/622539958cfd21c6497064df0b64aabc98ddbf98/rj_param_utils/src/ros2_param_provider.cpp#L123-L126

rj_constants/include/rj_constants/topic_names.hpp
```diff
@@ -35,6 +35,7 @@ constexpr auto kQuickRestartSrv = "referee/quick_restart";
 namespace vision_receiver::topics {
 constexpr auto kRawProtobufPub = "vision_receiver/raw_protobuf";
 constexpr auto kDetectionFramePub = "vision_receiver/detection_frame";
+constexpr auto kVisionReceiverParamModule = "vision_receiver";
```
kylestach: I think this already exists in the `rj_vision_receiver` source somewhere.

rj_param_utils/CMakeLists.txt
```diff
@@ -12,7 +12,7 @@ find_package(rclcpp REQUIRED)
 # ======================================================================
 # Define Targets
 # ======================================================================
-add_library(rj_param_utils SHARED)
+add_executable(global_params)
```
kylestach: We can still keep the old target. We still need to link to it from other nodes so we can actually use the parameter library.

rj_param_utils/CMakeLists.txt
```diff
@@ -24,28 +24,15 @@ set(RJ_PARAM_UTILS_DEPS_INCLUDE_DIRS include)
...
 # ======================================================================
 # Packaging
 # ======================================================================
 install(
-  TARGETS rj_param_utils
-  ARCHIVE DESTINATION lib
-  LIBRARY DESTINATION lib
-  RUNTIME DESTINATION bin)
+  TARGETS global_params
+  DESTINATION lib/${CMAKE_PROJECT_NAME})
```
kylestach: Again, you can keep the old install target (and the testing subdirectories above).

rj_param_utils/src/global_param.cpp
```diff
@@ -0,0 +1,28 @@
...
+
+    auto on_param_event_callback = 
+        [this](const rcl_interfaces::msg::ParameterEvent::SharedPtr event) {
+           return UpdateParameters(event->changed_parameters); 
+        };
+
+    params_client_sub_ = params_client_->on_parameter_event(on_param_event_callback);
+}
+
+int main(int argc, char**argv) {
```
kylestach: This main function should be in a `global_param_main.cpp`, which should only be included in the executable (not the library).

rj_param_utils/src/global_param.cpp
```diff
@@ -0,0 +1,28 @@
...
+    rclcpp::init(argc, argv);
+    rclcpp::Node* global_param_node = std::make_shared<rclcpp::Node>(params::kGlobalParamNodeName);
+    global_param_node->declare_parameter("global_bool", false);
+    global_param_node->declare_parameter("global_float", 3.14f);
+    global_param_node->declare_parameter("global_string", "I am a string.");
```
kylestach: TODO remove these once we verify from another node that things work.

kylestach: We should also be able to add an automated test case that spins up ROS, creates a global param node, creates another node and a few global params, and makes sure we can access it.

rj_vision_receiver/src/vision_receiver.cpp
```diff
@@ -11,9 +11,8 @@
 #include <rj_utils/logging_macros.hpp>
 #include <rj_vision_receiver/vision_receiver.hpp>
 
-constexpr auto kVisionReceiverParamModule = "vision_receiver";
```
kylestach: Ah, I see. I think this variable can stay here instead of moving to `rj_constants`.

kylestach (commented):
Almost done! A few things still need to be moved around, and I think an automated test would be good, but otherwise this looks pretty close.

rj_param_utils/CMakeLists.txt
```diff
@@ -30,6 +30,8 @@ target_link_libraries(rj_param_utils PUBLIC ${RJ_PARAM_UTILS_DEPS_LIBRARIES})
 
 ament_target_dependencies(rj_param_utils PUBLIC rclcpp)
 
+target_sources(rj_param_utils PRIVATE ${RJ_PARAM_UTILS_SRCS})
```
kylestach: You shouldn't need this here, only in the `src/CMakeLists.txt` file.

rj_param_utils/CMakeLists.txt
```diff
@@ -41,6 +43,9 @@ if(BUILD_TESTING)
   target_link_libraries(rj_param_utils_testing PUBLIC rj_param_utils ${GTEST_LIBRARIES})
 endif()
 
+# position-independent-code flag
+target_compile_options(rj_param_utils PRIVATE "-fPIC")
```
kylestach: Can you check again to see if this is really necessary? I'm kind of confused as to why it wasn't working without it...

rj_param_utils/CMakeLists.txt
```diff
@@ -12,7 +12,7 @@ find_package(rclcpp REQUIRED)
 # ======================================================================
 # Define Targets
 # ======================================================================
-add_library(rj_param_utils SHARED)
+add_library(rj_param_utils STATIC)
```
kylestach: I don't think you should need to change this.

rj_param_utils/include/rj_param_utils/global_param.hpp
```diff
@@ -0,0 +1,27 @@
...
+
+namespace params {
+/**
+ * Implementation of a global parameter provider used by every node.
+ */
+
+class ROS2GlobalParamProvider : public ::params::ROS2ParamProvider {
+public:
+    explicit ROS2GlobalParamProvider(rclcpp::Node* node, const std::string& global_node); 
+    rclcpp::AsyncParametersClient::SharedPtr params_client_;
```
kylestach: This variable should be private.

rj_param_utils/testing/CMakeLists.txt
```diff
@@ -1,7 +1,7 @@
 # ======================================================================
 # Set Testing Sources
 # ======================================================================
-set(RJ_PARAM_UTILS_TESTING_SRCS src/declare_test.cpp src/param_test.cpp)
+set(RJ_PARAM_UTILS_TESTING_SRCS src/declare_test.cpp src/param_test.cpp src/gpp_test.cpp)
```
kylestach: Nit: rename this file to `global_param_provider_test.cpp`.

rj_param_utils/src/global_param.cpp
```diff
@@ -0,0 +1,30 @@
+#include <rj_param_utils/param.hpp>
```
kylestach: Nit: this file should be `global_param_provider.hpp/cpp`.

rj_param_utils/testing/src/gpp_test.cpp
```diff
@@ -0,0 +1,156 @@
...
+
+    ASSERT_EQ(PARAM_bare_int64_vec.size(), kExampleIntVecValue.size());
+    for (size_t i = 0; i < PARAM_bare_int64_vec.size(); i++) {
+        EXPECT_EQ(PARAM_bare_int64_vec[i], kExampleIntVecValue[i]);
+    }
+
+    ASSERT_EQ(PARAM_bare_string_vec.size(), kExampleStringVecValue.size());
+    for (size_t i = 0; i < PARAM_bare_string_vec.size(); i++) {
+        EXPECT_EQ(PARAM_bare_string_vec[i], kExampleStringVecValue[i]);
+    }
```
kylestach: You need something in these tests that does something like:
```cpp
global_node_->set_parameter(rclcpp::Parameter("bare_bool", false);
```
and then asserts that the proper value is changed.

kylestach (commented):
Looks mostly good, a few nits and comments on the tests.

rj_param_utils/include/rj_param_utils/ros2_local_param_provider.hpp
```diff
@@ -0,0 +1,34 @@
...
+     * updates. The callback will update the registered parameters in
+     * ParamRegistry.
+     * @param node
+     */
+    void InitUpdateParamCallbacks(rclcpp::Node* node);
+
+    rclcpp::node_interfaces::OnSetParametersCallbackHandle::SharedPtr callback_handle_;
+};
+
+}  // namespace params
```
HussainGynai, line 34: Nit: Newline

HussainGynai (approved):
Apart from a few new line nits, LGTM

kylestach (approved):

PR #1625, opened 2021-02-07
Mock up of skills

HussainGynai:
## Description
Mock out of some skills, they use the stubs from that action stubs PR, so its a bit messy since that stuff is not in yet. Skill are based off a behavior tree of actions, most of them are just a sequence node with actions as children. Apart from skills I also tried to fix the fact that skills were the ones creating role requests instead of tactics like planned. Also created a general action behavior and one behavior for skills using cost functions to move.

## Design Documents
[Link](https://docs.google.com/document/d/1pf-aaHnvonTPC6ccI7RdHudpuev84fZslMVDq8lw59s/edit)

## Questions
Not quite sure how to handle behaviors for the trees, currently just have one for most actions and one other for skills that use a cost function to determine where a robot should move. It could also be possible that we have a thin behavior for each action.


kylestach:
Where does this stand in relation to #1631?

PR #1628, opened 2021-02-21
Fumble classifier

Yudai-8-om:
## Description
Draft of the fumble classifier (and pass classifier).
Classified based on the velocity vector of the ball.

has_ball attribute is not working at this time.


rj_gameplay/classifier/fumble_classifier.py
```diff
@@ -0,0 +1,96 @@
+import stp.rc as rc
+import rj_gameplay.gameplay_node as gameplay_node
+import rclpy
+
+
+def fumble_classifier(world_state: rc.WorldState, possess) -> None:
```
HussainGynai: You have this in a few places, so I'll mention it at the top, you should be putting a type next to most and putting the correct return type for each function. For example in this case the function declaration should be:
`def fumble_classifier(world_state: rc.WorldState, possess: int) -> bool`

rj_gameplay/classifier/fumble_classifier.py
```diff
@@ -0,0 +1,96 @@
...
+		if old_angle is not None and kicker is not None:
+			if old_angle != new_angle and possess != kicker:
+				passing = pass_classifier(gameplay_node.get_world_state(), possess, kicker)
+				fumbling = fumble_classifier(gameplay_node.get_world_state(), possess, kicker)
+				if passing:
+					kicker = possess
+				else: 
+					kicker = None
+	except NameError:
+		pass
```
HussainGynai: You may want to make this whole part a function, but just for testing purposes it should be fine.

rj_gameplay/classifier/fumble_classifier.py
```diff
@@ -0,0 +1,123 @@
...
+		vel = (ball_vel[0]**2 + ball_vel[1]**2) ** 0.5
+		if ball_vel[0] == 0:
+			angle = ball_vel[1] / (ball_vel[0]+0.00001)
+		else:
+			angle = ball_vel[1] / ball_vel[0]
+		return angle
+
+""" ball_possession returns the id of the robot who has the ball"""
+
+def ball_possession(world_state: rc.WorldState):
```
kylestach: This should work for robots on our team or theirs, right? In that case we probably need a way of representing a "tag" (i.e. robot+team). Also, it shouldn't just rely on ball sense.

rj_gameplay/classifier/fumble_classifier.py
```diff
@@ -0,0 +1,123 @@
...
+	if kicker is None:
+		kicker = possess
+
+	old_angle = new_angle
+	new_angle = get_velocity_angle(world_state)
+
+	#When the velocity angle changes, if the member possesses the ball, passing is succeeded, otherwise intercepted.
+	
+	if old_angle is not None and kicker is not None:
+		if old_angle != new_angle and possess != kicker:
```
kylestach: Purely due to noise and filtering, `old_angle` will almost never be equal to `new_angle`. You probably want `math.abs(old_angle - new_angle) < SOME_THRESHOLD`.

rj_gameplay/rj_gameplay/gameplay_node.py
```diff
@@ -0,0 +1,66 @@
+import rclpy
```
kylestach, line 1: This is from the other PR, right?

rj_gameplay/classifier/fumble_classifier.py
```diff
@@ -0,0 +1,123 @@
...
+		ball_pos_x = world_state.ball.pos[0]
+		ball_pos_y = world_state.ball.pos[1]
+		distance = ((ball_pos_x-pose_x)**2 + (ball_pos_y-pose_y)**2)**0.5
+		return distance
+
+"""
+get_velocity_angle returns the ratio of y component of the velocity vector to x component
+"""
+
+def get_velocity_angle(world_state: rc.WorldState) -> float:
```
kylestach, line 60: Based on how it's used, I would suggest making this an actual angle (see: `math.atan2` or `np.atan2`), and then dealing with angle wrap accordingly when comparing two angles. If it absolutely needs to be a ratio like this, please change the name.

rj_gameplay/classifier/fumble_classifier.py
```diff
@@ -0,0 +1,123 @@
...
+
+def fumble_classifier(world_state: rc.WorldState, possess) -> bool:
+	if possess is None:
+		return True
+	else: 
+		return False
+
+"""
+get_distance returns the distance between the ball and the kicker
+num: kicker's id
```
kylestach: Nit: `num` -> `kicker` (or rename the parameter and the doc to `kicker_id`, which is probably the ideal name)

rj_gameplay/classifier/fumble_classifier.py
```diff
@@ -0,0 +1,123 @@
...
+"""
+def pass_or_fumble(world_state: rc.WorldState, kicker, old_angle, new_angle):
+	possess = ball_possession(world_state)
+	if kicker is None:
+		kicker = possess
+
+	old_angle = new_angle
+	new_angle = get_velocity_angle(world_state)
+
+	#When the velocity angle changes, if the member possesses the ball, passing is succeeded, otherwise intercepted.
```
kylestach, line 93: I think you'll probably need to use a more complex heuristic than "velocity angle unchanged" to determine whether it's been fumbled. For example, if the ball goes straight past the intended recipient but isn't captured by an opponent, it should still count as a fumble.

rj_gameplay/classifier/fumble_classifier.py
```diff
@@ -0,0 +1,123 @@
...
+
+play_selector = gameplay_node.EmptyPlaySelector()
+gameplay_node = gameplay_node.GameplayNode(play_selector)
+kicker = None
+old_angle = None
+new_angle = None
+
+while (True):
+	rclpy.spin_once(gameplay_node)
+	[kicker, old_angle, new_angle] = pass_or_fumble(gameplay_node.get_world_state(), kicker, old_angle, new_angle)
```
kylestach: This is temporary for testing, right?

kylestach (changes requested):

rj_gameplay/classifier/fumble_classifier.py
HussainGynai: There is a possession classifier being worked on, so I think this is fine for now, but I do agree that we need a tag to represent robots.

rj_gameplay/rj_gameplay/gameplay_node.py
HussainGynai: Yeah they're using the gameplay entry point branch so its easy for them to test

rj_gameplay/classifier/fumble_classifier.py
```diff
@@ -0,0 +1,123 @@
...
+		if robot.has_ball:
+			num = robot.id
+			return num
+		else:
+			return None
+
+"""
+fumbling: boolean whether it is fumbled or not
+"""
+def pass_or_fumble(world_state: rc.WorldState, kicker, old_angle, new_angle):
```
HussainGynai: A bit of a small thing, but don't forget to include the types for the parameters as well as the return types for functions.

rj_gameplay/classifier/fumble_classifier.py
HussainGynai: I agree. I remembered that I wrote something similar for the previous version of gameplay in coordinated_pass. The function is called [ball_heading_to_recieve_point ](https://github.com/RoboJackets/robocup-software/blob/73c92878baf960844b5a4b34c72804093f1ea459/soccer/gameplay/tactics/coordinated_pass.py#L205-L220). It's not quite the same since this uses a known receive point, so you may want to find a way to guess where pass might be trying to get to.

rj_gameplay/classifier/fumble_classifier.py
```diff
@@ -0,0 +1,171 @@
...
+		distance = ((ball_pos_x-pose_x)**2 + (ball_pos_y-pose_y)**2)**0.5
+		return distance
+		
+
+#get_recipient guessing the intended recipient
+def get_recipient(world_state: rc.WorldState, team, vel_angle):
+	ball_pos_x = world_state.ball.pos[0]
+	ball_pos_y = world_state.ball.pos[1]
+	angle = 6.28
+	if team == 'our':
```
HussainGynai: probably just change team to be a boolean instead of a string. So instead of team it could be an our_team variable that is either True or False.

rj_gameplay/classifier/fumble_classifier.py
```diff
@@ -0,0 +1,171 @@
...
+		for robot in world_state.our_robots:
+			pose_x = robot.pose[0]
+			pose_y = robot.pose[1]
+			recp_angle = atan2(pose_y - ball_pos_y, pose_x - ball_pos_x)
+			angle_diff = abs(vel_angle - recp_angle)
+			if angle_diff < angle:
```
HussainGynai, line 34: Maybe I'm misunderstanding, but doesn't this always return true since your angle is 2pi, and all angles are going to be less than 2pi?

Yudai-8-om: The angle is supposed to be updated at line 42 in for loop, so I believe it would work?
Also, my concern is only considering angle is not enough to determine the recipient, what do you think?

HussainGynai: If it works from your testing then it should be good

rj_gameplay/classifier/fumble_classifier.py
```diff
@@ -0,0 +1,172 @@
+import stp.rc as rc
+import rj_gameplay.gameplay_node as gameplay_node
+import rclpy
+import math.atan2 as atan2
```
HussainGynai: Should be:
```from math import atan2```

HussainGynai:
I spent some time testing your code, and I found that your recipient finding is very good, it work quite well. For the actual determining if the pass is fumbled, it doesn't work as well. It has trouble detecting a successful pass. I think this is because you are checking if the recipient robot actually possesses the ball. This is not a bad first check, but there are a lot of times where the ball stops or slows down very close to the recipient, but the recipient doesn't actually capture it. We still want to consider this a successful pass since the recipient will likely easily drive up to and capture the ball. You may want to check if the ball is close to the recipient and going at a slow speed, and if it is, then you want to say its successful and not return fumble. But overall good work so far!

PR #1635, opened 2021-04-11
One touch pass

mamarcus64:
## Description
add the heurstic "cost_pass" into rj_gameplay


rj_gameplay/pass_costing/cost_heuristic.py
```diff
@@ -0,0 +1,73 @@
...
+
+def cost_heuristic(world_state: rc.WorldState, target_robot: rc.Robot) -> float:
+    our_robots = world_state.our_robots
+    their_robots = world_state.their_robots
+    robot_with_ball = our_robots[0]
+    for robot in our_robots:
+        if robot.ball_sense_triggered:
+            robot_with_ball = robot
+            if robot_with_ball == target_robot:
+                return 99999
```
HussainGynai, line 25: Bit of a nit but if you want to return a number larger than all others use float('inf') instead

rj_gameplay/pass_costing/cost_heuristic.py
```diff
@@ -0,0 +1,73 @@
...
+    # distance from passing robot to potential passer
+    running_cost += robot_distance_weight * \
+        normalize_cost(robot_distance(robot_with_ball, target_robot), \
+            lower_bound=0, upper_bound=MAX_ROBOT_DISTANCE)
+
+    # distance to nearest opponent robot
+    running_cost += closest_opponent_distance_weight * \
+        normalize_cost(MAX_CLOSEST_OPPONENT_DISTANCE - closest_opponent_robot_distance(target_robot, their_robots), \
+        lower_bound=0, upper_bound=MAX_CLOSEST_OPPONENT_DISTANCE)
+
```
HussainGynai, line 43: I think this is a very good start but I think some considerations are missing. For example you could have a the target robot at some adequate distance away and downfield with no opponents near it, but an opponent robot in the way of a potential pass. In this case your cost function might pass to a robot for which the pass is blocked by an opponent robot which is not close to the target.

HussainGynai: Another issue I found using your visualizer (which is very nice btw) is that it seems that you can end up preferring really short range passes which isn't really what we want, as very short passes can be sort of pointless.

PR #1636, opened 2021-04-13
Pass Success classifier

adithya-gv:
## Description
Draft of a pass success classifier between two robots.
Classifies based on position of ball relative to position and direction of robot (or if a robot has possession of a ball).

No ROS message created after a classification.

### DIFF
--- a/rj_gameplay/classifier/pass_success_classifier.py
+++ b/rj_gameplay/classifier/pass_success_classifier.py
@@ -15,6 +15,10 @@ def classifier(world_state: rc.WorldState) -> bool:
         if robot.has_ball:
             return True
 
+    for robot in opp_robots:
+        if robot.has_ball:
+            return False
+
     net_robot_list = our_robots + opp_robots
 
     robot = shortest_distance(rc.WorldState, net_robot_list, ball)
--- a/rj_gameplay/classifier/pass_success_classifier.py
+++ b/rj_gameplay/classifier/pass_success_classifier.py
@@ -1,0 +1,91 @@
+import stp.rc as rc
+import rj_gameplay.gameplay_node as gameplay_node
+import rclpy
+
+import numpy as np
+import math as m
+
+def classifier(world_state: rc.WorldState) -> bool:
+    our_robots = world_state.our_robots
+    opp_robots = world_state.their_robots
+    ball = world_state.ball
+
+    # check if any robot on our team has the ball 
+    for robot in our_robots:
+        if robot.has_ball:
+            return True
+
+    net_robot_list = our_robots + opp_robots
+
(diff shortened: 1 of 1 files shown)


rj_gameplay/classifier/pass_success_classifier.py
```diff
@@ -0,0 +1,95 @@
+import stp.rc as rc
+import rj_gameplay.gameplay_node as gameplay_node
+import rclpy
+
+import numpy as np
+import math as m
+
+def classifier(world_state: rc.WorldState) -> bool:
```
HussainGynai: Nitpick, but just call this function pass_success_classifier or something more descriptive

rj_gameplay/classifier/pass_success_classifier.py
```diff
@@ -0,0 +1,95 @@
...
+    ball = world_state.ball
+
+    # check if any robot on our team has the ball 
+    for robot in our_robots:
+        if robot.has_ball:
+            return True
+
+    for robot in opp_robots:
+        if robot.has_ball:
+            return False
```
HussainGynai: So for the classifiers, we actually won't have access to things like robot.has_ball (this is never true for opponent robots, since we don't have access to their statuses). Instead of using has_ball  try using something like a robots distance to the ball. You can check Yudai's PR, he did something similar.

rj_gameplay/classifier/pass_success_classifier.py
```diff
@@ -0,0 +1,95 @@
...
+    
+
+def shortest_distance(world_state: rc.WorldState, robot_list, ball) -> rc.Robot:
+    ball_x = ball.pos()[0]
+    ball_y = ball.pos()[1]
+
+    ball_vel_x = ball.vel()[0]
+    ball_vel_y = ball.vel()[1]
+
+    minDistance = 100000
```
HussainGynai: Another nitpick, but instead of using a very large number use float('inf') which will basically be infinity

rj_gameplay/classifier/pass_success_classifier.py
```diff
@@ -0,0 +1,95 @@
...
+
+    ball_ang = m.atan2(displacement_y, displacement_x)
+    
+    if ball_ang < 0:
+        ball_ang = ball_ang + (2 * m.pi)
+
+    if robot_ang < 0:
+        robot_ang = robot_ang + (2 * m.pi)
+    
+    if abs(ball_ang - robot_ang) < (m.pi / 2):
```
HussainGynai: For all these constants we usually don't just want to have the actual numbers around. You should create a constant variable called something like MIN_ACCEPTABLE_ANGLE (the all caps denotes it as a constant).

PR #1639, opened 2021-05-12
Pass Costing

anshulm07:
Initial commit for regular pass-costing heuristic


rj_gameplay/pass_costing/cost_heuristic.py
```diff
@@ -0,0 +1,73 @@
...
+
+def cost_heuristic(world_state: rc.WorldState, target_robot: rc.Robot) -> float:
+    our_robots = world_state.our_robots
+    their_robots = world_state.their_robots
+    robot_with_ball = our_robots[0]
+    for robot in our_robots:
+        if robot.ball_sense_triggered:
+            robot_with_ball = robot
+            if robot_with_ball == target_robot:
+                return 99999
```
HussainGynai, line 25: Nitpick: This should probably be fine always, since you wont get a cost this high, but you should use float('inf'), which is a number which is always bigger than any other number.

rj_gameplay/pass_costing/cost_heuristic.py
```diff
@@ -0,0 +1,73 @@
...
+    return passer.pose[1] - receiver.pose[1]
+
+# distance from passing robot to potential passer
+def robot_distance(first_robot: rc.Robot, second_robot: rc.Robot) -> float:
+    return math.sqrt(pow(first_robot.pose[0] - second_robot.pose[0], 2)
+        + pow(first_robot.pose[1] - second_robot.pose[1], 2))
+
+# distance to nearest opponent robot
+def closest_opponent_robot_distance(our_robot: rc.Robot, their_robots: list) -> float:
+    min_distance = 99999
```
HussainGynai, line 57: Same here as before use float('inf')

rj_gameplay/reg_pass_costing/cost_heuristic.py
```diff
@@ -0,0 +1,157 @@
...
+
+    our_robots = world_state.our_robots
+    their_robots = world_state.their_robots
+
+    robot_with_ball = our_robots[0]
+    for bot in our_robots:
+        if bot.has_ball_sense:
+            robot_with_ball = bot
+            if robot_with_ball == robot:
+                return -1
```
HussainGynai: This should be a really big number right? You wouldn't pass to a robot which already has the ball.

rj_gameplay/reg_pass_costing/cost_heuristic.py
```diff
@@ -0,0 +1,157 @@
...
+    # return 1 if robot is too far away from robot_with_ball
+    if (distance_from_ball(robot, robot_with_ball) > MAX_PASS_DIST):
+        return 1 
+
+    running_cost = 0
+    running_cost += distance_from_goal(robot, robot_with_ball, field)
+    running_cost += average_opponent_distance(robot, their_robots)
+    running_cost += nearest_opponent_distance(robot, their_robots)
+    running_cost += opponents_between_robot_goal(robot, their_robots, field)
+    running_cost += opponents_between_robot_ball(robot, their_robots, robot_with_ball)
```
HussainGynai: I don't know how much you've tested this, but it might be good to have some float coefficients by each of the variables. So it would look something like this:
'running_cost += goal_dist_coeff * distance_from_goal' and so forth. This would make it easier to tune in the future, and we can maybe even let people specify the weights of each if maybe someone wants to prioritize a robot super close to goal or something. For now, you can just define them all to 1.

rj_gameplay/reg_pass_costing/cost_heuristic.py
```diff
@@ -0,0 +1,157 @@
...
+import numpy as np
+
+MAX_PASS_DIST = 5
+MAX_GOAL_DIST = 7.5
+MAX_AVG_OPP_DIST = 6
+MAX_NEAREST_OPP_DIST = 2.5
+MAX_OPP_BETWEEN_ROBOT_GOAL = 5
+MAX_OPP_BETWEEN_ROBOT_BALL = 1
+MIN_GOAL_DIST_BONUS = 1
+MAX_GOAL_DIST_BONUS = 1.75
```
HussainGynai, line 13: What are the points of these maximums? The logic behind them might be sound, but I'm just wondering.

rj_gameplay/reg_pass_costing/cost_heuristic.py
```diff
@@ -0,0 +1,157 @@
...
+    AB_AE = np.dot(B - A, E - A)
+
+    if AB_BE < 0 and AB_AE > 0:
+        intercept_dist = abs(np.cross(B - A, E - A)) / np.linalg.norm(B - A)
+        if intercept_dist < target_dist / 6:
+            return True
+    return False
+
+# returns the cost for the robot's distance from goal
+def distance_from_goal(robot: rc.Robot, robot_with_ball: rc.Robot, field: rc.Field) -> float:
```
HussainGynai: You may want to rename this function, since it is not only distance from goal because of the normalization with the Max and the multiplication of the bonus. The function itself seems good, the name is just a bit misleading.

rj_gameplay/reg_pass_costing/cost_heuristic.py
```diff
@@ -0,0 +1,157 @@
...
+        bonus *= ((MAX_GOAL_DIST_BONUS - MIN_GOAL_DIST_BONUS) * (robot_with_ball_Ypos - BALL_MIN_yPOS) / (BALL_MAX_yPOS - BALL_MIN_yPOS)) + 1
+
+    ROBOT_MAX_yPOS = 2.5
+    if robot_Ypos < ROBOT_MAX_yPOS:
+        bonus *= 1.5
+
+    return bonus
+
+# returns the cost for the robot's average distance from opponent robots
+def average_opponent_distance(robot: rc.Robot, their_robots: list) -> float:
```
HussainGynai: Again the name here is a bit misleading because of the normalization at the end

HussainGynai: maybe explicitly calling it cost would help, i.e average_opponent_dist_cost

HussainGynai (changes requested):
The main thing is adding the coefficients for each of the factors going into the cost, but overall everything looks very good.

rj_gameplay/reg_pass_costing/cost_heuristic.py
anshulm07: The point of the maximums was to set the range of values that maps to the cost. For example, robots that are more than 7.5 units away from the goal should all receive the maximum cost, and robots that have the nearest opponent more than 2.5 units away should all receive the minimum cost.

rj_gameplay/reg_pass_costing/cost_heuristic.py
anshulm07: Those coefficients seem like they would essentially serve the same purpose as the cost values that I have defined at the beginning of my function, but I do the multiplication within each of the individual cost functions instead. I figured that if anyone wanted to tune the function in the future, they could just adjust those cost values.

PR #1640, opened 2021-05-18
Make target to run the new gameplay tests (doesn't work, pleas help)

liquidmetal9015:
## Description
I might be dumb, but I don't think there is a maketarget that actually runs all the new python tests that CI runs. It should be trivial to create one as I have done here, except it doesn't work. Running the command regularly by copy-pasting into the command prompt runs the tests as expected, but I can't get it to work from the makefile. I'm not super experienced with the build system but if someone could tell me why this doesn't work that would be great. More info blow.

Where I'm getting the command from in CI:
image

What the issue with the make target is:
image

### DIFF
--- a/makefile
+++ b/makefile
@@ -124,6 +124,9 @@ behavior-diagrams: all
 clean:
 	cd build-debug && ninja clean || true
 
+test-python-new:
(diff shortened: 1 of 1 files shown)


kylestach:
Fixed in #1642 (our new gameplay does fail pylint though).

PR #1645, opened 2021-05-24
Add the global param server node and use it for some common parameters.

kylestach:
## Description
This adds a new node (basically does nothing) to act as a global parameter server for the other nodes.

## Associated Issue
Issue #1586.

### All parameters declared
1. Run simulation
2. Run `ros2 param list /global_parameter_server`

Expected result: several parameters are listed.

### Changing parameters
1. Add some code to print `PARAM_*` for one of the global parameters
2. Change the parameter during runtime with `ros2 param set /global_parameter_server <parameter> <value>`

Expected result: the readout should change.


kylestach:
Note: I eventually intend to put everything from `GameSettings` (https://github.com/RoboJackets/robocup-software/blob/ros2/rj_msgs/msg/GameSettings.msg) into global parameters, but we do not have that for now.

kfu02 (changes requested):
All parameters were declared, but on trying to change a parameter I got the following error:
```
❯ ros2 param set /global_parameter_server PARAM_use_sim_time 42069
Setting parameter failed
```
Granted, I did merge in gameplay_node_update because sim kept crashing on me, perhaps I did it wrong? The only merge conflict was in `autonomy_interface.cpp`, which I handled by taking the code from this branch (gameplay_node_update had some code for "_old_soccer" nodes).

kylestach:
You should set `use_sim_time` rather than `PARAM_use_sim_time` - the `PARAM_` prefix is only in C++ code - but also, don't do `use_sim_time`. That's a ROS builtin, not something from our code. I would suggest testing `soccer.robot.max_kick_speed` (in C++ code, `soccer::robot::PARAM_max_kick_speed`) instead (remember that it's a double, so you need to set it to something like 6.0 rather than 6).

kfu02 (approved):
Expected behavior achieved, after a merge with gameplay_node_update.

rj_param_utils/src/ros2_local_param_provider.cpp
```diff
@@ -40,4 +42,4 @@ void LocalROS2ParamProvider::InitUpdateParamCallbacks(rclcpp::Node* node) {
     callback_handle_ = node->add_on_set_parameters_callback(on_update);
 }
 
-}  // namespace params
+}  // namespace params
```
HussainGynai: \n

soccer/src/soccer/global_param_server.cpp
```diff
@@ -0,0 +1,18 @@
...
+    auto node = std::make_shared<rclcpp::Node>(
+        kGlobalParamServerNode, rclcpp::NodeOptions{}
+                                    .allow_undeclared_parameters(true)
+                                    .automatically_declare_parameters_from_overrides(true));
+    auto provider = params::LocalROS2ParamProvider(node.get(), kGlobalParamModule);
+    rclcpp::spin(node);
+    rclcpp::shutdown();
+    global_params_server_dummy_function();
+    return 0;
+}
```
HussainGynai: \n

HussainGynai (approved):
LGTM, just newlines

Issue #1627, opened 2021-02-13
Joystick Control


kylestach:
## Description
Joystick control is removed in #1618 and needs to be reimplemented. This should happen in a new node for manual control as well as an action in the rest of the action server.

## Requirements
 - ROS joystick API
    - Can assign joystick to robot mapping
    - Assigning a joystick starts a joystick action for the robot and publishes a message indicating that the robot is manually controlled
 - Moving the joystick causes the robot to move accordingly

## Design
N/A

kylestach:
Closed by #1647

Issue #1147, opened 2018-06-16
Keyboard Controls for Manual Control


CollinAvidano:
Due to a controller failure during comp testing it was brought up that we should really have keyboard controls for manual control of the robots in soccer.

jgkamat:
We need to make sure that the existing shortcuts don't conflict with this.

Examples are:

| s          | stop       |
| left/right | log scroll |

Collin Avidano writes:

Jason27chan:
@jgkamat Is there a place where I could find the existing shortcuts?

JNeiger:
I don't think we have added anymore since this was updated last year
https://robojackets.github.io/robocup-software/md_doc__keyboard__shortcuts.html

Jason27chan:
@jgkamat How do we implement our existing shortcuts? I found a way that [QT handles key presses](http://doc.qt.io/qt-5/qobject.html#installEventFilter), but we should probably implement our shortcuts the same way?

jgkamat:
I'm not entirely sure, but I think these are implemented through qtcreator in
the mainwindow.ui file.

For example, here's (I think) the shortcut 'q' from quicksave in
mainwindow.ui:

  <action name="actionQuicksaveRobotLocations">
   <property name="text">
    <string>Quicksave Robot Locations</string>
   </property>
   <property name="shortcut">
    <string>Q</string>
   </property>
  </action>

Jason Chan writes:

ashaw596:
Hmm. They would conflict. Maybe make a mode or something. Otherwise jump
move the frame control to alt +

PR #1647, opened 2021-05-24
Manual control implementation

kylestach:
## Description
Very basic manual control.

This PR implements keyboard and gamepad controllers. It creates a new node for managing manual inputs and a new launch configuration for this case. Unlike the old system, we do not (currently) allow hybrid setups where some robots are in manual mode and some robots are controlled by gameplay.

The launch configuration is `manual.launch.py`. It is currently configured for simulation only. Once it is running, you can list the available joysticks as follows:
```
ros2 service call /list_joysticks rj_msgs/srv/ListJoysticks
```

Then, picking from the list, you can assign a joystick to a robot:
```
ros2 service call /select_manual rj_msgs/srv/SetManual '{robot_id: 0, controller_uuid: "keyboard-controller", connect: true}'
```
`"keyboard-controller"` is always available for now.

Eventually this functionality will have a proper UI, but for now, service calls it is.

## Associated Issue
Issues #1147, #1004, #1627

## Design Documents
[Link](link-to-design-doc)

### Keyboard input
1. Fire up `manual.launch.py`
2. Run the service call above for selecting a manual control
3. Make sure the black window has focus
4. Spam w/a/s/d for movement, j/k for kick/chip, up/down arrows to change dribbler power, right/left to change kicker power.

Expected result: the selected robot obeys commands


kfu02 (approved):
Works as expected, nicely done. I'd like a way to turn L/R in place (as trying to get smooth motion with only WASD was painful) but I assume that wouldn't be hard to add later.

kylestach:
@kfu02 q/e are turning in place. Also, shift is damped (slow) movement.

kfu02:
Ah okay, makes sense.

HussainGynai (approved):
LGTM, I guess put in a TODO or issue or something to test the gamepad control at some point.

PR #1648, opened 2021-05-26
Marking skill added

kfu02:
## Description
This PR adds the mark skill, which moves a robot between the ball and an enemy robot.

## Steps to test
Run sim. 

```
source /opt/ros/foxy/setup.bash
make
source install/setup.bash
ros2 launch rj_robocup sim.launch.py
```

Start play with green check in sim window.

## Expected behavior
Yellow robot 7 should position itself between blue robot 1 and the ball, facing ball at all steps. (Drag ball and blue robot 1 around to test.)

If ball is too close to blue robot 1 to get in-between, yellow robot 7 should position itself directly in front of the ball, on the same blue-robot-to-ball line as before.


rj_gameplay/rj_gameplay/action/move.py
```diff
@@ -49,7 +49,7 @@ def tick(self, intent: msg.RobotIntent) -> msg.RobotIntent:
             path_command.override_angle=[self.face_angle]
 
         if(self.face_point is not None):
-            path_command.override_face_point=[Point(self.face_point[0], self.face_point[1])]
+            path_command.override_face_point=[Point(x=self.face_point[0], y=self.face_point[1])]
```
kfu02: @HussainGynai this is a fix to the move action and should probably be pulled into gameplay_node_update

rj_gameplay/rj_gameplay/skill/mark.py
```diff
@@ -23,15 +50,40 @@ class IMark(skill.ISkill, ABC):
...
-        # TODO: change so this properly returns the actions intent messages
+    def tick(self, robot: rc.Robot, world_state: rc.WorldState) -> None:
+        self.robot = robot
+
+        # update target point every tick to match movement of ball & target robot
+        if world_state and world_state.ball.visible:
+            if self.target_robot is None:
+                mark_point = get_mark_point(1, world_state)
+            else:
+                mark_point = get_mark_point(self.target_robot.id, world_state)
```
kfu02, line 81: This if statement is because I don't know how to have the nmark tactic pick enemy robots to mark, so I hardcoded an id just to test the skill.

rj_gameplay/rj_gameplay/skill/mark.py
```diff
@@ -11,9 +11,36 @@
...
 import stp.action as action
+from rj_gameplay.action import move
 from stp.skill.action_behavior import ActionBehavior
-from stp.skill.cost_behavior import CostBehavior
 import stp.rc as rc
+import stp.utils.constants as constants
+
+from rj_geometry_msgs.msg import Point, Segment
+
+def get_mark_point(target_robot_id: int, world_state: rc.WorldState):
```
kfu02, line 21: I'm assuming this is the wrong place to put this "cost function", but I don't know where the right place is.

rj_gameplay/rj_gameplay/action/move.py
kylestach: Nit: we should have a conversion helper between numpy and Point/Pose/Twist types, as we will be doing that a lot.

rj_gameplay/rj_gameplay/gameplay_node.py
```diff
@@ -9,7 +9,8 @@
 import stp
 import numpy as np
 from rj_gameplay.action.move import Move
-from rj_gameplay.play import line_up
+from rj_gameplay.play import basic122 
+# from rj_gameplay.play import line_up
```
kylestach: Dead code

rj_gameplay/rj_gameplay/gameplay_node.py
```diff
@@ -22,7 +23,8 @@ def select(self, world_state: rc.WorldState) -> None:
 
 class TestPlaySelector(situation.IPlaySelector):
     def select(self, world_state: rc.WorldState) -> Tuple[situation.ISituation, stp.play.IPlay]:
-        return (None, line_up.LineUp())
+        return (None, basic122.Basic122())
+        # return (None, line_up.LineUp())
```
kylestach: Dead code

rj_gameplay/rj_gameplay/skill/mark.py
```diff
@@ -23,15 +50,40 @@ class IMark(skill.ISkill, ABC):
...
+
+        # update target point every tick to match movement of ball & target robot
+        if world_state and world_state.ball.visible:
+            if self.target_robot is None:
+                mark_point = get_mark_point(1, world_state)
+            else:
+                mark_point = get_mark_point(self.target_robot.id, world_state)
+
+            if mark_point is None: 
+                return None
```
kylestach: If this fails out, then the outer tactic fails to assign and new roles are generated, right? cc @HussainGynai. Just trying to make sure I understand the dataflow for the new gameplay system.

rj_gameplay/rj_gameplay/tactic/nmark.py
```diff
@@ -28,20 +28,24 @@ def __call__(
         world_state: rc.WorldState,
     ) -> float:
 
-        return robot.pose[1] - world_state.ball.pos[1]
+        # TODO: make it not just robot 7 that marks
```
kylestach: Blocker on this. Please move this up to a test play or something.

rj_gameplay/stp/utils/constants.py
```diff
@@ -0,0 +1,49 @@
+# import robocup
+import stp.rc as rc
+import math
+
+DegreesToRadians = math.pi / 180.0
```
kylestach: Is this the style guide for Python constants? I thought it was `DEGREES_TO_RADIANS` (although strictly speaking anywhere you're using this you should just be doing `np.deg2rad/np.rad2deg`).

rj_gameplay/stp/skill/cost_behavior.py
```diff
@@ -35,8 +37,10 @@ def update(self) -> py_trees.common.Status:
         Creates new move action to get robot to move there
         Always returns running
         """
-        move = Move.Move(np.array([0,1+self.robot.id]))
-        move.tick(self.robot)
+        # self.move = move.Move(target_point = np.array([0,1+self.robot.id]))
+        # self.move.tick(self.robot)
+        self.move = move.Move(self.robot.id, target_point = np.array([0.0,1.0+self.robot.id]))
```
kylestach: This should only set the target point, not create a new action.

kylestach (changes requested):

rj_gameplay/stp/utils/constants.py
kfu02: Fixed, left comment for the np thing. @hdemusg please pull this file into your branch when you get the chance.

rj_gameplay/rj_gameplay/skill/mark.py
HussainGynai: I don't think it should return None. It should probably return an empty list, since skill normally return a list of actions, but this shouldn't be able to get a None point anyway I think.

rj_gameplay/rj_gameplay/skill/mark.py
```diff
@@ -11,9 +11,36 @@
...
+    # find point between ball and target robot that leaves SAG_DIST between edges of robots
+    ball_pos = world_state.ball.pos
+    opp_pos = world_state.their_robots[target_robot_id].pose[0:2]
+
+    mark_dir = (ball_pos - opp_pos) / np.linalg.norm(ball_pos - opp_pos)
+    mark_dist = mark_dir * (2.0 * constants.Robot.RADIUS + SAG_DIST)
+
+    # if ball too close to robot, can't mark successfully
+    if np.linalg.norm(mark_dist) < 2.0 * constants.Robot.RADIUS:
+        return None
```
HussainGynai: I don't think it should be None, you should probably return a point in front of the robot, like trying to block a pass or shot coming from them.

HussainGynai (changes requested):
Looks pretty good, just make change to the get_mark_point.

rj_gameplay/stp/skill/cost_behavior.py
kfu02: CostBehavior is kinda broken right now, so I didn't end up using this code. I'll revert it back to what it was originally.

rj_gameplay/rj_gameplay/skill/mark.py
HussainGynai: Its probably fine here for now, or you can create a gameplay_cost functions folder and create a sub folder for like defense or something, but not required.

rj_gameplay/rj_gameplay/skill/mark.py
HussainGynai: The nmark tactic should pick the robots to mark itself, i.e the logic for choosing the robots should be on the tactic level, not skill level.  So you can just get rid of the conditional.

HussainGynai (approved):
LGTM

rj_gameplay/rj_gameplay/skill/mark.py
kfu02: Fixed

rj_gameplay/rj_gameplay/skill/mark.py
kfu02: Okay, we can refactor later.

rj_gameplay/rj_gameplay/skill/mark.py
kfu02: When world_state is none, removing this conditional crashes sim. Reverted to original

kasohrab (approved):
Looks good to me

Yudai-8-om (approved):
LGTM!

PR #1649, opened 2021-05-26
Fix vision filter timing bugs and synchronize to vision receiver

kylestach:
## Description
Synchronize the vision receiver to the incoming packet frequency.

This essentially inverts the triggering scheme for the vision receiver: instead of updating on a set frequency with all packets from an interval, we now update whenever we get a new packet and publish at a set frequency.

## Associated Issue
ClickUp: https://app.clickup.com/t/wbagen

## Design Documents
N/A

### Emperical verification
1. Run code in manual control mode
2. Open rqt and plot robot velocity
3. Move a robot around

Expected result: velocity is smooth

Before:
image

After:
image

### DIFF
--- a/soccer/include/rj_vision_filter/camera/world.hpp
+++ b/soccer/include/rj_vision_filter/camera/world.hpp
@@ -1,12 +1,13 @@
 #include <list>
+#include <vector>
+
 #include <rj_vision_filter/ball/world_ball.hpp>
 #include <rj_vision_filter/camera/camera.hpp>
 #include <rj_vision_filter/camera/camera_frame.hpp>
-#include <rj_vision_filter/kick/kick_event.hpp>
 #include <rj_vision_filter/kick/detector/fast_kick_detector.hpp>
 #include <rj_vision_filter/kick/detector/slow_kick_detector.hpp>
+#include <rj_vision_filter/kick/kick_event.hpp>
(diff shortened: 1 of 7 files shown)


JNeiger (approved):
Those graphs look super solid! All the code looks good. I assume you make the assumption that the frames come in fast enough that you don't need to worry about predicting between frames?

kylestach:
@JNeiger yes. An outstanding todo is to deal with inter-frame predictions properly (so eventually we'll want to support variable time steps), but at least for the simulator version (which is mostly all we need to worry about for this comp) all of the packets come in pretty consistently and at the same times so it's not a big deal.

JNeiger:
> @JNeiger yes. An outstanding todo is to deal with inter-frame predictions properly (so eventually we'll want to support variable time steps), but at least for the simulator version (which is mostly all we need to worry about for this comp) all of the packets come in pretty consistently and at the same times so it's not a big deal.

Gotcha, nice job!

kfu02 (commented):
Not sure if this is within the margin for error but my curve is somewhat less smooth than your "after" pic:

vis_filter

I'm not confident enough in my Kalman filter understanding to assess your code, but if the above graph is fine than consider this an approving review.

kylestach:
@kfu02 yes, that's fine - it's expected that there might still be some minor synchronization issues.

kfu02 (approved):

PR #1650, opened 2021-05-26
Python parameters

kylestach:
## Description
Allow use of ROS params from the gameplay code.

## Associated Issue
ClickUp: https://app.clickup.com/t/wbag93

### Global parameters
1. In the gameplay node, do `import stp.global_parameters as global_parameters` and then `print(global_parameters.soccer.robot.max_kick_speed)`

Expected result: the correct value is printed

### Local parameters
1. Add a local parameter:
```python
from stp.local_parameters import Param
my_parameter = Param("my_param_name", 1.0, Param.PARAMETER_DOUBLE)
```
2. Print the local parameter with `my_parameter.value`; it should be equal to 1.0
3. Run `ros2 param list /gameplay_node`; it should display `my_param_name`
4. Run `ros2 param set /gameplay_node my_param_name 2.0`; the previous display should now instead print 2.0


rj_gameplay/rj_gameplay/gameplay_node.py
```diff
@@ -25,19 +27,23 @@ class GameplayNode(Node):
...
+
         self.world_state = world_state
         self.partial_world_state: conv.PartialWorldState = None
         self.game_info: rc.GameInfo = None
         self.field: rc.Field = None
         self.robot_statuses: List[conv.RobotStatus] = []
 
+        self.global_parameter_client = GlobalParameterClient(
+            self, '/global_parameter_server')
+        local_parameters.register_parameters(self)
```
HussainGynai, line 45: So how would like an stp or action get access to a parameter?

HussainGynai (approved):
Looks pretty good, and the tests work. Just wondering how an stp gets access.

rj_gameplay/rj_gameplay/gameplay_node.py
kfu02: I tested this on my marking code. An stp gets params the same way as gameplay_node does:
```
import stp.global_parameters as global_parameters
[...]
def tick(...)
    print(global_parameters.soccer.robot.max_kick_speed)
```

kfu02: The local params also work like the testing, e.g. in `skill/mark.py`
```
sag_dist_param = Param("mark_sag_dist", RobotConstants.RADIUS * 0.5, Param.PARAMETER_DOUBLE)
[...]
mark_pos = opp_pos + mark_dir * (2 * RobotConstants.RADIUS + sag_dist_param.value)
```
This is slightly annoying because you have to reference the value with `<param_name>.value` rather than `PARAM_NAME`, as you would with a normal constant. Unless I'm missing something @kylestach ?

kfu02 (approved):
Looks good! Am I correct in assuming we're supposed to use local params in place of constants now?

PR #1653, opened 2021-06-08
(Partially) fix python static

oswinso:
## Description
This PR fixes the broken `make pylint` step on python static introduced in #1650 by adding the `rj_gameplay/__init__.py` file.

### Test Case 1
1. Observe CircleCI

Expected result: No long breaks on the `make pylint` step. Instead it fails `mypy` because of existing typechecking errors.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -37,8 +37,8 @@ jobs:
       - checkout
       - <<: *install_deps
       - run: |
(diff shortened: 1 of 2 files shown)


kylestach (approved):
Sure

HussainGynai (approved):
smh, failed python-static

kfu02 (approved):
Works as advertised, thanks!

PR #1661, opened 2021-06-13
Wall tactic

kfu02:
## Description
Creates wall tactic, which builds a wall of n robots between the ball and the goal.

## Associated Issue
https://app.clickup.com/t/v5cpe0

### Test Case 1
1. Start & run sim
2. Move ball slowly around the field (either with slow left-click + drag or kicks under 6 m/s)
3. Kick ball toward goal

Expected result: 3 robots (ids 0-2) should move to form a wall and track the ball (relatively) smoothly. They should block the ball from being scored. They should also never enter the penalty box or touch its bounding lines.

### Test Case 2
1. Start & run sim
2. Change the number in the `init()` method of gameplay's `play/wall_ball.py` that controls how many robots are in the wall tactic
3. Repeat Test Case 1

Expected result: N robots should move now, same behavior as before.

### Known Issues
 - Moving the ball too quickly will cause the robots to scramble
 - Sim will occasionally crash, complaining that:

```
[gameplay_node-11]     self.move.face_point = self.face_point
[gameplay_node-11] AttributeError: 'Move' object has no attribute 'face_point'
```

This is not true; both move action and move skill have a face_point. Relaunch sim if this error pops up.


rj_gameplay/rj_gameplay/gameplay_node.py
```diff
@@ -2,6 +2,7 @@
 from rclpy.node import Node
```
HussainGynai: You don't want your changes for the penalty box stuff in this PR I think.

rj_gameplay/rj_gameplay/tactic/wall_tactic.py
```diff
@@ -0,0 +1,199 @@
...
+        robot: rc.Robot,
+        prev_result: Optional["RoleResult"],
+        world_state: rc.WorldState
+    ) -> float:
+
+        # TODO: make closest robots form wall, rather than setting on init
+        # aka actually use this method
+        return 0.0
+
+def find_wall_pts(num_robots: int, ball_pt: np.ndarray, goal_pt: np.ndarray, world_state: rc.WorldState):
```
HussainGynai: Can you put return types on all your functions using -> ReturnType after each function

rj_gameplay/rj_gameplay/tactic/wall_tactic.py
```diff
@@ -0,0 +1,199 @@
...
+class WallTactic(tactic.ITactic):
+
+    def __init__(self, 
+            num_robots: int, 
+            ball_pt: np.ndarray = None, 
+            goal_pt: np.ndarray = None,
+        ):
+
+        self.num_robots = num_robots
+        self.ball_pt = ball_pt
```
HussainGynai: Why do you need to keep a ball_pt, shouldn't you just get this through world state?

rj_gameplay/rj_gameplay/tactic/wall_tactic.py
```diff
@@ -0,0 +1,199 @@
...
+
+        # create move SkillEntry for every robot
+        self.move_list = [
+            tactic.SkillEntry(move.Move()) 
+            for _ in range(num_robots)
+        ]
+
+        self.wall_pts = None
+        self.cost = wall_cost()
+        self.world_state = None
```
HussainGynai: Same with world_state, why does the class need its own world state?

HussainGynai (changes requested):
You probably don't need to maintain a copy of worldstate and ball_pt

rj_gameplay/rj_gameplay/gameplay_node.py
kfu02: Yeah that's my bad, I fixed it while you were reviewing I think

rj_gameplay/rj_gameplay/tactic/wall_tactic.py
kfu02: I don't, originally I had it named something like "mark_pt" because the old wall tactic took any two arbitrary points to form a wall between. Then I text replaced it all to ball_pt bc I didn't like that. I'll fix it.

rj_gameplay/rj_gameplay/tactic/wall_tactic.py
kfu02: Good question

rj_gameplay/rj_gameplay/tactic/wall_tactic.py
```diff
@@ -0,0 +1,194 @@
...
+        return None
+
+    # assign every wall pt a robot to move there
+    for wall_pt in wall_pts:
+        robot = robot_to_wall_pt(wall_pt, world_state, assignments)
+        assignments[robot.id] = (robot, wall_pt)
+
+    return assignments
+
+def robot_to_wall_pt(wall_pt: np.ndarray, world_state: rc.WorldState, assignments: Dict[int, Tuple[rc.Robot, np.ndarray]])-> rc.Robot:
```
kfu02:
```suggestion
def robot_to_wall_pt(wall_pt: np.ndarray, world_state: rc.WorldState, assignments: Dict[int, Tuple[rc.Robot, np.ndarray]]) -> rc.Robot:
```

rj_gameplay/rj_gameplay/tactic/wall_tactic.py
```diff
@@ -0,0 +1,193 @@
...
+            min_robot = robot
+
+    return min_robot
+
+class WallTactic(tactic.ITactic):
+
+    def __init__(self, 
+            num_robots: int, 
+            ball_pt: np.ndarray = None, 
+            goal_pt: np.ndarray = None,
```
HussainGynai: You can remove these init params now

HussainGynai (approved):
LGTM, just remove the bal_pt and goal_pt in the init of wall tactic

rj_gameplay/rj_gameplay/tactic/wall_tactic.py
kfu02: Ah good catch, thanks

rj_gameplay/rj_gameplay/tactic/wall_tactic.py
```diff
@@ -0,0 +1,189 @@
...
+    """
+    def __call__(
+        self,
+        robot: rc.Robot,
+        prev_result: Optional["RoleResult"],
+        world_state: rc.WorldState
+    ) -> float:
+
+        # TODO: make closest robots form wall, rather than setting on init
+        # aka actually use this method
```
kfu02:
```suggestion
        # aka actually use this method
        # can do this by creating mult cost instances, adding attr like "wall_pt" to this class, 
        # and giving each roleRequest the correct cost in tactic below
```

Yudai-8-om (approved):
LGTM!

rj_gameplay/rj_gameplay/tactic/wall_tactic.py
```diff
@@ -0,0 +1,191 @@
...
+
+    # dist is slightly greater than penalty box bounds
+    box_w = world_state.field.penalty_long_dist_m
+    box_h = world_state.field.penalty_short_dist_m
+    line_w = world_state.field.line_width_m
+    DIST_FROM_DEF = RobotConstants.RADIUS + line_w + np.hypot(box_w/2, box_h)
+
+    # check if vision is up and running
+    # (if it is these points should not be equal)
+    if (ball_pt == goal_pt).all():
```
kylestach: This doesn't really mean vision is up and running...I would just return None if you don't have `ball.visible`.

rj_gameplay/rj_gameplay/tactic/wall_tactic.py
```diff
@@ -0,0 +1,191 @@
...
+        # can do this by creating mult cost instances, adding attr like "wall_pt" to this class, 
+        # and giving each roleRequest the correct cost in tactic below
+        return 0.0
+
+def find_wall_pts(num_robots: int, ball_pt: np.ndarray, goal_pt: np.ndarray, world_state: rc.WorldState) -> List[np.ndarray]:
+    """Calculates num_robots points to form a wall between the ball and goal.
+    :return list of wall_pts (as numpy arrays)
+    """
+    # TODO: param server this const
+    WALL_SPACING = BallConstants.RADIUS * 1.9
```
kylestach: Could we make this a little smaller? Also, fix the TODO please. This can be a parameter now.

rj_gameplay/rj_gameplay/tactic/wall_tactic.py
```diff
@@ -0,0 +1,191 @@
...
+        return None
+
+    # assign every wall pt a robot to move there
+    for wall_pt in wall_pts:
+        robot = robot_to_wall_pt(wall_pt, world_state, assignments)
+        assignments[robot.id] = (robot, wall_pt)
+
+    return assignments
+
+def robot_to_wall_pt(wall_pt: np.ndarray, world_state: rc.WorldState, assignments: Dict[int, Tuple[rc.Robot, np.ndarray]]) -> rc.Robot:
```
kylestach: This should really be decided by role assignment, you should just produce a cost for each robot, right?

kylestach (changes requested):

rj_gameplay/rj_gameplay/tactic/wall_tactic.py
kfu02: will fix

rj_gameplay/rj_gameplay/tactic/wall_tactic.py
kfu02: #1666

rj_gameplay/rj_gameplay/tactic/wall_tactic.py
```diff
@@ -20,108 +20,72 @@
...
     ) -> float:
 
-        # TODO: make closest robots form wall, rather than setting on init
-        # aka actually use this method
-        # can do this by creating mult cost instances, adding attr like "wall_pt" to this class, 
-        # and giving each roleRequest the correct cost in tactic below
-        return 0.0
+        if robot is None or self.wall_pt is None:
+            return 0
+        return np.linalg.norm(robot.pose[0:2]-self.wall_pt)
```
kylestach: Nit: units for cost functions should be seconds, not distance. Divide by max robot velocity from global constants.

kylestach (approved):
One minor nit

rj_gameplay/rj_gameplay/tactic/wall_tactic.py
kfu02: Good to know

kfu02:
'fix invisible robot assignment' fixed the flaky wall assignment issues, as some robots that weren't visible were being assigned to be part of the wall

rj_gameplay/rj_gameplay/tactic/wall_tactic.py
```diff
@@ -0,0 +1,154 @@
...
+                self.move_list[i].skill.face_point = world_state.ball.pos
+                self.cost_list[i].wall_pt = wall_pts[i]
+
+        # create RoleRequest for each SkillEntry
+        role_requests = {
+            self.move_list[i]: [role.RoleRequest(role.Priority.HIGH, False, self.cost_list[i])]
+            for i in range(self.num_wallers)
+        }
+
+        for se, rr in role_requests.items():
```
kfu02, line 128: take out these debug prints

rj_gameplay/rj_gameplay/tactic/wall_tactic.py
```diff
@@ -32,7 +33,12 @@ def __call__(
 
         if robot is None or self.wall_pt is None:
             return 0
-        return np.linalg.norm(robot.pose[0:2]-self.wall_pt)
+
```
kylestach, line 36:
```suggestion

        # TODO(#1669): Remove this once role assignment no longer assigns non-visible robots
```

rj_gameplay/rj_gameplay/tactic/wall_tactic.py
kfu02: done in different PR, ignore for now

HussainGynai (approved):
LGTM

PR #1644, opened 2021-05-24
New gameplay passing

HussainGynai:
## Description
Everything needed for passing will go here (pivoting, pivot kicking, receiving, passing).
Current things in:
- Pivot action with test play
- Pivot kick skill with test play
- Receive Action
- Passing tactic with test play
With the pivot kick skill, I had to create our own separate sequence block for behavior trees, since we need to pass actions/intent back up the tree. When an action server is implemented we will no longer need this, since the action can just make a direct call to the server to start behavior, and won't need to pass any data back up.

### Test Case 1
1. Run sim
2. robot 1 should capture the ball, spin to face robot 7, kick the ball at 7, 7 settles the ball.

PR #1662, opened 2021-06-14
Pass seeking cost function

Yudai-8-om:
## Description
Add cost if:
- Opponents block the way (+999)
- Recipient is surrounded by opponents
- We pass toward downfield

Weight and normalization should be tweaked.


kfu02:
I'm going to hold off on reviewing this until #1644 is merged and you can actually test this cost function. Should work out of the box with pass_tactic.

rj_gameplay/stp/passing_cost.py
```diff
@@ -0,0 +1,54 @@
...
+		punish_amplifier = 1
+
+		our_robots = world_state.our_robots
+		their_robots = world_state.their_robots
+		recipient = robot
+		kicker = [robot for robot in our_robots if robot.has_ball_sense]
+		try:
+			kicker = kicker[0]
+		except:
+			print("Our team is not possessing the ball.")
```
HussainGynai: you should probably return a float('inf') if we don't posses the ball (this is a float which acts as infinity).

kfu02: For the wall_tactic, I got a ValueError when returning float('inf') for invisible robots. @Yudai-8-om you might have to change this back to `return 9999` or something similar. Once you can actually test this cost func this should be more obvious.

HussainGynai (approved):
LGTM

PR #1677, opened 2021-06-20
New Play selection

HussainGynai:
## Description
A basic new play selector, currently only works for one play per situation (which is probably all we'll have for this comp). Also created periodic publisher for gamestate.

### Test Case 1
1. Run sim
2. Watch pas play happen
3. Edit PLAY_DICT in the play selector file with whatever plays you want, and check it works properly


kfu02 (approved):
Seems good, hard for me to test play selection since I don't know what the situations are intuitively

kfu02 (approved):
With my slightly better intuition (thanks @HussainGynai) I've run through some scenarios manually and the play selector seems to be picking the right one based on the DICT and executing it. 

TLDR: LGTM

PR #1679, opened 2021-06-20
clear play created

HussainGynai:
## Description
Ne clear play, also fixed issue with pivot kick is_done()

### Test Case 1
1. run the sim

Expected result: The robot closest to the ball should chip the ball up the field

### DIFF
--- a/rj_gameplay/rj_gameplay/play/defensive_clear.py
+++ b/rj_gameplay/rj_gameplay/play/defensive_clear.py
@@ -13,6 +13,7 @@ class DefensiveClear(play.IPlay):
 
     def __init__(self):
         #TODO: Make these markers into seekers
+        # TODO: add goalie
         self.two_mark = nmark_tactic.NMarkTactic(2)
         self.clear = clear_tactic.Clear(np.array([0.0, 10.0]))
         self.role_assigner = NaiveRoleAssignment()
--- a/rj_gameplay/rj_gameplay/play/defensive_clear.py
+++ b/rj_gameplay/rj_gameplay/play/defensive_clear.py
@@ -1,23 +1,24 @@
 import stp.play as play
 import stp.tactic as tactic
 
-from rj_gameplay.tactic import stub_striker, nmark_tactic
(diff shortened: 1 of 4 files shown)


rj_gameplay/rj_gameplay/tactic/clear_tactic.py
```diff
@@ -0,0 +1,73 @@
...
+    def __call__(self,
+                robot:rc.Robot,
+                prev_result:Optional["RoleResult"],
+                world_state:rc.WorldState) -> float:
+        if not robot.visible:
+            return 99999
+        else:
+            return np.linalg.norm(world_state.ball.pos - np.array(robot.pose[0:2]))
+
+class Clear(tactic.ITactic):
```
kfu02, line 29:
```suggestion
# TODO: discuss if clear deserves its own tactic if we can just use pivot_kick skill for it
class Clear(tactic.ITactic):
```

rj_gameplay/rj_gameplay/play/defensive_clear.py
```diff
@@ -1,23 +1,24 @@
...
 import stp.rc as rc
+import numpy as np
 from typing import Dict, Generic, Iterator, List, Optional, Tuple, Type, TypeVar
 
 class DefensiveClear(play.IPlay):
 
     def __init__(self):
-        # TODO: add chipper tactic here
-
+        #TODO: Make these markers into seekers
```
kfu02:
```suggestion
        #TODO: Make these markers into seekers
        # TODO: add goalie
```

kfu02 (changes requested):
Behaves well, left TODO suggestions

rj_gameplay/rj_gameplay/tactic/clear_tactic.py
HussainGynai: I think it does, because ideally there would be some sort of heuristic function that chooses a good place to pass to.

kfu02: ah okay

PR #1685, opened 2021-06-20
created scramble/pileup play

HussainGynai:
## Description
A very basic pileup/scramble play, just captures with 2 markers and goalie.


kfu02 (changes requested):
Goalie stops on capture's end, this should be fine since we're just throwing the goalie tactic into every play but something to keep in mind for the future

Also, since we only have one play at a time for comp, should we just throw in a wall or more markers?

PR #1689, opened 2021-06-21
created basic defend restart play that is just a goalie and 5 markers

HussainGynai:
## Description
Just 5 markers and a goalie.


kylestach:
Maybe it would be good to have two or so robots dedicated to walling?

HussainGynai:
> Maybe it would be good to have two or so robots dedicated to walling?

My thought as that they can't directly kick at goal right?

kfu02:
should the is_done method be tied to the is_done of the markers?

kfu02:
LGTM

PR #1695, opened 2021-06-21
Hotfix def restart

kfu02:
## Description
not allowed to be close to ball on defending a restart, so @kylestach recommended making a local obstacle around the ball. this PR does that in the least maintainable way.


kfu02:
transition from restart to other plays not working, merging anyways

PR #1698, opened 2021-06-21
Bug fixes maybe

kfu02:
## Description
fouls from our first match
during stop we can't do more than 1 m/s
can't chip into goal
we got an aimless kick?

Can't actually test w grSim right now so I'm just guessing here


kfu02:
outdated

PR #1706, opened 2021-06-26
made static attack play for hardware challenge

HussainGynai:
Let me sleep


kfu02:
we didn't commit gameplay node

set TestPlaySelector to `static_attack.Attack`

PR #1707, opened 2021-06-28
New CI

kylestach:
New CI using GitHub actions. This should be much nicer than the old CircleCI config (and it lets us use annotations!)


kfu02 (changes requested):
I haven't actually looked through the code in this PR. That said, here are a few thoughts I had from looking at #1708:

1) I understand the reason this check exists, but can we change pylint to allow one-word variable names (which therefore don't need an underscore)?
Screen Shot 2021-07-02 at 1 50 35 AM

2) would it be possible to add a line like "run `make pretty-lines` in robocup-software to fix" to style error msgs like this?
Screen Shot 2021-07-02 at 1 51 51 AM

3) This one is a nit, but is there a reason all three of these checks have to print a very similar error msg?
Screen Shot 2021-07-02 at 1 54 46 AM

Overall though, nice work. This is way cleaner than CircleCI.

kylestach:
1. That check generally checks for extremely short variable names (1-2 characters). I think it is reasonable to disable the "too short" check and let us complain about that manually, as usually when you have one-letter variable names it's just transcribed from math.
2. Unfortunately that would be a little bit obnoxious to implement, and if people don't know how to run the autoformatter then that's a documentation issue.
3. That's just a quirk of the fact that we have several linters checking overlapping things. It's unfortunate but it would be a massive pain to fix (there are likely many such checks shared between linters).

kfu02 (approved):
In that case, LGTM!

kasohrab (approved):
I also did not look at the code. But the new CI in #1708 looks good to me.

adithya-gv (approved):

kasohrab (approved):

PR #1709, opened 2021-07-06
Comp 2021

kylestach:
Draft PR for comp 2021 code.

This should not be merged. Instead, we'll sort through it and do an interactive add. This PR is just so we can have a WIP place to comment.


kylestach:
Closing in favor of #1719

Issue #1539, opened 2020-06-29
Fix curvature issues with Bezier path and re-enable curvature limiting


kylestach:
## Description
Currently, our path smoothing code places Bezier waypoints at the endpoints when you have zero initial/final velocity. This creates numerical conditioning issues, which make curvature infinite near the endpoints in these case.

In general, we should switch to a path smoothing algorithm that explicitly considers curvature (or some related measure). For example, it should be pretty simple to use Hermite curves and place "velocity" (d/ds) values such that the sum of squared "acceleration" (d^2/ds^2) at curve endpoints is minimized.

This is currently reflected in BezierPathTest.cpp, in the `DISABLED` tests.

## Requirements
 - `DISABLED` tests from the Bezier path test are re-enabled and passing.
 - In velocity profiling code, curvature limiting is re-enabled (`VelocityProfiling.cpp`)

kylestach:
Fixed in comp 2021 (#1709)

PR #1719, opened 2021-07-10
Comp2021 merge

kylestach:
## Description
Merge all code from RoboCup 2021 competition.

A lot of this still needs to be cleaned up (most of gameplay will be rewritten soon).

List of changes:
 - Infrastructure
    - Docker image for competition
    - Upgrade googletest version
 - General changes
    - Add gameplay debug text to UI
    - **Referee rewrite,** including breaking out GameState -> PlayState + MatchState
    - Allow vision, sim radio, and ref interface to specify a particular network interface to use
    - Tick the radio faster (to receive return packets)
    - Fix drawing polygons (rectangles and alpha in fill were not working)
 - Planning
    - Line kick planner changes (WIP, hacks)
    - Pivot planner: cache target and pivot points and recalculate when necessary
    - Fix derivative calculation issue in Bezier path and re-enable curvature limiting (#1539)
    - Don't spin in place when the commanded velocity is small
    - Don't crash in replanner when doing a partial replan where the trajectories cannot be spliced
    - Reduce redundant subscriptions in planner node
    - Draw obstacles
    - Make everything move a lot faster (motion is a lot tighter with curvature limiting re-enabled)
 - Gameplay: there were lots of changes given that the gameplay system was basically incomplete heading into comp
    - New skills, tactics, plays (nearly all will be rewritten)
    - Added actions even though we are planning to combine those w/ skills
    - Add obstacles for goals/goal zones, and the ball during stop
    - Rules compliance in restarts, ball placement, stop state, penalties, etc.
    - New situations for some restarts we didn't catch
    - Situation analysis (preliminary implementation)


kfu02 (approved):
Made some light additions to the gameplay section of the PR description. Also, a note for posterity:

>    Allow vision, sim radio, and ref interface to specify a particular network interface to use

That would be in `rj_vision_receiver/src/vision_receiver.cpp`, `soccer/src/soccer/radio/sim_radio.cpp`, and `soccer/src/soccer/referee/external_referee.cpp`, respectively. 

Also, did we settle on setting external ref as default? If so, this link is necessary to use that external ref interface from comp: https://github.com/RoboCup-SSL/ssl-game-controller

kfu02:
Note: we need to run `make pretty-lines` on this branch before merging.

kasohrab (approved):
Approved cause it compiles and runs.

kfu02 (approved):
It build

PR #1723, opened 2021-07-31
[DRAFT] Goalie utility function

kasohrab:
## Description
Goalie tactic with utility functions for more dynamic behavior.
## Associated Issue
Issue #???

## Design Documents
(on clickup)

### Test Case 1
1. Make
2. Run sim
3. ???

Expected result: nothing new yet.

## Sub-issues (for drafts)
- [x] Mockup 
- [ ] Working test
- [ ] Better goalie behavior

### DIFF
--- a/rj_gameplay/rj_gameplay/tactic/goalie_tactic.py
+++ b/rj_gameplay/rj_gameplay/tactic/goalie_tactic.py
@@ -104,7 +104,6 @@ class GoalieTactic(tactic.ITactic):
         """
         :return: A list of role requests for move skills needed
         """
-
         def track_utility() -> float:
             """
             :return: the utility of track action
@@ -122,8 +121,8 @@ class GoalieTactic(tactic.ITactic):
             # this probably makes sense to stay hard coded like this
             # but maybe there is a better option?
             if ball_speed < .5 and (
-                abs(ball_pos[0]) < box_w / 2 + line_w + MAX_OOB
-                and ball_pos[1] < box_h + line_w + MAX_OOB
+                    abs(ball_pos[0]) < box_w / 2 + line_w + MAX_OOB
+                    and ball_pos[1] < box_h + line_w + MAX_OOB
             ) and not world_state.game_info.is_stopped():
                 return 1
             else:
@@ -192,8 +191,7 @@ class GoalieTactic(tactic.ITactic):
 
         ball_speed = np.linalg.norm(world_state.ball.vel)
         ball_pos = world_state.ball.pos
-        ball_dist = np.linalg.norm(world_state.field.our_goal_loc -
-                                   ball_pos)
+        ball_dist = np.linalg.norm(world_state.field.our_goal_loc - ball_pos)
         goal_pos = world_state.field.our_goal_loc
         towards_goal = goal_pos - ball_pos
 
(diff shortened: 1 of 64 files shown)


rj_gameplay/rj_gameplay/tactic/goalie_tactic.py
```diff
@@ -103,7 +105,82 @@ def get_requests(self, world_state: rc.WorldState,
         :return: A list of role requests for move skills needed
         """
 
+        def track_utility() -> float:
```
kasohrab, line 107: The new stuff starts around here. My questions are in python comments

rj_gameplay/rj_gameplay/tactic/goalie_tactic.py
```diff
@@ -103,7 +105,82 @@ def get_requests(self, world_state: rc.WorldState,
...
+                ball_to_goal_time = 100
+            return 1 - (ball_to_goal_time / 100)
+
+        def clear_utility() -> float:
+            """
+            :return: the utility of clear action
+            """
+            # should clear if no open pass available
+            # how do I check that?
+            return 0
```
kfu02: Our passing isn't working right now, leave it as is for now.

kfu02: Actually, set the value to 1 for "always clear" for now until we get passing.
```suggestion
            return 1
```

rj_gameplay/rj_gameplay/tactic/goalie_tactic.py
```diff
@@ -113,61 +190,80 @@ def get_requests(self, world_state: rc.WorldState,
...
-            ball_speed = np.linalg.norm(world_state.ball.vel)
-            ball_pos = world_state.ball.pos
-            ball_dist = np.linalg.norm(world_state.field.our_goal_loc - ball_pos)
-            goal_pos = world_state.field.our_goal_loc
-            towards_goal = goal_pos - ball_pos
+        best_utility = utility_functions()
+
+        if best_utility is not None:
 
+            # what is this?
```
kfu02: oh dear someone else found it

This is a hack I put in during comp2021 to make sure our goalie didn't try to intercept the ball in stop/penalty situations. (All robots are supposed to stop moving when a foul is called or when a penalty kick is being set up.)

rj_gameplay/rj_gameplay/tactic/goalie_tactic.py
```diff
@@ -103,7 +105,82 @@ def get_requests(self, world_state: rc.WorldState,
...
+            """
+            :return: the utility of clear action
+            """
+            # should pass if open pass available
+            # same problem as clear
+            return 0
+
+        def utility_functions():
+            """Compares utility functions of all actions.
+            :return: The best action based on the utility functions.
```
kfu02, line 161: Change the name of this function to something more descriptive. Maybe "get_best_action()"?

rj_gameplay/rj_gameplay/tactic/goalie_tactic.py
kfu02:
```suggestion
            # TODO (1724) remove this from the code
```
https://github.com/RoboJackets/robocup-software/issues/1724

rj_gameplay/rj_gameplay/tactic/goalie_tactic.py
```diff
@@ -113,61 +190,80 @@ def get_requests(self, world_state: rc.WorldState,
...
                     self.move_se.skill.face_point = world_state.ball.pos
                     role_requests[self.move_se] = [
                         role.RoleRequest(role.Priority.HIGH, True,
                                          self.role_cost)
                     ]
+                elif best_utility is pass_utility:
+                    # TODO : under current system, should assign pass tactic here
+                    pass
+
+        # what is this? and should I keep it?
```
kfu02: This is some more hacky stuff from comp2021. Keep it for now, but try removing it in testing and seeing what happens.

rj_gameplay/rj_gameplay/tactic/goalie_tactic.py
```diff
@@ -103,7 +105,82 @@ def get_requests(self, world_state: rc.WorldState,
...
+                ball_to_goal_time = ball_dist / ball_speed
+            return ball_to_goal_time / 100
+
+        def receive_utility() -> float:
+            """
+            :return: the utility of receive action
+            """
+            # this probably makes sense to stay hard coded like this
+            # but maybe there is a better option?
+            if ball_speed < .5 and (
```
kfu02, line 123: None of the utility() functions have params passed in, it would be better to pass them in explicitly. And for now hardcoded is fine, we can tune later.

kfu02 (commented):
Left some comments. You're on the right track, but currently it feels like you've just moved the logic of the original if-else branches into functions, which leaves some implicit in that section. The final should be more like:
```
if receive:
    # do receive skill
elif block:
    # do block skill
elif ...
```
where the interaction of utility scores ensures that the right action is performed given the situation.

rj_gameplay/rj_gameplay/tactic/goalie_tactic.py
kasohrab: Since it's an inner function, it can access those params (I think). So should I make all of these functions separate from get_requests?

kfu02: Ahh, I didn't notice that. My bad. It's fine as is then.

PR #1725, opened 2021-08-04
Add world_state to tick in tactics

kasohrab:
## Description
Title

## Associated Issue
Issue #???

## Design Documents
on clickup

### Test Case 1
0. Revert fully to grSim
1. Make
2. Run sim
3. Try different plays and run sim again

Expected result: Should still work

### DIFF
--- a/rj_gameplay/stp/tactic/__init__.py
+++ b/rj_gameplay/stp/tactic/__init__.py
@@ -147,8 +147,10 @@ class ITactic(Generic[PropT], ABC):
         return [self.create_request(**kwargs) for _ in range(num_requests)]
 
     @abstractmethod
-    def tick(self, role_results: RoleResults, props: PropT) -> List[skill.ISkill]:
+    def tick(self, world_state: rc.WorldState, role_results: RoleResults,
+             props: PropT) -> List[skill.ISkill]:
         """Ticks the tactic, returning a tuple of the skills and the skills executed.
+        :param world_state: Current world state.
         :param role_results: The results of role assignment.
         :param props: The state of the current tactic.
         :return: A list of skills to be executed.
--- a/config/sim.yaml
+++ b/config/sim.yaml
@@ -101,7 +101,8 @@
 
 /vision_receiver:
   ros__parameters:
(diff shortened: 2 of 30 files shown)


rj_gameplay/stp/tactic/__init__.py
```diff
@@ -147,8 +147,10 @@ def create_requests(self, num_requests: int, **kwargs) -> List[role.RoleRequest]
         return [self.create_request(**kwargs) for _ in range(num_requests)]
 
     @abstractmethod
-    def tick(self, role_results: RoleResults, props: PropT) -> List[skill.ISkill]:
+    def tick(self, world_state: rc.WorldState, role_results: RoleResults,
+             props: PropT) -> List[skill.ISkill]:
```
kasohrab, line 151: Just so I understand, how is props different than world_state? And how come none of the implementations of this interface take in props?

kfu02: The infrastructure of our gameplay system was designed by @oswinso and built out by Kyle, Hussain, Yudai and I. This is an example of an issue stemming from that: "props" is short for properties, which I think was intended to be a way for us to pass data from the previous tick of a tactic to the next, rather than recomputing everything every tick from world_state. We currently don't use props (see any tactic's [empty compute_props implementation](https://github.com/RoboJackets/robocup-software/blob/comp2021-merge/rj_gameplay/rj_gameplay/tactic/wall_tactic.py#L114-L115)).

kfu02: It might be worth removing the method entirely from the interface and making a note in GH issues that this is the way we can pass state around. For now, it seems to be overengineering.

kfu02 (approved):
Nice work, might have to remove compute_props(), we can discuss next Tues.

adithya-gv (approved):
Looks good to me!

Yudai-8-om (approved):
LGTM!

anshulm07 (approved):
Looks good to me!

RyanBard7 (approved):

PR #1728, opened 2021-08-10
add prototype

kfu02:
## Description
Draft PR to show possible new gameplay structure. (DO NOT MERGE.)


kfu02:
As discussed in 8/10 meeting: keeping role assignment is beneficial in ways that our current codebase doesn't properly show. Switching tactics to be single-robot behaviors but dropping the rest of these proposed changes.

PR #1735, opened 2021-09-28
add field setup doc

kfu02:
## Description
Add setup doc for using the field computer in the SCC.

## Steps to test
1. Ask newish software member to test, get feedback
2. Ask non-software member to test, get feedback

Expected result: Both understand and are able to launch the necessary programs.


Yudai-8-om (approved):
LGTM

RyanBard7 (approved):

kfu02:
note: field setup doc is correct, but issue with field computer currently means that external ref is not linking correctly. further investigation needed

PR #1740, opened 2021-10-20
Fix CI

kasohrab:
## Description
CI fails in some places. Should be fixed to make CI helpful.
Updates googletest and sets pip version (a better way would probably be to update the cmake instead but this works so)
only works on the action server right nowcbut not now on ros2. So i think we can just disable ci (as we are just using github actions now)


kfu02 (approved):
specifically, googletest external was updated to origin/master instead of the specific commit we had it checked out to, and pip was downgraded so it still had the --build flag

kfu02:
still need to fix the line too long and typing errors -Alex

Yudai-8-om (approved):

kfu02:
note: rerun `./util/ubuntu-setup` if you run into build errors now

PR #1742, opened 2021-11-02
Two robots passing

Yudai-8-om:
## Description
basic passing

##Test
1. Run sim
2. Spread robots
3. It will pass to the best robot (based on the preliminary cost function)

### DIFF
--- a/rj_gameplay/rj_gameplay/skill/pivot.py
+++ b/rj_gameplay/rj_gameplay/skill/pivot.py
@@ -50,7 +50,6 @@ class Pivot():
         intent.trigger_mode = intent.TRIGGER_MODE_STAND_DOWN
         intent.dribbler_speed = float(self.dribble_speed)
         intent.is_active = True
-        print(self.target_point)
         return {self.robot.id: intent}
 
 
--- a/rj_gameplay/rj_gameplay/skill/pivot.py
+++ b/rj_gameplay/rj_gameplay/skill/pivot.py
@@ -26,7 +26,7 @@ class Pivot():
                  pivot_point: np.ndarray=None,
                  target_point: np.ndarray=None,
                  dribble_speed: float = 1.0,
-                 threshold: float = 0.02,
+                 threshold: float = 0.05,
                  priority: int = 1):
         self.robot = robot
         self.pivot_point = pivot_point
@@ -42,8 +42,7 @@ class Pivot():
              intent: RobotIntent):
         self.robot = robot
         self.pivot_point = world_state.ball.pos[0:2]
-        self.target_point = pass_tactic.find_potential_target(world_state).pose[0:2]
-        print(self.target_point)
+        self.target_point = pass_tactic.Pass().find_potential_receiver(world_state).pose[0:2]
 
         pivot_command = PivotMotionCommand()
         pivot_command.pivot_point = Point(x=self.pivot_point[0], y=self.pivot_point[1])
@@ -67,7 +66,7 @@ class Pivot():
         heading_vect = np.array([np.cos(robot.pose[2]), np.sin(robot.pose[2])])
         dot_product = np.dot(heading_vect, robot_to_target_unit)
         angle = np.arccos(dot_product)
-        if (angle < angle_threshold) and (abs(world_state.our_robots[self.robot_id].twist[2]) < stopped_threshold):
+        if (angle < angle_threshold) and (abs(world_state.our_robots[self.robot.id].twist[2]) < stopped_threshold):
             return True
         else:
             return False
--- a/rj_gameplay/rj_gameplay/skill/pivot.py
+++ b/rj_gameplay/rj_gameplay/skill/pivot.py
@@ -10,28 +10,23 @@ import stp.skill as skill
 import stp.role as role
 from rj_msgs.msg import RobotIntent, PivotMotionCommand
 from rj_geometry_msgs.msg import Point
-from rj_gameplay.tactic import pass_tactic
 import stp.rc as rc
 import numpy as np
 
 
 
 class Pivot():
-    """
-    Pivot skill that robot aims at the receiver or the goal
-    """
-
     def __init__(self,
                  robot: rc.Robot = None,
-                 pivot_point: np.ndarray = None,
-                 target_point: np.ndarray = None,
-                 dribble_speed: float = 1.0,
+                 pivot_point: np.ndarray = np.array([0.0, 0.0]),
+                 target_point: np.ndarray = np.array([0.0, 0.0]),
+                 dribbler_speed: float = 1.0,
                  threshold: float = 0.05,
                  priority: int = 1):
         self.robot = robot
         self.pivot_point = pivot_point
         self.target_point = target_point
-        self.dribble_speed = dribble_speed
+        self.dribbler_speed = dribbler_speed
         self.threshold = threshold
 
         self.__name__ = 'pivot skill'
@@ -39,16 +34,13 @@ class Pivot():
     def tick(self, robot: rc.Robot, world_state: rc.WorldState,
              intent: RobotIntent):
         self.robot = robot
-        self.pivot_point = world_state.ball.pos[0:2]
-        self.target_point = pass_tactic.Pass().find_potential_receiver(
-            world_state).pose[0:2]
 
         pivot_command = PivotMotionCommand()
         pivot_command.pivot_point = Point(x=self.pivot_point[0], y=self.pivot_point[1])
         pivot_command.pivot_target = Point(x=self.target_point[0], y=self.target_point[1])
         intent.motion_command.pivot_command = [pivot_command]
         intent.trigger_mode = intent.TRIGGER_MODE_STAND_DOWN
-        intent.dribbler_speed = float(self.dribble_speed)
+        intent.dribbler_speed = float(self.dribbler_speed)
         intent.is_active = True
         return {self.robot.id: intent}
 
--- a/rj_gameplay/rj_gameplay/skill/pivot.py
+++ b/rj_gameplay/rj_gameplay/skill/pivot.py
@@ -50,6 +50,7 @@ class Pivot():
         intent.trigger_mode = intent.TRIGGER_MODE_STAND_DOWN
         intent.dribbler_speed = float(self.dribble_speed)
         intent.is_active = True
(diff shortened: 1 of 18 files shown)


rj_gameplay/rj_gameplay/skill/pivot.py
```diff
@@ -33,19 +36,19 @@ def __init__(self,
...
+        self.target_point = pass_tactic.Pass().find_potential_receiver(
+            world_state).pose[0:2]
 
         pivot_command = PivotMotionCommand()
         pivot_command.pivot_point = Point(x=self.pivot_point[0], y=self.pivot_point[1])
         pivot_command.pivot_target = Point(x=self.target_point[0], y=self.target_point[1])
         intent.motion_command.pivot_command = [pivot_command]
         intent.trigger_mode = intent.TRIGGER_MODE_STAND_DOWN
-        intent.dribbler_speed = self.dribble_speed
+        intent.dribbler_speed = float(self.dribble_speed)
```
kfu02: is the cast necessary? dribble_speed is already passed in as a float.

rj_gameplay/rj_gameplay/skill/pivot_kick.py
```diff
@@ -47,19 +47,17 @@ def __init__(self,
         self.capture = capture.Capture(robot)
 
 
-    def tick(self, 
-             robot: rc.Robot, 
-             world_state: rc.WorldState, 
+    def tick(self, robot: rc.Robot, world_state: rc.WorldState,
```
kfu02, line 50: add a comment here that describes the flow (which I believe from reading goes capture -> pivot -> kick?) since the if-else order doesn't match the sequence of actions

rj_gameplay/rj_gameplay/skill/receive.py
```diff
@@ -24,26 +24,23 @@
...
 
-    def tick(self, 
-             robot: rc.Robot, 
-             world_state: rc.WorldState, 
+    def tick(self, robot: rc.Robot, world_state: rc.WorldState,
              intent: RobotIntent):
         if self.settle.is_done(world_state):
-            return self.capture.tick(robot, world_state)
+            return self.capture.tick(robot, world_state, intent)
+        # elif np.linalg.norm(world_state.ball.vel) > RECEIVE_BALL_SPEED_THRESHOLD:
```
kfu02: is this intentionally commented out? if it's not being used I would advise deleting the THRESHOLD param

rj_gameplay/rj_gameplay/skill/settle.py
```diff
@@ -5,48 +5,66 @@
...
-
+        print(f'Passer: {self.passer.id}')
         return {self.robot.id: intent}
 
 
     def is_done(self, world_state) -> bool:
         if self.robot is None:
             return False
-        if world_state.our_robots[self.robot.id].has_ball_sense or np.linalg.norm(world_state.ball.vel) < SETTLE_BALL_SPEED_THRESHOLD:
+        if world_state.our_robots[self.robot.id].has_ball_sense or (
```
kfu02: could you put a comment explaining what this if statement checks? it looks like "if ball sense OR (ball closer to me than passer AND ball is slow)" but that took some time to figure out

rj_gameplay/stp/role/assignment/naive.py
```diff
@@ -109,11 +112,9 @@ def compute_costs_matrix(
 
             # Throw an exception if the returned cost is not finite.
             if not isfinite(unassigned_cost):
-                    raise ValueError(
-                        "Got a non-finite cost ({}) for request {} and unassinged robot".format(
-                            cost, request, robot
-                        )
-                    )
+                raise ValueError(
+                    "Got a non-finite cost ({}) for request {} and unassinged robot"
```
kfu02:
```suggestion
                    "Got a non-finite cost ({}) for request {} and unassigned robot"
```

rj_gameplay/stp/role/assignment/naive.py
```diff
@@ -71,28 +74,28 @@ def compute_costs_matrix(
             for robot_idx, robot in enumerate(free_robots):
 
                 # If the robot is not visible then do not consider its cost
+                # Cannot be infinte, will break linear_sum_assignment
```
kfu02:
```suggestion
                # Using 1e9 because float('inf') will break linear_sum_assignment
```

rj_gameplay/rj_gameplay/tactic/pass_tactic.py
```diff
@@ -106,44 +106,98 @@ def __call__(
     ) -> float:
 
         if robot is None or self.target_point is None:
-            return 99
+            return 1e9
```
kfu02, line 108: it would be nice if we made role assignment auto-cast float('inf') to 1e9 in the role assigner so we didn't have to remember this magic number

kfu02: added as new task to board, dw about it

rj_gameplay/rj_gameplay/tactic/pass_tactic.py
```diff
@@ -181,14 +234,25 @@ def create_request(self, **kwargs) -> role.RoleRequest:
         pass
 
     def find_potential_receiver(self, world_state: rc.WorldState) -> rc.Robot:
```
kfu02: any reason this one is named "find_potential_receiver" and the one below is "find_passer"? (why "potential"?)

kfu02: also, this doesn't have to be fixed for this PR, but these funcs are so similar that they could be one "find min" function that takes in a cost func as a param: https://www.geeksforgeeks.org/passing-function-as-an-argument-in-python/

rj_gameplay/rj_gameplay/tactic/pass_tactic.py
```diff
@@ -106,44 +106,98 @@ def __call__(
...
+            return 1e9
+        # if not robot.visible:
+        #     return 1e9
+        if self.passer_robot is not None and robot.id == self.passer_robot.id:
+            # can't pass to yourself
+            return 1e9
+        angle_threshold = 5
+        dist_threshold = 1.5
+        backpass_punish_weight = 0.5
+        if robot.id != self.passer_robot.id:
```
kfu02: isn't this if already covered in line 164 above?

rj_gameplay/rj_gameplay/tactic/pass_tactic.py
```diff
@@ -106,44 +106,98 @@ def __call__(
...
+            pass_dist = np.linalg.norm(self.passer_robot.pose[0:2] -
+                                       robot.pose[0:2])
+
+            cost = 0
+            goal_to_receiver = np.linalg.norm(robot.pose[0:2] -
```
kfu02: add a few comments explaining chunks of below code, e.g. here I would put 
"add to cost for backwards passes (scaled by backpass_punish_weight)"

rj_gameplay/rj_gameplay/tactic/pass_tactic.py
```diff
@@ -106,44 +106,98 @@ def __call__(
...
         if self.passer_robot is not None and robot.id == self.passer_robot.id:
             # can't pass to yourself
-            return 99
-        if self.chosen_receiver is not None and self.chosen_receiver.id == robot.id:
-            return -99
+            return 1e9
+        # if self.chosen_receiver is not None and self.chosen_receiver.id == robot.id:
+        #     return -1e9
 
         # TODO: pick "most open" pass
```
kfu02: this TODO is done, right?

rj_gameplay/rj_gameplay/tactic/pass_tactic.py
```diff
@@ -222,18 +283,27 @@ def tick(self, world_state: rc.WorldState,
...
-            self.pivot_kick.skill.target_point = np.array(
-                [potential_receiver.pose[0], potential_receiver.pose[1]])
+        if pivot_result and pivot_result[0].is_filled():
+            self.receiver_cost.passer_robot = pivot_result[0].role.robot
+            # self.pivot_kick.skill.target_point = np.array(receive_result[0].role.robot.pose[0:2])
+            self.pivot_kick.skill.target_point = self.find_potential_receiver(world_state).pose[0:2]
+            self.pivot_kick.skill.pivot_point = np.array(
+                pivot_result[0].role.robot.pose[0:2])
+
+            # self.receiver_cost.potential_receiver = receive_result[
```
kfu02: delete this commented out code if it's not being used, or leave comment explaining why it should stay

kfu02 (changes requested):
This is great, way better than what we've had so far. A few things:
 - when I put the ball in the top left and hit start, sim crashes often, giving this error: "[planner_node-7] terminate called after throwing an instance of 'std::invalid_argument'
[planner_node-7]   what():  Cannot splice trajectories a and b, where a.last() != b.first()" No idea what this means. I'm fine with solving this in a later PR.
 - when the pass doesn't crash sim (i.e. ball starts in top right) the receiver often passes it right back to the original passer. This issue should be fixed for this PR. My hypothesis is that the pass tactic doesn't return done because neither robot has captured the ball yet, so it reassigns the new closest robot to the ball as the passer and the old passer as the new receiver and does the pass again.
 - I left a few comments in the files. Most of them are just "please add explanatory comment here". Will make future dev easier.
Nice work, @Yudai-8-om @adithya-gv and @HussainGynai

rj_gameplay/rj_gameplay/skill/pivot.py
Yudai-8-om: I know, it is weird, but I got an error without float for some reason.

kfu02:
I think this is a pretty perfect example as to how our gameplay system is unnecessarily over-complicated. See my last commit msg:  
```
implement basic passing

 - buggy bc doesn't end after one pass (aka Receive never ends)
 - implemented as behavior tree and by circumventing RoleAssignment:
 - passer/receiver are picked on first get_requests() call
 - role costs are set to always pick same passer/receiver
 - first, passer is requested
 - when passer finishes PivotKick, receiver is requested
 - receiver then goes to ball (in theory, poorly implemented rn)
 - when receiver gets ball, Tactic should end (again in theory)
```

PR #1797, opened 2021-11-21
redo feat/speedup-compile

kasohrab:
## Description
Redo #1620 because when i tested it it didn't work (and I modified/fixed some additional things so that it should work now).

### Test Case 1
0. ./util/ubuntu-setup
1. make clean
2. make perf

only when switching to a new branch is make clean needed.

Expected result: should be faster
It was definitely faster on my machine, there appears to be a speedup on ci as well.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -7,11 +7,13 @@ project(rj_robocup LANGUAGES CXX)
 # ======================================================================
 # C++ Version and Compiler Flags
 # ======================================================================
-
 set(CMAKE_C_STANDARD 99)
 set(CMAKE_CXX_STANDARD 17)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 
+# ======================================================================
+# Set to use PThreads  
+# ======================================================================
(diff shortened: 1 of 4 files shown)


kfu02 (approved):
Ridiculous speedup when using `make clean` and `make perf` as opposed to the old `rm -rf build*` and `make perf`. 

However, I noticed that `make perf` produces the `build-release-debug/` folder, which `make clean` doesn't seem to take care of? I don't really know how CMake works so I'm not sure if that's an issue. Tagging @kylestach for review since he wrote the original speedup PR referenced above.

kylestach:
Looks good. If you want it to get rid of the `cmake-build-*` directories, replace the `ninja clean` lines in the makefile with an `rm -rf`.

kylestach (approved):

RyanBard7 (approved):

Yudai-8-om (approved):

adithya-gv (approved):
It definetly makes a lot faster. Looks good to me!

PR #1798, opened 2021-12-02
fix description of STP system

kfu02:
## Description
Updates README for rj_gameplay to bring it in line with our finalized STP system.

## Design Documents
[Hussain's Gameplay Doc](https://docs.google.com/document/d/1pf-aaHnvonTPC6ccI7RdHudpuev84fZslMVDq8lw59s/edit?usp=sharing)

## Steps to Review

See if explanations provided match Gameplay Doc above.

### DIFF
--- a/rj_gameplay/README.md
+++ b/rj_gameplay/README.md
@@ -11,7 +11,7 @@ Stable code for gameplay.
 
 **Implementation - `rj_gameplay/`**
 
-Frequently changing code of gameplay .
+Frequently changing code of gameplay.
 
 **Tests - `tests/stp`**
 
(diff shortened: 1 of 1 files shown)


DonovenDortilus (commented):
The README has a better flow but could still use better explanation on certain subjects such as the role costing. It also does not mention enough about role request/assignment and states which connects the layers. To further explain actions, an example behavior tree would help as well. The diagrams on the gameplay doc would also be a nice addition.

rj_gameplay/README.md
```diff
@@ -11,32 +11,44 @@ Stable code for gameplay.
...
 
-**Skill - `rj_gameplay/skill/`**
+**Play - `rj_gameplay/play/`**
 
-Atomic robot behavior which consists of a behavior tree and calls a ROS action.
-(Examples: Move, Kick)
+Handles full team behavior. Gets potential roles from a list of >=6 tactics,
+then assigns 6 of them at any time.
+
+(Examples: Basic Defense)
```
HussainGynai: Probably link to the file.

rj_gameplay/README.md
```diff
@@ -11,32 +11,44 @@ Stable code for gameplay.
...
 
 **Tactic - `rj_gameplay/tactic/`**
 
-Handles complex single robot behavior. Think of as one role in a play.
-(Examples: Goalie Receiver Waller)
+Handles role costing for one or more robots, such that Plays can choose
+higher-level ideas for each robot. Think of as abstraction layer between single-robot
+Skills and whole-team Plays.
```
HussainGynai, line 37: May want to mention that tactics typically are stateful or require some complex manipulation of worldstate info

HussainGynai (changes requested):
Minor changes, check comments.

kfu02:
@DonovenDortilus as clarified last meeting, I wanted the README included to be bare-bones. I will add a link to the longer doc as suggested by you and @HussainGynai.

rj_gameplay/README.md
kfu02: Adding links would work well when viewing on GH, but not so much when from the command line. I can add filepaths instead so it's easy to find regardless.

rj_gameplay/README.md
kfu02: I don't love either detail because skills are also stateful and take in worldstate info. "Complex manipulation" is the only differentiation and that's hard to define. That's why I chose to write that they "handle role costing" because it defines better what the complex manipulation is. Do you have a better way to differentiate those descriptors from skills?

echen0918: Just my suggestions/thoughts: If you're just trying to better differentiate tactics and skills here, I think it would be useful to emphasize how skills handle the specific actions of the robot through the behavior tree while tactics handle redistributing/organizing those skills. Tactics, to my understanding, are mainly used to handle the assignment of robots to skills for a given play, and don't directly "tell" any specific robot what actions they should be performing at any moment (which skills do with the behavior tree). Tactics instead use the worldstate info to reorganize the necessary skills among the robots with new role requests. You mention the behavior tree below in the skills section, so I'm not sure if its necessary to repeat that info for clarification. If anything is unclear or my current understanding is flawed, feel free to DM me on Slack.

PotentialError:
Two adjustments I would personally suggest.
1. I would prefer the structure of the elements of the framework to work from smallest to biggest (ex. first is action and then last is play) because then it seems easier to build off of the smaller elements when describing the larger/abstract elements. For example, by putting play first, I feel like I'm learning what play really is as I go down to descriptions of smaller elements such as tactics when the reader should know exactly what play is in its respective section (I'll show the exact part in comments).
2. Though I understand this README is supposed to be bare bones I feel like explaining how roles work is crucial because of its connection with the other elements (and also I struggle to understand the relationship between roles and skills both in this README and the document but that just might be me being new to this).

rj_gameplay/README.md
```diff
@@ -11,32 +11,44 @@ Stable code for gameplay.
...
-(Examples: Move, Kick)
+(Examples: `basic_defense.py` in folder above)
 
 **Tactic - `rj_gameplay/tactic/`**
 
-Handles complex single robot behavior. Think of as one role in a play.
-(Examples: Goalie Receiver Waller)
+Handles role costing for one or more robots, such that Plays can choose
+higher-level ideas for each robot. Think of as abstraction layer between single-robot
+Skills and whole-team Plays.
```
PotentialError, line 37: In this tactic description, I feel like I'm learning what plays are more compared to the actual plays description.

kfu02:
closed bc #1811 renders this unhelpful

PR #1741, opened 2021-10-25
foreach in all plays

hdemusg:
## Description
Replaced repetitive statements in plays with foreach loops

## Steps to test
Run simulator

Expected result: It should work

PR #1801, opened 2021-12-06
fix mypy errors in nmark_tactic.py

N8BWert:
## Description
Basically fixed (or attempted to fix) the mypy static typing errors in the file tactic/nmark_tactic.py.  For specific details of changes see below and in Files changed.

## Associated Issue
#1741 foreach in all plays (tactic/nmark_tactic.py)

## Steps to test
Find "Pull request checks / mypy" and click details.

Expected result: Should run

## Sub-issues (Still open things that I think aren't actually issues)
Error: rj_gameplay/rj_gameplay/tactic/nmark_tactic.py:64: error: Module has no attribute "soccer" -> unsolved? Mypy error due to ros2 

Error: rj_gameplay/rj_gameplay/tactic/nmark_tactic.py:92: error: Argument 1 to "SkillEntry" has incompatible type "Mark"; expected "Type[Any]" -> In the skillEntry class that this is inherited from, it uses the TypeVar, which for some weird reason causes this incompatible type error because the variable type is generated at runtime making mypy get angry.


rj_gameplay/rj_gameplay/tactic/nmark_tactic.py
```diff
@@ -35,7 +35,7 @@ def __init__(self, enemy_to_mark: rc.Robot=None):
     def __call__(
         self,
         robot: rc.Robot,
-        prev_result: Optional["RoleResult"],
+        prev_result: Optional[role.RoleResult],
```
N8BWert, line 44: Error: rj_gameplay/rj_gameplay/tactic/nmark_tactic.py:38: error: Name 'RoleResult' is not defined
-> Imported static type from stp.role and replaced optional static type with actual static type

rj_gameplay/rj_gameplay/tactic/nmark_tactic.py
```diff
@@ -65,7 +65,7 @@ def __call__(
 
     def unassigned_cost_fn(
         self,
-        prev_result: Optional["RoleResult"],
+        prev_result: Optional[role.RoleResult],
```
N8BWert, line 76: Error: rj_gameplay/rj_gameplay/tactic/nmark_tactic.py:68: error: Name 'RoleResult' is not defined
-> See Comment Above

rj_gameplay/rj_gameplay/tactic/nmark_tactic.py
```diff
@@ -79,7 +79,7 @@ def __init__(self, n: int):
         self.num_markers = n
 
         # create empty mark SkillEntry for each robot
-        self.mark_list = [
+        self.mark_list: type[mark.Mark] = [
```
N8BWert: Error: rj_gameplay/rj_gameplay/tactic/nmark_tactic.py:82: error: Need type annotation for 'mark_list'
-> Added type hinting from rj_gameplay.skill.mark.Mark

rj_gameplay/rj_gameplay/tactic/nmark_tactic.py
```diff
@@ -98,7 +98,7 @@ def create_request(self, **kwargs) -> role.RoleRequest:
 
     def get_requests(
         self, world_state: rc.WorldState, props
-    ) -> List[tactic.RoleRequests]:
+    ) -> tactic.RoleRequests:
```
N8BWert: Error: rj_gameplay/rj_gameplay/tactic/nmark_tactic.py:99: error: Return type "List[Dict[SkillEntry[Any], List[RoleRequest]]]" of "get_requests" incompatible with return type "Dict[SkillEntry[Any], List[RoleRequest]]" in supertype "ITactic"
+
Error: rj_gameplay/rj_gameplay/tactic/nmark_tactic.py:119: error: Incompatible return value type (got "Dict[SkillEntry[Any], List[RoleRequest]]", expected "List[Dict[SkillEntry[Any], List[RoleRequest]]]")
-> There doesn't seem to be any reason hinted in the code to return a list of dictionaries, so list hint was removed.

rj_gameplay/rj_gameplay/tactic/nmark_tactic.py
```diff
@@ -119,7 +119,7 @@ def get_requests(
         return role_requests
 
     def tick(self, world_state: rc.WorldState,
-             role_results: tactic.RoleResults) -> List[tactic.SkillEntry]:
+             role_results: tactic.RoleResults, props: Optional[tactic.PropT]) -> List[tactic.SkillEntry]:
```
N8BWert: Error: rj_gameplay/rj_gameplay/tactic/nmark_tactic.py:121: error: Signature of "tick" incompatible with supertype "ITactic"
-> If the method given (tick) was meant to override or overload the parent then it doesn't do that.  The solution was to just add the props as an optional parameter.
-> This shouldn't break anything, but if it does then I'll fix things.

kfu02 (approved):
Looks good! Nice work.

PR #1802, opened 2021-12-06
Fixed kickoff_play.py MyPy Errors

gordonm867:
I fixed MyPy errors in kickoff_play.py by replacing "RoleResult" with "role.RoleResult", replacing "calculations" with "wall_calculations" to match an existing class, and modifying the specified return types of some methods to match their actual return types.  I can confirm that the kickoff play still runs on my machine with this code and reports no MyPy errors within the kickoff_play.py file, but I had to change the IPlay class to avoid type incompatibility, and I am not sure if that is what I was supposed to do or if that will destroy other things.

### DIFF
--- a/rj_gameplay/rj_gameplay/gameplay_node.py
+++ b/rj_gameplay/rj_gameplay/gameplay_node.py
@@ -32,7 +32,7 @@ class TestPlaySelector(situation.IPlaySelector):
     """
     def select(self, world_state: rc.WorldState) -> Tuple[situation.ISituation, stp.play.IPlay]:
         self.curr_situation = None
-        return (None, penalty_defense.PenaltyDefense())
+        return (None, kickoff_play.PrepareKickoffPlay())
 
 
 class GameplayNode(Node):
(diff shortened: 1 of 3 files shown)


kfu02 (changes requested):
Resolve the two style checks complaining about "line too long". (See Files Changed tab.)

Not your fault because the style checks didn't pop up until I manually clicked run on them (link your account with my.robojackets.org to resolve this, I think). I will review the change to the IPlay class after style is fixed.

gordonm867:
> Resolve the two style checks complaining about "line too long". (See Files Changed tab.)
> 
> Not your fault because the style checks didn't pop up until I manually clicked run on them (link your account with my.robojackets.org to resolve this, I think). I will review the change to the IPlay class after style is fixed.

Oh, thanks for letting me know.  I should have resolved those issues.  I also connected my account with my.robojackets.org.

oddlyspookycherry (approved):
Works well on my end, good job!

kfu02 (approved):
Nice job!

PR #1813, opened 2022-01-07
add build-release-debug/ to git ignore

kfu02:
## Description
Makes it so that build-release-debug/ stops showing up when git status is run.

## Steps to test
Build repo, then run `git status`. Should not show build-release-debug in red.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,7 @@
 /.settings
 /api_docs
 *~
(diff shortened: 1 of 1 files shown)


Yudai-8-om (approved):

echen0918 (approved):
Approve change.

PR #1816, opened 2022-01-15
fixed wall_ball.py

ajaybati:
## Description
Fixing mypy issue in wall_ball.py


kfu02 (changes requested):
Nice work so far! Can you merge ros2 into this branch first so there are no merge conflicts? (You can google "how to merge branches git")

PR #1819, opened 2022-01-17
wall_calculations code moved here + adjusted code

azooz2003-bit:


kfu02 (changes requested):
Did you test these changes? If so, how can I also test them? Your description is blank so I'm not sure how to test.

Also, make sure you're merging into the correct base branch from your branch.

PR #1820, opened 2022-01-17
Apply hashable robot

Yudai-8-om:
## Description
robot hash is used in role assignment

### Test Case 1
1. Run sim
2. Make sure passing works

### Test Case 2
1. Run sim
2. Set `self.test_play` in gameplay_node.py to this: `self.test_play = basic_defense.BasicDefense()`
3. Make sure 6 robots do something in the play.

### DIFF
--- a/rj_gameplay/stp/play/__init__.py
+++ b/rj_gameplay/stp/play/__init__.py
@@ -76,8 +76,7 @@ class Play(ABC):
         """Given that all roles are in sorted order of priority, greedily assign the highest-priority role to the lowest-cost robot for that role. Instantiate tactics with the correct robots post-assignment.
         Satisfy constraint that all roles of a tactic must be assigned together.
         """
-
-        # TODO: use hashable Robots directly once PR #1815 merged
+        
         used_robots = set()
         for tactic in self.prioritized_tactics:
             # TODO: handle if tactic requests more roles than exist
--- a/rj_gameplay/stp/play/__init__.py
+++ b/rj_gameplay/stp/play/__init__.py
@@ -80,7 +80,7 @@ class Play(ABC):
 
         all_assigned_robots = []
         # TODO: use hashable Robots directly once PR #1815 merged
-        used_robot_ids = set()
+        used_robots = set()
         for tactic in self.prioritized_tactics:
             # TODO: handle if tactic requests more roles than exist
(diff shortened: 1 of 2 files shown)


rj_gameplay/stp/play/__init__.py
```diff
@@ -90,7 +90,7 @@ def assign_roles(
                 min_cost = 1e9
                 cheapest_robot = None
                 for robot in world_state.our_robots:
-                    if robot.id in used_robot_ids:
+                    if robot.__hash__() in used_robots:
```
kfu02: You should actually be able to just say this. Python will automatically call the hash function you wrote. (Typically all functions that start with __ are not meant to be called directly by the programmer in Python.)
```suggestion
                    if robot in used_robots:
```

rj_gameplay/stp/play/__init__.py
```diff
@@ -101,9 +101,11 @@ def assign_roles(
                     # TODO: properly error handle if cheapest_robot is None
                     print(f"RoleRequest ({role}, {cost_fn}) was not assigned")
 
-                used_robot_ids.add(cheapest_robot.id)
+                used_robots.add(cheapest_robot.__hash__())
```
kfu02: see comment above

rj_gameplay/stp/play/__init__.py
```diff
@@ -101,9 +101,11 @@ def assign_roles(
...
                 assigned_robots.append(cheapest_robot)
             tactic.assigned_robots = assigned_robots
+            print(f'used:{used_robots}')
```
kfu02: delete these prints when finalized

kfu02 (changes requested):
Merge in the base branch again, I made some changes to basic defense play that will be better to test these changes.
(Also set gameplay node to use basic_defense instead of keepaway play.)

EDIT: I created some merge conflicts with my changes, sorry about that. Key difference is I renamed a list called "assigned_robots" to "robots_for_tactic" to be more clear, and also created a "set_assigned_robots" method for Tactics.

kfu02 (approved):
delete the associated TODO, other than that looks good

N8BWert (approved):
Basic defense runs well.  I did not encounter any issues running the code and everything looks good.

PR #1821, opened 2022-01-17
move wall_calculations to wall_tactic

azooz2003-bit:
## Description
Code from wall_calculations file moved to wall_tactic, changes were also made to avoid unnecessary code and imports.

Steps showing it works:
1- Run the basic_defense file with no errors

### DIFF
--- a/rj_gameplay/rj_gameplay/tactic/wall_tactic.py
+++ b/rj_gameplay/rj_gameplay/tactic/wall_tactic.py
@@ -9,14 +9,23 @@ from rj_gameplay.skill import move
 import stp.skill as skill
 import numpy as np
 
+"""Holds functions used by plays that use the wall tactic."""
+
+from dataclasses import dataclass
+
+import stp.action as action
+import stp.rc as rc
+import stp.tactic as tactic
+import stp.role as role
+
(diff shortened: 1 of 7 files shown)


rj_gameplay/rj_gameplay/tactic/wall_tactic.py
```diff
@@ -9,14 +9,23 @@
...
+
 # TODO: replace w/ global param server
 from stp.utils.constants import RobotConstants, BallConstants
 import stp.global_parameters as global_parameters
 
 from rj_msgs.msg import RobotIntent
 
-# TODO: move this out of calculations file and into this tactic
-from rj_gameplay.calculations import wall_calculations
+# TODO: move this out of calculations file and into this tactic -> DONE
```
kfu02, line 28: you can just delete this TODO now
```suggestion
```

kfu02 (changes requested):
Looks good! I requested changes because there are a few style errors that should be cleaned up. (They are almost certainly from my branch as I didn't clean that up yet. However, if you would be willing to remove the unused imports on my behalf, I would appreciate it.)

PR #1822, opened 2022-01-18
update debugging cpp for ros2

kasohrab:
## Description
The steps to debug with ros2  were outdated. This pr adds a script to generate separate debugging launch files and describes how to debug. Also, it makes small changes to .gitignore and makefile related to the update.

### Test Case 1
you may need to chmod +x util/debug-cpp.sh
1. util/debug-cpp.sh -n node_to_debug (example: planner)
2. make perf / make debug
3. follow the steps echoed out by the script in step 1.
4. press r then enter in the debugger to begin running the executable (for planner the executable is called planner_node)

feel free to test any other debugging features, and the regular sim launch file should still work as well.

The todo to add debug-sim back can be deleted if we think it's not really worth thinking about. Also, there is a way to add to the script to both create the new launch files for debugging and launch and run the other node in a new terminal (but I couldn't figure it out in the short time I spent on this, what I tried is commented out at the bottom of the script). So if that is worth adding it can be done also (in a different pr I would think).

Expected result: should still work

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,7 @@
 /api_docs
 *~
 build-*
+cmake-*
 
 # sublime-text
 *.sublime-workspace
@@ -43,3 +44,6 @@ perf.data*
 .coverage
 .coverage.*
 coverage.xml
+
+# Debugging launch files
+launch/debug-*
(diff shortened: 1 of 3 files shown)


adithya-gv (changes requested):
Fix the description, some of the scripts to execute in the README are incorrect.

gordonm867 (approved):
The debugger seems to work, as long as the typo in the instructions is fixed to avoid confusion.  Good job!

adithya-gv (approved):

PR #1824, opened 2022-01-19
self.robot is updated in Skill superclass tick method

N8BWert:
## Description
TODO: make superclass force robot update in tick()
  (do this quietly with self.robot = world_state.our_robots[self.robot.id]

In the future, it will likely be necessary to update specific variables and/or states during all skill ticks.  To ensure DRY code the tick method in the skill superclass is no longer to be an abstract method and will now contain implementations of any shared code between ticks.  In the future, all skill subclasses will now include the lines "super().tick(world_state)" instead of the lines that are to be shared.

### Test Case 1
run any of the working plays that use capture

Expected result: Nothing bad


gordonm867 (approved):
The changes all seem to make sense (mostly just calling super().tick() to update the robot position) and do not seem to break the plays.  Good job!

rj_gameplay/stp/skill/__init__.py
```diff
@@ -21,10 +21,12 @@ class Skill(ABC):
     @abstractmethod
     def tick(self, world_state: stp.rc.WorldState) -> RobotIntent:
         """Logic for Skill goes here. RobotIntents obtained via Actions.
+        self.robot is updated forably and quietly.
         .msg
                 :param world_state: Current world state.
                 :return: A single RobotIntent.
         """
+        self.robot = world_state.our_robots[self.robot.id]
```
kfu02: When running BasicDef, I got an error claiming NoneType has no attr 'id', which suggests self.robot is None on a tick(). I believe the issue is actually that sometimes the init functions for certain skills take in None as their default param (e.g. skill/receive), but I haven't tested that hypothesis. Can you remove that default param and see if that fixes it?

kfu02 (changes requested):
See comment, I got an error on the new line

rj_gameplay/rj_gameplay/skill/capture.py
```diff
@@ -27,6 +27,10 @@ def __init__(self, robot: Optional[rc.Robot] = None):
         self.__name__ = "capture skill"
 
     def tick(self, world_state: rc.WorldState) -> RobotIntent:
+        """
```
kfu02: can you move this comment to the superclass

kfu02 (approved):
Looks good, but can you move that one comment to the superclass? Thanks

PR #1829, opened 2022-01-19
ROS2 Action Server

kasohrab:
## Description
Adding ROS2 action server and supporting infrastructure (see design doc).

Things that are not planned as part of this pr but will be done before comp:
- Eventually ServerIntent's should not be the monolithic goal. I will change them to be more specific later.
- Assure that planning should stop planning once cancel is received
- is_done should use callbacks from clients
- tick should use feedback from clients
- other changes wil be made over time
- such as using has_ball from RobotStatus to assist with recieving the ball (well it looks like we already can see robot statuses in gameplay, but using it from the serve via feedback instead would be better)

## Associated Issue
Issue #???

## Design Documents
[Link](https://docs.google.com/document/d/1P7VkJQtJrStnEk3-Z7tV20YO3QziLQibLHO8cqtJrCg/edit#)

### Test Case 1
1. Run Sim

Provide feedback on the design of the system and the classes themselves. If any code is unclear, I can change it to make it more obvious what the code is doing (but I would rather not add too many comments, instead try to make the code self explanatory). I have already kinda done that (by avoiding auto and creating separate methods anytime I felt like adding a comment), but let me know anywhere else worth clearing up.

Also I had some TODO's that if anyone know the answer to that would be great.

Expected result: still works


rj_gameplay/rj_gameplay/skill/mark.py
```diff
@@ -99,25 +102,35 @@ def tick(self, robot: rc.Robot, world_state: rc.WorldState, intent: RobotIntent)
...
+                    ** 2
+                )
+                < threshold
+            ):
+                return True
+            else:
+                return False
         else:
-            return False
+            return True
```
kasohrab, line 134: probably need to change this back too.

kasohrab: (and in move)

kasohrab:
resolved :)

PR #1836, opened 2022-01-26
Delete trajectory test that no longer works

gordonm867:
Because of recent changes to prevent a crash from trajectory splicing in the settle planner, there is a test here that is no longer applicable to our current code.  For now, we have commented it out.


kfu02 (changes requested):
Nice explanation! I'm not sure why the style formatter is yelling at you, can you run `make pretty-lines` to fix?

kfu02 (approved):
LGTM

PR #1839, opened 2022-02-01
Added pre-commit hook for Black python formatter

PotentialError:
## Description
Added pre-commit to auto add a pre-commit hook to the repo. Now whenever you try and commit, black will run and stop you from committing if there is a style issue. Black will reformat the code and you can re-commit with your new stylized code.

## Steps to test
1. Get pre-commit installed either by running `./util/ubuntu-setup` or run `pip install pre-commit`
2. Run `pre-commit install` if you ran `pip install pre-commit`
3. Change a file to be stylistically incorrect
4. Try and commit

Expected result: Pre-commit hook should stop you from committing and black should run to reformat code. Running it again, it should pass and your commit will be committed.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,6 @@
+#Run pre-commit autoupdate periodically to make sure version (rev) is correct
+repos:
+-   repo: https://github.com/psf/black
+    rev: 22.1.0
+    hooks:
+    -   id: black
--- a/rj_gameplay/rj_gameplay/play/defensive_clear.py
+++ b/rj_gameplay/rj_gameplay/play/defensive_clear.py
@@ -11,7 +11,6 @@ import numpy as np
 from typing import Dict, List, Tuple
 
 
-
 class DefensiveClear(play.IPlay):
     def __init__(self):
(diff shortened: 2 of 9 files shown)


kfu02 (commented):
Works great! Two questions:
1) Would including the `pre-commit install` line in `./util/ubuntu-setup` automatically install this for everyone when they run that?
2) The precommit hook says Black "fails" but then Black prints a message indicating that it succeeds (see screenshot). Is there a way to get the hook to match what Black says?
Screenshot from 2022-02-01 19-49-06

Re-request my review when needed

PotentialError:
@kfu02 
1. I can certainly look into the ubuntu-setup
2. Basically the way pre-commit handles formatters like Black is first run it is a test and then actually run the formatter. So I do not really see how to change it as my only interface is the config file which is limited. So all you have to do is recommit and it should work

Yudai-8-om (approved):
Worked nicely!

kfu02:
@PotentialError makes sense, ping me when 1) is done / if 1) is not feasible then.

kfu02 (approved):
Beautiful

PR #1840, opened 2022-02-02
switch to local host in grSim

kasohrab:
## Description
title

## Steps to test
run grsim 

Expected result: works and no more pollution
224.5.23.2 is only for comp.


kfu02 (approved):
g

Yudai-8-om (approved):
g

PR #1811, opened 2022-01-06
STP Refactor 2022

kfu02:
## Description
Implements gameplay changes, including 1) greedy role assignment, 2) redefinition of Tactics and Roles, and 3) simplification of existing architecture. See design doc below for more details.

## Design Documents
[STP Refactor 2022](https://docs.google.com/document/d/1gRAF--W7FwGoof0--1l_pyjM-N4RmMiV6FhBvyQIRgM/edit?usp=sharing)

## Steps to test
Launch sim. Run different plays by changing the `test_play` variable at the bottom of `gameplay_node.py`. (The only three plays written so far are the ones imported [at the top of that file](https://github.com/RoboJackets/robocup-software/pull/1811/files#diff-90a35bbaea2fb27fd2d7efbdd3896b7a09d65ca6b7054300c7df727e3b00ac02R33).) Ensure that the plays behave as their docstrings would suggest.

Note: sometimes world_state is not passed to gameplay correctly and everything breaks. Try launching a few times. The reason for this bug is unclear; I added a task to the board to address this.

## Other Review Notes

I intentionally left the unused imports at the top of gameplay_node and the "hanging indent" complaints from pylint. Unused imports are there so you know which plays are testable. Hanging indent was left to be fixed by changing either black auto-formatter or pylint preferences.

There are a lot of files in this PR. Here's one _suggested_ way to read them:

 1. SRTP structure: `stp/skill/__init__, stp/role/__init__, stp/tactic/__init__, stp/play/__init__, stp/role/cost.py, gameplay_node.py`
 2. line_up play: `play/line_up, tactic/line_tactic, role/dumb_move, skill/move`
 3. basic_defense play: `play/basic_defense, tactic/goalie_tactic, tactic/wall_tactic, role/goalie_rolewall_calculations` + pivot_kick skill (below)
 4. pivot_kick skill: `skill/capture, skill/kick, skill/pivot, skill/pivot_kick`
 5. keepaway play: `play/keepaway, tactic/pass_tactic, role/passer, role/receiver, skill/receive, skill/settle`
 6. least important files: `coordinator, move_tactic, role/__init__` (all currently unused or empty), `rc.py` (literally deleted one whitespace line), `stp/role/assignment/__init__, stp/situation/__init__` (typing changes), `basic_play_selector` (only ~10 of the 97 lines are actual logic changes, mostly formatting, and the logic change is dead simple)

Comment or Slack with any concerns!

PR #1843, opened 2022-02-04
Make FSMs in gameplay Enum-based

kfu02:
## Description
Title. This was done as opposed to doing string comparisons, which are slow (and ugly). The State Enums allow one to glance at the top of the file and know what the FSM will tick through.

## Steps to test
Run sim with 3 plays: `BasicDefense, LineUp, and Keepaway . 

Expected result: Behavior should match #1811 .


kasohrab (approved):
looks good. Hopefully in the future we can explore using something like this for our FSMs: https://github.com/ros-industrial/packml_ros2. Or create a new FSM class, I think you said there were issues with the current one.

HussainGynai (approved):
LGTM

PR #1845, opened 2022-02-09
Basic Offense Play

kfu02:
## Description

Fills in basic122 with a sensible default offense play: move two robots to the wings, have a third robot pass to one of those two, and shoot once the wings have received a pass.

Costs are now passed into the pass tactic, to allow different plays to choose how they want to pass differently.

Currently the passing is buggy because kick speed is wrong and capture planner is weak, need to fix in separate PRs.

## Steps to test
Run sim.

Expected result: Description of sensible default offense play matches observed behavior.

### DIFF
--- a/rj_gameplay/rj_gameplay/play/basic122.py
+++ b/rj_gameplay/rj_gameplay/play/basic122.py
@@ -1,116 +1,37 @@
-import stp.play as play
-import stp.tactic as tactic
+from typing import List
 
-from rj_gameplay.tactic import (
-    striker_tactic,
-    goalie_tactic,
-    pass_seek,
-    wall_tactic,
-)
-import stp.skill as skill
-import stp.role as role
-from stp.role.assignment.naive import NaiveRoleAssignment
-import stp.rc as rc
-from typing import (
-    Dict,
-    List,
-    Tuple,
-    Type,
-)
-import numpy as np
-from rj_gameplay.calculations import wall_calculations
+import stp
+
+from rj_gameplay.tactic import pass_tactic
+
+from rj_msgs.msg import RobotIntent
+
+from enum import Enum, auto
 
 
-class Basic122(play.IPlay):
-    """A basic offensive play. One robot is a striker and shoots as soon it gets the ball.
-    other two seek to specific spots (BAD HARDCODING)
+class State(Enum):
+    INIT = auto()
+    ACTIVE = auto()
+    ASSIGN_ROLES = auto()
+
+
+class Basic122(stp.play.Play):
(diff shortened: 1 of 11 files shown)


rj_gameplay/rj_gameplay/play/basic122.py
```diff
@@ -1,116 +1,138 @@
...
+
+            return self.get_robot_intents(world_state)
+
+        elif self._state == State.INIT_PASS:
+            init_passer_cost = stp.role.cost.PickClosestToPoint(world_state.ball.pos)
+            init_receiver_cost = stp.role.cost.PickClosestToPoint(self._seek_pts[0])
+            self.prioritized_tactics = [
+                goalie_tactic.GoalieTactic(world_state, 0),
+                pass_tactic.PassTactic(
+                    world_state, init_passer_cost, init_receiver_cost
```
Yudai-8-om, line 85: Is there a reason that we don't have seek tactic here?

rj_gameplay/rj_gameplay/tactic/striker_tactic.py
```diff
@@ -8,213 +8,59 @@
...
+            self.init_roles(world_state)
 
-        if capture_result and capture_result[0].is_filled():
-            return [self.capture]
-        if shoot_result and shoot_result[0].is_filled():
-            self.shoot.skill.target_point = find_target_point(
-                world_state, kick_speed=KICK_SPEED
-            )
-            return [self.shoot]
+        return [(role.robot.id, role.tick(world_state)) for role in self.assigned_roles]
```
Yudai-8-om, line 43: low priority, but we don't need for loop here

Yudai-8-om (approved):
Works great overall.
Just wanna note that the receiver starts capturing too early, so we might wanna fix that later if kick speed didn't help

rj_gameplay/rj_gameplay/play/basic122.py
kfu02: no, will add TODO

rj_gameplay/stp/play/__init__.py
```diff
@@ -38,6 +38,10 @@ class IPlay(ABC):
     pass
 
 
+class RoleAssignFailure(Exception):
+    pass
```
p-nayak11, line 42: Why are were we passing the role assign failure?

p-nayak11 (approved):
Sim ran fine yesterday, but it's currently giving me trouble. It is probably a local problem. I don't understand why we are passing the role assign failure. Other than that, it looks good.

PR #1852, opened 2022-02-21
Ball gets to target in amount of time specified

N8BWert:
## Description
Basically I used a quick linear approximation from [ZJUNlict's 2020 TDP](https://ssl.robocup.org/wp-content/uploads/2020/03/2020_ETDP_ZJUNlict.pdf) to allow the ball to end at the other opponent at a specified "velocity."  I'm not sure the units or dimensionality of this "velocity," but it works proportionally.  

The approximation seems to work pretty accurately, however it seems that the robot will not always pass the ball directly to the other robot.  This is because the threshold for pivoting is high and the other robot does not actually move to get the ball.  In the future if we wanted to increase the efficacy of our passes we could either lower the threshold of pivoting (this would make the robots take longer before passing) or increasing the effectiveness of our receiving robot receiving the ball.

### Test Case 1
1. Run keepaway play
2. Watch robots pass
3. profit

Expected result: Robots should make passes that reach their opponents in the amount of time specified in passer.py (by default this is 0.75 seconds)

### DIFF
--- a/rj_gameplay/rj_gameplay/role/passer.py
+++ b/rj_gameplay/rj_gameplay/role/passer.py
@@ -1,4 +1,3 @@
-from mimetypes import init
 import stp.role
 import stp.rc
 
@@ -71,8 +70,6 @@ class PasserRole(stp.role.Role):
             # kick_speed is modeled off of the ETDP of ZJUNlict, which can be found in section 5 of https://ssl.robocup.org/wp-content/uploads/2020/03/2020_ETDP_ZJUNlict.pdf
             distance = np.linalg.norm(self._target_point - self.robot.pose[0:2])
             initial_velocity = np.sqrt((FINAL_VELOCITY ** 2) - (2 * BALL_DECELERATION * distance))
-            print(distance)
-            print(initial_velocity)
             self.pivot_kick_skill = pivot_kick.PivotKick(
                 robot=self.robot,
                 target_point=self._target_point,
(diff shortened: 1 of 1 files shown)


kfu02 (approved):
Really long and really short passes behave strangely; however, in reasonable gameplay schemes we shouldn't see either. Fine for now, can update the time to pass later.

Yudai-8-om (approved):
I noticed that long passes behave weirdly too, but lgtm in general

rj_gameplay/rj_gameplay/role/passer.py
```diff
@@ -1,3 +1,4 @@
+from mimetypes import init
```
kfu02: what is this?

rj_gameplay/rj_gameplay/role/passer.py
```diff
@@ -60,11 +68,16 @@ def tick(self, world_state: stp.rc.WorldState) -> RobotIntent:
         # this state transition is done by the PassTactic, which is not canonical FSM
         elif self._state == State.INIT_EXECUTE_PASS:
             # TODO: make these params configurable
+            # kick_speed is modeled off of the ETDP of ZJUNlict, which can be found in section 5 of https://ssl.robocup.org/wp-content/uploads/2020/03/2020_ETDP_ZJUNlict.pdf
+            distance = np.linalg.norm(self._target_point - self.robot.pose[0:2])
+            initial_velocity = np.sqrt((FINAL_VELOCITY ** 2) - (2 * BALL_DECELERATION * distance))
+            print(distance)
+            print(initial_velocity)
```
kfu02: delete these prints

kfu02 (approved):
Couple of small things to fix, but overall much-needed improvement!

PR #1857, opened 2022-02-28
fix gameplay node and add warning comment

kfu02:
self.test_play was incorrectly overwritten at some point, fixed here


HussainGynai (approved):
LGTM

anshulm07 (approved):
LGTM

PR #1861, opened 2022-02-28
Fix ball_sense to be false

HussainGynai:
## Description
has_ball_sense was not a bool, and therefore always returned True when checked, so we were never settling.

### DIFF
--- a/rj_gameplay/stp/utils/world_state_converter.py
+++ b/rj_gameplay/stp/utils/world_state_converter.py
@@ -265,10 +265,10 @@ def robot_creator(
     else:
         is_ours = True
         robot_id = robot_state.id
-        ball_sense = Optional[robot_status.has_ball_sense]
-        kicker_charged = Optional[robot_status.kicker_charged]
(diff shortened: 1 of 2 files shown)


rj_gameplay/stp/utils/world_state_converter.py
```diff
@@ -265,10 +265,10 @@ def robot_creator(
     else:
         is_ours = True
         robot_id = robot_state.id
-        ball_sense = Optional[robot_status.has_ball_sense]
-        kicker_charged = Optional[robot_status.kicker_charged]
-        kicker_healthy = Optional[robot_status.kicker_healthy]
-        lethal_fault = Optional[robot_status.lethal_fault]
+        ball_sense = False
```
kfu02: actually I think this should be `ball_sense = robot_status.has_ball_sense`, right? setting False always ignores the input when robot status is not None

kfu02 (changes requested):
Not sure if I'm missing something here

PR #1864, opened 2022-03-01
SeekerRole

p-nayak11:
## Description
This role assigns a robot a region of space according to a preset formation. Then, the robot finds a point that is a given distance away from the opposition robots in this region of space. The robot moves to this target point while facing the ball position. 

## Associated Issue
The "getting open" section of the code seems to have a high time complexity so another algorithm will need to be used if it is an issue in the sim. 

I still need to update the basic_seek tactic with these new changes to the role. In the future, a Formations class also needs to be implemented in world_state so that the tactic can call both the region and formation type to the seeker role. 

- Update Seek Tactic
- Test on Sim 
- Debug Issues


N8BWert (approved):
Looks good to me.

anshulm07 (approved):
LGTM

PR #1865, opened 2022-03-01
Constants via constants class

CK-Christopher:
## Description
Moved all constants in rj_gameplay to stp/utils/constants.py and namespaced them. 

## Associated Issue
Issue #???

## Design Documents
[Link](link-to-design-doc)

### Test Case 1
1. Step 1
2. Step 2
3. Step 3

Expected result: ???

## (Optional) Sub-issues (for drafts)
_Note: if you find yourself breaking this PR into many smaller features, it may make sense to break up the PR into logical units based on these features._
- [ ] Step 1
- [ ] Step 2

PR #1866, opened 2022-03-07
Penalty play defense

CK-Christopher:
## Description
Play for penalty defense is added. The goalie needs to defend the penalty kick and the other robots need to stay on the side. 

## Associated Issue
Issue #???

## Design Documents
[Link](link-to-design-doc)

### Test Case 1
1. Step 1
2. Step 2
3. Step 3

Expected result: ???

- [ ] Step 1
- [ ] Step 2


sseye6 (approved):
Works for me

llvm.sh
```diff
@@ -0,0 +1,111 @@
+#!/bin/bash
```
HussainGynai: Do not commit changes to this file

llvm.sh.1
```diff
@@ -0,0 +1,111 @@
+#!/bin/bash
```
HussainGynai: Nor this one

rj_gameplay/rj_gameplay/gameplay_node.py
```diff
@@ -30,7 +30,13 @@
...
 
 # ignore "unused import" error
-from rj_gameplay.play import line_up, basic_defense, keepaway, basic122  # noqa: F401
+from rj_gameplay.play import (  # noqa: F401
+    line_up,  # noqa: F401
+    basic_defense,  # noqa: F401
+    keepaway,  # noqa: F401
+    penalty_defense,  # noqa: F401
+)  # noqa: F401
+from rj_gameplay.play import basic122  # noqa: F401
```
HussainGynai: What are these noqa: F401 comments?

rj_gameplay/rj_gameplay/play/penalty_defense.py
```diff
@@ -1,137 +1,75 @@
...
+    INIT = auto()
+    ACTIVE = auto()
+    DONE = auto()
 
 
-class PenaltyDefense(play.IPlay):
-    """For when we don't have the ball and are trying to stop the opponent from scoring."""
+class PenaltyDefense(stp.play.Play):
+    """
+    Play consistas of:
```
HussainGynai: Typo

rj_gameplay/rj_gameplay/play/penalty_defense.py
```diff
@@ -1,137 +1,75 @@
...
+            return self.get_robot_intents(world_state)
+        elif self._state == State.DONE:
+            # TODO: does this state need to exist?
+            return None
+
+
+class PreparePenaltyDefense(stp.play.Play):
+    def __init(self):
+        super().__init__()
+        self._state = State.Init
```
HussainGynai: Init should be all caps

rj_gameplay/rj_gameplay/play/penalty_defense.py
```diff
@@ -1,137 +1,75 @@
...
+        world_state: stp.rc.WorldState,
+    ) -> List[RobotIntent]:
+        if self._state == State.INIT:
+            self.prioritized_tactics.append(goalie_tactic.GoalieTactic(world_state, 0))
+            self.prioritized_tactics.append(line_tactic.LineTactic(world_state, 5))
+            self.assign_roles(world_state)
+            self._state = State.ACTIVE
+            return self.get_robot_intents(world_state)
+        elif self._state == State.ACTIVE:
+            if self.prioritized_tactics[-1].is_done(world_state):
```
HussainGynai: Why are you checking if one of the wallers is done? Shouldn't it be based on if the goalie has the ball or if a goal has been scored or something.

rj_gameplay/rj_gameplay/play/penalty_defense.py
```diff
@@ -1,137 +1,75 @@
...
+    ACTIVE = auto()
+    DONE = auto()
 
 
-class PenaltyDefense(play.IPlay):
-    """For when we don't have the ball and are trying to stop the opponent from scoring."""
+class PenaltyDefense(stp.play.Play):
+    """
+    Play consistas of:
+                    -5 wallers
```
HussainGynai: This should be 5 lined up robots or something, not wallers right?

rj_gameplay/rj_gameplay/play/penalty_defense.py
```diff
@@ -1,137 +1,75 @@
...
-        flat_requests = play.flatten_requests(role_requests)
-        flat_results = self.role_assigner.assign_roles(
-            flat_requests, world_state, prev_results
-        )
-        role_results = play.unflatten_results(flat_results)
+        world_state: stp.rc.WorldState,
+    ) -> List[RobotIntent]:
+        if self._state == State.INIT:
+            self.prioritized_tactics.append(goalie_tactic.GoalieTactic(world_state, 0))
+            self.prioritized_tactics.append(line_tactic.LineTactic(world_state, 5))
```
HussainGynai, line 38: Looking at line_tactic, I don't actually think how its currently written would be legal. The robots that aren't the goalie have to be 1 meter behind the ball, and I think line_tactic lines them up along the left middle. You could try to change line_tactic to take in a set of points or something so you have control over where the robots line up, or you could use 5 move commands. 
See penalty rules in section 5.3.5 here: https://robocup-ssl.github.io/ssl-rules/sslrules.pdf

rj_gameplay/rj_gameplay/tactic/line_tactic.py
```diff
@@ -1,22 +1,24 @@
...
         super().__init__(world_state)
 
         # compute move points
-        # TODO: make start on side of the field, so this Tactic is actually useful during penalty situations
-        start = (3.0, 0.0)
+        # access the length and the width of the field (plus the border), and starts robots there
+        start = (
+            (world_state.field.width_m + world_state.field.border_m) / 2,
+            (world_state.field.length_m - world_state.field.border_m) / 2,
+        )
```
HussainGynai: Again don't know if these set of points are legal for a penalty kick, maybe make a separate tactic for penalties, or allow this one to be adjustable somehow.

HussainGynai (changes requested):
Mostly looks good, just make sure line_tactic is legal for penalties, and change it or create a separate line up for penalties if not.

rj_gameplay/rj_gameplay/gameplay_node.py
CK-Christopher: flake8 gives unused import errors. flake8 ignores lines with #noqa F401. Since all plays are tested in the gamplay_node, a bunch of plays is imported.

rj_gameplay/rj_gameplay/play/penalty_defense.py
CK-Christopher: I fixed that

rj_gameplay/rj_gameplay/play/penalty_defense.py
CK-Christopher: yeah, that makes sense. I changed the comments.

rj_gameplay/rj_gameplay/play/penalty_defense.py
kfu02: Hey, just adding context: the line tactic should be the one that lines up our robots properly for penalty offense/defense. By the rules (in the section Hussain cited): "3. Throughout the penalty kick procedure, all other robots have to be 1m behind the ball such that they do not interfere the penalty kick procedure."

The best solution is to make line tactic take in a start and end point, and a number of robots, then put those robots evenly in-between start and end points. I will make a card for this.

rj_gameplay/rj_gameplay/play/basic_defense.py
```diff
@@ -2,9 +2,8 @@
 
 from rj_gameplay.tactic import wall_tactic, goalie_tactic, nmark_tactic
 import stp.role
-from stp.role.assignment.naive import NaiveRoleAssignment
 import stp.rc
-from typing import Dict, List, Tuple, Type
+from typing import Dict, List, Type  # noqa: F401
```
kfu02: If one of these imports is unused, just delete it rather than putting the noqa flag

rj_gameplay/rj_gameplay/play/penalty_defense.py
kfu02: ^ goalie has ball would be easiest to implement. To be honest, this play doesn't really need an end since the ref will signal a penalty end and the play selector will pick a different play regardless.

kfu02 (changes requested):
Hey Christina, nice work, but I noticed there are some changes from PR #1865 bundled into this one. Can you set this PR's base to be that branch so those changes don't show up twice?

With regards to the line tactic thing: don't worry about fixing them in this PR. Fixing line tactic can be its own PR.

rj_gameplay/rj_gameplay/play/penalty_defense.py
CK-Christopher: I fixed that

rj_gameplay/rj_gameplay/play/basic_defense.py
CK-Christopher: ok

rj_gameplay/rj_gameplay/play/penalty_defense.py
```diff
@@ -1,137 +1,68 @@
...
-        return all([tactic.is_done(world_state) for tactic in self.tactics])
+class State(Enum):
+    INIT = auto()
+    ACTIVE = auto()
+    DONE = auto()
 
 
-class PenaltyDefense(play.IPlay):
-    """For when we don't have the ball and are trying to stop the opponent from scoring."""
+class PenaltyDefense(stp.play.Play):
```
Yudai-8-om, line 20: Is there a reason for having two separate classes? I think they are pretty much the same?

rj_gameplay/rj_gameplay/play/penalty_defense.py
```diff
@@ -1,137 +1,68 @@
...
-    def is_done(self, world_state):
-        # returns done when all tactics are done
-        # TODO: change all is_done() to use all()?
-        return all([tactic.is_done(world_state) for tactic in self.tactics])
+    def tick(
+        self,
+        world_state: stp.rc.WorldState,
+    ) -> List[RobotIntent]:
+        if self._state == State.INIT:
+            self.prioritized_tactics.append(line_tactic.LineTactic(world_state, 5))
```
Yudai-8-om: I believe we can start assigning the goalie at this stage since the goalie has nothing to do with the "1m rule".  Correct me if I'm wrong

rj_gameplay/rj_gameplay/play/penalty_defense.py
```diff
@@ -1,137 +1,68 @@
...
+            self.assign_roles(world_state)
+            self._state = State.ACTIVE
+            return self.get_robot_intents(world_state)
+        elif self._state == State.ACTIVE:
+            if self.prioritized_tactics[-1].is_done(world_state):
+                self._state = State.DONE
+            return self.get_robot_intents(world_state)
+        elif self._state == State.DONE:
```
Yudai-8-om: We can omit this

Yudai-8-om (commented):
It worked well, great job.
I think we don't need PreparePenaltyDefense play

rj_gameplay/rj_gameplay/play/penalty_defense.py
CK-Christopher: I initially had it as a single class, but then basic play selector complained, so I wrote a similar class.

rj_gameplay/rj_gameplay/play/penalty_defense.py
CK-Christopher: That makes sense. I will also add goalieTactic to prioritized tactics.

rj_gameplay/rj_gameplay/play/penalty_defense.py
CK-Christopher: Actually, now basic play selector is changed in ros2, so the requirement to have a prepare penalty defense play is removed. Therefore, I removed prepare penalty play defense

rj_gameplay/rj_gameplay/play/penalty_defense.py
CK-Christopher: done

kfu02:
was used in #1887 , will be cleaned up by #1895  but I'm merging current state in now to give credit to @CK-Christopher

PR #1869, opened 2022-03-09
Fix settle's end condition

HussainGynai:
## Description
Originally, this PR was supposed to be about creating an intercept skill to make up for the lack of an intercept planner. However, along the way I realized that settle.is_done() always returned true due to some typing errors in create_robot() in world_state_converter.py. After fixing this and tuning the min speed param, settle seems to work pretty well for getting in front of the ball, could probably use some tuning at the planning level though. I left the intercept skill there, it just moves a robot to the closest point on the straight line trajectory of the ball.

### Testing Settle
1. Open gameplay_node, and set the test_play_selector to use some passing play
2. Open sim and let the robots play
3. Observe Passing

Expected result: The receiving robot should get in front of the ball and try to face it.

### Testing Intercept
1. Create a play that uses the intercept skill (preferably defense)
2. Open sim2play and run the defense play against a passing play
3. Observe

Expected Result: The intercepting robot should try to get along the path of the robot


rj_gameplay/rj_gameplay/skill/intercept.py
```diff
@@ -5,57 +5,82 @@
...
-        self.target_point = target_point
-
-        self.__name__ = "Intercept Skill"
-        if self.robot is not None:
-            self.intercept = intercept.Intercept(self.robot.id, target_point)
-        else:
-            self.intercept = intercept.Intercept(None, target_point)
-        self.intercept_behavior = ActionBehavior(
-            "Intercept", self.intercept, self.robot
+    def get_intercept_pt(self, world_state: rc.WorldState, my_robot: np.ndarray) -> np.ndarray:
```
kfu02, line 29: Can you add a docstring here explaining what this does? Fairly certain it's "find point perp to the ball-bot line" but some clarification would be good for posterity

rj_gameplay/rj_gameplay/skill/intercept.py
```diff
@@ -5,57 +5,82 @@
...
+        ball_dir = vel / (np.linalg.norm(vel) + 1e-6)
+
+        ball_to_bot = pos - my_robot.pose[:1]
+        # ball_to_bot_unt = ball_to_bot / (np.linalg.norm(ball_to_bot) + 1e-6)
+        intercept_pt = np.dot(ball_to_bot, ball_dir)*ball_dir
+
+        return intercept_pt
+
+
+    def tick(self, world_state: rc.WorldState) -> RobotIntent:
```
kfu02, line 44: fix style error

rj_gameplay/rj_gameplay/skill/intercept.py
```diff
@@ -5,57 +5,82 @@
...
+            Point(x=world_state.ball.pos[0], y=world_state.ball.pos[1])
+        ]
+
+        intent.motion_command.path_target_command = [path_command]
+        intent.is_active = True
+
+        # TODO: motion planning is a lot more stable when not being spammed with repeat intents, use Action Client/Server to avoid re-requests when the intent is the same
+        return intent
 
     def is_done(self, world_state) -> bool:
```
kfu02, line 66: Again, could you add a docstring here? Seems like "check if robot has ball sense, the ball is slow, or is within a certain angle of the ball -> done" (also, why is it slow -or- within angle?)

kfu02 (changes requested):
Settle looks great. I went to test the intercept skill, but it looks like the skill is not actually in any roles? If intercept turns out not to be necessary since calling Settle within Receive does that but better, you can just strike it from this PR and put Receive inside of the Goalie Role instead of the movement logic that lives there right now.

EDIT: if intercept is unnecessary you can disregard the intercept-related comments below as well.

rj_gameplay/rj_gameplay/skill/intercept.py
HussainGynai: You are correct, the slow should be anded with the angle.

kfu02 (approved):
Looks great. Sorry I forgot to re-review. Will force-merge this in as it will be useful for the upcoming scrim (in roughly a week)

PR #1855, opened 2022-02-23
Added unassigned_role and tactic role request edge case

PotentialError:
## Description
Added an unassigned_role and logic to handle unassigned robots and tactics which have too many roles relative to available robots.
**FOR FUTURE TACTIC WRITERS:** After creating new role requests, wait until len(self.assigned_roles) == # of needed roles to continue through your tactic!

## Steps to test
1. Run `ros2 launch rj_robocup sim.launch.py`with a play and tactics that will have robots unassigned
2. Run the play in the simulator

Expected result: In the command prompt, the message "UnassignedRole created for: {robot info}" should be printed as many times as there are unassigned robots

PR #1872, opened 2022-03-14
Penalty offense prep

Yudai-8-om:
## Description
The striker will be positioned at .5m behind from the ball as a prep.
Line tactics are commented out until they are ready to use

### Test Case 1
1. Run sim
2. check the striker's position


CK-Christopher (commented):
In basic_seek, it crashes unless stp.role.cost is imported. I am not sure why though. I might just be my VM. In the future, the commented out line tactic needs to match its method header. Other than that, it looks good to me.

CK-Christopher (approved):

rj_gameplay/rj_gameplay/gameplay_node.py
```diff
@@ -8,30 +8,29 @@
```
kfu02: Can you avoid committing to gameplay_node in this PR? You can reset this file to what's in ros2 with https://stackoverflow.com/questions/2364147/how-to-get-just-one-file-from-another-branch

rj_gameplay/rj_gameplay/play/penalty_offense.py
```diff
@@ -1,75 +1,61 @@
...
-        skill_dict = {}
-        for tactic in self.tactics:
-            skill_dict.update(role_results[tactic])
-        return (skill_dict, skills)
+    ) -> List[RobotIntent]:
+
+        if self._state == State.INIT:
+            self.prioritized_tactics = [
+                prep_move.PrepMove(world_state)
+                # assume line tactic is working
```
kfu02: can you put the working line tactic in after #1874 is merged?

rj_gameplay/rj_gameplay/play/penalty_offense.py
```diff
@@ -1,75 +1,61 @@
...
+                # line_tactic.LineTactic(world_state),
+                # line_tactic.LineTactic(world_state),
+                # line_tactic.LineTactic(world_state),
+                # line_tactic.LineTactic(world_state),
+                # line_tactic.LineTactic(world_state),
+            ]
+            self.assign_roles(world_state)
+            self._state = State.PREP
```
kfu02, line 40: you should add a while loop here that checks if the number of requested tactics is actually assigned before switching the state, see this line:

https://github.com/RoboJackets/robocup-software/pull/1855/files#diff-a0f92563cb784149768f1724a5d7e152b2ce3b8042d3477f11ce84f823ddb589R111

kfu02 (changes requested):
some small suggestions, overall behaves well

rj_gameplay/rj_gameplay/play/penalty_offense.py
kfu02: (this can be done after merging this PR if necessary)

p-nayak11 (approved):
All of the robots behaved in the proper manner and the striker appropriately approaches the ball in the offensive penalty situation.

kfu02:
was used in #1887 , will be cleaned up by #1895  but I'm merging current state in now to give credit to @Yudai-8-om

PR #1874, opened 2022-03-16
Fix line tactic

Yudai-8-om:
## Description
New line tactic.
Robots will line up based on the number of robots, the start point, and the end point.

Continuation of the penalty offense

### Test Case 1
1. run sim
2. play around with start and end point

### DIFF
--- a/rj_gameplay/rj_gameplay/tactic/line_tactic.py
+++ b/rj_gameplay/rj_gameplay/tactic/line_tactic.py
@@ -1,29 +1,50 @@
-import stp
-from rj_gameplay.role import dumb_move
 from typing import List, Tuple
 
+import numpy as np
+import stp
 from rj_msgs.msg import RobotIntent
+
+from rj_gameplay.role import dumb_move
 
 
 class LineTactic(stp.tactic.Tactic):
-    """Tactic for line up play that puts all six robots in a line on the left of the field."""
+    """Tactic for line up play that puts all six robots in a line on the left of the field.
+    Assuming that the liners are more than 2"""
 
-    def __init__(self, world_state: stp.rc.WorldState):
+    def __init__(
+        self,
+        world_state: stp.rc.WorldState,
+        num_liners: int,
+        start: np.ndarray,
+        end: np.ndarray,
+    ):
         super().__init__(world_state)
 
         # compute move points
-        # TODO: make start on side of the field, so this Tactic is actually useful during penalty situations
-        start = (3.0, 0.0)
-        dy = 0.5
-        # TODO: make the # here a param instead of hardcoding for same reason as above TODO
-        self.move_points = [(start[0], start[1] + i * dy) for i in range(6)]
+        self.start = start
+        self.end = end
+        xpts = np.linspace(self.start[0], self.end[0], num_liners)
+        ypts = np.linspace(self.start[1], self.end[1], num_liners)
+        self.target_points = []
 
-        # request closest robot every pt
-        for pt in self.move_points:
-            # for some reason stp.role doesn't need to be imported?
+        for x, y in zip(xpts, ypts):
+            target = np.array([x, y])
+            self.target_points.append(target)
             self._role_requests.append(
-                (stp.role.cost.PickClosestToPoint(pt), dumb_move.DumbMove)
+                (stp.role.cost.PickClosestToPoint(target), dumb_move.DumbMove)
             )
+
+        print(self.target_points)
+
+        # TODO: make the # here a param instead of hardcoding for same reason as above TODO
+        # self.move_points = [(start[0], start[1] + i * dy) for i in range(6)]
+
+        # # request closest robot every pt
+        # for pt in self.move_points:
+        #     # for some reason stp.role doesn't need to be imported?
+        #     self._role_requests.append(
+        #         (stp.role.cost.PickClosestToPoint(pt), dumb_move.DumbMove)
+        #     )
 
         # OR hardcode certain ids to go
         # for i, pt in enumerate(self.move_points):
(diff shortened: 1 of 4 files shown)


rj_gameplay/rj_gameplay/tactic/line_tactic.py
```diff
@@ -1,30 +1,51 @@
...
+import numpy as np
+import stp
 from rj_msgs.msg import RobotIntent
 
+from rj_gameplay.role import dumb_move
+
 
 class LineTactic(stp.tactic.Tactic):
-    """Tactic for line up play that puts all six robots in a line on the left of the field."""
+    """Tactic for line up play that puts all six robots in a line on the left of the field.
```
kfu02: all six robots?

rj_gameplay/rj_gameplay/tactic/line_tactic.py
```diff
@@ -1,30 +1,51 @@
...
+            target = np.array([x, y])
+            self.target_points.append(target)
             self._role_requests.append(
-                (stp.role.cost.PickClosestToPoint(pt), dumb_move.DumbMove)
+                (stp.role.cost.PickClosestToPoint(target), dumb_move.DumbMove)
             )
 
+        print(self.target_points)
+
+        # TODO: make the # here a param instead of hardcoding for same reason as above TODO
```
kfu02: you can delete these comments

rj_gameplay/rj_gameplay/tactic/line_tactic.py
```diff
@@ -1,30 +1,51 @@
...
 
-        # request closest robot every pt
-        for pt in self.move_points:
-            # for some reason stp.role doesn't need to be imported?
+        for x, y in zip(xpts, ypts):
+            target = np.array([x, y])
+            self.target_points.append(target)
             self._role_requests.append(
-                (stp.role.cost.PickClosestToPoint(pt), dumb_move.DumbMove)
+                (stp.role.cost.PickClosestToPoint(target), dumb_move.DumbMove)
```
kfu02, line 41: since this tactic is only useful for penalty situations, you could more simply pass in a list of robot ids from the penalty play and force those robots to move into the line. right now if I move the ball around, sometimes the robot that was supposed to be the striker moves into the line.

rj_gameplay/rj_gameplay/tactic/prep_move.py
```diff
@@ -0,0 +1,53 @@
...
+
+import numpy as np
+import stp
+from rj_msgs.msg import RobotIntent
+
+from rj_gameplay.role import dumb_move
+
+
+class PrepMove(stp.tactic.Tactic):
+    """Seeks to a single point, passed in on init."""
```
kfu02: maybe change this to clarify that this is basically just for penalty offense

rj_gameplay/rj_gameplay/tactic/prep_move.py
```diff
@@ -0,0 +1,53 @@
...
+
+    def __init__(self, world_state: stp.rc.WorldState):
+        super().__init__(world_state)
+
+        self._target_pt = np.array([0.0, 0.0])
+
+        self._role_requests.append(
+            (stp.role.cost.PickClosestToPoint(self._target_pt), dumb_move.DumbMove)
```
kfu02: again, you could pass in a specific ID here instead (but if the line is prioritized ahead of this one and that's given fixed robot_ids it wouldn't make a difference)

kfu02 (changes requested):
Nice job. Good use of np.linspace. 

I have a minor concern with the use of "ClosestToPoint" as the cost for the role request, see comment. Also, could you set base branch to be your penalty offense branch instead of ros2?

rj_gameplay/rj_gameplay/tactic/line_tactic.py
Yudai-8-om: Forgot to change that, now updated.

rj_gameplay/rj_gameplay/tactic/line_tactic.py
Yudai-8-om: I agree but PickRobotbyId doesn't work due to role assignment failure almost all the time for some reason.
I tick all tactics, so it should work properly now.

rj_gameplay/rj_gameplay/tactic/line_tactic.py
```diff
@@ -49,14 +51,15 @@ def is_done(
         self,
         world_state: stp.rc.WorldState,
     ) -> bool:
-        return all([role.is_done(world_state) for role in self.assigned_roles])
+        return False
+        # return all([role.is_done(world_state) for role in self.assigned_roles])
```
kfu02: delete this line

rj_gameplay/rj_gameplay/tactic/prep_move.py
```diff
@@ -39,7 +39,8 @@ def is_done(
         self,
         world_state: stp.rc.WorldState,
     ) -> bool:
-        return all([role.is_done(world_state) for role in self.assigned_roles])
+        return False
+        # return all([role.is_done(world_state) for role in self.assigned_roles])
```
kfu02: delete this line

kfu02 (approved):
LGTM, some small nits

rj_gameplay/rj_gameplay/tactic/line_tactic.py
```diff
@@ -1,37 +1,39 @@
...
-
-        # request closest robot every pt
-        for pt in self.move_points:
-            # for some reason stp.role doesn't need to be imported?
+        self.start = start
+        self.end = end
+        xpts = np.linspace(self.start[0], self.end[0], num_liners)
+        ypts = np.linspace(self.start[1], self.end[1], num_liners)
+        self.target_points = []
+
```
HussainGynai, line 36: Maybe have a check to make sure the robots can fit along the line, and throw an error if they can't. Like if the start and end point are too close or there are too many robots to fit between. You can use the diameter of the robots for calculating this.

HussainGynai (changes requested):
Looks mostly good, just add the check to make sure the lineup is possible

PR #1875, opened 2022-03-16
Analysis with score

oddlyspookycherry:
## Description
Includes score in situation analysis


rj_gameplay/rj_gameplay/gameplay_node.py
```diff
@@ -147,6 +156,12 @@ def set_play_state(self, play_state: msg.PlayState):
 
     def set_match_state(self, match_state: msg.MatchState):
         self.match_state = match_state
+    
+    def set_our_team_info(self, our_team_info: msg.TeamInfo):
```
kfu02, line 164: add docstring here

rj_gameplay/rj_gameplay/gameplay_node.py
```diff
@@ -147,6 +156,12 @@ def set_play_state(self, play_state: msg.PlayState):
...
+        self.our_team_info = our_team_info
+    
+    def set_their_team_info(self, their_team_info: msg.TeamInfo):
```
kfu02, line 170: add docstring here

kfu02 (changes requested):
Hey, looks like you have some style errors. Can you run `./util/ubuntu-setup` to make sure you have the pre-commit hooks installed? Then any time you commit new changes style autoformatter will run for you. (It will say "checks failed" and run the autoformatter. Just re-add and commit the same files to fix.)

Also, there are a couple places where you should add a docstring.

rj_gameplay/rj_gameplay/gameplay_node.py
```diff
@@ -153,10 +161,23 @@ def set_play_state(self, play_state: msg.PlayState):
     def set_match_state(self, match_state: msg.MatchState):
         self.match_state = match_state
 
+    """
+    Subscription callback for setting information for our team
+    """
```
kfu02: move the docstring to under the function declaration, e.g.

```
def set_our_team_info(...):
    """
    docstring here
    """
```

rj_gameplay/rj_gameplay/gameplay_node.py
```diff
@@ -153,10 +161,23 @@ def set_play_state(self, play_state: msg.PlayState):
...
+    def set_our_team_info(self, our_team_info: msg.TeamInfo):
+        self.our_team_info = our_team_info
+
+    """
+    Subscription callback for setting information for enemy team
+    """
+    def set_their_team_info(self, their_team_info: msg.TeamInfo):
```
kfu02, line 170: same docstring move as above

rj_gameplay/stp/utils/world_state_converter.py
```diff
@@ -179,6 +182,7 @@ def build_game_info(
                 play_state_msg.placement_point.y,
             ]
         ),
+        np.array([our_team_msg.score, their_team_msg.score]),
```
kfu02: this can just be a normal list (change the typing in stp/rc.py to match, too)
```suggestion
        [our_team_msg.score, their_team_msg.score],
```

rj_gameplay/stp/rc.py
```diff
@@ -671,6 +675,13 @@ def their_restart(self) -> bool:
 
         return not self.__our_restart
 
+    @property
+    def game_score(self) -> np.array:
+        """
+        :return: The game score
+        """
+        return self.__game_score  # 0 for our score, 1 for opponent score
```
kfu02: this comment is a bit confusing, can you instead clarify that it returns a 2-element array where the first element is our score and the second is their score? feel free to use your own wording but something to that effect would be helpful, like `:return: Score of each team in format [our_score, their_score]` would be good as well

kfu02: also, put it in the docstring above, not as a comment next to the return
```suggestion
        return self.__game_score
```

kfu02 (changes requested):
Great start! However, when I go to the bottom of gameplay node and print out the game_info object, it tells me game_info is None. I checked and it's because the "our_team_info" and "their_team_info" fields that build_game_info relies on are always None. Does this happen on your end too?

kasohrab (changes requested):
yeah looks like a good start! Adding to what kevin said, it would be better if instead of having to keep track of what index holds what, we should abstract that away for the rest of gameplay. So create a property `our_score` that returns `self.__game_score[0]` and one for `their_score`.

PR #1876, opened 2022-03-16
simplifying situation analysis.

kasohrab:
## Description
Title. Still needs to be cleaned up and finalized but should work

## Associated Issue
Issue #???

## Design Documents
[Link](link-to-design-doc)

### Test Case 1
1. Step 1
2. Step 2
3. Step 3

Expected result: ???

## (Optional) Sub-issues (for drafts)
_Note: if you find yourself breaking this PR into many smaller features, it may make sense to break up the PR into logical units based on these features._
- [ ] Step 1
- [ ] Step 2

PR #1877, opened 2022-03-16
Delete unused gameplay files + fix test play bug

kfu02:
## Description
Put gameplay in a clean state for upcoming comp. 

**Also, fix test_play/play_selector discrepancy introduced from #1870 . When given a test play, GameplayNode will now override play selector with that test play.**

- delete unused actions (waiting on ROS ActionServer)
- delete outdated plays, tactics, and skills
- add a TODO to update Mark skill (still on old interface)
**- also fixed a mark skill bug**
- fix typing in pass tactic 
- rm MAX_KICK_SPEED file 

## Steps to test
1. Run all working plays. As of 3/29, this includes basic_defense, basic_offense (was basic122 before #1876), and keepaway. (Can do this with test play var in GameplayNode.)

Expected result: All plays behave as expected. No new errors.


kfu02:
Unused STP files deleted, but not ready for review because I would like to leave comments explaining what some plays that we haven't rewritten should do (so we can rewrite more quickly)

N8BWert (dismissed):
Not sure why, but no matter what play I run it always runs basic_offense.

kfu02:
@N8BWert fixed now

p-nayak11 (approved):
All working plays ran the way they were supposed to.

kasohrab (approved):
lgtm

PR #1880, opened 2022-04-03
Formations Class

p-nayak11:
## Description
A formations class to divide the field into different regions for the seekers to go to. The current formation that is in the class is the X formation with 2 seekers up front, 1 seeker in the middle, and seekers in the back. The tactic only calls three seekers so the first two regions are being occupied. I will create the new diamond formation and commit to this branch.

**EDIT**: Diamond Formations class has been added. This formation creates a striker that is direct in front of the opponent goal, a left and right winger, a center, and a sweeper behind for a pass back. 4 seekers are being used for this formation currently. 

**EDIT**: Formations class is now an abstract superclass with abstract methods that are implemented by the subclasses X_formation and Diamond_formation. This is a cleaner implementation of the formations. Now formations are only used in basic_offense which calls the type of formation and number of seekers. 
### Test Case 1
1. Run sim to make sure nothing crashes

Expected result: Four robots should move after ball is moved.


.pre-commit-config.yaml
```diff
@@ -6,7 +6,7 @@ repos:
     name: isort (python)
```
HussainGynai: I don't think you meant to push changes to this file.

HussainGynai (approved):
LGTM!

.pre-commit-config.yaml
p-nayak11: Actually, I think it might be automatically updated after the Ubuntu utils was recently updated.

kfu02: ^ everyone has to commit this change, otherwise precommit won't let them commit

rj_gameplay/stp/rc.py
```diff
@@ -421,6 +421,30 @@ def opp_defense_area_coordinates(self) -> List:
         ]
         return opp_defense_area
 
+    @property
+    def our_goal_post_coordinates(self) -> List:
+        """
+        Conveniance function for getting our goal post coordinates
```
kfu02:
```suggestion
        Convenience function for getting our goal post coordinates.
```

rj_gameplay/stp/rc.py
```diff
@@ -421,6 +421,30 @@ def opp_defense_area_coordinates(self) -> List:
...
+        our_goal_post_coord = [
+            [-self.__goal_width_m / 2, 0],
+            [self.__goal_width_m / 2, 0],
+        ]
+        return our_goal_post_coord
+
+    @property
+    def their_goal_post_coordinates(self) -> List:
+        """
+        Conveniance function for getting their goal post coordinates
```
kfu02:
```suggestion
        Convenience function for getting their goal post coordinates.
```

rj_gameplay/stp/rc.py
kfu02: also could you add a nice return: here like below?

rj_gameplay/stp/utils/formations.py
```diff
@@ -0,0 +1,141 @@
...
+                0 + 2 * RobotConstants.RADIUS,
+                self.y_quarter,
+            ),
+        ]
+
+        return X_formation
+
+    def get_centroid(self, region: List) -> np.ndarray:
+        """
+        Conveniance function for getting center of a region
```
kfu02:
```suggestion
        Convenience function for getting center of a region.
```

rj_gameplay/rj_gameplay/tactic/basic_seek.py
```diff
@@ -1,100 +1,33 @@
...
+
+from rj_gameplay.role import seeker
 
 
 class BasicSeek(stp.tactic.Tactic):
     """Seeks to a single point, passed in on init."""
 
-    def __init__(self, world_state: stp.rc.WorldState, num_seekers: int):
+    def __init__(
+        self, world_state: stp.rc.WorldState, num_seekers: int, formations: List
```
kfu02: should pass in your new Formations class here

rj_gameplay/stp/utils/formations.py
```diff
@@ -0,0 +1,141 @@
+from typing import List
+
+import numpy as np
+import stp
+from stp.utils.constants import RobotConstants
+
+
+class Formations:
```
kfu02: this is written a bit like a static method, should have types of Formations subclass some abstract Formation class and pass that in instead

kfu02 (changes requested):
Requesting some Formation class design changes that we discussed IRL.

rj_gameplay/stp/formations/diamond_formation.py
```diff
@@ -0,0 +1,77 @@
+from typing import List
+
+import numpy as np
+
+import stp
+from stp.utils.constants import RobotConstants
+
+
+class Diamond_formation(stp.formations.Formations):
```
kfu02:
```suggestion
class DiamondFormation(stp.formations.Formations):
```

nit: this is how the rest of our codebase's Python classes are styled, would be nice to match (if you do this don't forget to change the references to this class)

rj_gameplay/stp/formations/diamond_formation.py
```diff
@@ -0,0 +1,77 @@
...
+                self.box_xright,
+                self.our_def_area_short_dist_m + 2 * RobotConstants.RADIUS,
+                self.y_quarter,
+            ),
+        ]
+
+        return diamond_formation
+
+    @property
+    def get_centroid(self) -> List:
```
kfu02:
```suggestion
    def get_centroids(self) -> List:
```

since it returns multiple centroids

kfu02: (again don't forget to change references above)

rj_gameplay/stp/formations/__init__.py
```diff
@@ -0,0 +1,50 @@
...
+        self.center_yup: float = (
+            world_state.field.center_field_loc[1] + world_state.field.center_diameter_m
+        )
+        self.center_ydown: float = (
+            world_state.field.center_field_loc[1] - world_state.field.center_diameter_m
+        )
+
+    @property
+    @abstractmethod
+    def get_centroid(self) -> List:
```
kfu02:
```suggestion
    def get_centroids(self) -> List:
```

since it returns multiple centroids

rj_gameplay/stp/formations/x_formation.py
```diff
@@ -0,0 +1,77 @@
+from typing import List
+
+import numpy as np
+
+import stp
+from stp.utils.constants import RobotConstants
+
+
+class X_formation(stp.formations.Formations):
```
kfu02:
```suggestion
class XFormation(stp.formations.Formations):
```

same nit as above with Diamond_formation vs DiamondFormation

kfu02 (changes requested):
Hey, this looks much cleaner than before, nice work. I left a few suggestions for how to name things that I believe should be done, but talk to me if you disagree; none of them affect the functionality or design. After we hash those out I can approve this.

In the future we should account for blocking our own seekers somehow, but that's a different PR. (See robot 5 blocking robot 3 below.)

Screenshot from 2022-04-06 15-03-08

p-nayak11:
> Hey, this looks much cleaner than before, nice work. I left a few suggestions for how to name things that I believe should be done, but talk to me if you disagree; none of them affect the functionality or design. After we hash those out I can approve this.
> 
> In the future we should account for blocking our own seekers somehow, but that's a different PR. (See robot 5 blocking robot 3 below.)
> 
> Screenshot from 2022-04-06 15-03-08

I saw this earlier. I will work on it later.

kfu02 (approved):
LGTM

PR #1881, opened 2022-04-03
Launch file enhancements

kasohrab:
## Description
basically just adds a bunch of different launch file configs. Long term we can build all and dynamically choose which ones to run using ifconditions, but for now this is better than commenting out nodes. 

To see how to use ifconditions, check out the use_grsim flag in sim.launch.py

Also sim2play will not use grSim anymore.

## Steps to test

Expected result: Can switch between different setups using the new make flags added in the makefile.


kfu02:
we may need to fix all of our launch files in general, @fourpenny and I just found out that vision_receiver.launch.py and soccer.launch.py take in hardcoded filenames for what config to use (meaning to go from sim-vision to real-vision we have to change both files)

p-nayak11 (approved):
LGTM

PR #1890, opened 2022-04-26
Rrt off lines

kfu02:
## Description

EDIT: fix one issue, create more issues...do not review this yet

During the last scrim we noticed issues where our RRT-based planner took unnecessarily wide lines around obstacles. This was especially obvious in our wall, when sometimes robots that only needed to move in a couple straight line segments would take very long paths, leaving holes in our wall to be scored on.

This PR addresses that by making a piecewise trajectory around the first obstacle before resorting to full RRT. Since this trajectory is found deterministically and tries to cut as narrow a path as possible, it should result in better motion planning behavior in crowded situations than RRT. [See this comment for more details.](https://github.com/RoboJackets/robocup-software/pull/1890/files#diff-dd3f19ffa77540702382cf850fceac2bb94109f100445678138af56441a8985fR12)

### Test Case 1
1. Run normal plays and see what happens.

Expected result: soccer should not crash, robots move to points on cleaner lines.

## Test Procedure

All timings are based on the time given by our UI (see screen recordings). Timer starts when blue UI trajectories are drawn and stops when yellow UI trails disappear (indicating no more movement). If a hole in the wall is left, stop the timer but note that.

1. Place ball on intersection of right sideline and midfield line, wait for robots to form wall.
3. Place ball on left sideline at x=3.0, y=2, time it
4. Repeat 1
5. Place opp robot at x=0, y=1.2
6. Repeat 2, time it

## Test Results

In general, the robots move more cleanly around the opponents, but are more likely to run into each other. A 10-man wall is used here to stress-test: with fewer robots both planners look better. The crashing issue (for wallers at least) can be resolved with a dedicated wall planner, where each robot is given the trajectory along the walling arc instead of a straight line to their destination.

[`ros2`](https://drive.google.com/file/d/13eyAMNUk51dmign3BCn-prh2NY1BfRfq/view?usp=sharing)
2) 7.57 s
4) 7.53 s
5) 9.40 s
**avg: 8.17 s**
every trial lost 1 robot from the wall

[`rrt-off-lines`](https://drive.google.com/file/d/15UBhRpwILHyaSh3us_86r78M4Z0FSmmn/view?usp=sharing)
2) 5.73 s
4) 3.66 s
5) 4.48 s
**avg: 4.62 s**
no loss of wallers in any trial


rj_geometry/include/rj_geometry/shape.hpp
```diff
@@ -10,6 +10,9 @@ class Segment;
 /**
  * The shape class provides the interface to all shapes that are subclasses.
  *
+ * TODO(Kevin): do we use boost python anymore? more likely this docstring
+ * was never updated (4/29/21 at time of reading)
```
kfu02:
```suggestion
 * was never updated (4/29/22 at time of reading)
```

I forgot the year...

soccer/src/soccer/planning/planner/line_kick_planner.cpp
```diff
@@ -164,8 +164,10 @@ Trajectory LineKickPlanner::plan(const PlanRequest& plan_request) {
             }
         }
 
+        // TODO: reenable this planner by modifying trajectory_hits_static() call
```
kfu02:
```suggestion
        // TODO(Kevin): reenable this planner by modifying trajectory_hits_static() call
```

soccer/src/soccer/planning/planner/path_target_planner.cpp
```diff
@@ -32,8 +32,14 @@ Trajectory PathTargetPlanner::plan(const PlanRequest& request) {
 
     // Debug drawing
     if (request.debug_drawer != nullptr) {
+        // TODO(Kevin): delete debug_drawer.hpp and see what happens, since we use
+        // ros_debug_drawer.hpp and that's incredibly confusing
         request.debug_drawer->draw_circle(Circle(goal_point, static_cast<float>(draw_radius)),
                                           draw_color);
+        // TODO: DELETE ME!
```
kfu02:
```suggestion
        // TODO(Kevin): DELETE ME!
```

soccer/src/soccer/planning/primitives/create_path.hpp
```diff
@@ -1,27 +1,43 @@
...
+ *   1. Straight line to goal.
+ *   2. Straight line to a point around the first obstacle, plus a straight
+ *   line from that point to the goal. Tightest line that fits this criteria
+ *   and doesn't collide with any other obstacles is chosen. (This is
+ *   iteratively generated: see the .cpp file for details.)
+ *   3. RRT-given path to goal.
+ *
+ * This is done under the assumption that straight trajectories are more
+ * efficient than curves (though technically, all our paths are smoothed to Bezier
+ * curves, it seems to be better for motion control).
```
kfu02, line 23:
```suggestion
 * curves, it seems to be better for motion control to start from straight lines).
```

kfu02: nvm

kfu02:
Split into #1906, #1907, and #1908, but without the whole "off lines" part of this...to be determined whether that's actually useful. I will adjust the dynamic obstacles first instead and see how that goes with our existing RRT/straight line stuff. (Luckily closed PRs can still be viewed.)

Issue #1889, opened 2022-04-20
Finalize new launch setup


kasohrab:
## Description

#1881 improves our setup to make it easier for new people, but I don't like all the extra launch files. We should be able to add all possible launch nodes to a single launch file and use ROS2 ifconditions to not include the ones we are not using (to make our setup more modular). Currently it would be a pain to add a new node.

I'm just making an issue so I don't forget to do this.

Issue #1592, opened 2020-10-26
Race condition in startup sequence


kylestach:
## Description
There is a race condition in the startup sequence for sim2play in the ROS version of our code. It comes from the propagation of the `blue` flag.

## Steps to reproduce
1. Run `make run-sim2play` on the ROS branch

Result: (intermittent) crash on startup with an error in the radio as it attempts to bind to a port already bound by the other instance of soccedr

- [X] This bug has been reproduced by someone else

## Additional information
The best approach to fixing this will likely be to make sure every node that needs to know if we're on blue or yellow either:
 - Waits for the result to come in from the config server
 - Gets the team color as a flag on the command line

PR #1892, opened 2022-04-29
Delete grSim and more launch file improvements

kasohrab:
## Description
title. I still kept internal ref default for now becuase that is what erForce does too. Also allow us to switch sides in sim launch file. And other improvements (see changes to launch files). Can choose external ref or network radio instead of defaults by using ros2 launch syntax use_*:=False. See launch file for what the flags are called or makefile for examples

You can use that or if anyone wants to add more presets using this format to the makefile go ahead.

And finally, this pr fixes the sim2play race condition.

## Associated Issues
#1889 #1592

## Steps to test
Try a bunch of different launch setups to make sure it still works

Expected result: still works, now it is also no longer possible to use grSim.


launch/sim2play.launch.py
```diff
@@ -46,9 +52,18 @@ def generate_launch_description():
...
     )
 
     blue = GroupAction([PushRosNamespace("blue"), soccer_blue])
 
-    return LaunchDescription([stdout_linebuf_envvar, yellow, blue])
+    return LaunchDescription(
+        [
+            stdout_linebuf_envvar,
+            yellow,
+            TimerAction(period=3.0, actions=[blue]),
```
kasohrab, line 76: dumb fix for the sim2play non-deterministic startup problem. A better solution would be to create a custom action that waits for yellow to launch, then blue can launch. But since this works I don't feel like making it better lol.

makefile
```diff
@@ -58,23 +58,18 @@ again:
...
-	
-run-external:
-	ros2 launch rj_robocup external_sim.launch.py use_grsim:=False
-	
+run-sim-ex:
+	ros2 launch rj_robocup sim.launch.py use_internal_ref:=False
+
 run-real:
-	ros2 launch rj_robocup real_soccer.launch.py
+	ros2 launch rj_robocup soccer.launch.py config_yaml:=real.yaml
```
kasohrab: I can't test this but I'm guessing it'll work

launch/sim.launch.py
```diff
@@ -11,32 +11,27 @@
...
-from launch.substitutions import LaunchConfiguration, PythonExpression
+from launch.substitutions import LaunchConfiguration
 
 
 def generate_launch_description():
     bringup_dir = Path(get_package_share_directory("rj_robocup"))
     launch_dir = bringup_dir / "launch"
 
+    config_yaml = LaunchConfiguration("config_yaml", default="sim.yaml")  # noqa: F841
+    use_internal_ref = LaunchConfiguration(  # noqa: F841
```
kfu02: tested the following:
`ros2 launch rj_robocup sim.launch.py use_internal_ref:=False`

and launched ssl-game-controller as usual, but this did not work for me. Switching False to True works fine so I know it's not a syntax thing.

makefile
```diff
@@ -65,7 +65,7 @@ run-sim-ex:
 	ros2 launch rj_robocup sim.launch.py use_internal_ref:=False
 
 run-real:
-	ros2 launch rj_robocup soccer.launch.py config_yaml:=real.yaml
+	ros2 launch rj_robocup soccer.launch.py config_yaml:=real.yaml use_internal_ref:=False use_sim_radio:=False
```
kasohrab, line 74: this should be right actually

launch/sim.launch.py
```diff
@@ -11,32 +11,27 @@
...
+        "use_internal_ref", default="True"
+    )
+    use_sim_radio = LaunchConfiguration("use_sim_radio", default="True")  # noqa: F841
     team_flag = LaunchConfiguration("team_flag", default="-b")
-    use_grsim = LaunchConfiguration("use_grsim", default="True")
     ref_flag = LaunchConfiguration("ref_flag", default="-noref")
-    headless_flag = LaunchConfiguration("headless_flag", default="")
+    direction_flag = LaunchConfiguration("direction_flag", default="plus")  # noqa: F841
```
kfu02: it would be nice if this could be translated to its semantic meaning, e.g.

`team_color_flag, default='yellow'`

which is then translated to the right direction. A checkbox in the UI would actually be fully ideal here.

launch/sim2play.launch.py
kfu02: honestly this works great, good-enough hack imo

kfu02 (changes requested):
I'm getting the "team color is nullptr" thing when running external ref, am I missing something?

(This occurs both on `run-sim-ex` and the full ros2 launch command)

launch/sim.launch.py
kfu02: (but you don't have to do the UI thing, we can bundle that with other UI changes)

launch/sim.launch.py
kasohrab: Should be fixed now

launch/sim.launch.py
kasohrab: We can do this with ui changes in the summer if that's ok with you

launch/soccer.launch.py
```diff
@@ -90,14 +118,24 @@ def generate_launch_description():
...
+        condition=IfCondition(PythonExpression([use_internal_ref])),
         package="rj_robocup",
         executable="internal_referee_node",
         output="screen",
         parameters=[config],
         on_exit=Shutdown(),
     )
 
+    external_ref_receiver = Node(
+        condition=IfCondition(PythonExpression(["not ", use_internal_ref])),
```
kasohrab, line 131: trial and error led me to discover this weird syntax (no documentation for this stuff)

kfu02 (approved):
Looks good. I can't test "real radio" remotely but the other flags check out.

launch/soccer.launch.py
kfu02: we're just like ROS!

PR #1887, opened 2022-04-18
scrim-hotfixes

kfu02:
## Description

Scrim 2021 hotfixes. ~~See commit log (I actually wrote decent commit messages for once).~~

Go through each commit and merge manually into ros2

## Associated Issue
Issue #???

## Design Documents
[Link](link-to-design-doc)

### Test Case 1
1. Step 1
2. Step 2
3. Step 3

Expected result: ???

## (Optional) Sub-issues (for drafts)
_Note: if you find yourself breaking this PR into many smaller features, it may make sense to break up the PR into logical units based on these features._
- [ ] Step 1
- [ ] Step 2

PR #1895, opened 2022-05-02
Scrim 2022 merge

kfu02:
## Description
Merges in all useful changes from #1887 . See that PR for [list of changes](https://github.com/RoboJackets/robocup-software/pull/1887#pullrequestreview-959588745) I chose to drop from the virtual scrim.

## Steps to test
Run ER-force sim + ssl-game-controller + `make run-sim-ex` and play through some different situations.

Expected result: Nothing crashes, no excess debug prints are shown, gameplay looks more or less as it did during scrim 2022.


soccer/src/soccer/radio/sim_radio.hpp
```diff
@@ -41,6 +42,7 @@ class SimRadio : public Radio {
     boost::asio::ip::udp::endpoint robot_control_endpoint_;
 
     std::vector<char> buffer_;
+    std::vector<RJ::Time> last_sent_diff;
```
kasohrab:
```suggestion
    std::vector<RJ::Time> last_sent_diff_;
```

config/real.yaml
```diff
@@ -38,9 +38,10 @@
         robot_radius: 0.09
     use_sim_time: false
 
-/internal_referee:
+/external_referee:
   ros__parameters:
     team_name: RoboJackets
+    interface: "172.19.0.0"
     use_sim_time: false
```
kasohrab, line 46:
```suggestion

/internal_referee:
  ros__parameters:
    team_name: RoboJackets
    use_sim_time: false
```

config/real.yaml
```diff
@@ -97,13 +98,15 @@
 /radio:
```
kasohrab: since I changed the name of this, maybe it should now be 2 of them? one sim_radio, one network_radio?

config/real.yaml
kasohrab: for same reason as below.

rj_gameplay/rj_gameplay/play/penalty_defense.py
```diff
@@ -18,26 +17,27 @@ class State(Enum):
...
-    """
-    Play consists of:
-                    -5 robots lined up on the side
-                    -1 goalie
-    The goalie will defend the opponents penalty kick.
-    """
-
     def __init__(self):
         super().__init__()
+
```
kasohrab:
```suggestion
```

soccer/src/soccer/radio/sim_radio.cpp
```diff
@@ -87,12 +93,23 @@ void SimRadio::send(int robot_id, const rj_msgs::msg::MotionSetpoint& motion,
     // Send a sim packet with a single robot. The simulator can handle many robots, but our commands
     // may come in at different times and it should be fine to just recalculate like this.
     // TODO(Kyle): Verify that this is okay.
+    if (RJ::now() - last_sent_diff.at(robot_id) < 5.1ms) {
+        return;
+    }
+    last_sent_diff.at(robot_id) = RJ::now();
```
kasohrab:
```suggestion
    last_sent_diff_.at(robot_id) = RJ::now();
```

soccer/src/soccer/radio/sim_radio.cpp
```diff
@@ -62,6 +65,9 @@ SimRadio::SimRadio(bool blue_team)
       blue_team_(blue_team),
       socket_(io_service_, ip::udp::endpoint(ip::udp::v4(), blue_team ? kSimBlueStatusPort
                                                                       : kSimYellowStatusPort)) {
+    for (int i = 0; i < kNumShells; ++i) {
+        last_sent_diff.emplace_back(RJ::now());
```
kasohrab:
```suggestion
        last_sent_diff_.emplace_back(RJ::now());
```

soccer/src/soccer/radio/sim_radio.cpp
```diff
@@ -87,12 +93,23 @@ void SimRadio::send(int robot_id, const rj_msgs::msg::MotionSetpoint& motion,
     // Send a sim packet with a single robot. The simulator can handle many robots, but our commands
     // may come in at different times and it should be fine to just recalculate like this.
     // TODO(Kyle): Verify that this is okay.
+    if (RJ::now() - last_sent_diff.at(robot_id) < 5.1ms) {
```
kasohrab:
```suggestion
    if (RJ::now() - last_sent_diff_.at(robot_id) < 5.1ms) {
```

rj_gameplay/rj_gameplay/play/basic_offense.py
```diff
@@ -79,6 +84,12 @@ def tick(
             if seek_tactic.is_done(world_state):
                 self._state = State.INIT_PASS
             """
+            if self._striker_st_time is None:
+                self._striker_st_time = time.time()
+            if (time.time() - self._striker_st_time) > 5:
+                print("striker failed, reassigning")
```
kasohrab, line 92: not part of this pr ofc, but we should move this to ui or actual logging instead of just printing later.

kasohrab (changes requested):
Yeah this is great! I just left some small changes.

rj_gameplay/rj_gameplay/play/basic_defense.py
```diff
@@ -36,8 +34,14 @@ def tick(
 
         if self._state == State.INIT:
             self.prioritized_tactics.append(goalie_tactic.GoalieTactic(world_state, 0))
-            self.prioritized_tactics.append(wall_tactic.WallTactic(world_state, 3))
-            self.prioritized_tactics.append(nmark_tactic.NMarkTactic(world_state, 2))
+            num_wallers = min(3, len(world_state.our_visible_robots) - 1)
+            self.prioritized_tactics.append(
+                wall_tactic.WallTactic(world_state, num_wallers)
```
p-nayak11, line 39: We still need to change the position of the line for this play, but that would be in a different PR.

rj_gameplay/rj_gameplay/play/penalty_defense.py
```diff
@@ -18,26 +17,27 @@ class State(Enum):
...
     def tick(
         self,
         world_state: stp.rc.WorldState,
     ) -> List[RobotIntent]:
+
         if self._state == State.INIT:
             self.prioritized_tactics.append(goalie_tactic.GoalieTactic(world_state, 0))
-            self.prioritized_tactics.append(line_tactic.LineTactic(world_state, 5))
+            num_liners = len(world_state.our_visible_robots) - 1
+            start_pt = np.array([-3.0, 0.5])
```
p-nayak11, line 32: We need to change the start point and end point for the line tactic, but that is in a different PR.

rj_gameplay/rj_gameplay/play/penalty_offense.py
```diff
@@ -3,55 +3,73 @@
...
         if self._state == State.INIT:
-            self.prioritized_tactics = [
-                prep_move.PrepMove(world_state),
-                line_tactic.LineTactic(
-                    world_state, 5, np.array([2.0, 2.0]), np.array([-2.0, 2.0])
-                ),
-            ]
+            self.prioritized_tactics.append(striker_tactic.StrikerTactic(world_state))
+            num_liners = len(world_state.our_visible_robots) - 1
+            start_pt = np.array([-2.0, 1.5])
```
p-nayak11, line 36: Similar issue as the penalty defense, where the start and end point need to changed so it depends on the ball rather than the side of the field.

rj_gameplay/rj_gameplay/play/penalty_offense.py
```diff
@@ -3,55 +3,73 @@
...
+    def tick(
+        self,
+        world_state: stp.rc.WorldState,
+    ) -> List[RobotIntent]:
+
+        if self._state == State.INIT:
+            self.prioritized_tactics.append(prep_move.PrepMove(world_state))
+            num_liners = len(world_state.our_visible_robots) - 1
+            start_pt = np.array([-2.0, 1.5])
+            end_pt = np.array([2.0, 1.5])
```
p-nayak11, line 66: Similar issue as the penalty defense, where the start and end point need to changed so it depends on the ball rather than the side of the field.

rj_gameplay/rj_gameplay/skill/pivot.py
```diff
@@ -53,6 +48,7 @@ def is_done(self, world_state: rc.WorldState) -> bool:
         if self.robot is None:
             return False
         angle_threshold = self.threshold
+        angle_threshold = 0.05
```
p-nayak11, line 51: Is there a reason why there are two different angle_thresholds?

p-nayak11 (approved):
I left a couple of reminders and a question. Overall, everything works the way it is supposed to. LGTM.

rj_gameplay/rj_gameplay/skill/pivot.py
kfu02: no, will fix! (in the future you can put all of your comments into one review btw, by leaving a bunch of comments and hitting "submit review" at the end)

kfu02:
note to self: drop the name "basic_" from play selector, defense, offense (the intent is to have multiple offense/defense plays but since we have exactly one, it's a bit of a useless designation)

makefile
```diff
@@ -166,7 +169,6 @@ endif
 # Restyles all C++ (Clang formatter) and Python files (Black formatter) excluding files in the external and build folders
```
kfu02: is there a way to only run black on python files? I had issues where black and clang-format disagreed, resulting in an inability to commit

otherwise this is an okay solution because out pre-commit hooks force us to conform to both black and pretty-lines separately (which works when they don't conflict)

```suggestion
# Restyles all C++ (Clang formatter) excluding files in the external and build folders. For Python, run black rj_gameplay.
```

rj_gameplay/rj_gameplay/play/basic_offense.py
kfu02: comment added

rj_gameplay/rj_gameplay/play/penalty_defense.py
kfu02: put it on the board!

rj_gameplay/rj_gameplay/play/penalty_offense.py
kfu02: put it on the board!

config/real.yaml
kfu02: added a TODO, wasn't confident exactly what you meant by this? plus I feel that this change belongs in a more radio-relevant PR

kasohrab (approved):
lgtm (assuming style will be fixed)

PR #1898, opened 2022-05-21
fix ubuntu-setup by moving git installs out

kfu02:
## Description
When installing our repo on a new machine, I realized why we had to git submodule update manually before after running ./util/ubuntu-setup. Turns out git stopped allowing their installs (including pre-commit) to be done by sudo. This PR fixes that by making a new util install script.

Updates:
 * changes default pip=22.1
 * requires new users to run ./util/git-setup after ubuntu-setup
 * fixes CI pipeline to account for this

## Steps to Test
 * delete local copy of this repo (MAKE SURE YOU DON'T HAVE ANY UNPUSHED CHANGES FIRST)
 * re-clone
 * checkout `fix-ubuntu-setup`
 * `./util/ubuntu-setup && ./util/git-setup`
 * `make clean && make perf`
 * `make run-sim`

Expected result: should work. (I have tested this on two machines, however, neither were fully clean when I tested. I suppose we'll find out when new members join in the fall.)


util/git-setup
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash -e
+
+# git does not appreciate installing things as sudo, so this script is for
+# a couple git-related installs
+
+# install pre-commit hooks
+echo "-- Installing pre-commit hook"
+pre-commit install
+pre-commit autoupdate
```
kasohrab:
```suggestion
# install pre-commit hooks if not already installed
if ! command -v pre-commit &> /dev/null
then
	echo "-- Installing pre-commit hook"
	pre-commit install
	pre-commit autoupdate
fi
```

kasohrab (changes requested):
looks good! just one small change, so that people can run the git setup script more than once without it failing (if they want to update submodules later down the line using the same script).

util/git-setup
kfu02: fancy

kasohrab (approved):

PR #1899, opened 2022-05-30
Test play improvements

### LINKED ISSUE #1896: More Configurable Test Play
## Description
Making an issue again cause it helps me remember to do it eventually. 

As we learned from the scrim, configurable params are great.

We should be able to swap between test plays via cli for testing. We already have a python side param client we can use for this or it can just be a gameplay node param since it doesn't need to be shared. ~~At the same time we should resolve #1666.~~ Actually following yagni, let's just make the test play configurable for now. 

## Requirements
 - Can switch between test plays with cli
 - Then add this to the UI
 - ~~(and other params like plus/minus and other configurable values).~~ <- future pr should do this

## Design -- Sources

The design should basically be:
Test play is a gameplay node param string.
Test play is a string topic which the ui will publish to gameplay node.
Must also add a callback to gameplay node to reject test plays if they do not exist in a set of valid plays.
This way we don't pollute global param server when only these two nodes need it. Also it allows the the ui to update with a gameplay node subscription callback.

User can change test play with ui or use cli (ui will just be a wrapper for that ros2 param cli cmd).

https://github.com/ros2/launch/blob/master/launch/launch/actions/opaque_function.py
https://github.com/jrgnicho/collaborative-robotic-sanding/blob/3902e4f0e76bde226b18a997fd60fc30e1961212/crs_application/launch/perception.launch.py#L21
https://docs.ros.org/en/foxy/Tutorials/Using-Parameters-In-A-Class-Python.html#write-the-python-node

kasohrab:
## Description
See issue below for description.

## Associated Issue
Resolves #1896 

### Test Case 1
~~1. note the required format of the plays (see the plays set close to the top of gameplay node in its constructor).~~
2. go to testing tab in ui, ~~add play in that format.~~
3. click the desired play in the top table, and hit run.
4. all of them will work (except lineup because it's broken rn)

I did not implement the next button. It would be easy to set up later though. 
I also did not use the python local param server (I commented it out, as I think it's overengineering for now, but it could be important later)
Later we can easily just at launch populate all valid plays into the selected plays list.

Also this pr deletes some unused ui stuff.
But at this point because it is functional, this pr is closed to new features.

Expected result: can change test play at runtime

Someone should double check trying to run invalid formats, non-existent plays, etc.. The code should handle all those cases.

Also you can use `ros2 param set /gameplay_node test_play formatted_name`


rj_gameplay/rj_gameplay/gameplay_node.py
```diff
@@ -479,6 +499,21 @@ def tick_override_actions(self, world_state) -> None:
...
+        """
+        Checks all gameplay node parameters to see if they are valid.
+        Right now test_play is the only one.
+        It should be a string in the plays set.
+        """
+        rejected_parameters = (
+            param
+            for param in parameter_list
+            if param.name == "test_play"
+            and param.type_ is Parameter.Type.String
```
kasohrab: make all-caps and add rclpy.

kfu02 (commented):
Once #1898 is merged, could you merge in ros2? I can't test this on my pip=22.1 machine :(

Also, I would really strongly prefer there to be a list of checkbox options to select valid plays, since we have that list already. Typing in the same play on every restart of the UI will be a pain when developing new gameplay features. 

Not sure what the best way to do this is--have both the ui code + GameplayNode read from some yaml? Publish a list of valid play strings as part of the msg on the test play topic from GN?

soccer/src/soccer/ui/main_window.cpp
```diff
@@ -1170,3 +1167,30 @@ void MainWindow::updateDebugLayers(const LogFrame& frame) {
         _ui.debugLayers->sortItems();
     }
 }
+
+void MainWindow::on_addToTable_clicked() {
+    auto to_add = (_ui.testInput->toPlainText().toStdString());
+    boost::trim(to_add);
+    auto* selectedTestsTable = _ui.selectedTestsTable;
```
kfu02: nit: should be snake case 

`selected_tests_table`

soccer/src/soccer/ui/main_window.hpp
```diff
@@ -73,6 +73,8 @@ class MainWindow : public QMainWindow {
 
     void setUseRefChecked(bool use_ref);
 
+    rclcpp::Publisher<std_msgs::msg::String>::SharedPtr test_play_pub_;
```
kfu02, line 76: nit: pub member shouldn't end in _

kfu02 (commented):
Following up on the last review, here are some nits.

soccer/src/soccer/ui/main_window.cpp
kasohrab: in the .ui files they use camelCase for some reason, so I was being consistent. I just got rid of this temp var though.

soccer/src/soccer/ui/main_window.hpp
kasohrab: all other pubs end in _ also (see planner_node.hpp for example) since they are private vars. I can't change that otherwise clang will complain.

p-nayak11 (approved):
The test tab works as intended. The keepaway play also does not work. Will need to check on it.

PR #1902, opened 2022-06-08
unassigned Role is Seeker Role

p-nayak11:
## Description
This makes the seeker role the default for all robots that are not being used by a tactic. This allows for robots to "get open" so they are viable for the receiver role in a pass tactic. 

### Test Case 1
1. Run the Offense play and ensure that all robots are not being assigned a tactic are "seeking". 
2. Make sure that the seeker role is NOT being assigned to a robot that is already assigned another role!
3. Run the Defense play and change the number of wallers to see how the seekers behave.

Expected result: All robots have some role to do either seeking or the ones assigned by the play. 

##Sub-issues
- [x] A whole Formation object is being passed into the Seek Tactic, ensure that its implementation makes sense


rj_gameplay/rj_gameplay/tactic/seek.py
```diff
@@ -13,20 +14,21 @@ def __init__(
...
         num_seekers: int,
-        formations: List,
-        centroids: List,
+        formations: stp.formations.Formations,
     ):
         super().__init__(world_state)
 
-        formation = formations
-        centroid_list = centroids
+        formation = formations.get_regions
```
kfu02: These names are confusing. The class being passed in is "DiamondFormation" but this param is named formations; the local var formation is then assigned to the param formations.get_regions?

It seems like you should change `formations` to `formation`, then change this var to `region = formation.get_regions` to avoid confusion.

rj_gameplay/rj_gameplay/tactic/seek.py
```diff
@@ -13,20 +14,21 @@ def __init__(
...
+        centroid_list = formations.get_centroids
         self._used_regions = []
         self._used_centroids = []
         self._num_seekers = num_seekers
-
+        # TODO: make seeker be able to handle more robots than formation regions available OR make formations have more regions
```
kfu02, line 25:
```suggestion
        # TODO (Prahbanjan): make seeker be able to handle more robots than formation regions available OR make formations have more regions
```
Can you make this TODO a part of this PR? If we're replacing all unassigned behavior with seekers, having some hardcoded limit of seekers based on the formation will make developing new plays more confusing.

kfu02 (changes requested):
Hey, this looks good on the offense play! Can you test it on the defense play (remove all the markers) to see if it works in general? I have some concerns about that based on your associated issues.

rj_gameplay/rj_gameplay/tactic/seek.py
p-nayak11: The current way that it's being handled is the only way to "handle more robots than there are regions". The end result is that multiple robots are being passed into the same region. The only viable solution is to either:

A) add more regions to the formations class that are smaller and more definite 
B) create some logic in the seeker role to avoid being a certain distance from our robots in that region

The first one is less related to this PR. The second one makes the seeker role clunky. Will need to discuss this issue further, which is why the TODO was added instead of fixing the issue.

rj_gameplay/rj_gameplay/play/defense.py
```diff
@@ -34,14 +34,16 @@ def tick(
 
         if self._state == State.INIT:
             self.prioritized_tactics.append(goalie_tactic.GoalieTactic(world_state, 0))
-            num_wallers = min(3, len(world_state.our_visible_robots) - 1)
+            num_wallers = 3
             self.prioritized_tactics.append(
                 wall_tactic.WallTactic(world_state, num_wallers)
             )
+            """
```
kfu02: any reason you took out the markers? was it just to see if seekers work on defense as well?

rj_gameplay/rj_gameplay/tactic/seek.py
kfu02: okay in that case, please add a comment after the TODO that explains why the clunky-looking reset code

rj_gameplay/stp/play/__init__.py
```diff
@@ -133,10 +134,19 @@ def assign_roles(
                 print(
                     f"Tactic {tactic} denied: {len(tactic.role_requests)} requested roles, but only {numRobotsAvailable} robots available"
                 )
-
+        # calculation needs to be done after robots and roles have finalized
```
kfu02:
```suggestion
        # seeker calculation needs to be done after robots and roles have finalized
```

rj_gameplay/stp/play/__init__.py
```diff
@@ -133,10 +134,19 @@ def assign_roles(
...
         for robot in world_state.our_robots:
-            if robot not in used_robots:
-                self.unassigned_roles.append(unassigned_role.UnassignedRole(robot))
+            if robot not in used_robots and robot.visible:
```
kfu02, line 139: I'm slightly concerned about the behavior on penalty plays--if our robots start seeking instead of lining up like they should it would be a little embarassing. Can you merge in `ros2` and test those penalty plays out to validate there's no conflict?

kfu02 (changes requested):
Hey, just a couple minor nits and one request for more testing info. Overall looks really good!

rj_gameplay/rj_gameplay/play/defense.py
p-nayak11: Yes, I wanted to see what would happen if some robots did not have a role in defense. They did what they were supposed to do, which is seek.

rj_gameplay/stp/play/__init__.py
p-nayak11: Theoretically speaking, this scenario will never happen. The seeker role is ONLY assigned if robots have nothing else assigned to them. In the penalty situation, they all have the dumb_move role assigned, which means that none of them are seeking. 
Based on the tests that I just ran, my understanding is correct. However, I also learned that the penalty offense is severely broken :(

kfu02: Lol in that case, could you add a card to the board that explains what's broken ab it and how to fix?

kfu02 (approved):
LGTM

PR #1903, opened 2022-06-10
make only 6 robots present

kfu02:
## Description
Pretty self-explanatory. Argument to be made that the number 6 should be put into a yaml or config or something. I have a feeling that that would be a much bigger PR, though (for instance our gameplay thinks NUM_ROBOTS=16, our topic generation follows that 16 as well, but the two variables are not set in the same place), so I left it at this for now.

## Steps to Test

1. build
2. launch sim
3. launch er-force (with `-g 2020B`)
4. click and drag every robot with id >= 6 to see them fly away!

Expected result: only 6v6 soccer by the end

@p-nayak11 if you want to add a card in the long-term that does this automatically instead of forcing us to click and drag every robot that'd be great

### DIFF
--- a/soccer/src/soccer/radio/sim_radio.cpp
+++ b/soccer/src/soccer/radio/sim_radio.cpp
@@ -63,7 +63,7 @@ static SimulatorCommand convert_placement_to_proto(
         //
         // to see this in effect, click and drag the robots expected to be not
         // present and they should disappear
-        if (robot.robot_id <= 6) {
+        if (robot.robot_id <= 5) {
             robot_proto->set_present(true);
         } else {
             robot_proto->set_present(false);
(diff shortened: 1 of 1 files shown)


soccer/src/soccer/radio/sim_radio.cpp
```diff
@@ -50,12 +50,24 @@ static SimulatorCommand convert_placement_to_proto(
...
         robot_proto->mutable_id()->set_team(robot.is_blue_team ? Team::BLUE : Team::YELLOW);
-        robot_proto->set_present(true);
+
+        // this line sets robots with an ID above 6 to "not present", meaning
+        // they will be removed on their next sim command
+        // see "rj_protos/proto/ssl_simulation_control.proto"
+        //
+        // to see this in effect, click and drag the robots expected to be not
+        // present and they should disappear
+        if (robot.robot_id <= 6) {
```
p-nayak11: This should be less than 6. Since we want 6 robots on the field, the IDs should be 0-5.

p-nayak11 (changes requested):
All robot IDs that are greater than 6 work, but currently there are 7 robots on the field still.

soccer/src/soccer/radio/sim_radio.cpp
kfu02: yes

will fix
```suggestion
        if (robot.robot_id <= 5) {
```

p-nayak11 (approved):
LGTM! Everything works!

PR #1904, opened 2022-06-10
Ball Placement

kasohrab:
## Description
Yes the infamous ball placement problem .... It definitely looks like it's doable, but this has exposed some TODO's in gameplay and makes me wonder if an agent model like erforce uses would be better :).

### Test Case 1
1. Step 1
2. Step 2
3. Step 3

Expected result: ???


kfu02:
Honestly I'm fully on board with switching to an agent model--STP is like 20 years old, and we don't even really follow it anymore since I introduced Roles. But not sure who would spearhead that large of a refactor since I'm stuck in motion planning right now and you have firmware to do. @p-nayak11 ?

PR #1906, opened 2022-06-16
Test play improvements + new motion planning plays

kfu02:
## Description
* fixes bug where new UI test play selector re-init every tick
* dynamically imports the test play list (instead of having to change the import list and the config .txt file)
* adds new motion planning test plays

### Test Case 1
1. Try both motion planning test plays (bottom of UI)

Expected result: Both work as docstrings promise.


kasohrab (approved):
lgtm. assuming merge conflicts will be resolved. Did we want to be able to dynamically update the QT test play tab too? It's not a big deal, so this pr is good to merge.

kfu02:
@kasohrab what do you mean by dynamically update the QT tab? I thought that was based on the plays.txt file, which is what I'm basing the imports off of, so they should be matching at all times now?

kasohrab:
> @kasohrab what do you mean by dynamically update the QT tab? I thought that was based on the plays.txt file, which is what I'm basing the imports off of, so they should be matching at all times now?

I just meant we could add a refresh button if we change plays.txt during runtime. Yes you're right each time you run soccer it takes in plays.txt to fill out the selectedTestsTable. Since it seems like the python side can dynamically update now? maybe that's not actually how it works though.

But to be clear I don't think what I suggested matters, I just brought it up. We can always still type in a play to add it at runtime, or just close and relaunch soccer after adding a new play to the txt.

kasohrab:
Oh also it looks like build failed? I can't see the logs right now, but we should probably check that out before merging this😅.

kfu02:
@kasohrab ahh. The Python side doesn't check plays.txt every tick, I just added some logic to import everything when it reads the file on launch. IMO we don't add new plays often enough for what you're describing to be an often-used feature.

Also, build-and-test failed on util/ubuntu-setup? I re-ran it but I promise the branch builds on my machine at least lol

kfu02:
build check passed, merging now

PR #1908, opened 2022-06-17
Various planning fixes

kfu02:
## Description
Various tweaks and fixes to motion planning. Tuning the dynamic obstacle radius currently.

 * when goal pt is blocked by an obstacle, rather than not planning anything, pick the closest point on the start-end line and use that as the new goal
 * inflate obstacle radius when obstacles have velocity
 * shift opp robot obstacle in direction of travel slightly

per @p-nayak11 comment below, new obstacle shift:
Figure_1

 * other comments/TODOs/style fixes

### Test Case 1
 * Run the motion planning test play at the same time (from #1906) with sim2play. You should see the robots don't move when their goal points are blocked.
 * Run the same plays on my branch. You should see they move close to the point, meaning they can correct more easily.
 
### Test Case 2
Run various other plays on my branch.

Expected result: soccer does not crash, no obviously worse motion planning than before.


soccer/src/soccer/planning/planner/plan_request.hpp
```diff
@@ -102,6 +102,31 @@ struct PlanRequest {
...
+ * 2 m/s = center shifted by 0.4 * robot rad
+ * ...
+ *
+ * obs_radius_inflation = 0.2:
+ * At 0 m/s, obs_radius = robot radius
+ * 1 m/s = obs_radius = 1.2 * robot radius
+ * 2 m/s = obs_radius = 1.4 * robot radius
+ * ...
+ *
+ * @param robot ptr to robot that needs obstacle made
```
kfu02, line 128:
```suggestion
 * @param robot ptr to robot that needs obstacle made
 * @return shared_ptr to new Circle obstacle with inflated radius and center
```

soccer/src/soccer/planning/primitives/replanner.cpp
```diff
@@ -87,7 +109,8 @@ Trajectory Replanner::check_better(const Replanner::PlanParams& params, Trajecto
 }
 
 Trajectory Replanner::create_plan(Replanner::PlanParams params, Trajectory previous) {
-    rj_geometry::Point goal_point = params.goal.position;
+    rj_geometry::Point goal_point = params.goal.position;  // TODO: why does this one not need ()
```
kfu02:
```suggestion
    rj_geometry::Point goal_point = params.goal.position;
```

soccer/src/soccer/planning/primitives/replanner.cpp
```diff
@@ -88,6 +110,7 @@ Trajectory Replanner::check_better(const Replanner::PlanParams& params, Trajecto
 
 Trajectory Replanner::create_plan(Replanner::PlanParams params, Trajectory previous) {
     rj_geometry::Point goal_point = params.goal.position;
+    rj_geometry::Point start_point = params.start.position();
```
kfu02:
```suggestion
```

soccer/src/soccer/planning/planner/plan_request.cpp
```diff
@@ -2,43 +2,62 @@
...
 void fill_obstacles(const PlanRequest& in, rj_geometry::ShapeSet* out_static,
                     std::vector<DynamicObstacle>* out_dynamic, bool avoid_ball,
                     Trajectory* out_ball_trajectory) {
     out_static->clear();
     out_static->add(in.field_obstacles);
     out_static->add(in.virtual_obstacles);
 
-    // Add opponent robots as static obstacles.
+    // Add their robots as static obstacles (inflated based on velocity).
+    // See calc_static_robot_obs() docstring for more info.
```
kasohrab, line 29: there still seem to be a lot of opp. bot crashes. Not sure if we can do anything about it in this pr, and it seems better than before (the method makes sense as well).

kasohrab (changes requested):
looks good, but I ran it 3 times. 2 times it crashed for different reasons:
Probably just need to merge in ros2 for the second one? the first one not sure.

image

image

soccer/src/soccer/planning/planner/plan_request.cpp
```diff
@@ -2,43 +2,62 @@
 
 namespace planning {
 
+std::shared_ptr<rj_geometry::Circle> calc_static_robot_obs(const RobotState& robot) {
+    // params for obstacle shift
+    double obs_center_shift = 0.5;
+    double obs_radius_inflation = 0.5;
```
p-nayak11:
```suggestion
    double obs_radius_inflation = 1;
```
One change I played around with which slightly helped reduce number of collisions

soccer/src/soccer/planning/primitives/replanner.cpp
```diff
@@ -57,6 +60,25 @@ Trajectory Replanner::full_replan(const Replanner::PlanParams& params) {
         CreatePath::rrt(params.start.linear_motion(), params.goal, params.constraints.mot,
                         params.start.stamp, params.static_obstacles, params.dynamic_obstacles);
 
+    // if the initial path is empty, the goal must be blocked
+    // try to shift the goal_point until it is no longer blocked
+    int max_tries = 10;  // try max this many times before giving up (and sending NOP)
+    LinearMotionInstant almost_goal = params.goal;
+    rj_geometry::Point shift_dir = (params.start.position() - params.goal.position).normalized();
+    double shift_size = 2.0 * kRobotRadius;
```
p-nayak11:
```suggestion
    double shift_size = 4.0 * kRobotRadius;
```
One change I played around with which slightly helped reduce number of collisions

p-nayak11 (changes requested):
Here's a couple of things that I changed and played around with to see if worked better.
- obs_radius_inflation to 1 instead of .5 (plan_request.cpp)
- shift_size = 4.0 * kRobotRadius instead of shift_size_ 2.0 * kRobotRadius (replanner.cpp)

These two reduced the number of collisions to about max of 2 collisions per switch. 

**Observations:**
The path that's planned often crosses over each other (at times it's almost mirroring the opposing plan). Possible fix: The current trajectory_utils is based off of position of dynamic obstacles instead of velocity. If possible, reading the velocity could help change the path. 

Another thing that I noticed is that when planning, our velocity profiler creates HUGE curvatures. Kyle commented to re-enable curvature limitations, which could be worth looking into. 

Overall, the planning is better than what we currently have. I don't know the feasibility of these suggestions, but something to think about.

kfu02:
@kasohrab I've been seeing the second crash a lot across several branches. I think play selector is not checking if `game_info` is None before using it (and as you know sometimes world_state/game_info is none to Python). It wasn't introduced in this branch though, I'm fairly certain.

The first one is more concerning, what did you run?

kfu02:
@p-nayak11 good suggestions. I definitely didn't know about the curvature limitations idea; can you put that as a new card?

So our planning is supposed to take in the trajectories of other robots into account before planning (that's the `hits_dynamic_obstacle()` in trajectory_utils.cpp that you noticed). Trajectories are a series of positions + expected timestamps, so if the lines cross over each other in theory that shouldn't be an issue so long as the timestamps don't conflict. Since some robots have to be planned first, there is [a priority system](https://github.com/RoboJackets/robocup-software/blob/6c432be11debae6e7c732d4ca52a9e9cb846bf73/soccer/src/soccer/planning/planner_node.cpp#L90-L95) in place that picks which robots get to plan first, and which robots have to plan around them.

I suspect the priority is either not used correctly or not implemented correctly, so our ability to navigate around our own robots is worse than it should be. I played with making robot id = priority but that didn't seem to change anything. This would also make a good follow-up card.

soccer/src/soccer/planning/planner/plan_request.cpp
```diff
@@ -2,43 +2,62 @@
 
 namespace planning {
 
+std::shared_ptr<rj_geometry::Circle> calc_static_robot_obs(const RobotState& robot) {
+    // params for obstacle shift
+    double obs_center_shift = 0.5;
+    double obs_radius_inflation = 1;
```
kfu02:
```suggestion
    double obs_radius_inflation = 1.0;
```

soccer/src/soccer/planning/primitives/replanner.cpp
kfu02: This is interesting. Two things: 
1) Can you change the start point (the definition of almost_goal on line 66) of the iteration to be `params.goal + 4.0 * kRobotRadius * shift_dir` but keep the shift_size at 2.0 * kRobotRadius? I think this will have the same effect but keep a higher resolution.
2) Did you run [autoref](https://github.com/robotics-erlangen/autoref) to count collision fouls or just count by hand? If you haven't run the autoref with our sim before, the installation is similar to er-force's simulator, and it will launch a GUI. Make sure you configure the vision port to 10020 because that's what the simulator publishes to. You shouldn't have to connect game controller to see the crash fouls.

soccer/src/soccer/planning/planner/plan_request.cpp
kfu02: there are, I'm hoping this PR is more of a starting point than a solution. there are a bunch of params to tune now across planning/control that should help if we can dial it in, and this was easier than scrapping RRT entirely

soccer/src/soccer/planning/primitives/replanner.cpp
```diff
@@ -57,6 +60,25 @@ Trajectory Replanner::full_replan(const Replanner::PlanParams& params) {
         CreatePath::rrt(params.start.linear_motion(), params.goal, params.constraints.mot,
                         params.start.stamp, params.static_obstacles, params.dynamic_obstacles);
 
+    // if the initial path is empty, the goal must be blocked
+    // try to shift the goal_point until it is no longer blocked
+    int max_tries = 10;  // try max this many times before giving up (and sending NOP)
+    LinearMotionInstant almost_goal = params.goal;
```
p-nayak11:
```suggestion
    LinearMotionInstant almost_goal = params.goal + 4.0 * kRobotRadius * shift_dir;
```

p-nayak11:
> @p-nayak11 good suggestions. I definitely didn't know about the curvature limitations idea; can you put that as a new card?
> 
> So our planning is supposed to take in the trajectories of other robots into account before planning (that's the `hits_dynamic_obstacle()` in trajectory_utils.cpp that you noticed). Trajectories are a series of positions + expected timestamps, so if the lines cross over each other in theory that shouldn't be an issue so long as the timestamps don't conflict. Since some robots have to be planned first, there is [a priority system](https://github.com/RoboJackets/robocup-software/blob/6c432be11debae6e7c732d4ca52a9e9cb846bf73/soccer/src/soccer/planning/planner_node.cpp#L90-L95) in place that picks which robots get to plan first, and which robots have to plan around them.
> 
> I suspect the priority is either not used correctly or not implemented correctly, so our ability to navigate around our own robots is worse than it should be. I played with making robot id = priority but that didn't seem to change anything. This would also make a good follow-up card.

Currently, we are crashing against the opponent right? So the priority system isn't really going to change anything since it's about planning for our robots. I could be wrong since I don't know the gist of how sim2play works.

kasohrab:
> The first one is more concerning, what did you run?

just soccer. It crashed on startup. To be fair, I think I have seen this one before. Probably not part of this pr. Probably just some weird non-deterministic edge case; it still needs to be handled at some point though. If we print the backtrace at that point (or use gdb), we could better determine the cause.

kfu02:
> > @p-nayak11 good suggestions. I definitely didn't know about the curvature limitations idea; can you put that as a new card?
> > So our planning is supposed to take in the trajectories of other robots into account before planning (that's the `hits_dynamic_obstacle()` in trajectory_utils.cpp that you noticed). Trajectories are a series of positions + expected timestamps, so if the lines cross over each other in theory that shouldn't be an issue so long as the timestamps don't conflict. Since some robots have to be planned first, there is [a priority system](https://github.com/RoboJackets/robocup-software/blob/6c432be11debae6e7c732d4ca52a9e9cb846bf73/soccer/src/soccer/planning/planner_node.cpp#L90-L95) in place that picks which robots get to plan first, and which robots have to plan around them.
> > I suspect the priority is either not used correctly or not implemented correctly, so our ability to navigate around our own robots is worse than it should be. I played with making robot id = priority but that didn't seem to change anything. This would also make a good follow-up card.
> 
> Currently, we are crashing against the opponent right? So the priority system isn't really going to change anything since it's about planning for our robots. I could be wrong since I don't know the gist of how sim2play works.

Oh sorry, I misunderstood then. I thought you were referring to the self-crashes we sometimes see. Yes, for avoiding opponent crashes, increasing the obs radius or limiting max curvature and velocity would be steps to take.

kasohrab (approved):
lgtm

PR #1909, opened 2022-06-18
ball placement with ROS2 actions

### LINKED ISSUE #1559: Enable ball placement
## Description
We are removing the ball placement feature as a part of the port to ROS. This will need to be redesigned and reimplemented into the architecture.

In the ball placement pipeline, the referee gives us a position at which the ball should be placed. We are then responsible for moving the ball to that position before gameplay can begin, by passing and dribbling the ball to the point.

## Requirements

 - [ ] We are able to receive a ball placement point from the referee
 - [ ] We can dribble the ball to the ball placement point and back off.

kasohrab:
## Description
Title. Supersedes #1904.

## Associated Issue
Resolves #1559

## Design Documents
TODO: attach that diagram kevin made.

### Test Case 1
1. Step 1
2. Step 2
3. Step 3

Expected result: ???

- [x] Set up action server and client somehow
- [x] make sure gameplay is "turned off temporarily"
- [x] plan in action server


rj_msgs/action/BallPlacement.action
```diff
@@ -0,0 +1,8 @@
+# Goal
+BallPlacement ball_placement
```
kfu02: does ball placement need anything besides desired final ball position?
```suggestion
rj_geometry_msgs/Point goal_pt
```

soccer/src/soccer/server/ball_placement_server.cpp
```diff
@@ -0,0 +1,63 @@
...
+    using namespace std::placeholders;
+    std::thread{std::bind(&BallPlacementActionServer::execute, this, _1), goal_handle}.detach();
+}
+
+void BallPlacementActionServer ::execute(
+    const std::shared_ptr<GoalHandleBallPlacement> goal_handle) {
+    std::shared_ptr<const BallPlacement::Goal> goal = goal_handle->get_goal();
+    rj_msgs::msg::ServerIntent server_intent = goal->server_intent;
+    rj_msgs::msg::RobotIntent robot_intent = server_intent.intent;
+    int robot_id = server_intent.robot_id;
```
kfu02: so for the execute loop, I think what needs to happen is:
 * find vector from current ball pt to goal
 * call path target motion planner to move behind that vector
 * call collect planner to get ball in dribbler
 * write custom planner to move at a slow rate while dribbler is on (not entirely sure about this part, should be an adaptation of either path-target or collect planner though) 
 * OR call path-target to a different robot to position its mouth at goal, and call line kick planner (or kick directly since in theory we are facing the right way?)

The collect planner tries to calculate which direction it should approach the ball based on the ball's current velocity, so for ball placement it would be better just to force it on the right line (since goal is known). We can probably shove all this logic into the ball placement server for now and separate it into a new class if needed later

kfu02: we should just be able to publish a trajectory directly from this node, that's what planner node does: https://github.com/RoboJackets/robocup-software/blob/ad8cb935bfc58a84337fe02be046234c4c9b368e/soccer/src/soccer/planning/planner_node.cpp#L49-L51

kfu02: if we decided to run ball placement through the planner node it might be more painful, since we would have to make a new planner that does all the steps above and then pub a RobotIntent to that new planner. it could be done, though

kfu02 (commented):
Some thoughts on ball placement. Also, you have some build errors, there seem to be files not  committed?

rj_gameplay/rj_gameplay/gameplay_node.py
```diff
@@ -289,7 +293,16 @@ def gameplay_tick(self) -> None:
             else:
                 self._curr_play = self._test_play
 
-            intents = self._curr_play.tick(self.world_state)
+            # TODO: bad design,
+            #  but unresolvable without refactoring gameplay again
```
kfu02, line 309: :(

rj_gameplay/rj_gameplay/gameplay_node.py
```diff
@@ -302,6 +315,9 @@ def gameplay_tick(self) -> None:
                         empty_command = EmptyMotionCommand()
                         empty_intent.motion_command.empty_command = [empty_command]
                         self.robot_intent_pubs[i].publish(empty_intent)
+            else:
+                # probably using action client instead
```
kfu02, line 340: can you add this as output to the debug panel where we usually see the play?

(or just output to console if that's too difficult)

rj_gameplay/rj_gameplay/play/ball_placement.py
```diff
@@ -0,0 +1,41 @@
...
+        self._state = State.INIT
+
+    def tick(
+        self, world_state: stp.rc.WorldState, client: Optional[Node]
+    ) -> Optional[List[RobotIntent]]:
+        game_info = world_state.game_info
+        goal_array: np.array = game_info.ball_placement()
+        goal_pt: Point = Point(x=goal_array[0], y=goal_array[1])
+        print(client)
+        client.send_goal(goal_pt)
```
kfu02: for my clarification, we can have the ActionClient refuse to resend if the goal point has not changed, right?

kfu02: oh I just saw the todo in the ActionServer. What's standard, having the client refuse to send or having the server refuse to accept?

kfu02 (changes requested):
EDIT: there is no error when running with external ref since it provides a ball placement point. However, internal ref does not, which causes the issue. Can you add some check where ball placement outputs some error msg if no ball placement point is given but it's in the ball placement play (and doesn't say some weird NoneType error that seems unrelated)

Getting this error on make run-sim?
`[gameplay_node-11] The following exception was never retrieved: 'NoneType' object has no attribute 'is_free_placement'`

which leads to `[gameplay_node-11] The following exception was never retrieved: 'NoneType' object is not subscriptable` on ball placement play selection

rj_gameplay/rj_gameplay/gameplay_node.py
kfu02: this is done already, nvm

rj_gameplay/rj_gameplay/play/ball_placement.py
kasohrab: I should delete this code from tick in the play, it's not actually used😅. See the client for all the actual logic, gameplay node sends the goal using that in the if statement. The part where we determine when it's ball placement time is from play selector, so we need a play only for that purpose (🙃 this was the easiest way I promise)

kasohrab: I think we should do both, making assumptions that the input is valid or will be handled somewhere else often leads to problems later. 

on the client side, see the TODO in the `send_goal` method

kasohrab: So following up on the first reply, I can also undo the changes I made to the play abstract class.

kasohrab: And following up on the second one, I think it's standard for the server to reject goals it doesn't want. The server can also defer goals if it wants. But I still feel it should be done in both places.

kfu02: both done, see below

kfu02:
Proof of concept done. [Screen recording](https://drive.google.com/file/d/1qcc37rrjPIG9r0ll1Z3NgKeyKWc5ASFi/view?usp=sharing)

This does not handle many edge cases, including:
 * when the ball is on the wall and must be dragged backwards
 * when the pass is not successful and the receiver must go grab the ball to center it
 * when our robots are in the way of proper ball placement (notice my choice to pick the play that moves them all out of the way first)
 
However, as a proof of concept it shows: 
 * ball placement is possible with our current planners
 * passing may be easier with an agent model
 * gameplay would definitely be better with access to the new is_dones from planners. 
 * (I also modified planner node to handle singular RobotIntents correctly in this PR, instead of expecting them to be sent every tick from gameplay.)

Next step is to figure out how we should change planning/gameplay to account for this new info.

PR #1911, opened 2022-06-21
Is done planners

kfu02:
## Description
Adds is_done methods to each planner, which return true if the planner is done with its job and false otherwise. This sets the groundwork for more advanced planner behavior (e.g. ball placement #1904 ) on the C++ side.

Also tuned the capture skill for ER-force simulator. Did not tune other skills as that would have taken too long.

Resolves # or [ClickUp card](https://app.clickup.com/t/2zpw9jp)

## Design Documents
None, based the new is_done() methods off of the existing planner logic + gameplay Skill equivalents.

### Test Case 1
1. Run motion planning test play named `test_motion_planning.AllBots` (second test_motion_planning play).
2. Notice the debug messages.

**Expected result:** Debug msgs say "robot x's PathTargetPlanner is_done"

Note: other planners have is_done, but are annoying to test because gameplay Skill is_dones override the planner is_dones. (Ask Kevin to provide more clarification if that doesn't make sense.)

## Key Files to Review
Gameplay Changes
 * `rj_gameplay/skill/*` - simple tweaks to existing is_dones for better performance
 * other `rj_gameplay/*` and `config/` - bugfixes or new motion planning test play

new nearly_equals()
 * `rj_common/*`
 * `rj_geometry/*`
 * `soccer/src/soccer/planning/instant.hpp`

Main Planning is_done changes
 * `soccer/src/soccer/planning/*`

## Review Checklist

- [x] **Remove extra print statements**: Any print statements used for debugging should be removed*

*intentionally left some debug prints to illustrate working behavior, can delete after


soccer/src/soccer/planning/planner/intercept_planner.hpp
```diff
@@ -18,5 +18,7 @@ class InterceptPlanner : public PlannerForCommandType<InterceptCommand> {
         : PlannerForCommandType<InterceptCommand>("InterceptPlanner"){};
 
     Trajectory plan(const PlanRequest& request) override;
+
+    [[nodiscard]] bool is_done() const override;
```
kasohrab, line 22: we should make all of these is_done methods also constexpr. You cannot make virtual functions constexpr in cpp17 which we use right now, but when I switched to cpp20, this change worked. Marking as constexpr offloads all the work to compile time. Since doing this requires changing our cpp version, we can do it in a different pr if you want. I still think it should be done at some point.

soccer/src/soccer/planning/planner/settle_planner.cpp
```diff
@@ -536,4 +536,18 @@ void SettlePlanner::reset() {
...
+    // FSM: Intercept -> Dampen
+    // (see process_state_transition())
+    if (current_state_ != SettlePlannerStates::Dampen) {
+        return false;
+    }
+
+    // Dampen is done when the ball is slow-ish, let collect handle actually
+    // taking possession
+    // TODO(Kevin): consider making settle + collect one planner
+    double SETTLE_BALL_SPEED_THRESHOLD = 0.75;
```
kasohrab, line 549: github actions is complaining about the case style, even though it eventually will be a param.

soccer/src/soccer/planning/planner_node.cpp
```diff
@@ -22,7 +22,7 @@ PlannerNode::PlannerNode()
     robots_planners_.reserve(kNumShells);
     for (int i = 0; i < kNumShells; i++) {
         auto planner =
-            std::make_unique<PlannerForRobot>(i, this, &robot_trajectories_, &shared_state_);
+            std::make_shared<PlannerForRobot>(i, this, &robot_trajectories_, &shared_state_);
```
kasohrab, line 25: Is there a reason this need to be a shared pointer now? My understanding is that shared pointers have a lot more overhead.

soccer/src/soccer/planning/planner/settle_planner.hpp
```diff
@@ -29,6 +29,7 @@ class SettlePlanner : public PlannerForCommandType<SettleCommand> {
     Trajectory plan(const PlanRequest& plan_request) override;
 
     void reset() override;
+    bool is_done() const override;
```
kasohrab:
```suggestion
    [[nodiscard]] bool is_done() const override;
```

soccer/src/soccer/planning/planner_node.cpp
```diff
@@ -158,11 +163,26 @@ Trajectory PlannerForRobot::plan_for_robot(const planning::PlanRequest& request)
 
     debug_draw_.publish();
 
+    // TODO(Kevin): delete this debug print
+    if (this->is_done()) {
+        SPDLOG_ERROR("robot {}'s {} is_done", robot_id_, current_planner_->name());
```
kasohrab, line 168: this was helpful for reviewing 👍🏽

soccer/src/soccer/planning/planner_node.hpp
```diff
@@ -139,13 +143,45 @@ class PlannerForRobot {
...
+    /*
+     * @brief Check that robot is visible in world_state and that world_state has been
+     * updated recently.
+     */
     [[nodiscard]] bool robot_alive() const;
 
+    /*
+     * @return true if current planner is done, false otherwise.
+     */
+    bool is_done() const;
```
kasohrab: this should also be marked?
```suggestion
    [[nodiscard]] bool is_done() const;
```

soccer/src/soccer/planning/planner/path_target_planner.hpp
```diff
@@ -21,6 +24,12 @@ class PathTargetPlanner : public PlannerForCommandType<PathTargetCommand> {
     Trajectory plan(const PlanRequest& request) override;
     void reset() override { previous_ = Trajectory(); }
 
+    bool is_done() const override;
```
kasohrab:
```suggestion
    [[nodiscard]] bool is_done() const override;
```

soccer/src/soccer/planning/planner/planner.hpp
```diff
@@ -42,6 +42,11 @@ class Planner {
      */
     virtual void reset() {}
 
+    /*
+     * @return true if planner is done, false otherwise.
+     */
+    virtual bool is_done() const = 0;
```
kasohrab:
```suggestion
    [[nodiscard]] virtual bool is_done() const = 0;
```

soccer/src/soccer/planning/planner/path_target_planner.hpp
```diff
@@ -21,6 +24,12 @@ class PathTargetPlanner : public PlannerForCommandType<PathTargetCommand> {
...
+
+    // vars to tell if is_done
+    std::optional<LinearMotionInstant> cached_start_instant_;
+    std::optional<LinearMotionInstant> cached_goal_instant_;
```
kasohrab: these are public vars, the underscore after the name should just be used on private vars.
I'm assuming though these should just be made private.

soccer/src/soccer/planning/planner/pivot_path_planner.hpp
```diff
@@ -21,12 +21,20 @@ class PivotPathPlanner : public PlannerForCommandType<PivotCommand> {
         cached_pivot_target_ = std::nullopt;
         cached_pivot_point_ = std::nullopt;
     }
+    bool is_done() const override;
```
kasohrab:
```suggestion
    [[nodiscard]] bool is_done() const override;
```

kasohrab (changes requested):
looks great! I left some small comments that should mostly be quick to address.

soccer/src/soccer/planning/planner_node.cpp
kfu02: If my understanding is correct, this needed to be shared so I could assign the `current_planner_` field to whichever planner is the current planner (see line 128). Let me know if that's incorrect usage

soccer/src/soccer/planning/planner/path_target_planner.hpp
kfu02: yes, they should be private, good catch

soccer/src/soccer/planning/planner/settle_planner.cpp
kfu02: hmm. I can fix all the style issues or let the warnings persist. what do you think? 

the reason I didn't make any of them into ROS params already is because I don't love our existing custom param server and wanted to experiment with using default ROS params first. but that seemed like its own PR

soccer/src/soccer/planning/planner/intercept_planner.hpp
kfu02: I'm a little hesitant to introduce large changes to our build system at the moment. I think we should focus on functionality rather than staying on top of the latest C++/ROS/Ubuntu version. Obviously we will have to move at some point but right now a move would prevent us from building a working product. Maybe a GH issue # for later?

soccer/src/soccer/planning/planner_node.cpp
kasohrab: yes that's correct. and looking into it a little more, it's not a big deal actually, since these pointers are only created at startup and destroyed on shutdown. And the overhead comes from those activities primarily.

kasohrab (approved):
LGTM. I would still make ClickUp cards/issues for the commented todos and small things I mentioned when reviewing.

soccer/src/soccer/planning/planner/intercept_planner.hpp
kfu02: referenced in issue #1920

soccer/src/soccer/planning/planner/settle_planner.cpp
kfu02: referenced in clickup card https://app.clickup.com/t/31g2yfy

rj_gameplay/rj_gameplay/role/seeker.py
```diff
@@ -32,7 +32,9 @@ def get_open_point(self, world_state, region: List, centroid) -> np.ndarray:
         # line-of-sight calculations
         # (LOS is maximized by making dot product of angle between unit vectors as close to 0 as possible)
         ball_dir = lambda pos: pos - world_state.ball.pos
-        ball_dir_norm = lambda pos: ball_dir(pos) / np.linalg.norm(ball_dir(pos))
+        ball_dir_norm = lambda pos: ball_dir(pos) / (
+            1e-9 + np.linalg.norm(ball_dir(pos))
```
p-nayak11, line 36: What difference did this addition make?

soccer/src/soccer/planning/planner_node.cpp
```diff
@@ -122,12 +125,14 @@ Trajectory PlannerForRobot::plan_for_robot(const planning::PlanRequest& request)
...
         // a plan.
         if (trajectory.empty() && planner->is_applicable(request.motion_command)) {
+            current_planner_ = planner;
             trajectory = planner->plan(request);
         }
 
         // If it fails, or if the planner was not used, the trajectory will
         // still be empty. Reset the planner.
         if (trajectory.empty()) {
+            current_planner_ = nullptr;
```
p-nayak11, line 135: I noticed that while running sim2play, either blue or yellow would often not move until the last minute. I am assuming that this could be stemming from this section right here. Could you reduce the number of resets from every tick to a range of ticks similar to how seeker is given a new target point.

p-nayak11: I'm not sure if this would work, but it could at least let the robots being the trajectory and let the replanner do its purpose rather than resetting the whole trajectory from null.

soccer/src/soccer/planning/planning_params.cpp
```diff
@@ -62,7 +62,7 @@ DEFINE_NS_FLOAT64(
 //
 // High number indicates that it will always try to choose a point nearest
 // to the current robot position
-DEFINE_NS_FLOAT64(kPlanningParamModule, collect, ball_speed_approach_direction_cutoff, 0.1,
+DEFINE_NS_FLOAT64(kPlanningParamModule, collect, ball_speed_approach_direction_cutoff, 1.0,
```
p-nayak11, line 65: Isn't there a higher chance of us bouncing into the ball if we increase the speed at which we approach for capture?

p-nayak11 (changes requested):
I left a couple comments about issues that I noticed while testing. They aren't major issues; however, if the suggestions help, it would avoid future problems. Overall, planners are working much better. Steady progress being made :)

rj_gameplay/rj_gameplay/role/seeker.py
kfu02: if the norm of the ball dir vector is exactly 0, this causes a div by 0 error without the + tiny value

soccer/src/soccer/planning/planner_node.cpp
kfu02: the `current_planner_` var doesn't actually do anything other than store a ptr to the planner selected in the for-loop above (lines 127-130). Notice there that the trajectory is taken directly from one of the planners in the for-loop.

kfu02: the replanner resetting from null is because it gets a fresh RobotIntent from gameplay every tick. another reason why we need to overhaul gameplay to use actions instead of a blind gameplay->RobotIntent->planning pipeline

soccer/src/soccer/planning/planning_params.cpp
kfu02: the relevant lines of code in the collect planner [are here](https://github.com/RoboJackets/robocup-software/blob/2a4bacfd1b3f4a01c4fcdcedc6c9060e71f436dd/soccer/src/soccer/planning/planner/collect_planner.cpp#L88-L96). you can see that when the ball speed is < this thresh, the robot will try to go directly at it, while if the speed is > this thresh, the robot will try to serve as a backstop.

what I noticed in my testing is that the robot didn't backstop successfully very often, so I tuned a bunch of params including this and the accel_scale ones below. those changes combined actually led to less bouncing, but it's hard to isolate any of them. if you want, you can make a card that's like "tune capture" but I think a) that's not in the scope of this PR and b) that's not easy to do when gameplay-controlled passing has some timing bugs. I think once ball placement is up (with passing) it will be easier to see, so I decided to wait on tuning this and any other skills

soccer/src/soccer/planning/planner_node.cpp
kfu02: (for clarification, the ptr is only necessary here for the debug prints below, where I need to get the name of whatever planner is saying it's done)

p-nayak11 (approved):
Your explanations make sense. LGTM

PR #1927, opened 2022-07-24
Make manual easier

kfu02:
## Description
Making it easier to use the manual control node and service. 
* Adds script to avoid launching rqt every time we want to test manual.
* Deletes manual.launch.py and replaces it with a flag in soccer.launch.py. 
* Adds comments and slight changes to makefile and soccer.launch.py.

Resolves # or [ClickUp card](https://app.clickup.com/t/34bj9jn)

### Test Case 1
1. Follow instructions in this PR's `doc/RunningOnFieldComp.md`
2. Make sure I didn't break any of the other `make run-sim` derivatives.

**Expected result:** Docs are clear and describe what's happening. `make run-sim, make run-sim-ex, etc.` still work as intended.


launch/soccer.launch.py
```diff
@@ -38,12 +35,15 @@ def generate_launch_description():
     bringup_dir = Path(get_package_share_directory("rj_robocup"))
     launch_dir = bringup_dir / "launch"
 
+    # TODO(Alex): why are there two defaults? here and at bottom of file in
```
kasohrab: You can delete this. Apparently you need both, one is for local values (launch configuration), the other exposes it to the user to change at launch time: https://docs.ros.org/en/galactic/Tutorials/Intermediate/Launch/Using-Substitutions.html#substitutions-example-launch-file. 

This answer explains it better than ros2 documentation:
https://answers.ros.org/question/322874/ros2-what-is-different-between-declarelaunchargument-and-launchconfiguration/

So maybe we only need the declarelaunchargument, but I'm pretty sure we need both.

kasohrab: Seems like a dumb ROS quirk to me lol

util/manual_control_connect.bash
```diff
@@ -0,0 +1,42 @@
...
+do
+    case "${flag}" in
+        h) display_help;;
+        r) robot_id=${OPTARG};;
+        c) controller_uuid=${OPTARG};;
+        C) connect=${OPTARG};;
+    esac
+done
+
+cmd="ros2 service call /select_manual rj_msgs/srv/SetManual \"{robot_id: $robot_id, controller_uuid: $controller_uuid, connect: $connect}\""
```
kasohrab, line 39: This is a big improvement from before, but I would add a card to have the ui call this script.

kasohrab: (When you input a manual id and launch in manual mode)

kasohrab (changes requested):
Looks great! Some comments before merging.

util/manual_control_connect.bash
kfu02: way ahead of you: https://app.clickup.com/t/2tdhtqv

kasohrab (approved):
Looks good to me.

kfu02:
currently blocked by #1928 , I would like to make the launch process for sim vs real much better before merging these changes.

on second thought, #1928 will take a while and this change is nice. merging.

PR #1930, opened 2022-08-17
Added Python venv to ubuntu-setup

fourpenny:
## Description
The `util/ubuntu-setup` script now creates a Python virtual environment (`rj_env`) and uses it for installing our repo's Python packages

* Helps prevent causing problems in users' global Python installation
* Virtual environment is added to the $PYTHONPATH variable and can be used when calling our build script without additional flags/arguments

## Steps to Test
1. Run ubuntu-setup script
2. Build soccer
3. Run soccer (will need to activate the virtual environment with an additional command first: `source rj_env/bin/activate`)

It might be good for me to add that last one to the existing README in the launch instructions- I'll do that as long as this works for everyone else.

**Expected result:**
Soccer and sim should launch without problems using packages from the virtual environment

## Key Files to Review
* util/ubuntu-setup

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,7 +23,9 @@ jobs:
         run: "sudo ./util/ubuntu-setup --yes"
 
       - name: Install git pre-commit hooks & update submodules
-        run: "./util/git-setup --yes"
+        run: | 
+          sudo apt-get install python3-venv
+          "./util/git-setup --yes"
 
       - name: Make ${{ github.base_ref }} branch exist
         run: |
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,6 +35,7 @@ jobs:
           echo "::add-matcher::ci/clang.json"
           source /opt/ros/foxy/setup.bash
           source install/setup.bash
+          source rj_env/bin/activate
           make all
           echo "::remove-matcher owner=clang::"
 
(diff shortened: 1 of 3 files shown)


fourpenny:
Looks like this PR causes our CI/CD scripts to fail since they create new virtual environments by default... is that accurate?

kfu02:
@fourpenny the reason our CI fails seems to be that we don't have python3-venv as a dependency:
```
The virtual environment was not created successfully because ensurepip is not
available.  On Debian/Ubuntu systems, you need to install the python3-venv
package using the following command.

    apt-get install python3-venv

You may need to use sudo with that command.  After installing the python3-venv
package, recreate your virtual environment.

Failing command: ['/__w/robocup-software/rj_env/bin/python3', '-Im', 'ensurepip', '--upgrade', '--default-pip']

Error: Process completed with exit code 1.
```

Do you mind adding that in the python3/pip3 dependency list? If the build-and-test check works then that means your changes to ubuntu-setup should work on any Ubuntu 20.04 machine.

fourpenny:
@kfu02 

Added it to the requirements3.txt file. Anywhere else it needs to go? (Do I need to modify anything for the Dockerfiles that are used for the CI.yml file?)

kfu02:
@fourpenny nvm, it seems I've led you astray. Take the python3-venv out of requirements3.txt and try [this](https://stackoverflow.com/questions/65893608/github-actions-unable-to-set-up-python-virtual-environment) instead

(the tests still fail since it's expecting to have access to a venv before installing the packages in requirements3.txt)

fourpenny:
@kfu02 I'm still working on getting venv properly installed on the CI container- but something else: the default public for key is wrong about 50% (or more) of the time (see this GitHub issues post: https://github.com/ros2/ros2/issues/742) when downloading and installing Foxy during our current CI script. Do you want me to try to add a fix?

PR #1946, opened 2022-10-03
Coach Node

N8BWert:
## Description
Created a basic coach node to take referee commands and translate them into useful information.

[create Coach Node](https://app.clickup.com/t/2zx0vhn)
[update Coach node's possession state from distances](https://app.clickup.com/t/30zhxwb)
[update Coach node's possession state using ball_sense](https://app.clickup.com/t/30zhycz)

### match_situation (enum)
ball_placement - ball placement on a restart
kickoff - ball kickoff
free_kick = either direct or indirect free kicks (direct and indirect are outdated terms)
penalty_kick - penalty kick restart
in_play - basically normal play

### global_overrides
if state.halt -> max_speed = 0 m/s; min_dist_from_ball = 0m
elif state.stop -> max_speed = 1.5 m/s; min_dist_from_ball = 0.5m
elif state.playing -> max_speed = -1m/s; min_distance_from_ball = 0m

### Test Case 1
1. run sim
2. check rqt for topic publish

**Expected result:** Changes in the referee state should be processed and published to the /strategy/coach topic

## Review Checklist

- [ :white_check_mark:] **Docstrings**: All methods and classes should have the file appropriate docstrings which follow the guidelines in the ["Contributing" page](https://rj-rc-software.readthedocs.io/en/latest/contributing.html) of our docs.
- [:white_check_mark:] **Remove extra print statements**: Any print statements used for debugging should be removed
- [ :white_check_mark:] **Tag reviewers**: Tag some people for review and ping them on Slack


soccer/src/soccer/coach/coach_node.hpp
```diff
@@ -0,0 +1,43 @@
...
+    ball_placement, // ball placement on a restart restart.ball_placement
+    kickoff, // simple kickoff (start game/match/play)
+    free_kick, // either direct or indirect free kicks (direct and indirect are outdated terms)
+    penalty_kick, // penalty kick restarts
+    in_play, // normal play
+};
+
+/**
+ * @brief The coach node subscribes to the /referee/playstate topic and translates this data
+ * into pertinant information before sending it to another topic (/strategy/coach) for the robots
```
kfu02:
```suggestion
 * into pertinent information before sending it to another topic (/strategy/coach) for the robots
```

rj_msgs/CMakeLists.txt
```diff
@@ -22,12 +22,14 @@ rosidl_generate_interfaces(
   # Messages
   msg/BallState.msg
   msg/BallPlacement.msg
+  msg/Coach.msg
```
kfu02: let's name this something more descriptive. "PlayCall" seems semantically close to what the Coach node is actually doing. @anshulm07 any ideas?

soccer/src/soccer/coach/coach_node.cpp
```diff
@@ -0,0 +1,76 @@
+#include "coach_node.hpp"
+
+using std::placeholders::_1;
+
+CoachNode::CoachNode(const rclcpp::NodeOptions& options) : Node("coach_node", options) {
+    coach_pub_ = this->create_publisher<rj_msgs::msg::Coach>("/strategy/coach", 10);
+    playstate_change_timer_ = this->create_wall_timer(
+        100ms, std::bind(&CoachNode::check_for_playstate_change, this));
```
kfu02: this warning seems to want you to use a lambda to call the method you want, like so: https://github.com/RoboJackets/robocup-software/blob/ros2/soccer/src/soccer/radio/radio.cpp#L39

the reason is a little confusing to me, but try this SO post: https://stackoverflow.com/questions/1930903/bind-vs-lambda

soccer/src/soccer/coach/coach_node.cpp
```diff
@@ -0,0 +1,76 @@
+#include "coach_node.hpp"
+
+using std::placeholders::_1;
+
+CoachNode::CoachNode(const rclcpp::NodeOptions& options) : Node("coach_node", options) {
+    coach_pub_ = this->create_publisher<rj_msgs::msg::Coach>("/strategy/coach", 10);
+    playstate_change_timer_ = this->create_wall_timer(
```
kfu02: run `make pretty-lines`

soccer/src/soccer/coach/coach_node.cpp
```diff
@@ -0,0 +1,76 @@
...
+    current_play_state_.restart = PlayState::Restart::Kickoff;
+    current_play_state_.our_restart = true;
+
+    rj_geometry_msgs::msg::Point temp_point;
+    temp_point.x = -1;
+    temp_point.y = -1;
+    current_play_state_.placement_point = temp_point;
+}
+
+void CoachNode::playstate_callback(rj_msgs::msg::PlayState::SharedPtr msg) {
```
kfu02: do what the warning suggests ([context](https://stackoverflow.com/questions/2627166/what-is-the-difference-between-a-const-reference-and-normal-parameter))

pretty sure it should look like this, but double-check with compiler
```suggestion
void CoachNode::playstate_callback(const PlayState::SharedPtr &msg) {
```

soccer/src/soccer/coach/coach_node.cpp
```diff
@@ -0,0 +1,76 @@
...
+        case PlayState::Restart::Penalty:
+            coach_message.match_situation = match_situation::penalty_kick;
+            break;
+        }
+
+        if (current_play_state_.state == PlayState::State::Playing) {
+            coach_message.match_situation = match_situation::in_play;
+        }
+
+        // TODO: get possession from wherever possession is from
```
kfu02: can you make a card on [ClickUp](https://app.clickup.com/2209401/v/b/5-103744422-2) for this and then you & Anshul take it next? doesn't have to be this PR but is necessary and urgent

soccer/src/soccer/coach/coach_node.hpp
```diff
@@ -0,0 +1,43 @@
...
+
+#include <rj_msgs/msg/play_state.hpp>
+#include <rj_msgs/msg/coach.hpp>
+#include <rj_msgs/msg/global_override.hpp>
+
+#include <rj_geometry_msgs/msg/point.hpp>
+
+#include "game_state.hpp"
+
+enum match_situation {
```
kfu02: I think it wants camel case 
```suggestion
enum MatchSituation {
```

soccer/src/soccer/coach/coach_node.cpp
kfu02: recall you have to rewrite this from https://github.com/RoboJackets/robocup-software/blob/ros2/rj_gameplay/rj_gameplay/play_selector.py#L80

kfu02 (changes requested):
Great work! Some comments to address (and I didn't actually run it yet) but this looks like a great start so far

rj_msgs/CMakeLists.txt
kfu02: actually, "PlayCall" should be reserved for the allocation of defenders/attackers that the coach node will output. (this is also urgent for the next PR @p-nayak11 ). maybe "CoachMatchState" on topic strategy/coach_match_state?

rj_msgs/msg/CoachStateInterpretation.msg
```diff
@@ -0,0 +1,4 @@
+# match_situation is an enum defined in coach_node.hpp
```
kfu02:
```suggestion
# match_situation is an enum defined in soccer/src/soccer/coach/coach_node.hpp
```

soccer/src/soccer/coach/coach_node.hpp
```diff
@@ -0,0 +1,53 @@
...
+#include <rj_msgs/msg/robot_state.hpp>
+#include <rj_msgs/msg/robot_status.hpp>
+#include <rj_msgs/msg/world_state.hpp>
+
+#include "game_state.hpp"
+
+enum MatchSituation {
+    ball_placement,  // ball placement on a restart restart.ball_placement
+    kickoff,         // simple kickoff (start game/match/play)
+    free_kick,     // either direct or indirect free kicks (direct and indirect are outdated terms)
```
kfu02:
```suggestion
    free_kick,     // either direct or indirect free kicks (direct and indirect are outdated as of 10/2022)
```

soccer/src/soccer/coach/coach_node.hpp
```diff
@@ -0,0 +1,53 @@
...
+    in_play,       // normal play
+};
+
+/**
+ * @brief The coach node subscribes to the /referee/playstate topic and translates this data
+ * into pertinent information before sending it to another topic (/strategy/coach) for the robots
+ * to use.
+ *
+ * In short, this node takes the referee information and standardizes it for consumption of the
+ * robots.
```
kfu02:
```suggestion
 * individual robot agents.
```

soccer/src/soccer/coach/coach_node.cpp
```diff
@@ -0,0 +1,140 @@
+#include "coach_node.hpp"
+
+CoachNode::CoachNode(const rclcpp::NodeOptions& options) : Node("coach_node", options) {
+    coach_pub_ =
+        this->create_publisher<rj_msgs::msg::CoachStateInterpretation>("/strategy/coach", 10);
```
kfu02: The coach will also eventually publish position assignments, so it might be better to have this topic name be `strategy/coach_state` or something like that. Then coach will pub `strategy/role_assignments` as well and it's unambiguous which is which.

kfu02: rename the coach_pub_ var to coach_state_pub_ if this is implemented too

soccer/src/soccer/coach/coach_node.cpp
```diff
@@ -0,0 +1,140 @@
...
+    // initialize our robots
+    for (int i = 0; i < kRobotsPerTeam; i++) {
+        robot_status_subs_[i] = this->create_subscription<rj_msgs::msg::RobotStatus>(
+            fmt::format("/radio/robot_status/robot_{}", i), 10,
+            [this](const rj_msgs::msg::RobotStatus::SharedPtr msg) {
+                ball_sense_callback(msg, true);
+            });
+    }
+
+    // initialize their robots
```
kfu02: we don't have robot status for their robots. robot_status 0-15 exists because we could have 16 uniquely identified robots in theory. but these would all be controlled by our team still (why would the other team send us ball_sense?)

soccer/src/soccer/coach/coach_node.cpp
```diff
@@ -0,0 +1,140 @@
...
+}
+
+void CoachNode::play_state_callback(rj_msgs::msg::PlayState::SharedPtr msg) {
+    current_play_state_ = *msg;
+    play_state_has_changed_ = true;
+}
+
+void CoachNode::world_state_callback(rj_msgs::msg::WorldState::SharedPtr msg) {
+    // EDGE-CASE NOTE: If robots from both teams are bordering the ball possession will likely
+    // switch repeatedly
```
kfu02, line 44: can you add a TODO (and ClickUp card) to make a new scenario where instead of switching rapidly, we return a "contested" state of some kind?

soccer/src/soccer/coach/coach_node.cpp
```diff
@@ -0,0 +1,140 @@
...
+    if (!possessing_) {
+        for (rj_msgs::msg::RobotState robot_state : msg->our_robots) {
+            // There definitely has to be a better way, but this works...
```
kfu02: there is, see the static `nearly_equals` here: https://github.com/RoboJackets/robocup-software/blob/efc4f3ae3540917e241a669a40cd997433446ed4/rj_geometry/include/rj_geometry/point.hpp#L400

or the non-static version here: https://github.com/RoboJackets/robocup-software/blob/efc4f3ae3540917e241a669a40cd997433446ed4/rj_geometry/testing/point_test.cpp#L205

kfu02: (if I'm not mistaken you should be able to just use robot_state.pose.position as an rj_geometry Point. maybe there is some conversion function I'm forgetting)

soccer/src/soccer/coach/coach_node.cpp
```diff
@@ -0,0 +1,140 @@
...
+                    .dist_to(rj_geometry::Point(msg->ball.position.x, msg->ball.position.y)) <
+                kRobotDiameter) {
+                possessing_ = true;
+                play_state_has_changed_ = true;
+                return;
+            }
+        }
+    } else {
+        for (rj_msgs::msg::RobotState robot_state : msg->their_robots) {
+            // See above...
```
kfu02: ^

soccer/src/soccer/coach/coach_node.cpp
```diff
@@ -0,0 +1,140 @@
...
+                kRobotDiameter) {
+                possessing_ = false;
+                play_state_has_changed_ = true;
+                return;
+            }
+        }
+    }
+}
+
+void CoachNode::ball_sense_callback(rj_msgs::msg::RobotStatus::SharedPtr msg, bool our_team) {
```
kfu02: listen to this check warning
```suggestion
void CoachNode::ball_sense_callback(const rj_msgs::msg::RobotStatus::SharedPtr &msg, bool our_team) {
```
https://stackoverflow.com/questions/2627166/what-is-the-difference-between-a-const-reference-and-normal-parameter

soccer/src/soccer/coach/coach_node.cpp
```diff
@@ -0,0 +1,140 @@
...
+    if (our_team && !possessing_) {
+        if (msg->has_ball_sense) {
+            possessing_ = true;
+            play_state_has_changed_ = true;
+        }
+    } else if (!our_team && possessing_) {
```
kfu02: again, we don't have the other team's ball sense, so this can be simplified

kfu02 (changes requested):
The testable stuff works great. However, we don't actually have the opponent's ball sense info. See my comments for more details.

soccer/src/soccer/coach/coach_node.cpp
N8BWert: I don't think we can do nearly_equals here because the robot position is its centroid which should never be nearly equal to the ball's location.

I'm also not entirely sure why, but even though both the robot's position and the ball's position are rj_geometry::points the rj_geometry point seems to be different than the one in the message so I had to make that ugly initializer.

kfu02: oh you can just change the tolerance on nearly equals to be the robot radius (or whatever threshold you had already)

kfu02: and yes, you have to use this converter function from ROS msg type to C++ type: 

https://github.com/RoboJackets/robocup-software/blob/ros2/rj_geometry/include/rj_geometry/geometry_conversions.hpp#L33

soccer/src/soccer/coach/coach_node.cpp
```diff
@@ -0,0 +1,127 @@
...
+
+CoachNode::CoachNode(const rclcpp::NodeOptions& options) : Node("coach_node", options) {
+    coach_state_pub_ =
+        this->create_publisher<rj_msgs::msg::CoachStateInterpretation>("/strategy/coach_state", 10);
+    play_state_change_timer_ =
+        this->create_wall_timer(100ms, [this]() { check_for_play_state_change(); });
+
+    play_state_sub_ = this->create_subscription<rj_msgs::msg::PlayState>(
+        "/referee/play_state", 10,
+        [this](const rj_msgs::msg::PlayState::SharedPtr msg) { play_state_callback(msg); });
```
kfu02, line 11:
```suggestion
        [this](const rj_msgs::msg::PlayState::SharedPtr &msg) { play_state_callback(msg); });
```
90% sure this will fix this check warning. don't forget to change the .hpp to match

soccer/src/soccer/coach/coach_node.cpp
```diff
@@ -0,0 +1,127 @@
...
+
+    world_state_sub_ = this->create_subscription<rj_msgs::msg::WorldState>(
+        "/vision_filter/world_state", 10,
+        [this](const rj_msgs::msg::WorldState::SharedPtr msg) { world_state_callback(msg); });
+
+    // initialize all of the robot status subscriptions
+    // initialize our robots
```
kfu02:
```suggestion
```

soccer/src/soccer/coach/coach_node.cpp
```diff
@@ -0,0 +1,127 @@
...
+    for (int i = 0; i < kRobotsPerTeam; i++) {
```
kfu02: can we change this to kNumShells? just in case we use the shells for robots >= 6

kfu02: not certain that's the exact syntax, btw, but it's in the same file as kRobotsPerTeam

soccer/src/soccer/coach/coach_node.cpp
kfu02: same comment for the other warnings of similar nature

kfu02 (changes requested):
Much better! GH Actions is giving you a few check warnings which should be cleaned up, and I have one real concern (see comment about kNumShells). Otherwise looking good!

kfu02 (approved):
nice work!

PR #1958, opened 2022-10-30
one goal at a time (WIP)

kasohrab:
## Description
Ensures Planning ActionServer only runs one goal per robot, and cancels previously running goals on receiving a new one. This is how SimpleActionServer in ROS 1 behaves. (In ROS2, by default any action server can handle any number of goals at once, which we found to be a bad model for our use case.)

Implemented via a classic "handshake agreement":
https://www.thefreedictionary.com/Handshake+(computing)


kfu02 (approved):
LGTM!

kfu02:
@kasohrab is the GOAT

PR #1960, opened 2022-11-02
Sumedh/tutorial

hdemusg:
Answer for 4 (from root of robocup-software): rj_msgs/msg/RobotIntent.msg


hdemusg:
Answer for part 5:
Publisher: /gameplay_node
Subscribers: /control and /planner

PR #1961, opened 2022-11-02
Deletion of Ready Checmark in Sim

RyanBard7:
## Description
Deletes the Yellow Checkmark in sim (ready) so it cannot be activated from the sim controls


kfu02 (commented):
is there a reason you chose to comment out the references to fastReady instead of just deleting them?


--Ryan - I left it in in case we ever needed it again, most recent commit deletes the references in the main_window cpp and hpp files

kfu02 (approved):
nice work, @rkbard !

PR #1969, opened 2022-11-21
Fix compiler warnings

cengman2017:
## Description
Made some minor changes to get rid of a couple compiler warnings. Did not change functionality of any code


rj_common/include/rj_common/field_dimensions.hpp
```diff
@@ -213,7 +213,7 @@ struct FieldDimensions {
     float goal_height_;
     float penalty_short_dist_;
     float penalty_long_dist_;
-    float arc_radius_;
+    //float arc_radius_;
```
kfu02: delete this line, don't just comment it out. we can always revert the commit if we want this code back :)

.vscode/settings.json
```diff
@@ -0,0 +1,4 @@
+{
```
kfu02: don't commit your vscode settings in the future. try this to remove it: https://stackoverflow.com/questions/18357511/git-remove-committed-file-after-push

soccer/src/soccer/planning/planner/plan_request.cpp
```diff
@@ -27,7 +27,7 @@ void fill_obstacles(const PlanRequest& in, rj_geometry::ShapeSet* out_static,
 
     // Add their robots as static obstacles (inflated based on velocity).
     // See calc_static_robot_obs() docstring for more info.
-    for (int shell = 0; shell < kNumShells; shell++) {
+    for (unsigned long shell = 0; shell < kNumShells; shell++) {
```
kfu02: kNumShells is actually of type size_t. This resolves to unsigned long, so this change removes the compiler warnings, but it would be better if it was something like 
```suggestion
    for (size_t shell = 0; shell < kNumShells; shell++) {
```

kfu02: (if that re-triggers compiler warnings let me know)

rj_geometry/include/rj_geometry/polygon.hpp
```diff
@@ -52,7 +52,7 @@ class Polygon : public Shape {
     std::string to_string() override {
         std::stringstream str;
         str << "Polygon<";
-        for (int i = 0; i < vertices.size(); i++) str << vertices[i] << ", ";
+        for (unsigned long i = 0; i < vertices.size(); i++) str << vertices[i] << ", ";
```
kfu02: try this: https://www.learncpp.com/cpp-tutorial/for-each-loops/

kfu02 (changes requested):
Nice work so far! I have a few comments. Also, the size_t advice applies to everywhere you have "unsigned long" because `size_t` is an alias for `unsigned long int`. So use `size_t` because it's shorter and clearer.

Reference: https://codeforces.com/blog/entry/44795?f0a28=1

soccer/src/rj_vision_filter/camera/camera.cpp
```diff
@@ -302,7 +302,7 @@ void Camera::update_robots_mhkf(RJ::Time calc_time, const std::list<CameraRobot>
     for (const CameraRobot& camera_robot : single_robot_list) {
         bool was_used = used_camera_robot.at(camera_robot_idx);
 
-        if (!was_used && single_kalman_robot_list.size() < PARAM_max_num_kalman_robots) {
+        if (!was_used && single_kalman_robot_list.size() < (unsigned long) PARAM_max_num_kalman_robots) {
```
kfu02: run `make pretty-lines` to get rid of this error

kfu02 (approved):
Looks good! Can you run `make pretty-lines`, commit that, and then merge?

PR #1953, opened 2022-10-22
Agent action client

kfu02:
## Description
Initial commit for new strategy framework, see the linked design doc. 

Immediate next steps:
 - assign positions via Coach node
 - make agent action client respect Coach node's velocity limits
 - apply useful behavior to Offense/Defense/Goalie

Long-term TODO:
 - why does the planner node crash? see #1970 
 - reconsider task/role differentiation and responsiveness, discuss (e.g. should "track ball" be a task, or is constantly reassigning a close PTMC with a different face point okay?) longer term

## Design Documents
[Link](https://docs.google.com/document/d/1qnf3ca5nMZWkTeyERrk6fsng_ZnlourK01f8Z9vLoPs/edit?usp=sharing)

### Test Case 1
1. Run sim
2. Offense should move up/down
3. Defense should move side/side
4. Goalie should be goalie

Expected result: no crash, works as intended

## Key Files to Review

Bulk of Changes
 * all files in `strategy/`
   * `agent/` controls the ROS interface, delegates actual creation of RobotIntents to `position/`'s classes to comply with Strategy Design Pattern (if this is not clear from docstrings, lmk)
 * `trajectory_collection.*pp` see #1970 
 * `planner_node.*pp`, `plan_request.*pp`

New Equality Checks
 * `motion_command.hpp`
 * `robot_intent.hpp`
 * `point.hpp` (makes floating-point RJ Points use approximate comparison rather than exact)

### Less Important Files to Review

Tiny Changes to Legacy Code
 * `field_dimensions.hpp`
 * `circle.hpp`
 * `trajectory_utils.*pp`
 * `trajectory.*pp` - changed some iterators to const iterators, added a new error check that should never get triggered

Build/Launch Files
 * `**/CMakeLists.txt`
 * `soccer.launch.py` no longer launches gameplay, launches associated strategy nodes instead
 * `agent_action_client_main.cpp`

Naming Changes to other PRs
 * `CoachState.msg` , `coach_node*.*pp` (#1946, #1954)
 * `**/robot_action_client/` -> agent_action_client (#1948)

Miscellaneous
 * `motion_control.cpp` changes planning param (will be overwritten whenever we tune PID)

## Review Checklist

- [x] **Docstrings**: All methods and classes should have the file appropriate docstrings which follow the guidelines in the ["Contributing" page](https://rj-rc-software.readthedocs.io/en/latest/contributing.html) of our docs.
- [x] **Remove extra print statements**: Any print statements used for debugging should be removed
[...]

PR #1970, opened 2022-11-21
Fix planner node

kfu02:
## Description
Fixes the planner node race condition by disabling use of the shared TrajectoryCollection object. (This is clearly not the best way to go about it, since now we don't account for our own planned trajectories when we plan new motion paths.)

For now, since my other PR #1953 relies on this and is blocking everything else, I am inclined to merge it (then merge #1953). Future steps include:
 * add our robots as dynamic obstacles (like we do for the opponent robots) to mitigate self-crashes
 * actually fix TrajectoryCollection instead of disabling it like I did

Some useful tips for actually fixing it:
1) run lldb on the planner node to get a backtrace like this:
```
(terminal 1) ros2 run rj_robocup planner_node
(terminal 2) sudo lldb-10 -n planner_node
(terminal 3) make run-sim
```
This order allows the debugger to catch the planner node crashing before the whole system launches.

2) the crash seemed to occur on the `robot_trajectories_->put(...)` call, not the `->get(...)` call

3) for some reason the mutex stuff tried in `trajectory_collection.hpp` did not fix this crash, unclear why

Fixes crashing in #1953 which stops that from blocking everything else.

## Design Documents
[Link](link-to-design-doc)

### Test Case 1
1. Run sim 

**Expected result:** no crash

## Review Checklist

### DIFF
--- a/soccer/src/soccer/planning/planner_node.cpp
+++ b/soccer/src/soccer/planning/planner_node.cpp
@@ -182,8 +182,7 @@ std::optional<RJ::Seconds> PlannerForRobot::get_time_left() const {
     // TODO(p-nayak): why does this say 3s even when the robot is on its point?
     // get the Traj out of the relevant [Trajectory, priority] tuple in
     // robot_trajectories_
-    const auto latest_trajs = robot_trajectories_->get();
-    const auto& [latest_traj, priority] = latest_trajs.at(robot_id_);
+    const auto& [latest_traj, priority] = robot_trajectories_->get(robot_id_);
     if (!latest_traj) {
         return std::nullopt;
     }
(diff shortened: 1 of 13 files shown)


soccer/src/soccer/planning/planner_node.cpp
```diff
@@ -182,8 +182,7 @@ std::optional<RJ::Seconds> PlannerForRobot::get_time_left() const {
     // TODO(p-nayak): why does this say 3s even when the robot is on its point?
     // get the Traj out of the relevant [Trajectory, priority] tuple in
     // robot_trajectories_
-    const auto latest_trajs = robot_trajectories_->get();
-    const auto& [latest_traj, priority] = latest_trajs.at(robot_id_);
+    const auto& [latest_traj, priority] = robot_trajectories_->get(robot_id_);
```
kfu02: TODO after break: try removing this call entirely (to see if it's this get() or the get() below that's causing concurrency issues)

kfu02: it's not this one

N8BWert (approved):
:1st_place_medal: It ran!

p-nayak11 (approved):
LGTM! Nice

kfu02:
this branch was originally squashed into #1953 but then we accidentally overwrote it, fixed via manual merge

PR #1971, opened 2022-11-30
Link Sim and Framework and add to Makefile

kfu02:
## Description
This PR creates a bash script to run framework (assuming simulator-cli is built and somewhere in the home directory) and the simulation with the same parent process, so they are killed at the same time. This is to limit instances of framework running on the same network. 

[ClickUp card](https://app.clickup.com/t/384g763)

### Test Case 1
1. `cd robocup-software && . source.bash`
2. `make run-sim`
3. Let sim open, then press ctrl-c twice in the shell.

**Expected result:**
When the sim runs, framework automatically runs in the background. You should see six robots on the sim field. When ctrl-c is pressed, first the sim is killed and then the framework is killed. `ps -a` should show no sim or framework related processes left

## Key Files to Review
 * launch/framework.bash
 * makefile

## Review Checklist


kfu02:
@sid-parikh great work! I made some light naming modifications to clarify the difference between run-sim and what you called "run-sim-only"

PR #1978, opened 2022-12-06
Revert "Agent action client"

kfu02:
Reverts RoboJackets/robocup-software#1953

Forgot to re-merge ros2 before merging in #1953 .


kfu02:
never mind, the latest commits from ros2 are still being overwritten even after doing it right. reverting this revert

PR #1980, opened 2022-12-06
New goalie idle task

kfu02:
## Description
Makes Goalie's idle behavior (what it does when not blocking shots) much smoother by turning it into a planner instead of repeated calls to PathTargetPlanner. 

Also exposes some key weaknesses of the current planning stack (see below).

[ClickUp card](https://app.clickup.com/t/396mwu6)

### Test Case 1
1. Build, run sim.
2. Move ball around (slowly).
3. Take some shots on goal.
4. Compare with what is currently on `ros2`.

**Expected result:** Goalie should idle when ball is slow and intercept when ball is fast. The idle behavior, specifically, should look much cleaner than it does on `ros2`.

## Key Files to Review
short PR, all of them

## Review Checklist

## Addendum: Planning Gripes

1. It's hard to add new planners (the changes made in `planning/tests/conversion_tests.cpp` are vital to get it to link properly, which is not intuitive at all)
2. Most tasks are just some extension of PathTargetPlanner, including this one, but since PathTargetPlanner is a templated class linked to PathTargetMotionCommand.msg, I can't use it in another planner very easily.
3. The [equality check](https://github.com/RoboJackets/robocup-software/pull/1980/files#diff-7a4b491bf7b4c369564c340fdcc53ef9ceb30d1668a41e4063ac34ee174704b4R67) in `agent_action_client.cpp` currently uses the built-in equality check autogenerated for `RobotIntent.msg`, which doesn't correctly handle the way we've formatted `MotionCommand.msg` as a series of 1-element arrays of different msg types. (`MotionCommand.msg` is a bit janky in and of itself for that reason, btw.) 
 * The workaround for this is to use `rj_convert::convert_from_ros(...)` to convert the msg to a struct (as defined in `robot_intent.hpp`, which will then use that struct's operator== overload. 

In my opinion, it would be nicer to separate the C++ struct and the ROS msg type,  rather than having both refer to the same thing. Something like a pure C++ TaskPlanner superclass, where the subclasses are other classes representing different task types (GoalieIdle, PathTarget, etc.) and can call each other. The superclass should then translate different task types to a single simple `PlanRequest.msg` which has just `robot_id`, `goal_instant`, `face_pt`, `ignore_ball`.

This would simplify our current PlannerNode to what we currently consider PathTargetPlanner, and increase the complexity of the strategy side to do some translation of semantic "tasks" into concrete motion planning requests before sending them to PlannerNode via ROS.

### DIFF
--- a/docs/source/getting_started.rst
+++ b/docs/source/getting_started.rst
@@ -76,7 +76,8 @@ more information).
 
 .. code-block:: bash
 
-    git pull git checkout ros2
+    git pull 
+    git checkout ros2
 
 Then, source the ROS setup file. This allows your shell to use ROS commands.
 
@@ -143,10 +144,15 @@ terminal, you can launch sim with:
 
 .. code-block:: bash
 
-   . ./source.bash make run-sim
+   . ./source.bash 
+   make run-sim
 
 ``source.bash`` is an alias for the two source commands you saw above, and
-``make run-sim`` is an alias for ``ros2 launch rj_robocup sim.launch.py``.
+``make run-sim`` will launch both ER-Force's Framework (the physics simulator)
+and our stack (``ros2 launch rj_robocup sim.launch.py``). 
+
+To stop this process (like any other) press CTRL-C in the command line. You may
+have to press CTRL-C twice.
 
 Since Python is not compiled, if you're exclusively working on Python files, and
 staying in one terminal, it's likely that you'll mostly only need:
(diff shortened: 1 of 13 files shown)


kasohrab (approved):
Looks great! an improvement although it's stuck working with our current design

kfu02:
@kasohrab to be fair, planning is very performant in its current state. Just a bit hard to add to 😃

PR #1983, opened 2022-12-07
Planner ball obstacle

sid-parikh:
## Description

This PR updates the planning system to consider the ball a dynamic obstacle, and plan paths that keep a certain distance away from the ball as published by the coach node

Resolves [make planner respect ball on stop](https://app.clickup.com/t/34c4duz)

## Design Documents
[Link](link-to-design-doc)

### Test Case 1
1. Run the sim (green check/normal mode). Currently, six robots move in a predefined manner. Robot 1 (the defender) is the only one set to pay any attention to the ball. 
2. Place the ball in/near the path of robot 1
3. Give the ball some velocity in a collision course with robot 1's path

**Expected result:** Robot 1 avoids touching the ball

### Test Case 2: STOP
1. Run the sim in STOP. (Coach node should be publishing a min_dist_from_ball at 0.5)
2. Place the ball in/near the path of robot 1
4. Give the ball some velocity in a collision course with robot 1's path

**Expected result:** Robot 1 should keep a larger radius (0.5 in some unit?) around the ball.

## Key Files to Review
Plan Request: struct changed to keep a `min_dist_from_ball` field. also changed `fill_obstacles` (utility called by planners) to use the field.
 * `plan_request.hpp`
 * `plan_request.cpp`

Planner Node: now maintains a `coach_state` subscription and inserts the `min_dist_from_ball` data into each plan request when creating them from RobotIntents.
 * `planner_node.hpp`
 * `planner_node.cpp`

## Review Checklist

### DIFF
--- a/soccer/src/soccer/planning/planner/plan_request.cpp
+++ b/soccer/src/soccer/planning/planner/plan_request.cpp
@@ -63,7 +63,7 @@ void fill_obstacles(const PlanRequest& in, rj_geometry::ShapeSet* out_static,
     if (avoid_ball && out_dynamic != nullptr && out_ball_trajectory != nullptr) {
         // Where should we store the ball trajectory?
         *out_ball_trajectory = in.world_state->ball.make_trajectory();
-        out_dynamic->emplace_back(kBallRadius, out_ball_trajectory);
+        out_dynamic->emplace_back(kBallRadius + in.min_dist_from_ball, out_ball_trajectory);
     }
 }
 
--- a/soccer/src/soccer/planning/planner/plan_request.cpp
+++ b/soccer/src/soccer/planning/planner/plan_request.cpp
@@ -69,8 +69,9 @@ void fill_obstacles(const PlanRequest& in, rj_geometry::ShapeSet* out_static,
 
         if (in.debug_drawer != nullptr) {
             QColor draw_color = Qt::red;
-            in.debug_drawer->draw_circle(rj_geometry::Circle(in.world_state->ball.position, static_cast<float>(radius)),
-                                              draw_color);
+            in.debug_drawer->draw_circle(
+                rj_geometry::Circle(in.world_state->ball.position, static_cast<float>(radius)),
+                draw_color);
         }
     }
 }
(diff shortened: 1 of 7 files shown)


soccer/src/soccer/planning/planner_node.hpp
```diff
@@ -91,6 +98,10 @@ class SharedStateInfo {
         auto lock = std::lock_guard(mutex_);
         return &last_world_state_;
     }
+    [[nodiscard]] float min_dist_from_ball() const {
+        auto lock = std::lock_guard(mutex_);
```
kfu02, line 102: nice

soccer/src/soccer/planning/planner/plan_request.cpp
```diff
@@ -63,7 +63,7 @@ void fill_obstacles(const PlanRequest& in, rj_geometry::ShapeSet* out_static,
     if (avoid_ball && out_dynamic != nullptr && out_ball_trajectory != nullptr) {
         // Where should we store the ball trajectory?
         *out_ball_trajectory = in.world_state->ball.make_trajectory();
-        out_dynamic->emplace_back(kBallRadius, out_ball_trajectory);
+        out_dynamic->emplace_back(kBallRadius + in.min_dist_from_ball, out_ball_trajectory);
```
kfu02: can you add a call to the debug drawer here that draws a red circle around the ball when min_dist>0? (like you saw the old system do)

here is an example: https://github.com/RoboJackets/robocup-software/blob/726fba6e4c6559b93704d28368d73f26d7adc9e9/soccer/src/soccer/planning/planner/path_target_planner.cpp#L41

kfu02 (changes requested):
This is a really good PR, @sid-parikh ! I just have one small request: can you reenable drawing the circle around the ball? (I provided an example in PR comment)

(also, can you enable for all robots instead of just robot 1?)

kfu02:
Oh, to answer your question about units: everything is in meters. So our robots are 0.09 meters in radius, the STOP dist is 0.5 m, etc.

soccer/src/soccer/planning/planner/plan_request.cpp
sid-parikh: Added in the latest commit. Should be visible in all previously described tests. I wanted to do this earlier, but couldn't figure it out. Thanks for helping!

kasohrab (approved):
LGTM

PR #1982, opened 2022-12-07
Add angle comparison

cengman2017:
## Description
Added angle comparison to overloaded equality operator for PathTargetoCommand struct in soccer/src/soccer/planning/planner/motion_command.hpp 

## Associated / Resolved Issue
[ClickUp card](https://app.clickup.com/t/35zpjtf)

**Expected result:**???
Operator will return true if position, velocity, ignore_ball field, and angle, which is a variant data type that can hold one of 3 types of structs, are all equal.

PR #1985, opened 2022-12-09
New add angle comparison

cengman2017:
New version of add angle comparison: https://github.com/RoboJackets/robocup-software/pull/1982

### DIFF
--- a/soccer/src/soccer/planning/planner/motion_command.hpp
+++ b/soccer/src/soccer/planning/planner/motion_command.hpp
@@ -62,15 +62,15 @@ struct PathTargetCommand {
             std::holds_alternative<TargetFaceTangent>(ptc.angle_override)) {
             angle_eq = &(angle_override) == &(ptc.angle_override);
         } else if (std::holds_alternative<TargetFaceAngle>(angle_override) &&
-            std::holds_alternative<TargetFaceAngle>(ptc.angle_override)) {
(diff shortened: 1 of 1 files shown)


soccer/src/soccer/planning/planner/motion_command.hpp
```diff
@@ -56,9 +56,19 @@ struct PathTargetCommand {
     bool operator==(const PathTargetCommand& ptc) {
         bool pos_eq = goal.position == ptc.goal.position;
         bool vel_eq = goal.velocity == ptc.goal.velocity;
-        // TODO(Kevin): fix this to actually compare std::variants
-        bool angle_eq = true;
-        /* bool angle_eq = goal.velocity == ptc.goal.velocity; */
+        bool angle_eq = false;
+        if (std::holds_alternative<TargetFaceTangent>(angle_override) &&
+            std::holds_alternative<TargetFaceTangent>(ptc.angle_override)) {
+            angle_eq = &(angle_override) == &(ptc.angle_override);
```
kfu02: small nit, but usually the & won't have parens around the target variable
```suggestion
            angle_eq = (&angle_override == &ptc.angle_override);
```

kfu02 (approved):
Looks good. I would ask you to prove that this works in a test, but I recently discovered while working on #1980 that this comparison is actually unused right now in our agent system, so that would be hard for you to do.

PR #1987, opened 2022-12-19
Goalie block task

kfu02:
## Description
Follow-up to #1980 that makes goalie's block shots behavior use the Intercept planner. (Review #1980 first please!)

Also gives Position's get_task() an optional return type so that the position can return nullopt when no new task is requested.

### Test Case 1
1. Run sim
2. Take some shots on goal (left click to place ball, right click and drag to shoot ball)

**Expected result:** When shots are on target, goalie moves to try to intercept it. Otherwise, it should idle. (Clearing/passing yet to be implemented.) 

## Key Files to Review
**optional get_task()**
 * position.*pp
 * offense.*pp
 * defense.*pp
 * agent_action_client.*pp

**Goalie Intercept**
 * all else

## Review Checklist

### DIFF
--- a/soccer/src/soccer/planning/planner/intercept_planner.hpp
+++ b/soccer/src/soccer/planning/planner/intercept_planner.hpp
@@ -9,7 +9,8 @@ namespace planning {
  * Whether this means moving and stopping in the path of the ball
  * or completely driving through and "slapping" the ball.
  *
- * Mostly used for the goalie to block shots (w/ a target point of 0,0).
+ * Mostly used for the goalie to block shots (by giving it a target point of
+ * our goal).
  */
 
 class InterceptPlanner : public PlannerForCommandType<InterceptCommand> {
--- a/soccer/src/soccer/planning/planner/intercept_planner.hpp
+++ b/soccer/src/soccer/planning/planner/intercept_planner.hpp
@@ -1,52 +1,24 @@
 #pragma once
 
-#include <spdlog/spdlog.h>
-
-#include <rj_geometry/point.hpp>
-
 #include "planner.hpp"
-#include "planning/primitives/replanner.hpp"
 
 namespace planning {
 
 /**
- * Planner which tries to block a shot on goal as quickly as possible
+ * Planner which tries to intercept the path ball as quickly as possible
  * Whether this means moving and stopping in the path of the ball
  * or completely driving through and "slapping" the ball.
  *
- * (More accurate name would be "GoalieBlock", but Intercept kept for legacy
- * reasons.)
+ * Mostly used for the goalie to block shots (w/ a target point of 0,0).
  */
 
 class InterceptPlanner : public PlannerForCommandType<InterceptCommand> {
 public:
-    InterceptPlanner() : PlannerForCommandType<InterceptCommand>("intercept"){};
+    InterceptPlanner()
+        : PlannerForCommandType<InterceptCommand>("InterceptPlanner"){};
 
     Trajectory plan(const PlanRequest& request) override;
 
     [[nodiscard]] bool is_done() const override;
-
-    /*
-     * @return if a shot on goal is coming
-     *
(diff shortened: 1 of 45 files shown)


soccer/src/soccer/planning/planner/intercept_planner.hpp
```diff
@@ -9,7 +9,7 @@ namespace planning {
  * Whether this means moving and stopping in the path of the ball
  * or completely driving through and "slapping" the ball.
  *
- * Mostly used for the goalie / defenders to block shots
+ * Mostly used for the goalie to block shots (w/ a target point of 0,0).
```
kasohrab: Is it really always going to have a target of 0,0? Even if we are the other team?

soccer/src/soccer/strategy/agent/position/goalie.cpp
```diff
@@ -16,70 +16,44 @@ rj_msgs::msg::RobotIntent Goalie::get_task() {
...
-        // update intent
-        intent.motion_command.path_target_command = {ptmc};
+    if (shot_on_goal_detected(world_state)) {
+        // TODO(Kevin): fix intercept planner's is_done, then add in logic to
+        // clear/pass ball once done intercepting
+        auto intercept_mc = rj_msgs::msg::InterceptMotionCommand{};
+        intercept_mc.target.x = 0.0;
+        intercept_mc.target.y = 0.1;
+        intent.motion_command.intercept_command = {intercept_mc};
+        intent.motion_command.name = "intercept";
```
kasohrab, line 27: I like the .name addition. It makes the factory design pattern more viable also.

soccer/src/soccer/strategy/agent/agent_action_client.cpp
```diff
@@ -62,11 +62,26 @@ void AgentActionClient::get_task() {
...
+        // so any custom operator== overloads written don't apply
+        // TODO(Kevin): fix this by making Positions return RobotIntent structs, not msgs
+        if (task != last_task_) {
+            /* if (robot_id_ == 0) { */
+            /*     SPDLOG_INFO("sending new task {}", task.motion_command.name); */
+            /* } */
+            last_task_ = task;
+            send_new_goal();
+        } else {
+            /* if (robot_id_ == 0) { */
```
kasohrab, line 79: we can delete these commented out lines when merging.

kasohrab (approved):
Looks good! I approved but also left some small questions/requests

soccer/src/soccer/strategy/agent/position/goalie.cpp
kfu02: In #1988 I actually moved the .name field to RobotIntent.msg instead of MotionCommand.msg, for easier conversion to C++ struct (see this [commit](https://github.com/RoboJackets/robocup-software/pull/1988/commits/ccd0b504c706a0991092940699db829862c4dbb2)).

I can move it back and be less lazy with the ROS->C++ conversion methods. But while working on the PR I realized the current planner architecture is a bit less easy to convert to use a Factory method than I first thought, so I decided to go with whatever was easiest for me to add for debugging purposes.

Where do you think it should go?

soccer/src/soccer/strategy/agent/agent_action_client.cpp
kfu02: ah, this is superseded by the same lines in #1988 (see [here](https://github.com/RoboJackets/robocup-software/pull/1988/files#diff-7a4b491bf7b4c369564c340fdcc53ef9ceb30d1668a41e4063ac34ee174704b4R72-R75)) so it won't last long on ros2 anyways 😆

soccer/src/soccer/planning/planner/intercept_planner.hpp
kfu02: you are right, should say "w a target point of our goal", I'll fix that

PR #1988, opened 2022-12-23
Planning/strategy interface refactor

kfu02:
## Description

A refactoring PR to make it easier to add more complex strategy later. No visible on-field changes. Firmly establishes tasks/RobotIntents as long-running actions, as the ROS Action model requires. 

Big changes:
 * Add FaceBall option for PathTargetMotionCommand, instead of having to pass in an updated ball position to FacePoint. This makes the task "move here and face the ball" consistent in the eyes of the ActionClient, unlike before.
 * Confines use of ROS msg types to the ROS nodes they are sent/received from, to alleviate confusion over when our classes vs. ROS' autogenerated classes are used (see PR #1980 )
 * Uses the [Template Method + non-virtual interface](https://www.sandordargo.com/blog/2022/08/24/tmp-and-nvi) pattern to simplify get_task() in Position.
 * Moves the comparison operator overloads from `conversion_tests.cpp` (??) to `motion_command.cpp`, so they're next to their class definitions.
 * Debug drawer will now draw robot endpoint for all planner types (not just PathTargetPlanner)
 * RobotIntent.msg and RobotIntent struct now have a `motion_command_name` string for ease of introspection (warning: not enforced internally)
 * Standardize naming across ROS/motion_command.hpp

### Test Case 1
1. Run sim
2. Compare with #1987 

**Expected result:** Should look identical, except now intercept planner should draw a black circle too (not true in #1987 ).

Again (like #1811 ) my main goal was to ease future development, not introduce new changes, so review the code (especially the subclasses of Position).

## Key Files to Review
Reviewing by commit should be easiest here, I did a fairly good job of giving meaningful commit msgs. 

(**Except** in "use aggregate initialization for MotionCommands to be concise" I also gave RobotIntent.msg a motion_command_name field, which I forgot about when committing.)

## Review Checklist

### DIFF
--- a/rj_geometry/include/rj_geometry/point.hpp
+++ b/rj_geometry/include/rj_geometry/point.hpp
@@ -7,6 +7,7 @@
 #include <Eigen/Dense>
 #include <QtCore/QPointF>
 #include <boost/functional/hash.hpp>
+#include <spdlog/spdlog.h>
 
 #include <rj_geometry/util.hpp>
 #include <rj_geometry_msgs/msg/point.hpp>
@@ -173,9 +174,9 @@ public:
 
     /**
      * compares two points to see if both x and y are the same
-     * adds the == operator
+     * with the == operator and approx. double equality
      */
-    [[nodiscard]] bool operator==(Point other) const { return this->nearly_equals(other); }
+    [[nodiscard]] bool operator==(const Point& other) const { return this->nearly_equals(other); }
 
     /**
      * this is the negation of operator operator !=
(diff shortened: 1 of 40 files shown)


p-nayak11 (changes requested):
All hierarchical changes to the planner refactor makes sense. There are some style checks that are not being passed so fix those. Other than that, LGTM.

kfu02:
> All hierarchical changes to the planner refactor makes sense. There are some style checks that are not being passed so fix those. Other than that, LGTM.

@p-nayak11 What's weird is that a call to `make pretty-lines` doesn't change anything. So I'm not sure why the style checks are failing, but I'm inclined to merge regardless because all the "style issues" are in `motion_command.hpp`. If you run `make pretty-lines` on this branch and it changes something please commit directly for me!

p-nayak11 (approved):
Make pretty-lines is broken again. :(

PR #1990, opened 2023-01-05
Goalie clear

kfu02:
## Description
Re-enables kicking by adding a manipulator-setpoint publisher to planner node. (Also deletes the old node that published manipulator setpoints.) Now goalie clears!

### Test Case 1
1. Run sim.

**Expected result:** Goalie clears ball now when it is slow inside the goalie box. (Direction still a work in progress.)

## Review Checklist


kasohrab (approved):
LGTM

PR #1991, opened 2023-01-11
Hotfix: Make ROS2 run again (far trajectory bug)

N8BWert:
## Description
ros2 was broken because the velocity_profiling.cpp file had a line that was adding incredibly far away points to the trajectory (points with x and y in the magnitude of 10^17).  This caused the time to get to the point to be incredibly high, so high in fact that it seems that all of the times were approximately equal.  This change makes the planner ignore points that would take an extreme amount of time to reach.

I created a clickup card for stopping the out-of-bounds paths from being planned in the first place (https://app.clickup.com/t/86773tf5n)


anshulm07 (approved):
LGTM

PR #1997, opened 2023-01-30
Fix Framework Script 

sid-parikh:
## Description
The `make run-sim` script was sometimes leaving the ER-Force simulator running. When the script was later run again, this would result in two instances of ER-Force and create strange issues.
This PR attempts to correct this problem by manually killing `simulator-cli` instances before and after the script runs.

Resolves [ClickUp card](https://app.clickup.com/t/86776zrap)

### Test Case 1
1. Run `simulator-cli` manually
2. Run framework.bash in a separate shell.

**Expected result:** the original `simulator-cli` should be terminated and a new one will be created. one way to check is the PID should change. 

## Key Files to Review
 * framework.bash

## Review Checklist

### DIFF
--- a/launch/framework.bash
+++ b/launch/framework.bash
@@ -6,6 +6,6 @@ echo "launching ER-Force framework and our UI"
 # first trap SIGINT to kill the entire subshell
 # then find the location of simulator-cli in the home directory and run it with args -g 2020B (in background)
 # then run sim through makefile shortcut (in background)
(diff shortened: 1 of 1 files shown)


kfu02 (approved):
Nice! I don't know what you did to the TRAP instruction, but CTRL-C seems to actually kill the processes correctly now too.

PR #1996, opened 2023-01-25
Defense to waller refactor

Ninjatest321:
## Description
This PR changes the behavior of any robot assigned as a defensive robot to act as a waller between the ball and the goal. The wall is not cooperative between the robots at the time of this PR, but multiple defenders will form a clump as seen in the gif later in this PR.

## Associated / Resolved Issue
https://app.clickup.com/t/86773u1ke
https://app.clickup.com/t/389zmma

## Steps to Test
1. Run Simulator
2. See any robot set as a defensive robot follow the ball outside of the perimeter, as shown by robots 1 and 2 in the gif below. You can move the ball with the mouse and watch the defensive robots follow it.

ezgif-3-cdf34271c2

## Key Files to Review
Defense.cpp
Defense.hpp

PR #2007, opened 2023-02-22
Blockers

dwu359:
## Description
This is a defense strategy to move in between the ball and the goal to block anyone from shooting the ball into the goal.

Resolves [ClickUp card](https://app.clickup.com/t/86777u61x)

### Test Case 1
1. make run-sim
2. place a ball somewhere on the same side as the defenders' goal. 
**Expected result:**
Defender robots should move in between the ball and the goal.

## Key Files to Review
Group 1
 * defense.cpp
 * defense.hpp
 * blockers.cpp
 * blockers.hpp

## Review Checklist

- [ ] Step 1
- [ ] Step 2


kfu02 (changes requested):
hey Daniel, can you run `make pretty-lines` to fix the style errors?

also, let's wait on Amogh's PR (#1996 ) so you can use that interface, work with him to ensure your role will work with it

soccer/src/soccer/strategy/agent/position/blocker.cpp
```diff
@@ -0,0 +1,19 @@
+#include "blocker.hpp"
+
+namespace strategy {
+Blocker::Blocker(double factor) : factor{factor} {}
+std::optional<RobotIntent> Blocker::get_task(RobotIntent intent,
+                                             const WorldState* const world_state) {
+    // face ball
+    planning::PathTargetFaceOption face_option = planning::FaceBall{};
+    // avoid ball
+    constexpr bool ignore_ball = false;
```
kfu02: should be IGNORE_BALL

soccer/src/soccer/strategy/agent/position/blocker.cpp
```diff
@@ -0,0 +1,19 @@
...
+    auto goal_center = rj_geometry::Point{0.0, 0.0};  // TODO: replace with field coordinates
+    planning::LinearMotionInstant goal{
+        (world_state->ball.position - goal_center) * factor + goal_center,
```
kfu02: so right now it looks like the blocker has to block between the goal and the ball, which is the same behavior as the waller. can you change this to block between any two arbitrary points (e.g. two robots' positions)?

kfu02 (changes requested):
hey Daniel, can you fix all the check warnings? also, see my comment about making the marker work between any two points (also let's rename Blocker to Marker)

kfu02 (commented):
can you resolve the merge conflicts with ros2? I think they come from your recent changes in #2015

kfu02 (changes requested):
hey Daniel, you have a few check warning complaints to fix, shouldn't take too long. (google it if any are confusing, these are pretty common warnings.) nice work so far!

soccer/src/soccer/strategy/agent/position/marker.cpp
```diff
@@ -0,0 +1,32 @@
...
+std::optional<RobotIntent> Marker::get_point_task(RobotIntent intent,
+                                                  const rj_geometry::Point& point1,
+                                                  const rj_geometry::Point& point2) {
+    // face ball
+    planning::PathTargetFaceOption face_option = planning::FaceBall{};
+    // avoid ball
+    constexpr bool IGNORE_BALL = false;
+    planning::LinearMotionInstant goal{(point2 - point1) * factor + point1,
+                                       rj_geometry::Point{0.0, 0.0}};
+    intent.motion_command = planning::PathTargetMotionCommand{goal, face_option, IGNORE_BALL};
```
kfu02: merge in ros2. thanks to the changes in #2020 , this should be much easier to work with now. see[ these changes](https://github.com/RoboJackets/robocup-software/pull/2020/files#diff-ca908df5b2761a194bedc670ac10709d077ac28b53ec7d8a0abeb8596d775380R38) in particular

kfu02 (changes requested):
hey Daniel, the behavior looks pretty good! I will say the one thing missing is that when the ball is kicked at the marker, the marker will back up rather than stay in place to steal the ball. can you add a bit of logic to say "if the ball is being kicked in my vicinity, go try to intercept it?" the latest passing PR #2037 should help with that

sid-parikh:
Working wallers code can be found in comp 23 branches.

PR #2020, opened 2023-03-15
Rewrite planner node

kfu02:
## Description
Refactored PlannerNode to make it more usable and robust. Also set up the groundwork for making planners composable (now called PathPlanners to clarify their usage).

Detailed changes:
 * PlannerNode will now default to some behavior when PathPlanners fail, rather than crashing soccer
     * will also log WARNs
 * `*MotionCommand.msg` (e.g. `PathTargetMotionCommand.msg`) all collapsed into a single `MotionCommand.msg`
 * removed PlannerForCommandType, so Planners no longer need an associated ROS msg
     * this opens the door for composability and simplifies the workflow for adding a new planner
 * PathPlanners now referenced directly by name rather than the annoying is_applicable() interface

It used to be really hard to add new planners (see #1980). Now it should be less hard, which allows PathPlanners to serve as complex single-robot behaviors. For instance, walling (#1996) should be a planner, rather than repeated calls to PTMC. This can be coordinated by higher level strategy, but making the actual motion defined by planning is cleaner and on a faster refresh rate.

### Test Case 1
1. Run sim.

**Expected result:** no crashes! WARN logs when default planner active.

## Key Files to Review
`strategy/*`, `planning/planner/*`
 * In strategy/, notice how much cleaner the constructors of MotionCommand look than before.
 * Ditto in planning/planner

`rj_msgs/msg/*`
 * See the new `MotionCommand.msg` and how it unions the old `*MotionCommands`

`planner_node.*pp`
 * take the most time with this one

## Review Checklist


N8BWert (approved):
lgtm

soccer/src/soccer/planning/planner/escape_obstacles_path_planner.hpp
```diff
@@ -17,10 +17,16 @@ namespace planning {
...
+ * Params taken from MotionCommand:
+ *   None
  */
-class EscapeObstaclesPathPlanner : public Planner {
+class EscapeObstaclesPathPlanner : public PathPlanner {
 public:
-    EscapeObstaclesPathPlanner() : Planner("EscapeObstaclesPathPlanner"){};
+    // TODO(Kevin): think of better way to convey halted behavior than "EscapeObstaclesPathPlanner"
+    // actually this is not HALT, control node does HALT for us
+    // or maybe use PathTargetPlanner with no real velocity?
```
p-nayak11, line 29: This makes more sense intuitively. We should be stopping on HALT rather than trying to escape obstacles.

soccer/src/soccer/planning/planner/escape_obstacles_path_planner.hpp
```diff
@@ -53,7 +52,8 @@ class EscapeObstaclesPathPlanner : public Planner {
     [[nodiscard]] bool is_done() const override;
 
 private:
-    PathTargetPlanner planner_;
+    // TODO(Kevin): delete this if unused
```
p-nayak11: Did you end up using the variable?

soccer/src/soccer/planning/planner/planner.hpp
```diff
@@ -6,39 +6,42 @@
...
-class Planner {
-public:
-    explicit Planner(std::string name) : name_(std::move(name)) {}
-    virtual ~Planner() = default;
 
-    Planner(Planner&&) noexcept = default;
-    Planner& operator=(Planner&&) noexcept = default;
-    Planner(const Planner&) = default;
-    Planner& operator=(const Planner&) = default;
+// TODO(Kevin): change this filename to path_planner.hpp!!
```
p-nayak11: Change the file name.

soccer/src/soccer/planning/planner_node.cpp
```diff
@@ -144,24 +145,30 @@ void PlannerNode::execute(const std::shared_ptr<GoalHandleRobotMove> goal_handle
...
+    planners_[PathTargetPathPlanner().name()] = std::make_shared<PathTargetPathPlanner>();
+    planners_[SettlePathPlanner().name()] = std::make_shared<SettlePathPlanner>();
+    planners_[CollectPathPlanner().name()] = std::make_shared<CollectPathPlanner>();
+    planners_[LineKickPathPlanner().name()] = std::make_shared<LineKickPathPlanner>();
+    planners_[PivotPathPlanner().name()] = std::make_shared<PivotPathPlanner>();
+    planners_[EscapeObstaclesPathPlanner().name()] = std::make_shared<EscapeObstaclesPathPlanner>();
+
+    // EscapeObstaclesPathPlanner is our default path planner
+    // because when robots start inside of an obstacle, all other planners will fail
+    // TODO(Kevin): make EscapeObstaclesPathPlanner default start of all planner FSM so this doesn't
```
p-nayak11: From what I can tell while going through the PR, this hasn't been done yet. I'm going to make this a low priority card.

soccer/src/soccer/planning/planner_node.cpp
```diff
@@ -237,19 +244,22 @@ std::optional<RJ::Seconds> PlannerForRobot::get_time_left() const {
...
-    const auto def_area_obstacles = shared_state_->def_area_obstacles();
-    const auto goalie_id = shared_state_->goalie_id();
-    const auto play_state = shared_state_->play_state();
-    const bool is_goalie = goalie_id == robot_id_;
-    const auto min_dist_from_ball = shared_state_->coach_state().global_override.min_dist_from_ball;
-    const auto max_robot_speed = shared_state_->coach_state().global_override.max_speed;
+    // pass global_state_ directly
+    const auto* world_state = global_state_.world_state();
+    const auto goalie_id = global_state_.goalie_id();
+    const auto play_state = global_state_.play_state();
```
p-nayak11, line 245: Set it to null? Address the github action below.

soccer/src/soccer/planning/planner_node.cpp
```diff
@@ -307,63 +317,75 @@ PlanRequest PlannerForRobot::make_request(const RobotIntent& intent) {
...
+    }
+
+    return trajectory;
+}
+
+Trajectory PlannerForRobot::safe_plan_for_robot(const planning::PlanRequest& request) {
+    Trajectory trajectory;
+    try {
+        trajectory = unsafe_plan_for_robot(request);
+    } catch (std::runtime_error exception) {
```
p-nayak11, line 350: Catch by reference instead of by value according to the github action below.

soccer/src/soccer/planning/tests/conversion_tests.cpp
```diff
@@ -13,62 +13,65 @@ namespace planning {
...
-    PathTargetMotionCommand path_target{LinearMotionInstant(Point(1.0, 2.0), Point(3.0, 4.0)),
-                                        FacePoint{Point(1.0, 2.0)}};
-    test_lossless_convert_cpp_value(MotionCommand{EmptyMotionCommand{}});
-    test_lossless_convert_cpp_value(MotionCommand{path_target});
-    test_lossless_convert_cpp_value(MotionCommand{WorldVelMotionCommand{Point(1.0, 2.0)}});
-    test_lossless_convert_cpp_value(
-        MotionCommand{PivotMotionCommand{Point(1.0, 2.0), Point(3.0, 4.0)}});
-    test_lossless_convert_cpp_value(MotionCommand{InterceptMotionCommand{Point(1.0, 2.0)}});
-}
+// TODO(Kevin): add conversion test for new MotionCommand
```
p-nayak11, line 16: I'm guessing that this is a low priority task right now?

p-nayak11 (changes requested):
I left some comments about what needs to change. In terms of the functionality of the PR, there seems to be a bug that crashes the stack. Based on the terminal, it is "throwing an instance of std::bad_alloc" which is why it is currently terminating. It's occurring after PathPlanner couldn't create a valid Trajectory for line_kick. However, it also happened while I was trying to move the ball towards the offense role robot. (I tried this a few more times and it did not break so I'm not sure if this is valid). I'm not sure what the root of the problem is right now.

soccer/src/soccer/planning/planner/escape_obstacles_path_planner.hpp
kfu02: HALT is actually implemented control-side, the control node simply gives every HALT robot 0 velocity. It's annoying because we have to give some non-empty Trajectory for that to happen though (hence why EscapeObstacles is the default)

soccer/src/soccer/planning/planner/escape_obstacles_path_planner.hpp
kfu02: no I deleted it already, my b

kfu02:
```suggestion
```

soccer/src/soccer/planning/planner_node.cpp
kfu02: don't worry about this, it requires composable planners to work

soccer/src/soccer/planning/tests/conversion_tests.cpp
kfu02: yes if you could card this that'd be great

soccer/src/soccer/planning/planner/escape_obstacles_path_planner.hpp
kfu02: https://github.com/RoboJackets/robocup-software/pull/2020/files#diff-0dc67989d0c9e16403a6d3b3bb626886e3a97dc0d8eb1460aeb9f5938ec22065R361
related

kfu02:
> I left some comments about what needs to change. In terms of the functionality of the PR, there seems to be a bug that crashes the stack. Based on the terminal, it is "throwing an instance of std::bad_alloc" which is why it is currently terminating. It's occurring after PathPlanner couldn't create a valid Trajectory for line_kick. However, it also happened while I was trying to move the ball towards the offense role robot. (I tried this a few more times and it did not break so I'm not sure if this is valid). I'm not sure what the root of the problem is right now.

I turned a bunch of shared_ptrs into unique_ptrs, which should prevent accidental memory leaks from incorrectly transferring ownership (though really shared_ptrs are designed to handle what I was doing previously, so idk if that really changed much). 

After that, I was able to recreate the planner node crashing after around 20 minutes of leaving soccer running, but when I tried to reproduce with LLDB on I was unable to. So I really don't know why it does that. I noticed near the end before it crashed that the Offense robot stopped moving back and forth, so my guess is it's something weird with ROS's action lib and [long-running actions](https://github.com/ros2/rcl/issues/926). (Once again, I hate ROS...) 

Or maybe it's like: the continuous PathTarget spam from Defense fills up the AS queue, which we/ROS don't correctly clear, and eventually it runs out of queue space and crashes? (Another reason to have all RobotIntents represent long-running tasks, and make planners composable.)

Anyways, on the bright side, a SSL RoboCup match is only 5 minutes per half, and we get 4 timeouts totaling 5 minutes of timeout time per match. So if our stack crashes after 5 minutes it will still last to halftime. If not, with 4 timeouts, it only really has to last 5/3 minutes without crashing :)

soccer/src/soccer/planning/planner_node.cpp
kfu02: sorry I thought I did this but clearly I did not

PR #2032, opened 2023-03-29
Waller Communication

N8BWert:
## Description
This PR sets up communication between wallers so that wallers will line up in order on the field.

[coordinate wallers](https://app.clickup.com/t/8677qektb)

### Test Case 1
1. make perf && make run-sim

**Expected result:**
Robots 1, 3, and/or 5 (depends on who is on the field) will form a wall to block the goal from the ball.

## Key Files to Review
Communication
 * JoinWallRequest.msg
 * LeaveWallRequest.msg
 * JoinWallResponse.msg
 * LeaveWallResponse.msg
 * soccer/src/soccer/strategy/agent/communication/communication.*pp

Defense + Waller
 * soccer/src/soccer/strategy/agent/position/defense.*pp
 * soccer/src/soccer/strategy/agent/position/waller.*pp

## Review Checklist


kfu02 (commented):
I defer this until #2037 is addressed, since a lot of my higher level concerns are the same. That said the wall looks good!

kfu02:
as discovered in real-world testing:
 * the sim separation gap is too small, on field 2.5 * robot radius seems to work well
 * the robots should not send wall comms if the other robots aren't "alive" (I think this concept exists in world state, I know for sure it exists in planning), meaning visible && connected via radio
 * robots should not run out of the way when the ball is coming towards them (I've made this mistake before too dw)
 * (nit) robots shouldn't face ball, rather they should be perp to dir of travel

PR #2035, opened 2023-03-29
penalty kicker

sid-parikh:
## Description
Creates two new special positions: PenaltyPlayer and GoalKicker.

PenaltyPlayer runs when we are awarded a penalty kick during the setup phase, and lines up behind the ball.

GoalKicker is a simple position that immediately and always attempts to kick the ball into a corner of the goal. It is assigned during the penalty kick phase, and should be automatically un-assigned once the ball is kicked, as the referee will move the state to Normal Start (or Free Kick, depending on the outcome of the shot).

The same positions are applied to Kickoff.

This PR also refactors coach node's position assignment to accommodate special states. 

Resolves [ClickUp card](https://app.clickup.com/t/8677c0jq4)

### Penalty Kicking
1. Run sim with external referee
2. In external referee, award our team a Penalty Kick
3. It is easier to see behavior if "Continue Automatically" is turned off in the external referee. (Under the "Match" tab in the new referee UI). In this case, you have to award the penalty kick, and then once the robots are lined up, you have to manually move to "normal start".

**Expected result:** Robot 1 lines up behind the ball and then kicks it.

## Key Files to Review
New Positions
 * `penalty_player`
 * `goal_kicker`
 * `agent_action_client` (takes new positions)

Position Assignment Refactor
 * `coach_node`

## Review Checklist


sid-parikh:
Hey, this PR still works, but the planner node appears to break more often after rebasing onto the latest ros2.

sid-parikh:
The last commit was unrelated, but a tiny fix that should have been done in my field dims PR a while ago. I can move to a different branch if you prefer

p-nayak11 (changes requested):
All functionality works! Add the same position assignments for free-kick as the kickoff situation, as well. Delete all of the commented-out code and fix the TODOs

p-nayak11 (approved):
Currently, the penalty and kickoff work. The free kick does not completely work since it does go through the setup state. Therefore, we cannot immediately call the GoalKicker position. Currently calling the offense position as a placeholder. Eventually, this position should just move to the ball and kick it.

PR #2037, opened 2023-03-29
Direct Passing (Trimmed)

N8BWert:
## Description
This PR describes the basic format for direct passing.  Robot 2 should pass to Robot 4.

[develop direct passing](https://app.clickup.com/t/8677c0q36)

### Test Case 1
1. make perf
2. make run-sim
3. play

**Expected result:** Robot 2 should pass the ball to Robot 4 

## Key Files to Review (other files can be ignored I merged in the planner rewrite before it was officially merged)
Group 1 (Position Files)
 * soccer/src/soccer/strategy/agent/position/position.*pp
 * soccer/src/soccer/strategy/agent/position/offense.*pp

Group 2 (ROS2 Messages)
 * soccer/src/soccer/strategy/communication/communication.*pp
 * rj_msgs/request/BallInTransitRequest.msg
 * rj_msgs/request/PassRequest.msg
 * rj_msgs/requestIncomingPassRequest.msg
 * rj_msgs/response/PassResponse.msg

## Review Checklist


rj_msgs/request/BallInTransitRequest.msg
```diff
@@ -0,0 +1,2 @@
+uint32 request_uid
```
kfu02, line 1: I think some of the ideas in #2020 are applicable here--creating a new ROS msg that holds roughly the same values as all the other communication ROS msgs presents friction to devs, which makes it less likely to be extended and improved in the future.

However, as a first pass this is fine.

soccer/src/soccer/strategy/agent/agent_action_client.cpp
```diff
@@ -56,6 +56,14 @@ AgentActionClient::AgentActionClient(int r_id)
             get_communication();
             check_communication_timeout();
         });
+
+    if (r_id == 0) {
```
kfu02: why did you need to add this?

kfu02: ah never mind I see the nullptr commented out below, can you just delete those instead

and add a comment like //default positions

soccer/src/soccer/strategy/agent/communication/communication.hpp
```diff
@@ -194,6 +258,12 @@ struct RosConverter<strategy::communication::AgentRequest, rj_msgs::msg::AgentRe
         } else if (const auto* pass_request =
                        std::get_if<strategy::communication::PassRequest>(&from)) {
             result.pass_request.emplace_back(convert_to_ros(*pass_request));
+        } else if (const auto* incoming_ass_request =
```
kfu02: :thinking: 
```suggestion
        } else if (const auto* incoming_pass_request =
```

soccer/src/soccer/strategy/agent/communication/communication.hpp
```diff
@@ -25,6 +26,8 @@ namespace strategy::communication {
 
 struct PassRequest {
     u_int32_t request_uid;
+    bool direct;
+    u_int8_t from_robot_id;
```
kfu02, line 38: there needs to be docstrings for each of these structs explaining what they do and why they exist

you can be more wordy than you would be in a production codebase. imagine that someone 3 years down the line will be trying to read your code and understand what it does, and they will refuse to read the docs.

soccer/src/soccer/strategy/agent/position/defense.cpp
```diff
@@ -5,69 +5,112 @@ namespace strategy {
...
 
-    // Create Waller
-    Waller waller{0};
-    return waller.get_task(intent, world_state);
+    rj_geometry::Point robot_position = world_state->get_robot(true, robot_id_).pose.position();
+    rj_geometry::Point ball_position = world_state->ball.position;
+    double distance_to_ball = robot_position.dist_to(ball_position);
+
+    switch (current_state_) {
+        case IDLING:
```
kfu02, line 22: why are the first two cases empty?

soccer/src/soccer/strategy/agent/position/defense.cpp
```diff
@@ -5,69 +5,112 @@ namespace strategy {
...
+        planning::LinearMotionInstant target{target_robot_pos};
+        auto line_kick_cmd = planning::MotionCommand{"line_kick", target};
+        intent.motion_command = line_kick_cmd;
+        intent.shoot_mode = RobotIntent::ShootMode::KICK;
+        // NOTE: Check we can actually use break beams
+        intent.trigger_mode = RobotIntent::TriggerMode::ON_BREAK_BEAM;
+        // TODO: Adjust the kick speed based on distance
+        intent.kick_speed = 4.0;
+        intent.is_active = true;
+        return intent;
```
kfu02, line 91: I don't think this belongs in this PR, but in general I think we should hide the low-level MC from this level of position. Ideally the positions would call a high-level PathPlanner that handles a complex task (e.g. "Receive" as a planner) but as you already know, our current planner stack makes composing planners challenging. 

So the workaround right now is "RoleInterface" (see #1996 ), and ideally this passing and receiving behavior you have here belong in a RoleInterface implementation. Then it would be simply:
```
if (state == PASSING)
  Passer passer{};
  return passer.get_task(...);
else if (...)
  ...
```

kfu02: something to think about for next PR maybe

soccer/src/soccer/strategy/agent/position/defense.cpp
```diff
@@ -5,69 +5,112 @@ namespace strategy {
...
+    // Call to super
+    Position::pass_ball(robot_id);
+    // Update current state
+    current_state_ = PASSING;
+}
 
-    return comm_response;
+communication::Acknowledge Defense::acknowledge_ball_in_transit(
+    communication::BallInTransitRequest ball_in_transit_request) {
+    // Call to super
```
kfu02: this is kind of a weak comment, I would prefer no comment over this, or an explanation of what to expect from acknowledge_response

soccer/src/soccer/strategy/agent/position/goalie.cpp
```diff
@@ -35,8 +35,20 @@ Goalie::State Goalie::update_state() {
...
+        }
+    } else if (latest_state_ == RECEIVING) {
+        if (distance_to_ball < 0.1) {
+            return IDLING;
+        }
+    }
+
     // otherwise, default to idling
-    return IDLING;
+    return latest_state_;
```
kfu02, line 48: this return no longer matches the comment above, update comment

soccer/src/soccer/strategy/agent/position/goalie.cpp
```diff
@@ -35,8 +35,20 @@ Goalie::State Goalie::update_state() {
...
+    double distance_to_ball = robot_position.dist_to(ball_pt);
+    if (latest_state_ == PASSING) {
+        if (!ball_in_box) {
+            return IDLING;
+        }
+    } else if (latest_state_ == RECEIVING) {
+        if (distance_to_ball < 0.1) {
+            return IDLING;
+        }
+    }
```
kfu02, line 46: does this goalie passing actually work?

soccer/src/soccer/strategy/agent/position/goalie.cpp
```diff
@@ -77,6 +89,28 @@ std::optional<RobotIntent> Goalie::state_to_task(RobotIntent intent) {
...
+        return intent;
+    } else if (latest_state_ == PASSING) {
+        // attempt to pass the ball to the target robot
+        rj_geometry::Point target_robot_pos =
+            world_state()->get_robot(true, target_robot_id).pose.position();
+        planning::LinearMotionInstant target{target_robot_pos};
+        auto pass_kick_cmd = planning::MotionCommand{"line_kick", target};
+        intent.motion_command = pass_kick_cmd;
+        intent.shoot_mode = RobotIntent::ShootMode::KICK;
+        // NOTE: Check we can actually use break beams
```
kfu02: we can use BB now

soccer/src/soccer/strategy/agent/position/goalie.cpp
```diff
@@ -77,6 +89,28 @@ std::optional<RobotIntent> Goalie::state_to_task(RobotIntent intent) {
...
+        intent.trigger_mode = RobotIntent::TriggerMode::ON_BREAK_BEAM;
+        // TODO: Adjust the kick speed based on distance
+        intent.kick_speed = 4.0;
+        intent.is_active = true;
+        return intent;
```
kfu02, line 111: related to my comments above, if there was a RoleInterface to encapsulate a Passer role, there would be less duplicate code here :smile:

soccer/src/soccer/strategy/agent/position/position.hpp
```diff
@@ -114,6 +117,56 @@ class Position {
...
+     *
+     * @param incoming_pass_request the request that a ball will be coming to this robot
+     * @return communication::Acknowledge acknowledgement that the other robot may pass to this
+     * robot
+     */
+    virtual communication::Acknowledge acknowledge_pass(
+        communication::IncomingPassRequest incoming_pass_request);
+
+    /**
+     * @brief physically update the state of the robot to be passing the ball
```
kfu02: physically?

soccer/src/soccer/strategy/agent/position/position.cpp
```diff
@@ -71,22 +71,167 @@ bool Position::assert_world_state_valid() {
...
-void Position::receive_communication_response([
-    [maybe_unused]] communication::AgentPosResponseWrapper response) {}
+communication::PosAgentResponseWrapper Position::receive_communication_request(
+    communication::AgentPosRequestWrapper request) {
+    communication::PosAgentResponseWrapper comm_response{};
+    if (const communication::PassRequest* pass_request =
+            std::get_if<communication::PassRequest>(&request.request)) {
+        communication::PassResponse pass_response = receive_pass_request(*pass_request);
+        comm_response.response = pass_response;
+        // TODO: "IncomingPassRequest" => "IncomingBallRequest" (or smth)
```
kfu02: can you do this in this PR?

soccer/src/soccer/strategy/agent/position/position.cpp
```diff
@@ -71,22 +71,167 @@ bool Position::assert_world_state_valid() {
...
+        communication::Acknowledge ball_in_transit_acknowledge =
+            acknowledge_ball_in_transit(*ball_in_transit_request);
+        comm_response.response = ball_in_transit_acknowledge;
+    } else {
+        communication::Acknowledge acknowledge{};
+        communication::generate_uid(acknowledge);
+        comm_response.response = acknowledge;
+    }
+
+    // TEST CODE: UNCOMMENT TO TEST
```
kfu02: delete

soccer/src/soccer/strategy/agent/position/position.cpp
```diff
@@ -71,22 +71,167 @@ bool Position::assert_world_state_valid() {
...
+void Position::send_direct_pass_request(std::vector<u_int8_t> target_robots) {
+    communication::PassRequest pass_request{};
+    communication::generate_uid(pass_request);
+    pass_request.direct = true;
+    pass_request.from_robot_id = robot_id_;
+
+    communication::PosAgentRequestWrapper communication_request{};
+    communication_request.request = pass_request;
+    communication_request.target_agents = target_robots;
+    // For testing
```
kfu02: is it still for testing?

soccer/src/soccer/strategy/agent/position/position.hpp
```diff
@@ -162,8 +215,22 @@ class Position {
...
 
+    // the robot our robot is going to be passing to
+    int target_robot_id;
+
+    // the robot our robot will be facing (useful for passing)
+    int face_robot_id;
+
+    // the maximum distance from the robot to the ball for the robot to begin
+    // chasing the ball
+    const double max_receive_distance = 1.0;
```
kfu02:
```suggestion
    const double max_receive_distance = 1.0; // m
```

soccer/src/soccer/strategy/coach/coach_node.cpp
```diff
@@ -150,48 +150,48 @@ void CoachNode::check_for_play_state_change() {
...
-        for (int i = 0; i < kNumShells; i++) {
-            if (i != goalie_id_ && current_overrides_[i] == 1 || current_overrides_[i] == 2) {
-                positions[i] = current_overrides_[i];
-            }
-        }
-    }
-
-    positions_message.client_positions = positions;
-    positions_pub_->publish(positions_message);
+    // rj_msgs::msg::PositionAssignment positions_message;
```
kfu02: why is this all commented out?

soccer/src/soccer/strategy/agent/position/position.hpp
```diff
@@ -114,6 +117,56 @@ class Position {
...
+     */
+    virtual communication::Acknowledge acknowledge_pass(
+        communication::IncomingPassRequest incoming_pass_request);
+
+    /**
+     * @brief physically update the state of the robot to be passing the ball
+     *
+     * @param robot_id the robot id of the robot to pass the ball to
+     */
+    virtual void pass_ball(int robot_id);
```
kfu02, line 166: after reading the usage of this, I understand what you're going for here. however, currently you have it where the position implementations have to call the superclass implementation IN their own implementation of this virtual method, which is a little weird. you see this weirdness in being forced to add a Position::pass_ball() to force the superclass version.

instead, I would use the tmp-nvi template (see Position's get_task for more info). Position should have a concrete method (pass_ball()) that does all the shared messaging all derived classes need, then call a pure virtual function to force the positions to implement their own state update. something like:
```
void pass_ball(...) {
  // make pass request
  // make comm request
  derived_pass_ball(...);
}

virtual void pass_ball(...) = 0;
```
then in Defense/Off/Goalie:
```
void pass_ball() override ...
```

soccer/src/soccer/strategy/agent/position/position.cpp
```diff
@@ -71,22 +71,167 @@ bool Position::assert_world_state_valid() {
...
+std::optional<communication::PosAgentRequestWrapper> Position::send_communication_request() {
+    if (communication_request_ != std::nullopt) {
+        auto saved_comm_req = communication_request_;
+        communication_request_ = std::nullopt;
+        return saved_comm_req;
+    }
+    return std::nullopt;
+}
+
+void Position::receive_communication_response(communication::AgentPosResponseWrapper response) {
```
kfu02, line 89: also, upon doing some deep diving, I can't figure out why you chose to bundle fields in a struct (AgentPosResponseWrapper) instead of making them params to this method. why did you do that?

kfu02 (changes requested):
First off: congrats on re-implementing passing. I know firsthand that that's not easy to do.

I have some minor concerns that can be pushed to later PRs, but my two major concerns are:
(1) the way pass_ball() is implemented, see my comment about that
(2) the existence of the AgentPosRequestWrapper struct. I feel that this has potential to become a [Data Class](https://refactoring.guru/smells/data-class) like PlanRequest, and passing the fields of this struct as params wouldn't be bad (since there are only 4 fields total)

Let me know what you think! I did this review during core elections so sorry if I missed something obvious

soccer/src/soccer/strategy/agent/position/defense.cpp
N8BWert: There was no default behavior for defense before I made the state machine, so I left the idling and searching states empty for future PRs to fill in.

soccer/src/soccer/strategy/agent/position/goalie.cpp
N8BWert: Yes, if another agent/robot wants to pass to the goalie (and vice versa) the logic exists and is the same as for passing and receiving for offense and defense.

soccer/src/soccer/strategy/coach/coach_node.cpp
N8BWert: The coach node was overriding the default position assignments so I commented this out for testing.  Now that everything works I'm going to un-comment this, but then the visual test case won't work.  I'll also get rid of offense defaulting to passing when the pr is approved.

soccer/src/soccer/strategy/agent/position/position.cpp
N8BWert: The main reason there are the wrappers instead of method parameters is for the relaying of requests from the position to the agent and the relaying of responses from the agent to the position.

For the position to agent request case, although there are only 4 parameters (the request, the target agents, broadcast, and urgent), 3 of the 4 fields are not required in every case (i.e. target agents, broadcast, and urgent).  When I was writing everything I also forgot that C++ has the ability to give default parameters to variables (Rust does not) so I figured it would be easier to wrap all of the fields together instead of writing a different method for each case. 

For the agent the position response case, however, there is a lot of information to relay (some of which may be not always be used) so it felt like it would be annoying for a user to look at a function with 7 arguments of which only 1 is always used.  This wrapper is also used by the agent action client to handle the aggregation of responses to a request before relaying them to the current position and it made sense to do this in a singular aggregated data type.

kfu02: okay, fine by me then

soccer/src/soccer/strategy/coach/coach_node.cpp
kfu02: can you actually check that your code works properly with the new position assignment UI tab? based on the on-field testing today (4/2) there may be some issues, though that also might be due to a variety of other factors (external ref, bad day, etc.)

kfu02:
oh also, I noticed the passer sometimes tries to pass the ball after the receiver has already received the ball. can you either (1) ensure only one pass happens or (2) cleanly switch which one is passer and receiver after the pass so they pass indefinitely, similar to old Keepaway?

soccer/src/soccer/strategy/agent/position/goalie.cpp
```diff
@@ -35,8 +35,19 @@ Goalie::State Goalie::update_state() {
...
-    // otherwise, default to idling
-    return IDLING;
+    rj_geometry::Point robot_position = world_state->get_robot(true, robot_id_).pose.position();
+    double distance_to_ball = robot_position.dist_to(ball_pt);
+    if (latest_state_ == PASSING) {
+        if (!ball_in_box) {
+            return IDLING;
+        }
+    } else if (latest_state_ == RECEIVING) {
+        if (distance_to_ball < 0.1) {
```
kfu02, line 43: this should use the `check_is_done()` method from Position superclass, which links the notion of done-ness directly to the planner (and prevents strategy from having to know how planners are implemented). I think that's why you're seeing poor behavior from the receiving robot.

see here: https://github.com/RoboJackets/robocup-software/blob/ros2/soccer/src/soccer/strategy/agent/position/offense.cpp#L9

soccer/src/soccer/strategy/agent/position/offense.cpp
```diff
@@ -2,93 +2,161 @@
...
+        return intent;
+    } else if (current_state_ == SHOOTING) {
+        // TODO: Shoot the ball at the goal
+    } else if (current_state_ == RECEIVING) {
+        // check how far we are from the ball
+        rj_geometry::Point robot_position =
+            world_state()->get_robot(true, robot_id_).pose.position();
+        rj_geometry::Point ball_position = world_state()->ball.position;
+        double distance_to_ball = robot_position.dist_to(ball_position);
+        if (distance_to_ball > max_receive_distance && !chasing_ball) {
```
kfu02, line 119: what does this if check for? (add a comment or set a named boolean to clarify)
e.g. `bool too_far_to_intercept = ...`

kfu02 (changes requested):
EDIT: never mind, the pass looks much better now! I think I found the root of your receive issue, though
EDIT 2: never mind, my suggested issue does not fully solve the receiver not doing what is expected. that said, I do stand by the comment that strategy should not impose too much on the planner is_done

soccer/src/soccer/strategy/agent/position/goalie.cpp
kfu02: not sure if this is how you implemented it in the other Positions, but double-check

kfu02 (approved):
looks good for first passing, there are some things we could clean up but I think that can come later. imo future directions include:
1. making it easy for all positions to pass, not just offense
2. fixing up our receiver so it doesn't give up on the ball so much
3. (long term) passing to points off of robots

but for now, LGTM! nice job

PR #2061, opened 2023-04-12
Update global overrides

p-nayak11:
## Description
Make the dibbler respect the states of the external referee. 

[ClickUp card](https://app.clickup.com/t/8677w00dc)

### Test Case 1
1. Run the external sim 
2. Echo the strategy/coach_state topic in another terminal tab
3. Check the dribbler speed change with the different stages
4. Run it on the actual field

**Expected result:** The dribbler should not run during stop and halt. It should only run while it is in the playing state. 

## Key Files to Review
* coach_node.cpp
* manipulator_control.cpp


kfu02 (approved):
comp tested

PR #2063, opened 2023-04-12
Shoot only

N8BWert:
## Description
This PR outlines the basics of a shoot only offense.  In this offense strategy, one robot (the closest to the ball that is playing offense) will attempt to obtain the ball and shoot the ball at the opponents goal.

EDIT: PR #2069 was merged into this branch. It adds a pivot step before any shot on goal to improve line kick consistency, but only does so at the Position level, which is not ideal (should be a planner, but planners are hard to compose).

[coordinate shoot-only offense](https://app.clickup.com/t/8677u14px)

### Test Case 1
1. make perf && make run-sim
2. assign any number of robots to offense
3. click start (the robots will reassign is switched to offense and closer to the ball while sim is playing if you want to see that just start then assign robots to offense)

**Expected result:**The robot closest to the ball should get the ball and shoot it into the net.  Currently the robot keeps doing this on loop which may be undesired, but I would assume the coach will reassign him so I have the scorer on loop.

## Key Files to Review
Messages
 * rj_msgs/request/ScorerRequest.msg
 * rj_msgs/response/ScorerResponse.msg
 * soccer/src/soccer/strategy/agent/communication.*pp

Offense
 * soccer/src/soccer/strategy/agent/position/offense.*pp

## Review Checklist

### DIFF
--- a/soccer/src/soccer/strategy/agent/position/offense.cpp
+++ b/soccer/src/soccer/strategy/agent/position/offense.cpp
@@ -50,7 +50,8 @@ Offense::State Offense::update_state() {
         case SHOOTING:
             // transition to idling if we no longer have the ball (i.e. it was passed or it was
             // stolen)
-            if (check_is_done()) {
+            if (check_is_done() || distance_to_ball > ball_lost_distance_) {
+                send_reset_scorer_request();
                 next_state = SEARCHING;
             }
             break;
@@ -179,6 +180,10 @@ communication::PosAgentResponseWrapper Offense::receive_communication_request(
             std::get_if<communication::ScorerRequest>(&request.request)) {
         communication::ScorerResponse scorer_response = receive_scorer_request(*scorer_request);
         comm_response.response = scorer_response;
+    } else if (const communication::ResetScorerRequest* reset_scorer_request =
+                   std::get_if<communication::ResetScorerRequest>(&request.request)) {
+        communication::Acknowledge response = receive_reset_scorer_request(*reset_scorer_request);
+        comm_response.response = response;
     }
 
     return comm_response;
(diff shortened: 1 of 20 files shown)


soccer/src/soccer/strategy/agent/position/offense.cpp
```diff
@@ -65,16 +61,16 @@ Offense::State Offense::update_state() {
             }
             break;
         case STEALING:
-            // transition to idling if we are close enough to the ball
-            if (check_is_done()) {
-                /* SPDLOG_INFO("\033[92m ball pos {}{} \033[0m", ball_position.x(),
-                 * ball_position.y()); */
-
+            // The collect planner check_is_done() is wonky so I added a second clause to check
```
kfu02: is there a way to fix that planner's is_done? (e.g. just add the second clause into that is_done directly)

soccer/src/soccer/strategy/agent/position/offense.cpp
```diff
@@ -150,6 +158,90 @@ std::optional<RobotIntent> Offense::state_to_task(RobotIntent intent) {
...
+    // Switch scorers if better scorer
+    if (scorer && scorer_request.ball_distance < ball_distance) {
+        scorer = false;
+        current_state_ = FACING;
+    }
+
+    return scorer_response;
+}
+
+void Offense::handle_scorer_response(std::vector<communication::AgentResponseVariant> responses) {
```
kfu02: check warning

soccer/src/soccer/strategy/agent/position/offense.hpp
```diff
@@ -51,6 +56,32 @@ class Offense : public Position {
 
     // current state of the offensive agent (state machine)
     State current_state_ = IDLING;
+
+    bool scorer = false;
```
kfu02:
```suggestion
    bool scorer_ = false;
```
(will require changes everywhere else in .cpp)

kfu02 (changes requested):
This is looking good so far, but is there a way to reset which robot is the scorer after a shot is taken? Right now the looping behavior will probably net us double touch fouls. I imagine it would be something like once the scorer has attempted a kick, send a broadcast message that resets the value of scorer for every Offense player.

Sdssourcing:
#

PR #2064, opened 2023-04-12
New Seeking Behavior

rishiso:
## Description
Adding seek behavior to offense.

Resolves [ClickUp](https://app.clickup.com/t/389zmk)

## Design Documents
[Link](link-to-design-doc)

## Steps to Test
make run-sim (forced offense to always search)

**Expected result:** Travels to open position

## Key Files to Review
 * offense.cpp
 * offense.hpp
 * seeking.cpp
 * seeking.hpp

## Review Checklist

- [ ] Step 1
- [ ] Step 2

### DIFF
--- a/soccer/src/soccer/strategy/agent/position/seeker.cpp
+++ b/soccer/src/soccer/strategy/agent/position/seeker.cpp
@@ -4,7 +4,7 @@ namespace strategy {
 
 Seeker::Seeker(int seeker_num) {
     offense_type_ = "Seeker";
-    seeker_pos_ = seeker_num;
+    seeker_pos_ = seeker_num;  // Equal to robot_id
 }
 
 std::optional<RobotIntent> Seeker::get_task(RobotIntent intent, const WorldState* world_state,
@@ -82,13 +82,14 @@ rj_geometry::Point Seeker::calculate_open_point(double current_prec, double min_
  * Generates random noise for potential point.
  */
 rj_geometry::Point Seeker::random_noise(double prec) {
-    double x = (double)rand() / RAND_MAX * prec * .33;
-    double y = (double)rand() / RAND_MAX * prec * .33;
+    double x = (double)rand() / RAND_MAX * prec;
+    double y = (double)rand() / RAND_MAX * prec;
     return rj_geometry::Point{x, y};
 }
 
 /**
- * Corrects point to prevent it from being out of bounds or in the box.
+ * Corrects point to prevent it from being out of bounds or in the box. Also, prevents too many
+ * robots from being in one area.
  */
 rj_geometry::Point Seeker::correct_point(rj_geometry::Point p, FieldDimensions field_dimensions) {
     double BORDER_BUFFER = .2;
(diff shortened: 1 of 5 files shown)


soccer/src/soccer/strategy/agent/position/seeker.cpp
```diff
@@ -0,0 +1,152 @@
...
+                                                rj_geometry::Point current_point,
+                                                const WorldState* world_state,
+                                                FieldDimensions field_dimensions) {
+    if (current_prec < min_prec) {
+        return current_point;
+    }
+    rj_geometry::Point ball_pos = world_state->ball.position;
+    rj_geometry::Point min = current_point;
+    double min_val = max_los(ball_pos, current_point, world_state);
+    double curr_val;
```
kfu02:
```suggestion
    double curr_val{};
```
is safer to initialize variables, as C++ has unexpected behavior if this is ever uninitialized

kfu02 (changes requested):
image


This is what the seekers converged to after I let them stay put for a second, in general we don't want a clump of 3 robots in the same quadrant ever. I like the effort, but can you try going with a simpler model for now? Just something that spaces the Offense robots out roughly evenly around the opponent goal will do. For this upcoming comp we didn't have enough time to test and tune passing, so don't worry too much about all this movement to create open space.

If you want, you can leave this PR as a draft, but make a new one with the above recommendation implemented

rishiso:
Screenshot 2023-04-16 at 12 07 37 PM

I made some changes to reduce clumping. One of these changes is assigning individual bots to stay in a certain third based on position number. I also added more random noise and changed the model to incorporate max distance from friendly bots. The image above shows where the robots converge.

kfu02 (changes requested):
Looks better, nice! A few more things:

1. I had to modify your code to see the seeking behavior actually take place. Is this the case on your local branch? (If so, commit that change)
image

2. Docstrings should be placed in the .hpp files, this is just a convention with C++. It makes it easier for someone to simply read the header and understand what is happening. You have a few nice docstrings that are in the cpp files.

3. See the check warnings, in general we would prefer non-recursive calls if possible, so make that iterative instead (with some limit like 10 iterations, it won't improve much beyond that anyways). There's a couple minor ones about const/static, just listen to the warnings and look it up if you have time to learn.

Finally, I think the seek behavior is nice in sim, but we should refrain from confirming that it looks good without testing on real robots. So let's do that tonight.

rishiso:
I have made the changes to the docstrings, moving them from the cpp to the header file. Also, I changed the recursive function to iterative. I am not seeing any more warnings for the seeker files, so I think that they have been resolved.

As for making the seeking behavior take place, I am forcing the current_state to always be searching similar to your code. I am not exactly sure how the offensive decision tree works, but it doesn't seem to be assigning the role of searching very often.

I won't be able to make it to tonight's meeting, but please let me know if the seeking works properly in real life.

sid-parikh:
@rishiso does this PR work still? can we clean it up and merge it?

rishiso:
@sid-parikh It is somewhat working. I was recreating the seeking behavior from the Python codebase in C++. However, there are a few issues since I had to recreate many Python functions from scratch like scipy.optimize. I can look and see if I can get it working fully.

sid-parikh:
@sanatd33 @shourikb You two have taken a card that involved finishing this PR. @rishiso if you think you can help them, or finish it quickly, you can do that as well. Let's aim to have this done in a couple meetings.

sid-parikh:
I have marked this as a draft for until it is done.

PR #2069, opened 2023-04-16
Improve kick consistency

kfu02:
## Description
Improved shooting consistency by adding pivot to the Offense FSM.

Ideally this would be a part of the kick planner itself, but our planner stack sucks. It is not easy to compose planners (I tried in #2068 ) and even when you are able to, often one of the composed planners will fail to plan. This is mitigated at the Position level thanks to the safeguards on PlannerNode put in place by #2020 .

### Test Case 1
1. Run sim
2. Set as many robots as you'd like to Offense
3. Run

**Expected result:** the closest Offense shoots, per #2063 . this kick is smoother than the kicks in #2063 

## Review Checklist

### DIFF
--- a/soccer/src/soccer/planning/planner/plan_request.cpp
+++ b/soccer/src/soccer/planning/planner/plan_request.cpp
@@ -42,20 +42,27 @@ void fill_obstacles(const PlanRequest& in, rj_geometry::ShapeSet* out_static,
     // Add our robots, either static or dynamic depending on whether they have
     // already been planned. In both cases, radius is based on velocity like
     // above for opp robots.
+    // TODO: reenable dynamic obstacles for our robots (currently
+    // TrajectoryCollection is never filled at planner level)
     for (size_t shell = 0; shell < kNumShells; shell++) {
-        const auto& robot = in.world_state->our_robots.at(shell);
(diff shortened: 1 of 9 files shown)


soccer/src/soccer/planning/planner/plan_request.cpp
```diff
@@ -42,20 +42,27 @@ void fill_obstacles(const PlanRequest& in, rj_geometry::ShapeSet* out_static,
...
+        /*     // Dynamic obstacle */
+        /*     out_dynamic->emplace_back(obs_radius, ptr_to_traj); */
+        /* } else { */
+        /*     // Static obstacle */
+        /*     out_static->add(std::make_shared<rj_geometry::Circle>(obs_center, obs_radius)); */
+        /* } */
+
+        // Static obstacle
+        fill_robot_obstacle(our_robot, obs_center, obs_radius);
+        out_static->add(std::make_shared<rj_geometry::Circle>(obs_center, obs_radius));
```
kfu02, line 65: before this, our robots weren't respecting each other as obstacles AT ALL
(since we disabled trajectory sharing)

now they will leave the same buffer with each other as they do with opponent robots, which probably breaks walling

need to test

N8BWert (approved):
lgtm

PR #2083, opened 2023-07-12
Competition Changes [should we merge or just cherry-pick important parts as needed?]

sid-parikh:


kfu02:
probably would be better to cherry pick / separate into multiple PRs

sid-parikh:
Yeah, let's not do this. We'll do a thorough audit and make a trimmed down PR at some point

PR #1670, opened 2021-06-17
Rework CMake configuration

kylestach:
## Description
Rework CMake configuration and directory structure. This should speed up incremental builds, as we no longer need to relink the entirety of soccer (and what was previously the robocup.so library) for any change to any of the node.

While we're at it, might as well move soccer/src/soccer -> soccer so we don't have to deal with an extra layer of nested directories for no reason.

## Associated Issue
N/A, I was just annoyed

## Design Documents
N/A

### Do it build?
1. Do it build?

Expected result: It do build.

PR #2085, opened 2023-07-30
Rework CMake Configuration

sid-parikh:
Built from PR #1670, with some changes mostly to account for new files/structures in our codebase since that PR

Current state: building! ... but sim doesn't run and testing was deleted


sid-parikh:
don't merge, doesn't build!

sid-parikh:
It builds! but sim crashes repeatedly...

sid-parikh:
oh, i also completely deleted googletest out of linker frustration. should probably bring that back at some point (before merging this)

PR #2092, opened 2023-09-15
change to 4 wallers

petergarud:
Completes RC SW tutorials


petergarud:
double_spacing

petergarud:
For the rqt question, there is only one publisher for /planning/trajector/robo_0 and that is /planner

petergarud:
Screenshot 2023-09-18 at 4 21 48 PM
Screenshot 2023-09-18 at 4 21 38 PM

petergarud:
I understand the style checker isn't going through but if you look at the details it seems to have run through all the correct style but failed because an output function is being deprecated. The tutorial was overall pretty good but I still don't feel like I have a great grasp on ros or on the stack overall.

.clangd/index/Control.pb.h.8AF475F657FC3F97.idx
sid-parikh, line 1: hi Peter, not sure what happened (is this a weird LSP thing? or WSL?) but please uncommit these files and be careful in the future. you can try one of these solutions to tell git to ignore this directory without modifying our .gitignore  https://stackoverflow.com/questions/653454/how-do-you-make-git-ignore-files-without-using-gitignore

compile_commands.json
sid-parikh, line 1: same thing here. I know you need this file for your environment, but please don't commit it into our repository

soccer/src/soccer/soccer_mom.cpp
```diff
@@ -0,0 +1,28 @@
...
+#include <spdlog/spdlog.h>
+
+#include "std_msgs/msg/string.hpp"
+
+using namespace std::chrono_literals;
+
+namespace soccer_mom {
+SoccerMom::SoccerMom() : Node{"soccer_mom"} {
+    team_color_pub_ = create_publisher<std_msgs::msg::String>("/team_fruit", 10);
+    tick_timer_ = this->create_wall_timer(500ms, std::bind(&SoccerMom::timer_callback, this));
```
sid-parikh, line 14: this wall timer implementation is cool! it's good that you learned about timers as they are a useful ros feature. 

FYI though, the intent was for you to publish the message within the subscription callback itself. You don't need to redo it, but take a second to understand what the difference might be and why one might be preferred over the other.

sid-parikh: There's also a warning from GH actions here about preferring a lambda to std::bind. Not a huge difference, but would improve readability. You used a lambda for the subscription, so wanted to encourage you to read up on them and understand what you did

soccer/src/soccer/soccer_mom.cpp
```diff
@@ -0,0 +1,28 @@
...
+        referee::topics::kTeamColorTopic, rclcpp::QoS(1).transient_local(),
+        [this](rj_msgs::msg::TeamColor::SharedPtr color) {  // NOLINT
+            blue_team_ = color->is_blue;
+        });
+}
+
+void SoccerMom::timer_callback() {
+    auto message = std_msgs::msg::String();
+    message.data = blue_team_ ? "blueberries" : "banana";
+    RCLCPP_INFO(this->get_logger(), "Publishing: '%s'", message.data.c_str());
```
sid-parikh, line 25: you should be able to use `SPDLOG_INFO("Publishing: {}", message.data)` instead, for a cleaner approach and to use our normal logger. your code might have the same result but i'm not sure

soccer/src/soccer/strategy/agent/position/runner.cpp
```diff
@@ -0,0 +1,81 @@
+#include "runner.hpp"
+#include "planning/instant.hpp"
+
+namespace strategy {
+
+Runner::Runner(int r_id) : Position(r_id) {
```
sid-parikh, line 6: This class is really well done!

sid-parikh (approved):
Great work finishing this up! Take a look at my comments for some learning. I'll look for some work for you to do sooN!

PR #2101, opened 2023-09-27
Remove coach node

Ninjatest321:
This pull request adds a Referee PlayState subscription to all the files that use CoachState and don't have play state already (which is just Agent Action Client). This is part of the plan to remove CoachNode.


soccer/src/soccer/strategy/agent/agent_action_client.cpp
```diff
@@ -95,6 +99,10 @@ void AgentActionClient::world_state_callback(const rj_msgs::msg::WorldState::Sha
     last_world_state_ = std::move(world_state);
 }
 
+void AgentActionClient::play_state_callback(const rj_msgs::msg::PlayState::SharedPtr& msg) {
+    return;
```
sid-parikh: Hey amogh, can you also change the AC & Position to get play state data from this topic? the callback is empty right now; we should be forwarding it to position. I know position doesn't use play state data right now, but that's okay

sid-parikh (changes requested):

sid-parikh (approved):
Looks good, thanks!

Ninjatest321:
This last commit calculates global overrides in the Planner Node (planner_node.cpp)

p-nayak11:
Look at PR #2131

PR #2105, opened 2023-10-02
Aalind tyagi/robocup sw tutorial

Squid5678:
Did SoccerMom


Squid5678:
Screenshot 2023-10-01 at 8 17 13 PM
Screenshot 2023-10-01 at 8 17 25 PM

llvm.sh.1
sid-parikh, line 1: Hey, please don't commit these extra files to git.

llvm.sh.2
sid-parikh, line 1: same comment here

soccer/src/soccer/radio/.vscode/c_cpp_properties.json
sid-parikh, line 1: and here. you can look up various ways to tell git to ignore files and folders. if you think an ignore is global enough that everyone needs it, we can update our .gitignore file.

sid-parikh (approved):
This is wonderful work Aalind!

PR #2132, opened 2023-10-30
Making ball a static obstacle

rishiso:
## Description
The STOP play state is currently not working because the robots are still able to go within the zone of the ball. We resolved this issue by making the ball a static obstacle instead of a dynamic obstacle.

Resolves [ClickUp card](https://app.clickup.com/t/86ayj3tzd)

## Steps to Test
Run the simulator in the STOP state. The robots should not be in the zone of the ball.

## Review Checklist

### DIFF
--- a/soccer/src/soccer/planning/planner/plan_request.cpp
+++ b/soccer/src/soccer/planning/planner/plan_request.cpp
@@ -68,7 +68,8 @@ void fill_obstacles(const PlanRequest& in, rj_geometry::ShapeSet* out_static,
     // Finally, add the ball as a dynamic obstacle.
     // (This is for when the other team is trying to do ball placement, so we
     // don't interfere with them.)
-    out_static->add(std::make_shared<rj_geometry::Circle>(in.world_state->ball.position, kBallRadius + in.min_dist_from_ball));
+    out_static->add(std::make_shared<rj_geometry::Circle>(in.world_state->ball.position,
+                                                          kBallRadius + in.min_dist_from_ball));
     if (avoid_ball && out_dynamic != nullptr && out_ball_trajectory != nullptr) {
         // Where should we store the ball trajectory?
         *out_ball_trajectory = in.world_state->ball.make_trajectory();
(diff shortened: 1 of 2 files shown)


sid-parikh (changes requested):
Can you update this to inflate the ball radius based on ball velocity the way we do for `their_robots`?

rishiso:
Don't merge this yet. Just realized a potential issue

soccer/src/soccer/planning/planner/plan_request.cpp
```diff
@@ -65,14 +65,21 @@ void fill_obstacles(const PlanRequest& in, rj_geometry::ShapeSet* out_static,
         out_static->add(std::make_shared<rj_geometry::Circle>(obs_center, obs_radius));
     }
 
+    // Adding ball as a static obstacle (because dynamic obstacles are not working)
+    // Only added when STOP state is enabled
+    double ball_vel_scaled = in.world_state->ball.velocity.mag() * .1;
+    double radius = kBallRadius + in.min_dist_from_ball * (1.0 + ball_vel_scaled);
```
sid-parikh: can you leave a comment explaining this math? are you sure it's equivalent to the math used in fill_robot_obstacle? 

I know we add a min_dist, but we should start with that min_dist and then inflate+offset using the same math as fill_robot_obstacle (you can honestly just call that function), this ensures we stay min_dist away from a moving target by adding extra buffer proportional to velocity.

sid-parikh (changes requested):

soccer/src/soccer/planning/planner/plan_request.cpp
rishiso: The math I was using previously was an equivalent version. However, I changed it to use the same math as the other function.

sid-parikh (approved):

sid-parikh:
I approved PR to waive my "requested changes". However, since I authored a major portion of it, please do not merge until somebody else has reviewed.

p-nayak11 (approved):
Fix the GitHub Actions, but other than that LGTM

PR #2139, opened 2023-11-06
Split up classes in planner_node into 3 files

petergarud:
## Description
Splits up GlobalState, PlannerForRobot, PlannerNode into 3 files

Resolves https://app.clickup.com/t/863h9cthk

### Test Case 1
1. Step 1
2. Step 2
3. Step 3

**Expected result:**???

## Key Files to Review
Group 1
 * File 1
 * File 2

Group 2
 * File 3
 * File 4

## Review Checklist

- [ ] Step 1
- [ ] Step 2


soccer/src/soccer/planning/global_state.hpp
```diff
@@ -0,0 +1,86 @@
+#pragma once
+
+#include <rj_constants/topic_names.hpp>
+#include <rj_msgs/msg/coach_state.hpp>
+#include <rj_msgs/msg/goalie.hpp>
+
+#include "planning/planner/escape_obstacles_path_planner.hpp"
```
sid-parikh: can you remove this line? (as long as code still compiles) it doesn't seem necessary

sid-parikh (changes requested):
it looks good, i left one small comment and i'll test it before approving just in case

soccer/src/soccer/planning/global_state.hpp
petergarud: If I remove any of those includes, the code no longer compiles so I think all the includes are necessary. If you're talking about the blank line, that was added by the style-checker

sid-parikh: escape obstacles planner definitely isn't needed. compiling probably fails because it's using some include(s) within escape obstacles, but try to track down which ones we actually need. if not, i can try to run a tool I have detects detects necessary includes, in the shop on tuesday, remjnd me.

the reason i care so much is because in the next phase of this EscapeObsatclesPlanner will need to include this file, and we'll have a circular problem.

petergarud: Ok, I'll change the imports to only include necessary files inside escape_obstacles_path_planner

petergarud: I think I've removed all the unnecessary includes now so everything should be good.

sid-parikh (approved):

sid-parikh:
Congrats! Go ahead and hit the merge button

PR #2148, opened 2023-11-14
Global state return values

sid-parikh:
A continuation of the planner rewriting, let's not return unprotected pointers to mutex-locked resources.


p-nayak11 (changes requested):
Fix all the Github Actions. Makes sense to change how world_state is being passed around. After fixes, I can approve.

PR #2151, opened 2023-11-20
added play state to plan request

petergarud:
Adds PlayState to PlanRequest


sid-parikh (approved):
Going to merge this in without worrying about #2148 for now, because its critical for #2163

PR #2176, opened 2024-01-29
Added behavior for marking enemy robots

jvogt23:
## Description
Describe your pull request.
Added a defensive behavior to mark enemy robots by following them around the field and blocking them from receiving the ball. Currently, it assumes that one marker will be created, and it creates the marker after there are 3 Wallers in play. If more markers are used in the future, logic will have to be added that gives each marker a different target.

#86ayuj73h
## Design Documents
[Link](link-to-design-doc)

### Test Case 1
1. Step 1 - Start a match
2. Step 2 - There should be 3 Wallers, and a Marker should be following around the lowest enemy robot.
3. Step 3 - Move the ball and the enemy robot around, and move enemy robots in and out of our half of the field.

**Expected result:**???
The waller with the highest ID should now be a Marker, and it will follow around the lowest ID'd robot on the other team that is on our side of the field. The Marker will enter a planning state whenever it is searching for a robot to mark, and it will not move during this time. When it finds a robot to mark, it will immediately target that robot. 

## Key Files to Review
Group 1: Marker classes
 * File 1 marker.cpp
 * File 2 marker.hpp

Group 2 Defense
 * File 3 defense.cpp
 * File 4 defense.hpp

## Review Checklist

- [ ] Step 1
- [ ] Step 2

### DIFF
--- a/soccer/src/soccer/strategy/agent/position/defense.cpp
+++ b/soccer/src/soccer/strategy/agent/position/defense.cpp
@@ -2,9 +2,7 @@
 
 namespace strategy {
 
-Defense::Defense(int r_id) : 
-    Position(r_id, "Defense"), 
-    marker_{field_dimensions_} {}
+Defense::Defense(int r_id) : Position(r_id, "Defense"), marker_{field_dimensions_} {}
 
 std::optional<RobotIntent> Defense::derived_get_task(RobotIntent intent) {
     current_state_ = update_state();
@@ -42,7 +40,7 @@ Defense::State Defense::update_state() {
             // If a wall is already full,
             // Remove the robot with the highest ID from a wall
             // and make them a marker instead.
-            if (walling_robots_.size() > kMaxWallers && 
+            if (walling_robots_.size() > kMaxWallers &&
                 this->robot_id_ == *max_element(walling_robots_.begin(), walling_robots_.end())) {
                 send_leave_wall_request();
                 next_state = ENTERING_MARKING;
--- a/soccer/src/soccer/strategy/agent/position/defense.cpp
+++ b/soccer/src/soccer/strategy/agent/position/defense.cpp
@@ -3,7 +3,9 @@
 
 namespace strategy {
 
-Defense::Defense(int r_id) : Position(r_id, "Defense") {}
+Defense::Defense(int r_id) : 
+    Position(r_id, "Defense"), 
+    marker_{static_cast<uint8_t>(r_id)} {}
 
 std::optional<RobotIntent> Defense::derived_get_task(RobotIntent intent) {
     current_state_ = update_state();
@@ -31,6 +33,8 @@ Defense::State Defense::update_state() {
 
     switch (current_state_) {
         case IDLING:
+            if (robot_id_ == 2) 
+            SPDLOG_INFO("IN IDLING");
             break;
         case JOINING_WALL:
             send_join_wall_request();
@@ -63,11 +67,25 @@ Defense::State Defense::update_state() {
                 next_state = IDLING;
             }
         case MARKING:
+            if (marker_.get_target() == -1 || marker_.target_out_of_bounds(world_state)) {
+                SPDLOG_INFO("TRANSITION INTO IDLING FROM MARKING");
+                next_state = ENTERING_MARKING;
+            }
             break;
-    }
-
-    if (robot_id_ == 2) {
-        next_state = MARKING;
+        case ENTERING_MARKING:
+            SPDLOG_INFO("IN ENTERING MARKING");
+            int target_id = marker_.choose_target(world_state);
+            if (target_id == -1) {
+                SPDLOG_INFO("THE ID IS -1, ENTERING IDLING");
+                next_state = ENTERING_MARKING;
+            } else {
+                next_state = MARKING;
+            }
+        
+        // if (robot_id_ == 2 && has_already_run_enter_mark == false) {
+        //     next_state = ENTERING_MARKING;
+        //     has_already_run_enter_mark = true;
+        // }
     }
 
     return next_state;
@@ -124,11 +142,10 @@ std::optional<RobotIntent> Defense::state_to_task(RobotIntent intent) {
             Waller waller{waller_id_, (int)walling_robots_.size()};
             return waller.get_task(intent, last_world_state_, this->field_dimensions_);
         }
-    } 
-    else if (current_state_ == MARKING) {
-        Marker marker = Marker((u_int8_t) robot_id_);
-        return marker.get_task(intent, last_world_state_, this->field_dimensions_);
-    }else if (current_state_ == FACING) {
+    } else if (current_state_ == MARKING) {
+        //Marker marker = Marker((u_int8_t) robot_id_);
+        return marker_.get_task(intent, last_world_state_, this->field_dimensions_);
+    } else if (current_state_ == FACING) {
         rj_geometry::Point robot_position =
             last_world_state_->get_robot(true, robot_id_).pose.position();
         auto current_location_instant =
@@ -137,6 +154,10 @@ std::optional<RobotIntent> Defense::state_to_task(RobotIntent intent) {
         auto face_ball_cmd =
             planning::MotionCommand{"path_target", current_location_instant, face_ball};
         intent.motion_command = face_ball_cmd;
+        return intent;
+    } else if (current_state_ == ENTERING_MARKING) {
+        auto empty_motion_cmd = planning::MotionCommand{};
+        intent.motion_command = empty_motion_cmd;
         return intent;
     }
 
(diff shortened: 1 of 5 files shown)


soccer/src/soccer/strategy/agent/position/defense.cpp
```diff
@@ -33,6 +35,14 @@ Defense::State Defense::update_state() {
             walling_robots_ = {(u_int8_t)robot_id_};
             break;
         case WALLING:
+            // If a wall is already full,
+            // Remove the robot with the highest ID from a wall
+            // and make them a marker instead.
+            if (this->robot_id_ == *max_element(walling_robots_.begin(), walling_robots_.end()) &&
```
sid-parikh: it's really minor but can we flip the order of this &&? just logically, the size() check is more likely to be false, and also faster to execute

soccer/src/soccer/strategy/agent/position/defense.hpp
```diff
@@ -42,6 +43,8 @@ class Defense : public Position {
 private:
     int move_ct_ = 0;
 
+    static constexpr int MAX_WALLERS{3};
```
sid-parikh: kMaxWallers

soccer/src/soccer/strategy/agent/position/marker.cpp
```diff
@@ -0,0 +1,44 @@
+#include "marker.hpp"
+
+namespace strategy {
+Marker::Marker(u_int8_t robot_id) {}
```
sid-parikh: if you no longer need robot_id feel free to get rid of it

soccer/src/soccer/strategy/agent/position/marker.hpp
```diff
@@ -0,0 +1,34 @@
...
+#include "rj_geometry/point.hpp"
+#include "role_interface.hpp"
+
+namespace strategy {
+
+/**
+ * Represents a Marker, a defensive role that targets an enemy robot
+ * and follows it around the field while it is on our side, blocking passes.
+ */
+class Marker : public RoleInterface {
```
sid-parikh, line 18: note the GH action warning here. check the "rule of five" under https://en.cppreference.com/w/cpp/language/rule_of_three. you can explicitly set all of those functions to `=default;`

soccer/src/soccer/strategy/agent/position/marker.hpp
```diff
@@ -0,0 +1,34 @@
...
+private:
+    int target{-1};
```
sid-parikh:
```suggestion
    int target_{-1};
```

soccer/src/soccer/strategy/agent/position/marker.hpp
```diff
@@ -0,0 +1,34 @@
...
+
+    static constexpr double Y_BOUND{4.5};
```
sid-parikh:
```suggestion
    static constexpr double kYBound{4.5};
```

soccer/src/soccer/strategy/agent/position/marker.hpp
```diff
@@ -0,0 +1,34 @@
...
+
+public:
+    Marker(u_int8_t robot_id);
+    ~Marker() = default;
+    std::optional<RobotIntent> get_task(RobotIntent intent, const WorldState* const world_state,
```
sid-parikh:
```suggestion
    std::optional<RobotIntent> get_task(RobotIntent intent, const WorldState* world_state,
```

soccer/src/soccer/strategy/agent/position/offense.cpp
```diff
@@ -87,7 +87,10 @@ Offense::State Offense::update_state() {
 }
 
 std::optional<RobotIntent> Offense::state_to_task(RobotIntent intent) {
-    SPDLOG_INFO(current_state_);
+    // This was added to our code when we merged from ros2.
+    // Please let us know whether this log is needed when
+    // reviewing.
+    // SPDLOG_INFO(current_state_);
```
sid-parikh: feel free to delete

soccer/src/soccer/strategy/agent/position/marker.cpp
```diff
@@ -0,0 +1,44 @@
...
+        planning::MotionCommand{"path_target", goal, planning::FaceBall{}, true};
+    return intent;
+}
+
+int Marker::choose_target(WorldState* ws) {
+    // TODO: (James Vogt, github: jvogt23)
+    // If we ever use multiple Markers, they should choose different
+    // robots to track from each other. Logic for this operation must be
+    // added because multiple markers currently mark the same robot.
+    for (int i = 0; i < 11; i++) {
```
sid-parikh: don't hard code constants like 11. try using `kNumShells`, or better yet, use `for (const auto& robot : ws->their_robots)`

soccer/src/soccer/strategy/agent/position/marker.cpp
```diff
@@ -0,0 +1,44 @@
+#include "marker.hpp"
+
+namespace strategy {
+Marker::Marker(u_int8_t robot_id) {}
+
+std::optional<RobotIntent> Marker::get_task(RobotIntent intent, const WorldState* world_state,
+                                            FieldDimensions field_dimensions) {
```
sid-parikh:
```suggestion
                                            [[maybe_unused]] FieldDimensions field_dimensions) {
```

soccer/src/soccer/strategy/agent/position/marker.cpp
```diff
@@ -0,0 +1,44 @@
...
+    return intent;
+}
+
+int Marker::choose_target(WorldState* ws) {
+    // TODO: (James Vogt, github: jvogt23)
+    // If we ever use multiple Markers, they should choose different
+    // robots to track from each other. Logic for this operation must be
+    // added because multiple markers currently mark the same robot.
+    for (int i = 0; i < 11; i++) {
+        if (std::fabs(ws->get_robot(false, i).pose.position().x()) < 2.5 &&
```
sid-parikh: what is the `2.5` constant? are we choosing targets only on the edge of the field?

soccer/src/soccer/strategy/agent/position/marker.cpp
```diff
@@ -0,0 +1,44 @@
...
+            ws->get_robot(false, i).pose.position().y() < Y_BOUND &&
+            (ws->ball.position - ws->get_robot(false, i).pose.position()).mag() > .25) {
+            target = i;
+            return i;
+        }
+    }
+    target = -1;
```
sid-parikh: either return the value and allow the caller to save it into the member, or save it into the member and return void. doing both leads to bugs (especially with multiple return statements like here)

soccer/src/soccer/strategy/agent/position/marker.hpp
```diff
@@ -0,0 +1,34 @@
...
+                                        FieldDimensions field_dimensions) override;
+
+    int choose_target(WorldState* ws);
```
sid-parikh:
```suggestion
    int choose_target(const WorldState* ws);
```

soccer/src/soccer/strategy/agent/position/marker.hpp
```diff
@@ -0,0 +1,34 @@
...
+    int get_target();
+    bool target_out_of_bounds(WorldState* ws);
```
sid-parikh:
```suggestion
    bool target_out_of_bounds(const WorldState* ws);
```

sid-parikh (changes requested):
The logic is sound and the demo was impressive! I left a few comments for code quality purposes

soccer/src/soccer/strategy/agent/position/marker.cpp
jvogt23: This is actually to ignore targets that are right on the edge of the field. This is mostly so we don't accidentally mark a robot on the sideline or something like that.

jvogt23: Also note that it only does this calculation when choosing a target. Once it has a target, it only cares about whether the robot is on our side of the field

soccer/src/soccer/strategy/agent/position/marker.hpp
```diff
@@ -0,0 +1,41 @@
...
+
+/**
+ * Represents a Marker, a defensive role that targets an enemy robot
+ * and follows it around the field while it is on our side, blocking passes.
+ */
+class Marker : public RoleInterface {
+private:
+    int target_{-1};
+
+    static constexpr double kYBound{4.5};
```
sid-parikh: Hey, I appreciate you making things into named constants. I might've even asked you to make these, I don't remember. However, since these are field-related dimensions, can you calculate them using `field_dimensions` inside your function instead? 

The reason for that is the size of our field changes often, and hard-coded constants related to field size don't work. 

Our sim field, for example, is twice the size of our shop field.

sid-parikh (changes requested):
sorry for not catching this on my first review. i'll help you with merge conflicts on Tuesday

sid-parikh (approved):

PR #2194, opened 2024-02-21
Position Monitoring Framework

WillDeinzer:
Created new AgentState message type with one string field called state.
Made a pure virtual function in the position class that returns current state, each position returns the position name, and offense and defense also return integer value of state enum.
RobotFactoryPosition forwards this function call.


soccer/src/soccer/strategy/agent/position/position.hpp
```diff
@@ -75,6 +76,9 @@ class Position {
     void update_alive_robots(std::vector<u_int8_t> alive_robots);
     const std::string get_name();
 
+    //returns the current state of the robot
+    virtual std::string return_current_state() = 0;
```
sid-parikh:
```suggestion
    virtual std::string get_current_state() = 0;
```
Just a quick suggestion. It's not common to name a function "return...". I would suggest either `get_current_state` or even simply `current_state`

soccer/src/soccer/strategy/agent/position/defense.cpp
```diff
@@ -9,6 +10,10 @@ std::optional<RobotIntent> Defense::derived_get_task(RobotIntent intent) {
     return state_to_task(intent);
 }
 
+std::string Defense::return_current_state() {
+    return "Defense" + static_cast<int>(current_state_);
```
sid-parikh: Yeah, this doesn't work the way you think it does. Try this instead:

```suggestion
    return std::string{"Defense"} + std::to_string(static_cast<int>(current_state_));
```


The basic problem you have is that `"Offense"` is not of type `std::string` but rather `const char*`. It does get implicitly converted to an `std::string` but not before you add an integer to the pointer.

soccer/src/soccer/strategy/agent/position/offense.cpp
```diff
@@ -14,6 +14,10 @@ std::optional<RobotIntent> Offense::derived_get_task(RobotIntent intent) {
     return state_to_task(intent);
 }
 
+std::string Offense::return_current_state() {
+    return "Offense" + static_cast<int>(current_state_);
```
sid-parikh: Same problem as in Defense, the same fix should work. Note the clang-tidy warning present in Github. In the future, please check these warnings before requesting a review. I am happy to help you fix them if you don't know how.

sid-parikh (changes requested):
Looks pretty good! There's one tiny bug due to C++ annoyances, and I recommended a quick rename. Make those fixes and then re-request my review and you should be good.

sid-parikh:
Hey also, it doesn't matter too too much, but there's no need for you use a fork of our repo. Feel free to make a new branch and push them directly to our repo.

sid-parikh (approved):
Looks great. Going to push a style-fix commit and then merge

PR #2197, opened 2024-02-26
Test software

shourikb:
This pull request contains a testing class (hpp and cpp). This is a position just like Offense, Defense, and Goalie which can be set in robotfactoryposition.cpp. The code functions best when set to one robot being Testing at a time, but it does not break if multiple are set.

The testing code has the robots line up and move down, while facing the center (this simulates rotation). Then, one robot will go to the center in a "STEALING" state from the offense.

When having offense and testing running simultaneously, make sure offensive robot does not run into the testing robot.

Currently the states SPIN1 and SPIN2 are non-functional, we suspect this is due to the planner's check is done function not properly working.

The state LINEKICK also refers to the STEALING state; we did not change the name because we are hoping for this to actually become LINEKICK in the future version of this class.

### DIFF
--- a/soccer/src/soccer/planning/planner/line_kick_path_planner.hpp
+++ b/soccer/src/soccer/planning/planner/line_kick_path_planner.hpp
@@ -50,7 +50,7 @@ private:
     Trajectory prev_path_;
 
     // These constants could be tuned more
-    static constexpr double kIsDoneBallVel{1.5};
+    static constexpr double kIsDoneBallVel{0.5};
     static constexpr double kFinalRobotSpeed{1.0};
     static constexpr double kPredictIn{0.5};  // seconds
     static constexpr double kAvoidBallBy{0.05};
--- a/soccer/src/soccer/strategy/agent/position/testing.cpp
+++ b/soccer/src/soccer/strategy/agent/position/testing.cpp
@@ -2,7 +2,7 @@
 
 namespace strategy {
 
-Testing::Testing(int r_id) : Position(r_id) {
+Testing::Testing(int r_id) : Position(r_id), off_(r_id_){
     position_name_ = "Testing";
     current_state_ = IDLING;
     move_on_ = false;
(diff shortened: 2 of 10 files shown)


sid-parikh:
@shourikb @petergarud @shrikar (what's ur GH username?) this PR doesn't compile?

sid-parikh (changes requested):
Please do not commit ER-Force's framework repo as a submodule. I would not recommend storing it within the `robocup-software` directory. If you insist on this, at least un-commit it.

It also makes our GH build fail.

soccer/src/soccer/planning/planner/line_kick_path_planner.hpp
```diff
@@ -50,7 +50,7 @@ class LineKickPathPlanner : public PathPlanner {
     Trajectory prev_path_;
 
     // These constants could be tuned more
-    static constexpr double kIsDoneBallVel{1.5};
```
sid-parikh, line 53: why was this changed?

soccer/src/soccer/strategy/agent/position/testing.cpp
```diff
@@ -0,0 +1,188 @@
+#include "testing.hpp"
+
+namespace strategy {
+
+Testing::Testing(int r_id) : Position(r_id), off_(r_id_) {
+    position_name_ = "Testing";
+    current_state_ = IDLING;
+    move_on_ = false;
+    r_id_ = counter++;
```
sid-parikh, line 9: what does this do? a constructor is only called once

soccer/src/soccer/strategy/agent/position/testing.cpp
```diff
@@ -0,0 +1,188 @@
+#include "testing.hpp"
+
+namespace strategy {
+
+Testing::Testing(int r_id) : Position(r_id), off_(r_id_) {
+    position_name_ = "Testing";
+    current_state_ = IDLING;
```
sid-parikh, line 7: initialize all these members using a member initializer list instead of overwriting them in the constructor

soccer/src/soccer/strategy/agent/position/testing.cpp
```diff
@@ -0,0 +1,188 @@
...
+    return intent;
+}
+
+bool Testing::proceed() {
+    bool done = check_is_done();
+    if (done || move_on_) {
+        is_running_.at(r_id_) = false;
+        move_on_ = true;
+    } else {
+        is_running_.at(r_id_) = true;
```
sid-parikh, line 191: ohh i see what you're doing... let's talk about this i don't think it's the best idea

sid-parikh (changes requested):
i think we need to massively simplify this. just focus on testing some basic movements. have it only think about one robot (itself).

don't try to manually test offense, etc. we can do that in other ways.

i appreciate all the work you guys have done to create the testing procedure, but I'm realizing it is impractical to try to test other positions from a test position without way too much more code.

PR #2199, opened 2024-02-26
Run style checks for forks, but don't make the fixing PR

sid-parikh:
I noticed that when members make a PR from a fork, the style checks don't run.

I think this is because we couldn't make a PR to fix checks on a fork.

So I moved the if condition so we still check style on forks, just don't try to change it. 

What do you think? @N8BWert you have some GH action experience do you think this will break stuff


N8BWert (approved):
Looks good to me

PR #2207, opened 2024-03-04
New Position Constructor

oddlyspookycherry:
## Description
Use new Position constructor in subclasses.

Resolves # or [ClickUp card](https://app.clickup.com/t/86azjmthr)

## Test
Code compiles and runs with `make perf` and `make run-sim`.


sid-parikh (approved):
Looks great, thanks!

PR #2200, opened 2024-02-28
Free Kick / Penalty / Kickoff

sanatd33:
## Description
Writes free kick, kickoff, penalty code

## Associated / Resolved Issue
Resolves https://app.clickup.com/t/86azbvek6

## Key Files to Review
 * robot_factory_position.cpp

## Review Checklist

- [ ] **Docstrings**: All methods and classes should have the file appropriate docstrings which follow the guidelines in the ["Contributing" page](https://rj-rc-software.readthedocs.io/en/latest/contributing.html) of our docs.
- [ ] **Remove extra print statements**: Any print statements used for debugging should be removed
- [ ] **Tag reviewers**: Tag some people for review and ping them on Slack

PR #2212, opened 2024-03-06
Alive Robots in robot_factory_position

shourikb:
Changed a few things so that robot_factory_position and a couple other classes use the alive_robots_ array properly. Had to change the type of alive_robots_ from vector to array in a few classes.

In the loop in robot_factory_position, there is a variable j that keeps track of the count and the for each loop gets all of the booleans from alive_robots_ and uses that value in whether to execute the code in the loop body or not.


soccer/src/soccer/strategy/agent/position/robot_factory_position.hpp
```diff
@@ -69,7 +69,7 @@ class RobotFactoryPosition : public Position {
         current_position_->update_field_dimensions(field_dimensions);
     }
 
-    void update_alive_robots(std::vector<u_int8_t> alive_robots) override {
+    void update_alive_robots(std::array<bool, kNumShells> alive_robots) override {
```
sid-parikh, line 72:
```suggestion
    void update_alive_robots(std::array<bool, kNumShells> alive_robots) override {
        Position::update_alive_robots(alive_robots);
        current_position->update_alive_robots(alive_robots);
```
You need to call super function

soccer/src/soccer/strategy/agent/position/robot_factory_position.cpp
```diff
@@ -123,4 +128,8 @@ std::string RobotFactoryPosition::get_current_state() {
     return current_position_->get_current_state();
 }
 
+// void RobotFactoryPosition::update_alive_robots(std::array<bool, kNumShells> alive_robots) {
```
sid-parikh: delete this commented out code

soccer/src/soccer/strategy/agent/position/robot_factory_position.cpp
```diff
@@ -26,14 +26,19 @@ std::optional<RobotIntent> RobotFactoryPosition::get_task(WorldState& world_stat
     using RobotPos = std::pair<int, double>;  // (robotId, yPosition)
 
     std::vector<RobotPos> robots_copy;
-    for (int i = 0; i < 6; i++) {
+    int j = 0;
+    for (bool alive : alive_robots_) {
```
sid-parikh:
```suggestion
    for (int i = 0; i < kNumShells; ++i) {
```

soccer/src/soccer/strategy/agent/position/robot_factory_position.cpp
```diff
@@ -26,14 +26,19 @@ std::optional<RobotIntent> RobotFactoryPosition::get_task(WorldState& world_stat
...
         // Ignore goalie
-        if (i == 0) {
+        if (j == goalie_id_) {
             continue;
         }
-        robots_copy.emplace_back(i, world_state.our_robots[i].pose.position().y());
+        if (alive) {
```
sid-parikh:
```suggestion
        if (alive_robots_[i]) {
```

sid-parikh: this is a lot cleaner

sid-parikh (changes requested):
You also need to update the `position.cpp` implementation of alive robots.  It currently searches the list for the robot id, but it should instead check if `alive_robots[robot_id] == true` or something like that, because you changed it from a vector to an array

soccer/src/soccer/strategy/agent/position/robot_factory_position.hpp
sid-parikh: no, that's not correct. you need to call the super function _for robot factory position_, and then call the current_position function _without Position::_

sid-parikh: I've updated my GH suggestion to show what I mean

soccer/src/soccer/strategy/agent/position/robot_factory_position.hpp
```diff
@@ -69,8 +69,8 @@ class RobotFactoryPosition : public Position {
         current_position_->update_field_dimensions(field_dimensions);
     }
 
-    void update_alive_robots(std::vector<u_int8_t> alive_robots) override {
-        current_position_->update_alive_robots(alive_robots);
+    void update_alive_robots(std::array<bool, kNumShells> alive_robots) override {
+        current_position_->Position::update_alive_robots(alive_robots);
```
sid-parikh:
```suggestion

```

sid-parikh (changes requested):

soccer/src/soccer/strategy/agent/agent_action_client.cpp
```diff
@@ -116,6 +116,8 @@ void AgentActionClient::goalie_id_callback(int goalie_id) {
 
 void AgentActionClient::alive_robots_callback(const rj_msgs::msg::AliveRobots::SharedPtr& msg) {
     alive_robots_ = msg->alive_robots;
+
+    current_position_->Position::update_alive_robots(alive_robots_);
```
sid-parikh:
```suggestion
    current_position_->update_alive_robots(alive_robots_);
```

soccer/src/soccer/radio/sim_radio.cpp
```diff
@@ -94,18 +94,19 @@ SimRadio::SimRadio(bool blue_team)
         ip::udp::endpoint(address_, blue_team_ ? kSimBlueCommandPort : kSimYellowCommandPort);
     sim_control_endpoint_ = ip::udp::endpoint(address_, kSimCommandPort);
 
+    // assume robots 0-6 are always alive in sim
```
rishiso, line 97:
```suggestion
    // assume robots 0-5 are always alive in sim
```

soccer/src/soccer/strategy/agent/position/robot_factory_position.cpp
```diff
@@ -49,29 +50,29 @@ std::optional<RobotIntent> RobotFactoryPosition::get_task(WorldState& world_stat
     // Assigning new position
     // Checking whether we have possesion or if the ball is on their half (using 1.99 to avoid
```
rishiso, line 51:
```suggestion
    // Checking whether we have possesion or if the ball is on their half
```

rishiso, line 51:
```suggestion
    // Checking whether we have possesion or if the ball is on their half
```

soccer/src/soccer/strategy/agent/position/robot_factory_position.cpp
```diff
@@ -49,29 +50,29 @@ std::optional<RobotIntent> RobotFactoryPosition::get_task(WorldState& world_stat
...
     // rounding issues on midline)
```
rishiso, line 52:
```suggestion
```

soccer/src/soccer/strategy/agent/position/robot_factory_position.cpp
```diff
@@ -6,7 +6,6 @@ RobotFactoryPosition::RobotFactoryPosition(int r_id) : Position(r_id, "RobotFact
     if (robot_id_ == 0) {
```
rishiso, line 6:
```suggestion
    if (robot_id_ == goalie_id_) {
```

rishiso: Goalie is never reassigned inside get_task(), so it is important to do this here

soccer/src/soccer/strategy/agent/position/robot_factory_position.cpp
```diff
@@ -16,7 +15,7 @@ RobotFactoryPosition::RobotFactoryPosition(int r_id) : Position(r_id, "RobotFact
 std::optional<RobotIntent> RobotFactoryPosition::get_task(WorldState& world_state,
                                                           FieldDimensions& field_dimensions) {
     // If keeper, make no changes
-    if (robot_id_ == 0) {
+    if (robot_id_ == goalie_id_) {
```
rishiso, line 18: I will need to double-check with Sid about this, but if the goalie can change, we need to add functionality here to update the goalie if it is not assigned correctly.

soccer/src/soccer/strategy/agent/position/robot_factory_position.cpp
```diff
@@ -49,29 +50,29 @@ std::optional<RobotIntent> RobotFactoryPosition::get_task(WorldState& world_stat
     // Assigning new position
     // Checking whether we have possesion or if the ball is on their half (using 1.99 to avoid
     // rounding issues on midline)
-    if (Position::our_possession_ ||
-        world_state.ball.position.y() > field_dimensions.length() / 1.99) {
+    if (our_possession_ ||
+        world_state.ball.position.y() > field_dimensions.center_field_loc().y() - kBallDiameter) {
         // Offensive mode
         // Closest 2 robots on defense, rest on offense
         if (i <= 1) {
```
jacksherling, line 57: While we're working on this file, could we change the i variable name to something more descriptive? Perhaps `robot_y_index`?
Variable defined on line 42 but it's collapsed so GitHub won't let me comment there

soccer/src/soccer/strategy/agent/position/robot_factory_position.cpp
jacksherling: good catch

jacksherling (changes requested):

soccer/src/soccer/strategy/agent/position/robot_factory_position.cpp
jacksherling: good catch

sid-parikh:
Merged into kicker-picker see #2200

PR #2215, opened 2024-03-06
Sim Alive Robots Publish

oddlyspookycherry:
## Description
Publish robots 0-6 as alive in `sim_radio`.
Declare `strategy/alive_robots` as a global topic name for alive robots and make all radios use it.

## Steps to Test
1. Run sim.
2. `ros2 topic echo strategy/alive_robots`

### DIFF
--- a/soccer/src/soccer/radio/radio.cpp
+++ b/soccer/src/soccer/radio/radio.cpp
@@ -39,7 +39,7 @@ Radio::Radio()
     }
 
     alive_robots_pub_ =
-        create_publisher<rj_msgs::msg::AliveRobots>("/alive_robots", rclcpp::QoS(1));
+        create_publisher<rj_msgs::msg::AliveRobots>(strategy::topics::kAliveRobots, rclcpp::QoS(1));
 
     tick_timer_ = create_wall_timer(tick_period_, [this]() { tick(); });
 }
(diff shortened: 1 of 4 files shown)


soccer/src/soccer/radio/radio.cpp
```diff
@@ -39,7 +39,7 @@ Radio::Radio()
     }
 
     alive_robots_pub_ =
-        create_publisher<rj_msgs::msg::AliveRobots>("/alive_robots", rclcpp::QoS(1));
```
oddlyspookycherry, line 42: For some reason base Radio class (used by network_radio) and sim_radio used different topic names for alive robots. I went with the one used by sim.

soccer/src/soccer/radio/sim_radio.cpp
```diff
@@ -94,17 +94,19 @@ SimRadio::SimRadio(bool blue_team)
         ip::udp::endpoint(address_, blue_team_ ? kSimBlueCommandPort : kSimYellowCommandPort);
     sim_control_endpoint_ = ip::udp::endpoint(address_, kSimCommandPort);
 
+    // assume robots 0-6 are always alive in sim
+    for (uint8_t robot_id = 0; robot_id < 7; robot_id++) {
```
sid-parikh:
```suggestion
    for (uint8_t robot_id = 0; robot_id <= 6; robot_id++) {
``` 
is slightly more semantic

sid-parikh (changes requested):
Can we actually use the other `"/alive_robots"` topic? Can you also update `agent_action_client` to subscribe to the other one? We want everything to be the same. No need for the `"strategy/"` one. Thank you!

sid-parikh:
Code was merged into #2212. @oddlyspookycherry , feel free to leave a review there to ensure your changes were transferred the way you expect.

PR #2053, opened 2023-04-10
Simplify networking

kfu02:
## Description
Previously, correcting the interfaces we receive ref/vision data from required a lot of hardcoding in two separate files. Now it should not require any knowledge of interfaces or networking at all!

## Design Documents
IMG_4585

### Test Case 1

1. Build and run `run-sim-ex` (**NOT `run-sim`**)
2. Launch external referee (ssl-game-controller)
3. Launch ER-Force sim manually (cd to framework, then `./build/bin/simulator-cli -g 2020B`)

**Expected result:** Works the same as ROS2 (robot 2 struggles to pass)

## Review Checklist

- [x] **Docstrings**: All methods and classes should have the file appropriate docstrings which follow the guidelines in the ["Contributing" page](https://rj-rc-software.readthedocs.io/en/latest/contributing.html) of our docs.
- [x] **Remove extra print statements**: Any print statements used for debugging should be removed
- [x] **Tag reviewers**: Tag some people for review and ping them on Slack

PR #2221, opened 2024-03-10
Accept referee commands from all physical interfaces

sid-parikh:
## Description

Previously, an IP address for the external referee was specified. This would regularly be localhost when running ssl-game-controller on the same machine (i.e. for testing in sim), but at competitions this IP address would vary often. 

This forced us to first manually identify the physical IP address of the game controller, modify the constant, and then recompile. 

However, by default, boost/setsockopt allows us to listen on any/all physical interfaces (when binding to a multicast address). So, by simply removing this specification, we should be able to receive referee commands on any setup without any manual configuraiton.

Relevant paragraphs from the setsockopt docs: [1](https://arc.net/l/quote/rixrrtnz) and [2](https://arc.net/l/quote/ezlvtaia)

Resolves [ClickUp card](https://app.clickup.com/t/863h9cq3v)

### Localhost external ref
1. Run ssl-game-controller
2. Run ER-Force `simulator-cli`
3. Run `make run-sim-ex`

**Expected result:** soccer responds to referee commands (note that behavior is irrelevant; ensure that commands are visible in the left panel of the UI)

### Networked external ref
1. Run ssl-game-controller on some other machine on the same network as your own
- Note: it is possible at comp that this machine is connected directly to yours via a link-layer switch and there may be no router assigning IP addresses.
3. Run ER-Force `simulator-cli`
4. Run `make run-sim-ex`

**Expected result:** soccer responds to referee commands (note that behavior is irrelevant; ensure that commands are visible in the left panel of the UI)

## Key Files to Review
 * `network.hpp`
 * `external_referee.cpp`

## Review Checklist

### DIFF
--- a/rj_common/include/rj_common/network.hpp
+++ b/rj_common/include/rj_common/network.hpp
@@ -37,32 +37,9 @@
  * In networking terms, the referee packets' source address should match
  * kRefereeSourceAddress, and same thing for SharedVisionAddress.
  *
- * The physical interface we expect to receive these packets from is defined
- * below.
  */
 static const std::string kRefereeSourceAddress = "224.5.23.1";
 static const std::string kSharedVisionSourceAddress = "224.5.23.2";
-
-/*
- * UPDATE (4/9/2023): these are no longer necessary. Both VisionReceiver and
- * ExternalReferee now simply join the multicast group and listen for their
- * respective source addresses above.
- *
- * These IP addresses are the interfaces (e.g. Ethernet plugged into this
- * laptop) where we expect ref/vision data to come from. They are (likely)
- * physical Ethernet links to the network, so this address won't show up in
- * Wireshark.
- *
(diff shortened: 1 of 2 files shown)


kfu02 (commented):
This looks right, and when I originally added that comment in network.hpp [in this PR](https://github.com/RoboJackets/robocup-software/pull/2053/files) I applied these exact changes to both external ref and vision receiver. However at some point [I reverted those changes](https://github.com/RoboJackets/robocup-software/commit/cce1174232d68a764b3dc2d75454a459a561a7c7) in _only_ external ref, seemingly after on-field testing. I do not recall why, my guess is the ref program was old on-field and didn't support multicast? It's strange that seemingly the vision changes worked fine at comp (as I don't remember fiddling with that address, but I could be wrong, I was sleep-deprived).

sid-parikh:
Just like Kevin said, this somehow broke everything on the field computer. Needs some further testing.

PR #2234, opened 2024-03-13
Pk

shourikb:
Functionality for the Penalty state given by referee. Limited testing has been applied to this code so there may be some bugs.

When it is their penalty, the Goalie will got to the baseline and the other robots will be SmartIdle, which means they get away from the ball and do not do anything.

When it is our penalty, one robot turns into the PenaltyPlayer (the shooter) while the others become SmartIdle. PenaltyPlayer waits for the referee instructions such that it lines up, and then gets the signal to shoot. Uses code from offense to shoot the ball.

In the future: determine PenaltyPlayer by closest robot to penalty spot instead of just first robot on the list. Also add smart logic to goalie lining up.


sid-parikh:
merged into #2236

PR #2263, opened 2024-07-01
Position Override created to allow position testing

jvogt23:
## Description
Describe your pull request.
Created a way to override a robot's current position for testing purposes.

## Design Documents
[Link](link-to-design-doc)

### Test Case 1
1. Step 1 - Run the stack, and under the tab Positions on the right of the main window, set any robot to any position other than auto
2. Step 2 - Hit the reset button for that robot
3. Step 3 - Change the goalie number
4. Step 4 - Set any robot to any position other than auto, then set the goalie number to that robot

**Expected result:**???
1. The robot selected should start behaving consistent with the position selected
2. The robot should go back to whatever its default position is
3. The robot moving out of "Goalie" status should change to Auto, and the robot selected should become a goalie.
4. The robot should start by changing its position, then  when it becomes the Goalie, the previous goalie should become auto, the UI should allow you to select that robot's position, and the UI should switch the new goalie's override to "Auto" and gray out the combo box so you can no longer change that robot's position.

## Key Files to Review
Group 1 - namespace strategy
 * File 1 robot_factory_position.hpp and cpp
 * File 2 overriding_positions.hpp

Group 2 - UI
 * File 3 main_window.cpp
 * File 4 MainWindow.ui

## Review Checklist

- [ ] Step 1
- [ ] Step 2

### DIFF
--- a/rj_msgs/msg/OverridePosition.msg
+++ b/rj_msgs/msg/OverridePosition.msg
@@ -1,0 +1,6 @@
+# Contains two unsigned ints - The robot ID to override and the position
+# to set for that ID, based on the enum OverridingPositions,
+# which is contained in soccer/src/soccer/strategy/agent/position/overriding_positions.hpp
+
+uint32 robot_id
+uint32 overriding_position
--- a/rj_msgs/msg/OverridePosition.msg
+++ b/rj_msgs/msg/OverridePosition.msg
@@ -1,4 +1,4 @@
-# Contains two unsigned ints - The robot ID to override and the position
+# Contains unsigned int overriding_position -  the position
 # to set for that ID, based on the enum OverridingPositions,
 # which is contained in soccer/src/soccer/strategy/agent/position/overriding_positions.hpp
 
--- a/rj_msgs/msg/OverridePosition.msg
+++ b/rj_msgs/msg/OverridePosition.msg
@@ -2,5 +2,4 @@
 # to set for that ID, based on the enum OverridingPositions,
 # which is contained in soccer/src/soccer/strategy/agent/position/overriding_positions.hpp
 
-uint32 robot_id
 uint32 overriding_position
--- a/soccer/src/soccer/strategy/agent/position/overriding_positions.hpp
+++ b/soccer/src/soccer/strategy/agent/position/overriding_positions.hpp
@@ -3,26 +3,26 @@
 #include <string>
 #include <vector>
 
-namespace Strategy
-{
-    /*
-        OverridingPositions refers to the positions that can be manually set in the UI.
-        Normally, all robots are set to Auto. If you want to add a new position, add a new value
-        to this enum before LENGTH, add a string to the overriding_position_labels vector in main_window.hpp,
-        and add a case to the check_for_position_override method in RobotFactoryPosition.
-    */
-    enum OverridingPositions {
-        AUTO,
-        OFFENSE,
-        DEFENSE,
-        FREE_KICKER,
-        PENALTY_PLAYER,
-        PENALTY_NON_KICKER,
-        SOLO_OFFENSE,
-        SMART_IDLE,
-        ZONER,
-        IDLE,
-        LENGTH, //Do not remove
-    };
+namespace Strategy {
+/*
+    OverridingPositions refers to the positions that can be manually set in the UI.
+    Normally, all robots are set to Auto. If you want to add a new position, add a new value
+    to this enum before LENGTH, add a string to the overriding_position_labels vector in
+   main_window.hpp, and add a case to the check_for_position_override method in
+   RobotFactoryPosition.
+*/
+enum OverridingPositions {
+    AUTO,
+    OFFENSE,
+    DEFENSE,
+    FREE_KICKER,
+    PENALTY_PLAYER,
+    PENALTY_NON_KICKER,
+    SOLO_OFFENSE,
+    SMART_IDLE,
+    ZONER,
+    IDLE,
+    LENGTH,  // Do not remove
+};
 
-} // namespace Strategy
+}  // namespace Strategy
--- a/soccer/src/soccer/strategy/agent/position/overriding_positions.hpp
+++ b/soccer/src/soccer/strategy/agent/position/overriding_positions.hpp
@@ -1,0 +1,15 @@
+#pragma once
+
+namespace Strategy
+{
+    enum OverridingPositions {
+        AUTO = 0,
+        OFFENSE = 1,
+        DEFENSE = 2,
+        GOAL_KICKER = 3,
+        MARKER = 4,
+        PENALTY_PLAYER = 5,
+        SEEKER = 6,
+        WALLER = 7,
+    };
+} // namespace Strategy
--- a/soccer/src/soccer/strategy/agent/position/overriding_positions.hpp
+++ b/soccer/src/soccer/strategy/agent/position/overriding_positions.hpp
@@ -1,4 +1,7 @@
 #pragma once
+#include <map>
+#include <vector>
+#include <string>
 
 namespace Strategy
 {
@@ -6,10 +9,13 @@ namespace Strategy
         AUTO = 0,
         OFFENSE = 1,
         DEFENSE = 2,
-        GOAL_KICKER = 3,
-        MARKER = 4,
-        PENALTY_PLAYER = 5,
-        SEEKER = 6,
-        WALLER = 7,
+        FREE_KICKER = 3,
+        PENALTY_PLAYER = 4,
+        PENALTY_NON_KICKER = 5,
+        SOLO_OFFENSE = 6,
+        SMART_IDLE = 7,
+        ZONER = 8,
+        IDLE = 9,
     };
+
 } // namespace Strategy
--- a/soccer/src/soccer/strategy/agent/position/overriding_positions.hpp
+++ b/soccer/src/soccer/strategy/agent/position/overriding_positions.hpp
@@ -1,21 +1,28 @@
 #pragma once
 #include <map>
-#include <vector>
 #include <string>
+#include <vector>
 
 namespace Strategy
 {
+    /*
+        OverridingPositions refers to the positions that can be manually set in the UI.
+        Normally, all robots are set to Auto. If you want to add a new position, add a new value
+        to this enum before LENGTH, add a string to the overriding_position_labels vector in main_window.hpp,
+        and add a case to the check_for_position_override method in RobotFactoryPosition.
+    */
     enum OverridingPositions {
-        AUTO = 0,
-        OFFENSE = 1,
-        DEFENSE = 2,
-        FREE_KICKER = 3,
-        PENALTY_PLAYER = 4,
-        PENALTY_NON_KICKER = 5,
-        SOLO_OFFENSE = 6,
-        SMART_IDLE = 7,
-        ZONER = 8,
-        IDLE = 9,
+        AUTO,
+        OFFENSE,
+        DEFENSE,
+        FREE_KICKER,
+        PENALTY_PLAYER,
+        PENALTY_NON_KICKER,
+        SOLO_OFFENSE,
+        SMART_IDLE,
+        ZONER,
+        IDLE,
+        LENGTH, //Do not remove
     };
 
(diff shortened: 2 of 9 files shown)


soccer/src/soccer/strategy/agent/position/robot_factory_position.cpp
```diff
@@ -119,6 +127,14 @@ void RobotFactoryPosition::handle_ready() {
 }
 
 void RobotFactoryPosition::update_position() {
+    // if (test_play_position_ == Strategy::OverridingPositions::OFFENSE) {
```
shourikb: Can probably remove these comments

soccer/src/soccer/strategy/agent/position/robot_factory_position.hpp
```diff
@@ -109,6 +112,14 @@ class RobotFactoryPosition : public Position {
 private:
     std::unique_ptr<Position> current_position_;
 
+    // subscription for test mode - Allows position overriding
+    rclcpp::Subscription<rj_msgs::msg::OverridePosition>::SharedPtr test_play_sub_;
+    rclcpp::executors::SingleThreadedExecutor _executor;
+    std::thread _executor_thread;
+    void test_play_callback(const rj_msgs::msg::OverridePosition::SharedPtr message);
+    Strategy::OverridingPositions test_play_position_{Strategy::OverridingPositions::AUTO};
```
shourikb: Name variable "override_play_position" for consistency

soccer/src/soccer/ui/main_window.cpp
```diff
@@ -1146,3 +1170,154 @@ void MainWindow::updateDebugLayers(const LogFrame& frame) {
...
+
+void MainWindow::on_positionReset_12_clicked() { onResetButtonClicked(12); }
+
+void MainWindow::on_positionReset_13_clicked() { onResetButtonClicked(13); }
+
+void MainWindow::on_positionReset_14_clicked() { onResetButtonClicked(14); }
+
+void MainWindow::on_positionReset_15_clicked() { onResetButtonClicked(15); }
+
+void MainWindow::on_robotPosition_0_currentIndexChanged(int value) {
```
shourikb, line 1218: Why do we need a separate function for each of these instead of 1 function that takes in the robotPosition as an argument? Does the UI require these to all be separate functions?

Same for the onpositionreset.

If the UI does require them to all be separate, leave a comment at the top stating this.

soccer/src/soccer/strategy/agent/position/overriding_positions.hpp
```diff
@@ -0,0 +1,28 @@
...
+    OFFENSE,
+    DEFENSE,
+    FREE_KICKER,
+    PENALTY_PLAYER,
+    PENALTY_NON_KICKER,
+    SOLO_OFFENSE,
+    SMART_IDLE,
+    ZONER,
+    IDLE,
+    LENGTH,  // Do not remove
```
jacksherling: Why do we have this LENGTH?

soccer/src/soccer/ui/main_window.cpp
```diff
@@ -1146,3 +1170,154 @@ void MainWindow::updateDebugLayers(const LogFrame& frame) {
...
+        test_play_pub_->publish(message);
+        if (position_number != 0) {
+            position_reset_buttons.at(robot)->setEnabled(true);
+        } else {
+            position_reset_buttons.at(robot)->setEnabled(false);
+        }
+    }
+}
+
+void MainWindow::setGoalieDropdown(int robot) {
```
shourikb, line 1305: Is it possible to somehow store the value we set the Goalie to here so we don't need the scheme above to find out who the goalie is? And by default assume it is robot 0 if we have not set anything since robot 0 is the default set to goalie?

shourikb (changes requested):

soccer/src/soccer/strategy/agent/position/robot_factory_position.cpp
```diff
@@ -15,6 +15,14 @@ RobotFactoryPosition::RobotFactoryPosition(int r_id) : Position(r_id, "RobotFact
     } else {
         current_position_ = std::make_unique<Defense>(robot_id_);
     }
+
+    std::string node_name{"robot_factory_position_"};
+    _node = std::make_shared<rclcpp::Node>(node_name.append(std::to_string(robot_id_)));
+    test_play_sub_ = _node->create_subscription<rj_msgs::msg::OverridePosition>(
```
jacksherling: I think we can make our naming more consistent. Maybe override_position_sub_?

soccer/src/soccer/ui/main_window.cpp
```diff
@@ -49,6 +52,7 @@ void calcMinimumWidth(QWidget* widget, const QString& text) {
     widget->setMinimumWidth(rect.width());
 }
 
+// TODO: Set up the UI objects and bind them to methods
```
jacksherling: is this later done?

jacksherling (commented):
Really great work!

soccer/src/soccer/strategy/agent/position/overriding_positions.hpp
jvogt23: Length is a placeholder that should always be the last in the enum. Its index gives the "length" of the enum, or the number of items in it (minus the LENGTH) element. Updating the comment to reflect this

soccer/src/soccer/strategy/agent/position/robot_factory_position.cpp
jvogt23: Was renamed override_play_sub_ based on an existing conversation by another reviewer.

soccer/src/soccer/ui/main_window.cpp
jvogt23: Yes, starting on line 1174. Forgot to remove this comment.

soccer/src/soccer/strategy/agent/position/robot_factory_position.cpp
```diff
@@ -119,6 +127,9 @@ void RobotFactoryPosition::handle_ready() {
 }
 
 void RobotFactoryPosition::update_position() {
+    bool manual_position_set = check_for_position_override();
+    if (manual_position_set) return;
```
sid-parikh: from clang-tidy: 
```suggestion
    if (manual_position_set) {
        return
    };
```

rj_msgs/msg/OverridePosition.msg
```diff
@@ -0,0 +1,6 @@
+# Contains two unsigned ints - The robot ID to override and the position
+# to set for that ID, based on the enum OverridingPositions,
+# which is contained in soccer/src/soccer/strategy/agent/position/overriding_positions.hpp
+
+uint32 robot_id
```
sid-parikh: (open to discussion) would it be cleaner to create topics for each robot_id instead of adding a field?
pros:
- fits the pattern of other parts of our code
- calls only the necessary callback function instead of one for each Robot
cons:
- more topics, more noise in ros CLI tools
- need to make sure it doesn't break when robots are missing
- more UI code possibly
- -> this would probably be fixable with subtopics though, e.g. override_position/robot_1, override_position/robot_2, ... (would have to go through 16 probably)

soccer/src/soccer/strategy/agent/position/solo_offense.cpp
```diff
@@ -123,7 +123,7 @@ std::optional<RobotIntent> SoloOffense::state_to_task(RobotIntent intent) {
             planning::LinearMotionInstant target{calculate_best_shot()};
             // planning::LinearMotionInstant target{last_world_state_->ball.position};
             auto kick_cmd =
-                planning::MotionCommand{"path_target", target, planning::FaceTarget{}, true};
```
sid-parikh, line 126: why are you making this change?

soccer/src/soccer/ui/main_window.cpp
```diff
@@ -1033,6 +1049,14 @@ void MainWindow::on_actionUse_Multiple_Joysticks_toggled(bool value) {
 
 void MainWindow::on_goalieID_currentIndexChanged(int value) {
     update_cache(_game_settings.request_goalie_id, value - 1, &_game_settings_valid);
+    QString goalieNum = _ui.goalieID->currentText();
+    bool goalieNumIsInt{false};
+    int goalieInt = goalieNum.toInt(&goalieNumIsInt);
+    current_goalie_num_ = goalieInt;
+    if (goalieNumIsInt)
```
sid-parikh: use braces in if statement

soccer/src/soccer/ui/main_window.cpp
```diff
@@ -1033,6 +1049,14 @@ void MainWindow::on_actionUse_Multiple_Joysticks_toggled(bool value) {
 
 void MainWindow::on_goalieID_currentIndexChanged(int value) {
     update_cache(_game_settings.request_goalie_id, value - 1, &_game_settings_valid);
+    QString goalieNum = _ui.goalieID->currentText();
+    bool goalieNumIsInt{false};
+    int goalieInt = goalieNum.toInt(&goalieNumIsInt);
```
sid-parikh, line 1057: wow this might the worst type-conversion API i've ever seen LOL

sid-parikh (changes requested):
looks good. the main thing i'd want to discuss is the choice of topics. everything else was just me being nitpicky, feel free to ignore

soccer/src/soccer/strategy/agent/position/solo_offense.cpp
jvogt23: This change was to fix a problem where the solo offense robot would freeze up after approaching the ball. This change is reflected in ROS2 AFAIK

rj_msgs/msg/OverridePosition.msg
jvogt23: made requested changes using vector of publishers

soccer/src/soccer/strategy/agent/position/robot_factory_position.cpp
```diff
@@ -19,7 +19,7 @@ RobotFactoryPosition::RobotFactoryPosition(int r_id) : Position(r_id, "RobotFact
     std::string node_name{"robot_factory_position_"};
     _node = std::make_shared<rclcpp::Node>(node_name.append(std::to_string(robot_id_)));
     override_play_sub_ = _node->create_subscription<rj_msgs::msg::OverridePosition>(
-        "override_position_for_robot", 1,
+        "override_position_for_robot_" + std::to_string(robot_id_), 1,
```
sid-parikh:
```suggestion
        "override_position/robot_" + std::to_string(robot_id_), 1,
```
I recommend doing this so in ros tools we get a hierarchy

sid-parikh: ideally we also put this in some constants header but i know we're not great at that

sid-parikh (changes requested):

sid-parikh:
Thanks for making that change. I'm real excited for this feature. just a minor nitpick. also @shourikb @jacksherling has someone tested this? i only reviewed code

rj_msgs/msg/OverridePosition.msg
```diff
@@ -0,0 +1,5 @@
+# Contains two unsigned ints - The robot ID to override and the position
```
jacksherling: one int now right? Let's update this comment with the new design

soccer/src/soccer/strategy/agent/position/robot_factory_position.cpp
```diff
@@ -338,4 +351,58 @@ std::string RobotFactoryPosition::get_current_state() {
...
+    override_play_position_ =
+        static_cast<Strategy::OverridingPositions>(message->overriding_position);
+}
+
+/**
+ * Checks override_play_position_, which automatically updates when an override is set.
+ * If it is anything but auto, set the current position to that position and return true.
+ */
+bool RobotFactoryPosition::check_for_position_override() {
+    switch (override_play_position_) {
```
jacksherling, line 365: Not sure about this from a code design perspective. Method name is "check" but it also updates state. Really if we just want to know if a position is being override, that should be a separate method from actually updating that position here

PR #2288, opened 2024-10-02
add ros env vars to source script

sid-parikh:
## Description
ROS2 enviornment was messed up -- `ros2 topic list` would not result in anything in a new terminal. I identified that ros expects certain env vars to be the same in all shells that are talking about the "same" ros processes

Resolves [ClickUp card](https://app.clickup.com/t/86b2ca857)

### Test Case 1
1. Ideally on a fresh install
2. make run-sim in one terminal
3. source && ros2 topic list in another terminal

**Expected result:** you see all the topics

## Key Files to Review
 * env.sh has the new envs
 * do you like where I put the source command? (source.bash)
 
## Questions
* should we think about our naming convention with .zsh .bash etc?
* source.bash is pretty shell-agnostic

## Review Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,7 +6,6 @@
 
 # build directories
 /build
-/install
 /gmon.out
 *.swp
 /.settings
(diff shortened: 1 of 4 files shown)


jacksherling (approved):

shourikb (approved):
Switching to this branch worked for making all the topics show up when you run ros2 topic list after sourcing.

PR #2290, opened 2024-10-20
double wall spacing

obiceen2:
Completes RC SW tutorials.


obiceen2:
Screenshot 2024-09-29 195411

obiceen2:
For the topic /radio/robot_status/robot_0, /radio is the publisher and /planner is the subscriber

PR #2293, opened 2024-10-21
Waller Planner

sanatd33:
Changes behavior of the wallers to more properly move to a new position eg. be more constrained to the arc. This is done by the wallers following the one in front of them, allowing them to move more smoohtly


jacksherling (approved):

shourikb (approved):
Looks good, I'll take your word for it on the math based on how I saw it working

PR #2307, opened 2024-11-04
Valid intermediate Points

sanatd33:
Passes in field dimensions into the create path so that the chosen intermediate point is within the field borders so we dont get stuck

### DIFF
--- a/soccer/src/soccer/planning/global_state.cpp
+++ b/soccer/src/soccer/planning/global_state.cpp
@@ -74,6 +74,10 @@ GlobalState::GlobalState(rclcpp::Node* node) {
     auto lock = std::lock_guard(last_world_state_mutex_);
     return &last_world_state_;
 }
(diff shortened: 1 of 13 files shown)


soccer/src/soccer/planning/global_state.cpp
```diff
@@ -74,6 +74,10 @@ GlobalState::GlobalState(rclcpp::Node* node) {
     auto lock = std::lock_guard(last_world_state_mutex_);
     return &last_world_state_;
 }
+[[nodiscard]] FieldDimensions GlobalState::field_dimensions() const {
+    auto lock = std::lock_guard(last_field_dimensions_mutex_);
+    return last_field_dimensions_;
```
jacksherling: can we return a reference like with world state?

jacksherling (approved):

PR #2352, opened 2025-02-24
Added autoref to the docker

jvogt23:
## Description
Describe your pull request.
Added autoref to docker.

## Design Documents
[Link](link-to-design-doc)

## Steps to Test
1. Run a manual docker build from the branch. The file is Dockerfile.dev so the command should be:
docker build -t <name of build> -f Dockerfile.dev --build-arg TARGETARCH=amd64 .
2. After it finishes building, run it with commands similar to those in our docker hub site.
3. Note that the autoref is in a directory by the same name in the root directory, and the binaries should be stored in a folder called "build" within that directory.
4. After it starts up, log into it through vnc and run simulator-cli, the external ref (again using the commands listed on docker hub), then autoref, then our stack using run-sim-ex
5. The autoref should give occasional feedback about the game state.
### Test Case 1
1. Step 1
6. Step 2
7. Step 3

**Expected result:**???

## Key Files to Review
Group 1
 * File 1
 * File 2

Group 2
 * File 3
 * File 4

## Review Checklist

- [ ] Step 1
- [ ] Step 2

### DIFF
--- a/Dockerfile.dev
+++ b/Dockerfile.dev
@@ -10,6 +10,13 @@ ENV DISPLAY=:1
 ARG TARGETARCH
 ENV TARGETARCH=${TARGETARCH}
 
+# Warn and exit if build arguments are not given
+RUN if [ -z TARGETARCH] then \
+    echo "****************************** ERROR ******************************"; \
+    echo "No target architecture was given. TARGETARCH is a required argument."; \
+    exit 1; \
+    fi
+
 # Update the package list and install necessary packages, including the missing ones
 RUN apt-get update && \
     apt-get install -y \
@@ -67,6 +74,14 @@ RUN git clone https://github.com/robotics-erlangen/framework.git /root/framework
     mkdir build && cd build && \
     cmake .. && \
     make simulator-cli
+
+# Clone, build ER-Force's autoref
+RUN git clone https://github.com/robotics-erlangen/autoref.git /root/autoref && \
+    cd /root/autoref && \
+    git submodule update --init && \
+    mkdir build && cd build && \
+    cmake .. && \
+    make
 
 # Use bash to source the ROS2 setup file and run make perf
 RUN bash -c "source /opt/ros/humble/setup.bash && \
--- a/Dockerfile.dev
+++ b/Dockerfile.dev
@@ -10,12 +10,6 @@ ENV DISPLAY=:1
 ARG TARGETARCH
 ENV TARGETARCH=${TARGETARCH}
 
-# Warn and exit if build arguments are not given
-RUN if [ -z TARGETARCH] then \
-    echo "****************************** ERROR ******************************"; \
-    echo "No target architecture was given. TARGETARCH is a required argument."; \
-    exit 1; \
-    fi
 
 # Update the package list and install necessary packages, including the missing ones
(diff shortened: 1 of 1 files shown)


rishiso:
Is adding an extra check for `TARGETARCH` necessary? It is passed in as an arg to the build by default: [Docker Docs](https://docs.docker.com/reference/dockerfile/#automatic-platform-args-in-the-global-scope)

rishiso:
Can you show a screenshot or something of autoref running in the Docker container?

rishiso (commented):
Alternatively you can just show me that it works in person today

rishiso (approved):
Resolved in-person

jacksherling:
Could you add steps to test (e.g., what commands to run) to the PR description?

jvogt23:
> Could you add steps to test (e.g., what commands to run) to the PR description?

Just added some steps. Let me know if they need to be more detailed.

Dockerfile.dev
```diff
@@ -10,6 +10,13 @@ ENV DISPLAY=:1
 ARG TARGETARCH
 ENV TARGETARCH=${TARGETARCH}
 
+# Warn and exit if build arguments are not given
+RUN if [ -z TARGETARCH] then \
```
sid-parikh:
```suggestion
RUN if [ -z "$TARGETARCH"]; then \
```

I got the following error until I made the above change:

```
=> ERROR [ 2/14] RUN if [ -z TARGETARCH] then     echo "****************************** ERROR ******************************";     echo "No target architecture was given. TARGETARCH is a required argument.";  0.7s
------
 > [ 2/14] RUN if [ -z TARGETARCH] then     echo "****************************** ERROR ******************************";     echo "No target architecture was given. TARGETARCH is a required argument.";     exit 1;     fi:
0.206 /bin/sh: 1: Syntax error: "fi" unexpected (expecting "then")
------
Dockerfile.dev:14
--------------------
  13 |     # Warn and exit if build arguments are not given
  14 | >>> RUN if [ -z TARGETARCH] then \
  15 | >>>     echo "****************************** ERROR ******************************"; \
  16 | >>>     echo "No target architecture was given. TARGETARCH is a required argument."; \
  17 | >>>     exit 1; \
  18 | >>>     fi
  19 |
--------------------
ERROR: failed to solve: process "/bin/sh -c if [ -z TARGETARCH] then     echo \"****************************** ERROR ******************************\";     echo \"No target architecture was given. TARGETARCH is a required argument.\";     exit 1;     fi" did not complete successfully: exit code: 2
```

sid-parikh: though i'm not convinced about the necessity of this layer either

sid-parikh (changes requested):
i'm building it still, but this definitely needs to be fixed first

Dockerfile.dev
rishiso: Yeah it isn't necessary on updated versions of Docker since the param is passed in automatically. Considering we now have Docker CI, we should remove this check

rishiso (changes requested):
Please remove targetarch check and update Docker version locally

rishiso (approved):
LGTM

sid-parikh (approved):

PR #2353, opened 2025-02-26
Fix Angle Planning Crash

sanatd33:
Fix the bug with angle planning crashing by setting jacobian to 0 if the robot is at the point it wants to turn to.


rishiso (approved):
LGTM

jacksherling (approved):

PR #2356, opened 2025-03-04
fix custom ros2 message types

sid-parikh:
## Description
previously, since the humble upgrade, `ros2 topic echo`, `ros2 service call`, and any other CLI command that requires knowledge of a custom type that we declare in `rj_msgs` did not work. Now, they do work, because I've fixed a broken part of the environment variables script.

resolves [clickup card](https://app.clickup.com/t/86b45b57m)

### Test Case 1
1. make perf/debug/whatever
2. `. ./source.bash`
3. `ros2 topic echo /config/field_dimensions` (for example)

**Expected result**: output as usual, no warnings or errors

## Review Checklist

### DIFF
--- a/install/setup.bash
+++ b/install/setup.bash
@@ -34,12 +34,13 @@ _pythonpath_add() {
 # No direct replacement for the command exists, and in Ubuntu 22.04,
 # this script will produce a warning about this. We may need to find a
 # replacement library in the future.
-_PYTHON_LIB_PATH=$(python3 -c "import sysconfig; print(sysconfig.get_path('purelib'))")
-_pythonpath_add "${_PYTHON_LIB_PATH}"
-unset _PYTHON_LIB_PATH
+# _PYTHON_LIB_PATH=$(python3 -c "import sysconfig; print(sysconfig.get_path('purelib'))")
+_pythonpath_add "${_INSTALL_PATH}/local/lib/python3.10/dist-packages"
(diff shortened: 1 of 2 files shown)


install/setup.bash
```diff
@@ -34,12 +34,13 @@ _pythonpath_add() {
...
-_PYTHON_LIB_PATH=$(python3 -c "import sysconfig; print(sysconfig.get_path('purelib'))")
-_pythonpath_add "${_PYTHON_LIB_PATH}"
-unset _PYTHON_LIB_PATH
+# _PYTHON_LIB_PATH=$(python3 -c "import sysconfig; print(sysconfig.get_path('purelib'))")
+_pythonpath_add "${_INSTALL_PATH}/local/lib/python3.10/dist-packages"
+# unset _PYTHON_LIB_PATH
 
 _path_add "${_INSTALL_PATH}/bin"
 _ld_library_path_add "${_INSTALL_PATH}/lib"
+echo "${_INSTALL_PATH}"
```
jacksherling: are we keeping this `echo`?

rishiso: +1

rishiso (approved):
Tested on my end and the ros2 commands are working. Good work!

install/setup.bash
sid-parikh: ty

rishiso (approved):

jacksherling (approved):

PR #2357, opened 2025-03-05
Planner Manipulator

sanatd33:
We can now do manipulator in planners.
New robot intent commands:
TriggerMode: AT_END = at the end of the planner, kick (currently only supported by rotate, but can be expanded to any planner)
DribblerMode: OFF = dribbler off the whole planner
DribblerMode: ON = dribbler on the whole planner
DribblerMode: NEUTRAL = planners decide dribbler behavior

currently, shoot mode is done by strategy but we can have it overwriten by planners
kick speed is set by strategy
dribbler speed set by planners

### DIFF
--- a/config/sim_params.yaml
+++ b/config/sim_params.yaml
@@ -47,7 +47,7 @@
   ros__parameters:
     collect:
       approach_accel_scale: 0.7
-      approach_dist_target: 0.5
+      approach_dist_target: 0.1
       ball_speed_approach_direction_cutoff: 0.1
       control_accel_scale: 0.5
       dist_cutoff_to_approach: 1.0
--- a/config/sim_params.yaml
+++ b/config/sim_params.yaml
@@ -47,7 +47,7 @@
   ros__parameters:
     collect:
       approach_accel_scale: 0.7
-      approach_dist_target: 0.04
+      approach_dist_target: 0.5
       ball_speed_approach_direction_cutoff: 0.1
       control_accel_scale: 0.5
       dist_cutoff_to_approach: 1.0
--- a/rj_msgs/msg/RobotIntent.msg
+++ b/rj_msgs/msg/RobotIntent.msg
@@ -3,14 +3,21 @@ uint8 robot_id
 uint8 SHOOT_MODE_KICK = 0
 uint8 SHOOT_MODE_CHIP = 1
 
+# Stand Down - Never Kick
+# Immediate - Instantly Kick
+# On Break Beam - Kick once break
+# At End - Signals Planner to Kick after it is finished (planners do not support this by default)
 uint8 TRIGGER_MODE_STAND_DOWN = 0
 uint8 TRIGGER_MODE_IMMEDIATE = 1
 uint8 TRIGGER_MODE_ON_BREAK_BEAM = 2
 uint8 TRIGGER_MODE_AT_END = 3
 
+# Off - Off throughout entire planner
+# On - On throughout entire planner
+# Default - Planner decides when to turn on dribbler
 uint8 DRIBBLER_MODE_OFF = 0
 uint8 DRIBBLER_MODE_ON = 1
-uint8 DRIBBLER_MODE_NEUTRAL = 2
+uint8 DRIBBLER_MODE_DEFAULT = 2
 
 MotionCommand motion_command
 
--- a/rj_msgs/msg/RobotIntent.msg
+++ b/rj_msgs/msg/RobotIntent.msg
@@ -6,6 +6,11 @@ uint8 SHOOT_MODE_CHIP = 1
 uint8 TRIGGER_MODE_STAND_DOWN = 0
 uint8 TRIGGER_MODE_IMMEDIATE = 1
 uint8 TRIGGER_MODE_ON_BREAK_BEAM = 2
+uint8 TRIGGER_MODE_AT_END = 3
+
+uint8 DRIBBLER_MODE_OFF = 0
+uint8 DRIBBLER_MODE_ON = 1
+uint8 DRIBBLER_MODE_NEUTRAL = 2
 
 MotionCommand motion_command
 
@@ -14,10 +19,8 @@ rj_geometry_msgs/ShapeSet local_obstacles
 
 uint8 shoot_mode
 uint8 trigger_mode
+uint8 dribbler_mode
 float32 kick_speed
-
-# In range [0, 1]
-float32 dribbler_speed
 
 bool is_active
 
--- a/soccer/src/soccer/logger.cpp
+++ b/soccer/src/soccer/logger.cpp
@@ -206,7 +206,7 @@ std::shared_ptr<Packet::LogFrame> Logger::create_log_frame(Context* context) {
         ConvertRx::status_to_proto(status, rx);
 
         Packet::Robot* tx = log_frame->mutable_radio_tx()->add_robots();
-        ConvertTx::to_proto(trajectory, intent, setpoint, shell, tx);
+        ConvertTx::to_proto(trajectory, intent, setpoint, static_cast<int>(shell), tx);
     }
 
     // Opponent robots
--- a/soccer/src/soccer/logger.cpp
+++ b/soccer/src/soccer/logger.cpp
@@ -198,14 +198,14 @@ std::shared_ptr<Packet::LogFrame> Logger::create_log_frame(Context* context) {
         if (RJ::now() - status.timestamp > RJ::Seconds(0.5)) {
             continue;
         }
-        const auto& intent = context->robot_intents.at(shell);
+        const auto& trajectory = context->trajectories.at(shell);
         const auto& setpoint = context->motion_setpoints.at(shell);
 
         RadioRx* rx = log_frame->add_radio_rx();
         ConvertRx::status_to_proto(status, rx);
 
         Packet::Robot* tx = log_frame->mutable_radio_tx()->add_robots();
-        ConvertTx::to_proto(intent, setpoint, shell, tx);
+        ConvertTx::to_proto(trajectory, setpoint, shell, tx);
     }
 
     // Opponent robots
--- a/soccer/src/soccer/logger.cpp
+++ b/soccer/src/soccer/logger.cpp
@@ -199,13 +199,14 @@ std::shared_ptr<Packet::LogFrame> Logger::create_log_frame(Context* context) {
             continue;
         }
         const auto& trajectory = context->trajectories.at(shell);
+        const auto& intent = context->robot_intents.at(shell);
         const auto& setpoint = context->motion_setpoints.at(shell);
 
         RadioRx* rx = log_frame->add_radio_rx();
         ConvertRx::status_to_proto(status, rx);
 
         Packet::Robot* tx = log_frame->mutable_radio_tx()->add_robots();
-        ConvertTx::to_proto(trajectory, setpoint, shell, tx);
+        ConvertTx::to_proto(trajectory, intent, setpoint, shell, tx);
     }
 
     // Opponent robots
--- a/soccer/src/soccer/planning/planner/collect_path_planner.cpp
+++ b/soccer/src/soccer/planning/planner/collect_path_planner.cpp
@@ -603,6 +603,7 @@ Trajectory CollectPathPlanner::dampen(const PlanRequest& plan_request, RobotInst
     plan_angles(&dampen_end, start_instant, AngleFns::face_point(face_pos),
                 plan_request.constraints.rot);
     dampen_end.stamp(RJ::now());
+    dampen_end.dribbler_speed = 255;
     return dampen_end;
 }
 
@@ -663,7 +664,11 @@ Trajectory CollectPathPlanner::fine_approach(
                     start_instant.position() +
                         Point::direction(AngleFns::face_point(ball.position)(
                             start_instant.linear_motion(), start_instant.heading(), nullptr))));
+        plan_request.debug_drawer->draw_text("Fine Approach", start_instant.position() + Point(.1, .1),
+                                             QColor(255, 255, 255));
     }
+
+    path_hit.dribbler_speed = 255;
 
     return path_hit;
 }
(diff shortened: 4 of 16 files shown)


soccer/src/soccer/planning/planner_for_robot.cpp
```diff
@@ -69,12 +69,40 @@ void PlannerForRobot::execute_intent(const RobotIntent& intent) {
         auto trajectory = safe_plan_for_robot(plan_request);
         trajectory_topic_->publish(rj_convert::convert_to_ros(trajectory));
 
+        if (intent.dribbler_mode != RobotIntent::DribblerMode::NEUTRAL) {
+            trajectory.dribbler_speed =
+                (intent.dribbler_mode == RobotIntent::DribblerMode::ON) ? 255.0 : 0.0;
+        }
+
+        if (intent.trigger_mode != RobotIntent::TriggerMode::AT_END) {
```
jacksherling: no need for this if statement

soccer/src/soccer/radio/packet_convert.cpp
```diff
@@ -172,8 +172,10 @@ void to_rtp(const RobotIntent& intent, const MotionSetpoint& setpoint, int shell
         static_cast<int16_t>(setpoint.yvelocity * rtp::ControlMessage::VELOCITY_SCALE_FACTOR);
     rtp_message->body_w =
         static_cast<int16_t>(setpoint.avelocity * rtp::ControlMessage::VELOCITY_SCALE_FACTOR);
-    rtp_message->dribbler_speed =
-        std::clamp<uint16_t>(static_cast<uint16_t>(intent.dribbler_speed * kMaxDribble), 0, 255);
+    rtp_message->dribbler_speed = std::clamp<uint16_t>(
+        static_cast<uint16_t>((intent.dribbler_mode == RobotIntent::DribblerMode::ON ? 1.0 : 0.0) *
```
jacksherling: `kMaxDribble` or 0

jacksherling (changes requested):

soccer/src/soccer/planning/planner_for_robot.cpp
```diff
@@ -69,12 +69,40 @@ void PlannerForRobot::execute_intent(const RobotIntent& intent) {
         auto trajectory = safe_plan_for_robot(plan_request);
         trajectory_topic_->publish(rj_convert::convert_to_ros(trajectory));
 
+        if (intent.dribbler_mode != RobotIntent::DribblerMode::NEUTRAL) {
+            trajectory.dribbler_speed =
+                (intent.dribbler_mode == RobotIntent::DribblerMode::ON) ? 255.0 : 0.0;
```
rishiso, line 74: Trajectories have a default dribbler speed of 0 I believe, so can we just do a simple if statement setting dribbler speed to 255 when dribbler mode is ON

soccer/src/soccer/strategy/agent/position/offense.cpp
```diff
@@ -267,7 +267,6 @@ std::optional<RobotIntent> Offense::state_to_task(RobotIntent intent) {
 
             auto collect_cmd = planning::MotionCommand{"collect"};
             intent.motion_command = collect_cmd;
-            intent.dribbler_speed = 255.0;
```
rishiso, line 270: Why do we not set dribbler mode here?

jacksherling (approved):

soccer/src/soccer/radio/packet_convert.cpp
```diff
@@ -196,8 +198,8 @@ void to_rtp(const RobotIntent& intent, const MotionSetpoint& setpoint, int shell
     }
```
rishiso, line 198: Should we add AT_END to this switch statement?

soccer/src/soccer/planning/planner_for_robot.cpp
```diff
@@ -69,12 +69,40 @@ void PlannerForRobot::execute_intent(const RobotIntent& intent) {
...
+            case RobotIntent::TriggerMode::STAND_DOWN:
+                trajectory.trigger_mode = planning::Trajectory::TriggerMode::STAND_DOWN;
+                break;
+            case RobotIntent::TriggerMode::IMMEDIATE:
+                trajectory.trigger_mode = planning::Trajectory::TriggerMode::IMMEDIATE;
+                break;
+            case RobotIntent::TriggerMode::ON_BREAK_BEAM:
+                trajectory.trigger_mode = planning::Trajectory::TriggerMode::ON_BREAK_BEAM;
+                break;
+            default:
```
rishiso, line 87: Why do we not include AT_END?

sanatd33: we only want to directly copy the intent's trigger mode into the trajectory if it is one of the "strategy-dominated ones" (not at end). At end is like an internal trigger mode which just tells the planner to add a trajectory at the end of it to kick the ball.

soccer/src/soccer/radio/packet_convert.cpp
sanatd33: we dont want to use at end here becaseu at end isn't supported by the radio, its just an internal trigger mode for planners. the only thing we can actually send to radio is stand down, immediate, and on break beam. those are also the only ones supported by trajectory's trigger mode

soccer/src/soccer/planning/planner/rotate_path_planner.cpp
```diff
@@ -19,17 +19,29 @@ namespace planning {
...
+        case PIVOT:
+            return pivot(request);
+        case KICK:
+            return kick(request);
+    }
+    return {};
 }
 
-bool RotatePathPlanner::is_done() const {
+void RotatePathPlanner::update_state() {
```
rishiso, line 32: Not related to your PR, but it seems that a lot of the planners use states, so it would be nice to create some standardization around these state machines. Like maybe modular methods etc @jacksherling

soccer/src/soccer/planning/planner/rotate_path_planner.cpp
```diff
@@ -19,17 +19,29 @@ namespace planning {
...
     }
-    return abs(cached_angle_change_.value()) <
-           degrees_to_radians(static_cast<float>(kIsDoneAngleChangeThresh));
+    current_state_ = abs(cached_angle_change_.value()) <
+                             degrees_to_radians(static_cast<float>(kIsDoneAngleChangeThresh))
+                         ? KICK
+                         : PIVOT;
 }
 
+bool RotatePathPlanner::is_done() const { return current_state_ == KICK; }
```
rishiso: Are you sure about this is_done()? I am confused how being in the kick state alone indicates this

config/sim_params.yaml
```diff
@@ -47,7 +47,7 @@
   ros__parameters:
     collect:
       approach_accel_scale: 0.7
-      approach_dist_target: 0.04
+      approach_dist_target: 0.5
```
sid-parikh: what does this do?

rj_msgs/msg/RobotIntent.msg
```diff
@@ -6,6 +6,11 @@ uint8 SHOOT_MODE_CHIP = 1
 uint8 TRIGGER_MODE_STAND_DOWN = 0
 uint8 TRIGGER_MODE_IMMEDIATE = 1
 uint8 TRIGGER_MODE_ON_BREAK_BEAM = 2
+uint8 TRIGGER_MODE_AT_END = 3
+
+uint8 DRIBBLER_MODE_OFF = 0
+uint8 DRIBBLER_MODE_ON = 1
```
sid-parikh, line 19: can we add some documentation on what this stuff means? i assume "neutral" means let the planner decide. perhaps "Default" or "unspecified" would be better names? but either way, i'd like comments explaining.

soccer/src/soccer/logger.cpp
```diff
@@ -198,14 +198,15 @@ std::shared_ptr<Packet::LogFrame> Logger::create_log_frame(Context* context) {
...
+        const auto& trajectory = context->trajectories.at(shell);
         const auto& intent = context->robot_intents.at(shell);
         const auto& setpoint = context->motion_setpoints.at(shell);
 
         RadioRx* rx = log_frame->add_radio_rx();
         ConvertRx::status_to_proto(status, rx);
 
         Packet::Robot* tx = log_frame->mutable_radio_tx()->add_robots();
-        ConvertTx::to_proto(intent, setpoint, shell, tx);
+        ConvertTx::to_proto(trajectory, intent, setpoint, shell, tx);
```
sid-parikh: to get rid of this warning add an explicit `static_cast`

soccer/src/soccer/planning/planner/collect_path_planner.cpp
```diff
@@ -663,8 +664,12 @@ Trajectory CollectPathPlanner::fine_approach(
                     start_instant.position() +
                         Point::direction(AngleFns::face_point(ball.position)(
                             start_instant.linear_motion(), start_instant.heading(), nullptr))));
+        plan_request.debug_drawer->draw_text(
+            "Fine Approach", start_instant.position() + Point(.1, .1), QColor(255, 255, 255));
```
sid-parikh, line 668: neat

soccer/src/soccer/robot_intent.hpp
```diff
@@ -13,7 +13,8 @@ struct RobotIntent {
 
     using Msg = rj_msgs::msg::RobotIntent;
     enum ShootMode { KICK, CHIP };
-    enum TriggerMode { STAND_DOWN, IMMEDIATE, ON_BREAK_BEAM };
+    enum TriggerMode { STAND_DOWN, IMMEDIATE, ON_BREAK_BEAM, AT_END };
```
sid-parikh:
```suggestion
    enum TriggerMode { STAND_DOWN = 0, IMMEDIATE, ON_BREAK_BEAM, AT_END };
```
nit: technically you're not guaranteed it will start at 0

soccer/src/soccer/robot_intent.hpp
```diff
@@ -13,7 +13,8 @@ struct RobotIntent {
...
+    enum DribblerMode { OFF, ON, NEUTRAL };
```
sid-parikh: same

soccer/src/soccer/strategy/agent/position/solo_offense.cpp
```diff
@@ -101,15 +101,16 @@ std::optional<RobotIntent> SoloOffense::state_to_task(RobotIntent intent) {
...
         }
         case ROTATE: {
             planning::LinearMotionInstant target{calculate_best_shot()};
             auto pivot_cmd =
                 planning::MotionCommand{"rotate", target, planning::FaceTarget{}, false};
             intent.motion_command = pivot_cmd;
-            intent.dribbler_speed = 255;
+            intent.dribbler_mode = RobotIntent::DribblerMode::ON;
+            intent.trigger_mode = RobotIntent::TriggerMode::AT_END;
+            intent.kick_speed = 4.0;
```
sid-parikh, line 113: curious: what is kick speed measured in?

soccer/src/soccer/planning/planner_for_robot.cpp
```diff
@@ -302,7 +336,11 @@ bool PlannerForRobot::is_done() const {
         return false;
     }
 
-    return current_path_planner_->is_done();
+    if (current_path_planner_->is_done()) {
```
sid-parikh, line 339: crazy to me that we missed this. did you notice any differences by putting it in?

sid-parikh (commented):
looks solid. mostly have clarifying questions. will test soon

soccer/src/soccer/planning/planner/rotate_path_planner.cpp
sanatd33: My idea was that because the kick state trajectory is either empty (if we don't kick) or contains just one action, it will be very fast. So, once we're in the kick state, because the plan is always called before is done (i think), by the time it has gotten to the is done it should be done kicking (if it wants to). It seems to be working pretty well, but i'm open to changing this to like some cached bool that represents (we want to kick and we have kicked) or (we don't want to kick)

soccer/src/soccer/robot_intent.hpp
sanatd33: wait actually? thats crazy

soccer/src/soccer/strategy/agent/position/solo_offense.cpp
sanatd33: I think its m/s, rules specify the max kick speed as 6.5 m/s and usually in our stack we use values around 4-5, so it checks out

soccer/src/soccer/planning/planner_for_robot.cpp
sanatd33: Yeah I'm not sure how this worked previously for all the other planners which are stateful, but once I added it it seemed to reset the state properly

config/sim_params.yaml
sanatd33: Basically its the distance to switch from the coarse approach to the fine approach in collect, so its how far we want to be from the ball before we start slowing down and turning on the dribbler. 0.04 seemed really small for me because I didn't think we'd be able to slow down and turn on the dribbler in real life that fast. It doesn't really make a difference too much in sim, so this is probably something we would want to tune on the field for comp

sanatd33 (commented):
Somewhat unrelated but while testing this I noticed that our aiming to shoot is really bad, like we miss the goal by a lot pretty often, @jacksherling should we have someone look into that?

soccer/src/soccer/planning/planner/rotate_path_planner.cpp
sid-parikh: i think the assumption that it will be checked after calling plan() is OK. but maybe put it in some relatively prominent docstring that it should always be called like that. perhaps renaming KICK to END or something like that is better (since we don't always kick)

sid-parikh:
how can i set up sim to test this?

sanatd33:
> how can i set up sim to test this?

@sid-parikh Switch one of the robots to solo offense, it should go collect ball, pivot to goal and instantly shoot
also why doesn't github let me reply to like general comments

rishiso:
> Somewhat unrelated but while testing this I noticed that our aiming to shoot is really bad, like we miss the goal by a lot pretty often, @jacksherling should we have someone look into that?

I am working on offense related fixes right now, so I can fix this as well. Doesn't seem to be a consistent issue but have noticed that robot will shoot wide sometimes

soccer/src/soccer/planning/planner/rotate_path_planner.cpp
rishiso: Created a ticket

sid-parikh (approved):

rishiso (approved):

PR #2361, opened 2025-03-14
Investigated bootup crashes, found array bug in Waller

jvogt23:
## Description
The Waller get_task function previously crashed due to a vector index out of bounds. This PR adds additional in an attempt to avoid this. The bug was caused by undefined behavior, so I cannot be 100% certain that the changes I have made have removed the bug entirely, but I do know that they have made the software stack more reliable. After around 30 tests with the latest changes, I have not had a crash.

## Design Documents
[Link](link-to-design-doc)

## Steps to Test
1. Step 1: Run the current software stack until you see a crash on startup, and observe the behavior. Any time that it crashes, the initial message from the agent action client node should say something about a failed bounds check.
2. Step 2: Check out this branch and run the stack several times. You should not see a crash. If you do, contact @jvogt23 for further testing. You can also run a debugger on the agent action client node to get a backtrace if you need to.
3. Step 3: You can try adding these lines after line 58 of waller.cpp (the end of the first block that contains if (target_point.x() < robot_pos.x())):
bool parent_id_in_bounds = parent_id >= 0 && parent_id < 6;
SPDLOG_INFO("Parent ID In Bounds: {}", parent_id_in_bounds);
These lines should never produce a "false" in the logs. If they ever do, contact @jvogt23. Most likely, if you see a false, the stack will crash soon after.

**Expected result:**
The stack should not be crashing on startup (or at least with significantly less frequency and hopefully not due to the same bug).

## Key Files to Review
Group 1
 * File 1 waller.cpp

## Review Checklist

- [ ] Step 1
- [ ] Step 2

### DIFF
--- a/soccer/src/soccer/strategy/agent/position/waller.cpp
+++ b/soccer/src/soccer/strategy/agent/position/waller.cpp
@@ -47,11 +47,11 @@ std::optional<RobotIntent> Waller::get_task(RobotIntent intent, const WorldState
 
     if (abs(robot_pos.dist_to(goal_pos) - min_wall_rad) < kRobotRadius &&
         robot_pos.dist_to(target_point) > kRobotRadius) {
-        uint8_t parent_id;
+        uint8_t parent_id = robot_id; // Assigning a value to avoid any undefined behavior; will be changed
 
-        if (target_point.x() < robot_pos.x() && waller_pos_ != 1) {
+        if (target_point.x() < robot_pos.x() && waller_pos_ > 1) {
             parent_id = walling_robots_[waller_pos_ - 2];
-        } else if (target_point.x() >= robot_pos.x() && waller_pos_ != num_wallers) {
+        } else if (target_point.x() >= robot_pos.x() && waller_pos_ < num_wallers) {
             parent_id = walling_robots_[waller_pos_];
         } else {
             parent_id = robot_id;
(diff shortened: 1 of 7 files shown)


soccer/src/soccer/strategy/agent/position/waller.cpp
```diff
@@ -47,12 +47,16 @@ std::optional<RobotIntent> Waller::get_task(RobotIntent intent, const WorldState
 
     if (abs(robot_pos.dist_to(goal_pos) - min_wall_rad) < kRobotRadius &&
         robot_pos.dist_to(target_point) > kRobotRadius) {
-        uint8_t parent_id;
+        uint8_t parent_id =
```
sid-parikh, line 50: mind following through with your comment and taking this back out?
if you don't like the declaration and then assignment in the if-else (and there are good arguments against it)
one nice c++ pattern is an immediately invoked lambda: 

something like this:
```
uint8_t robot_id = [&](){
   if (...) return walling_robots_[waller_pos_ - 2];
   else return robot_id;
}();
```

sid-parikh: this forces the programmer to always return something so it's guaranteed robot_id will always have a value, but it can be based on a complex conditional.

sid-parikh: you can also just leave it with the previous pattern tho

sid-parikh (commented):
looks great thanks for finding this!

sid-parikh (approved):
LGTM. nice find and fix James!

sanatd33 (approved):

PR #2401, opened 2025-08-20
Build Stack with Colcon

N8BWert:
## Description
This PR refactors our build system to use colcon as the new build system.  It also rearranges some of the files to fit more into the typical ROS 2 workspace format.  Ideally, this PR shouldn't break any of the existing functionality as it shouldn't delete any files and, instead, move them around.

### Build the Stack
1. run `colcon build`
* Sometimes you might want to add the `--parallel-workers` argument because colcon will use all of your CPU if you let it

**Expected result:** Stack should build

### Run Sim / Real
2. The make commands should still work so you can run `make run-sim` or `make run-real`.
**Expected result:** Stack should run as normal

## Review Checklist


src/rj_planning/src/planner_for_robot.cpp
rishiso, line 1: Glad that you removed duplicate imports from several files

src/rj_strategy/include/rj_strategy/agent/position/smartidling.hpp
```diff
@@ -1,6 +1,8 @@
+#pragma once
```
rishiso, line 1: Good catch

sanatd33 (approved):
works on shop computer

Squid5678 (approved):
works on my vm

PR #2414, opened 2025-09-17
Changes from 2025 comp to SoloOffense

rhkoulen:
## TODO
- Someone more savvy than me needs to update the PID parameters. Sid wrote several pieces of code that allowed us to serve individual kP, kI, kD, W (rotational and translational) parameters to each robot. As it stands, we can only tune one robot and all robots share those parameters, which is unrealistic. However, I have no idea how to do this. Sid had some external program serving the parameters on the base station? I know that they were saved onto his local machine, and I know there were changes to motion_control.hpp and motion_control.cpp, but these changes have the potential to break all movement, and I can't figure out how to make this happen.
- Rocio and Cameron should comment on any other files that they may have changed, whose changes are worth merging to ros2.

## Description
This is a partial import of the changes made on comp-2025 that are worth keeping.

During comp 2025, the biggest change we made on the software side was to that of SoloOffense.
Due to technical issues, we really only had 1 well-functioning robot, and that meant SoloOffense was the strategy we pursued improvements to the most. Most of the time was spent chasing tiny bugs, and so we ended up scrapping most complicated planning and just rammed into the ball over and over. This behavior is the minimum viable strategy for a soccer team. Aalind confirmed that I should update the behavior of SoloOffense to this MVP, since ideally we'll never use SoloOffense again anyways. It can act as a simple debug position.

There are serious problems with Defense, and the subposition Waller. There are at least 3 existing communication deadlocks all of which happened in match. This includes a Waller failing to broadcast leave wall requests upon promotion to FreeKicker, a Waller failing to broadcast join requests and joining regardless, and one that I forgot about. Regardless, our motion planning here is terrible. Instead of merging in a bunch of terrible patches to Defense that ruined a bunch of functionality, I plan to just rewrite Defense after this task.

Similarly, @CameronLyon put a significant amount of work into the path planner. He could speak to this more, but again, I'm not going to include a bunch of patches. Cameron is actively working on improvements to the path planner on another branch.

We largely left other positions alone, as they functioned fine (I believe @rociopv06 may have made some changes to Goalie and PenaltyKicker). We also hard-coded a bunch of content in RobotFactoryPosition, but that needs to be changed for every test run anyways, so I see no need to patch in those details.

## Review Checklist
-

### DIFF
--- a/src/rj_strategy/include/rj_strategy/agent/position/solo_offense.hpp
+++ b/src/rj_strategy/include/rj_strategy/agent/position/solo_offense.hpp
@@ -31,34 +31,49 @@ public:
     std::string get_current_state() override;
 
 private:
+
+    // State space.
+    enum State { DEFAULT, TO_BALL, KICK };
+    State current_state_ = DEFAULT;
+    static constexpr std::string_view state_to_name(State s) {
+        switch (s) {
+            case DEFAULT:
+                return "DEFAULT";
+            case TO_BALL:
+                return "TO_BALL";
+            case KICK:
+                return "KICK";
+        }
+        return "NULL";
+    }
+
     /**
      * @brief Overriden from Position. Calls next_state and then state_to_task on each tick.
      */
     std::optional<RobotIntent> derived_get_task(RobotIntent intent) override;
-
-    enum State { TO_BALL, KICK, MARKER, ROTATE };
-
-    State current_state_ = TO_BALL;
-
-    rj_geometry::Point target_;
-
-    int marking_id_;
-
-    bool kick_ = false;
-    int counter_ = 0;
-
-    /**
-     * @return what the state should be right now. called on each get_task tick
-     */
-    State next_state();
-
-    /**
-     * @return the task to execute. called on each get_task tick AFTER next_state()
-     */
+    SoloOffense::State next_state();
     std::optional<RobotIntent> state_to_task(RobotIntent intent);
+    
+    // Timeout forcing.
+    RJ::Time last_time_;
+    void reset_timeout() { last_time_ = RJ::now(); }
+    bool timed_out() const {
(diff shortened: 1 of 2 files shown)


CameronLyon (commented):
Took a look through your code, most of it seems good. Here are two things I noticed:
- Usually in other offense roles, the kick target is usually calculated based on the point furthest away from the goalie and it still going in. Although, the target here is always set to field_dimensions_.their_goal_loc(), causing the target to always hit the center. I see why you commented that out, and I agree for the time being. Just know that as soon as motion is working, we have to plug that target planning back in.
- SPDLOG_INFO running in a loop for the derived_get_task, which might flood the terminal depending on how often this is triggered. If you think that's not going to be an issue for future testing and you want to leave it, I'm fine with that as well. 

Overall I don't see any errors or critical failures, but since my laptop can't run a linux VM well, I'll test run this as soon as I can on Sunday and approve the pull.

rhkoulen:
Aalind moved the robot-wise PID parameters to a separate task, so this is just waiting on anything @rociopv06 would like to add (and then the style merge).

rhkoulen:
> Although, the target here is always set to field_dimensions_.their_goal_loc(), causing the target to always hit the center.

I propose that this should be intended behavior until our shot standard deviation is less than the width of the goal... It's just an EV maximizing choice we made at comp.

CameronLyon (approved):

src/rj_strategy/include/rj_strategy/agent/position/solo_offense.hpp
Squid5678, line 1: We've decided to keep our current way of doing offense. Hopefully we do not have to use solo offense for next comp.

Squid5678 (changes requested):

PR #2418, opened 2025-09-22
removed rqt section, updated directories

Squid5678:
removing rqt section for now since I already went over it + not working on docker image currently. also some directory updates I forgot to do because of colcon.


rishiso (approved):
LGTM

shourikb (approved):

PR #2419, opened 2025-09-22
double wall spacing

woeiurkl:
## Description
doubled the wall_spacing
Screenshot 2025-09-21 201258

### DIFF
--- a/install/setup.bash
+++ b/install/setup.bash
@@ -17,11 +17,6 @@ _colcon_prefix_chain_bash_source_script() {
   fi
 }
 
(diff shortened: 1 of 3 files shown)


install/setup.bash
Squid5678, line 1: Make sure you aren't committing unnecessary files (best way to do this is to commit on a file-by-file basis)

install/setup.zsh
Squid5678, line 1: Same thing here

Squid5678 (changes requested):

PR #2420, opened 2025-09-22
Refactor Shape class to be pure virtual

rishiso:
## Description
Refactoring the `Shape` class to be pure virtual. It was previously not possible due to some dependencies on Python bindings. Those bindings are gone, so we can make it pure virtual.

## Testing
Nothing broken

### DIFF
--- a/src/rj_geometry/include/rj_geometry/shape.hpp
+++ b/src/rj_geometry/include/rj_geometry/shape.hpp
@@ -9,15 +9,6 @@ class Segment;
 
 /**
  * The shape class provides the interface to all shapes that are subclasses.
- *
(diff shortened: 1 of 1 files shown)


src/rj_geometry/include/rj_geometry/shape.hpp
```diff
@@ -9,15 +9,6 @@ class Segment;
 
 /**
  * The shape class provides the interface to all shapes that are subclasses.
- *
- * TODO(Kevin): do we use boost python anymore? more likely this docstring
```
Squid5678, line 13: finally

Squid5678 (approved):
lgtm

jvogt23 (approved):
LGTM

PR #2424, opened 2025-09-28
Make Offense use Pivot Kick

sanatd33:
Change Offense to use pivot kick (rotate path planner) instead of path target and line kick.

This also eliminates the second stage of Passing and Shooting, as the rotate and kick occurs in one step.

Testing Steps:
- `colcon build`
- `. ./source.bash`
- `make run-sim`

### DIFF
--- a/src/rj_strategy/src/agent/position/solo_offense.cpp
+++ b/src/rj_strategy/src/agent/position/solo_offense.cpp
@@ -91,14 +91,6 @@ std::optional<RobotIntent> SoloOffense::state_to_task(RobotIntent intent) {
             return intent;
         }
         case TO_BALL: {
-            // rj_geometry::Point robotToBall =
-            //     (last_world_state_->ball.position -
-            //      last_world_state_->get_robot(true, robot_id_).pose.position());
-            // double slowDown = 1.0;
-            // double length = robotToBall.mag() - kRobotRadius * slowDown;
-            // robotToBall = robotToBall.normalized(length);
-            // planning::LinearMotionInstant target{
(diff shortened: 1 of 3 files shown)


Squid5678 (approved):
LGTM

src/rj_strategy/src/agent/position/solo_offense.cpp
```diff
@@ -91,14 +91,14 @@ std::optional<RobotIntent> SoloOffense::state_to_task(RobotIntent intent) {
...
-                 last_world_state_->get_robot(true, robot_id_).pose.position());
-            double slowDown = 1.0;
-            double length = robotToBall.mag() - kRobotRadius * slowDown;
-            robotToBall = robotToBall.normalized(length);
-            planning::LinearMotionInstant target{
-                last_world_state_->get_robot(true, robot_id_).pose.position() + robotToBall};
+            // rj_geometry::Point robotToBall =
+            //     (last_world_state_->ball.position -
+            //      last_world_state_->get_robot(true, robot_id_).pose.position());
+            // double slowDown = 1.0;
```
rishiso: I would prefer us to delete these lines rather than comment them out. This behavior should be handled by the collect command anyway

rishiso (changes requested):

src/rj_strategy/src/agent/position/solo_offense.cpp
sanatd33: fixed

rishiso (approved):
LGTM

PR #2428, opened 2025-10-13
Yuvraj dhadwal/robocup sw tutorial

yuvrajdhadwal:
image

i finished!!!!


yuvrajdhadwal:
@Squid5678 i really appreciated the publisher we made in the mini tutorial because it helped me understand a lot better the format for the tutorial. I also liked that we had to dig into the repo code to find answers to some questions, but I wish there was a little bit of guidance of what kinds of files to look at. Also it would be nice for every section to have like a tldr that you can look at so that if you leave for a week (or two for fall break) you can reread that and be caught up to speed because it became hard to find important information in really large instructions when returning to a problem after a week

src/rj_strategy/src/agent/position/runner.cpp
```diff
@@ -0,0 +1,37 @@
...
+
+    if (distance_to_point > 0.3) {
+        return states_[current_state_];
+    }
+    return states_[(current_state_ + 1) % 4];
+}
+
+std::optional<RobotIntent> Runner::state_to_task(RobotIntent intent) {
+    rj_geometry::Point target_point = corners_[current_state_];
+    rj_geometry::Point target_vel{1.0, 1.0};
```
sanatd33, line 27: So this `target_vel` is basically the velocity we want to end the motion at, so like 99.9% of the time it would just be (0, 0), as we want to end our motion stationary.

src/rj_radio/src/soccer_mom.cpp
```diff
@@ -0,0 +1,68 @@
...
+    {
+        publisher_ = this->create_publisher<std_msgs::msg::String>("/soccer_mom", 10);
+
+        team_color_sub_ = create_subscription<rj_msgs::msg::TeamColor>(
+        referee::topics::kTeamColorTopic, rclcpp::QoS(1).transient_local(),
+        [this](rj_msgs::msg::TeamColor::SharedPtr color) {  // NOLINT
+            last_is_blue_ = color->is_blue;
+        });
+        
+        timer_ = this->create_wall_timer(
```
sanatd33, line 40: this is an interesting way to do this but I wouldn't personally recommend doing this.

a better solution is just calling this callback inside the subscription callback
i.e.
```team_color_sub_ = create_subscription<rj_msgs::msg::TeamColor>(
        referee::topics::kTeamColorTopic, rclcpp::QoS(1).transient_local(),
        [this](rj_msgs::msg::TeamColor::SharedPtr color) {  // NOLINT
            std_msgs::msg::String message;
            message.data = (color->is_blue) ? "blueberries" : "bananas";
            publisher_->publish(message);
        });

File: src/rj_radio/src/soccer_mom.cpp [hunk 3]
```diff
@@ -0,0 +1,68 @@
...
+
+        RCLCPP_INFO(this->get_logger(), "Soccer Mom has been started.");
+    }
+
+private:
+    void timer_callback()
+    {
+        std_msgs::msg::String message;
+
+        message.data = (*last_is_blue_) ? "blueberries" : "bananas";
```
sanatd33, line 51: calling `*last_is_blue_` is potentially risky here if the optional isn't initialized. typically when we use optionals we're acknowledging that there is a chance the optional is empty so should probably check for that using `has_value()` or something

src/rj_radio/src/soccer_mom.cpp
```diff
@@ -0,0 +1,68 @@
...
+#include "rj_radio/radio.hpp"
+
+using namespace std::chrono_literals;
+
+class SoccerMom : public rclcpp::Node
+{
+public:
+    SoccerMom() : Node("soccer_mom")
+    {
+        publisher_ = this->create_publisher<std_msgs::msg::String>("/soccer_mom", 10);
```
sanatd33, line 32: should probably be 1 instead of 10 for the publisher qos history depth

src/rj_radio/CMakeLists.txt
```diff
@@ -37,6 +37,11 @@ add_executable(sim_radio_node
   src/radio.cpp
 )
 
+add_executable(soccer_mom_node
+  src/radio.cpp
```
sanatd33, line 41: little weird that this is in the radio package?

src/rj_strategy/include/rj_strategy/agent/position/runner.hpp
```diff
@@ -0,0 +1,50 @@
...
+    Runner(const Position& other);
+    Runner(Runner&& other) = default;
+    Runner& operator=(const Runner& other) = default;
+    Runner& operator=(Runner&& other) = default;
+
+private:
+    std::optional<RobotIntent> derived_get_task(RobotIntent intent) override;
+
+    // enum State { SIDE1, SIDE2, SIDE3, SIDE4 };
+    vector<int> states_ = {0, 1, 2, 3};
```
sanatd33, line 35: typically we don't use vectors for our states, we usually use enums. I see how using a vector can make it a little easier in this scenario (can just index into your points from your state) but using a vector/raw numbers to encode state makes it much harder to read, especially in the large positions that exist in our stack

src/rj_strategy/include/rj_strategy/agent/position/runner.hpp
```diff
@@ -0,0 +1,50 @@
...
+
+    rj_geometry::Point point1{1.0, 5.0};
+    rj_geometry::Point point2{2.0, 5.0};
+    rj_geometry::Point point3{2.0, 6.0};
+    rj_geometry::Point point4{1.0, 6.0};
+
+    vector<rj_geometry::Point> corners_ = {point1, point2, point3, point4};
```
sanatd33, line 42: minor thing but if there isn't a good reason why something should be a vector (if it doesn't change in size) i personally lean more towards using arrays

src/rj_radio/src/soccer_mom.cpp
```diff
@@ -0,0 +1,68 @@
...
+#include <boost/asio.hpp>
+#include <spdlog/spdlog.h>
+
+#include <rj_common/network.hpp>
+#include <rj_common/radio/packet_convert.hpp>
+#include <rj_common/time.hpp>
+#include <rj_msgs/msg/alive_robots.hpp>
+#include <rj_msgs/srv/sim_placement.hpp>
+#include <rj_param_utils/global_params.hpp>
+#include <rj_protos/ssl_simulation_control.pb.h>
```
sanatd33, line 18: try not to add `#include` files that aren't used

src/rj_radio/src/soccer_mom.cpp
```diff
@@ -0,0 +1,68 @@
...
+#include <rj_protos/ssl_simulation_robot_control.pb.h>
+#include <rj_protos/ssl_simulation_robot_feedback.pb.h>
+#include <rj_utils/logging.hpp>
+
+#include "rj_radio/radio.hpp"
+
+using namespace std::chrono_literals;
```
sanatd33, line 25: again kind of weird this being under `chrono_literals` lol

sanatd33 (commented):
Overall, good work! You should get @Squid5678 review on this too and then you can close this PR once you get his review too

src/rj_radio/CMakeLists.txt
Squid5678: i had him do this, colcon switch made the files dirs from the tutorial no longer exist and I just had him make it in a random place for now

src/rj_radio/src/soccer_mom.cpp
Squid5678: +1

Squid5678 (commented):
Nice work, thank you for all your feedback on this. Agreed with Sanat on the changes.

PR #2433, opened 2025-10-17
Modernize constants for C++ 17

rishiso:
## Description
Using modern practice for C++17 by inlining constants. Inlining them ensures that they are only defined once and shared among all the translation units, meaning reduced binary size.


sanatd33 (approved):
lgtm

Squid5678 (approved):
lgtm

PR #2440, opened 2025-10-28
Obstacle rewrite

rishiso:
## Motivation
The previous obstacle system had several issues:
- Use of dynamic obstacles (deprecated)
- Use of globally published obstacles (deprecated)
- Poor obstacle avoidance for moving obstacles (ie. enemy robots)
- No clear separation between obstacle cores (hard physical collisions) and padding zones (safety margins)
- Lack of extensibility for obstacle functionality

## Changes
Did a complete rewrite of the current obstacles system:
- New consolidated `Obstacle`/`ObstacleSet`
  - Updated all use cases accordingly
  - Better visualization functionality for obstacles
- Improved creation of moving obstacles via stadium shapes
- Removed deprecated features
- Optimized `EscapeObstaclesPathPlanner`
- General cleanup of obstacle-related code

[ClickUp card](https://app.clickup.com/t/86b6e8ne2)

## Testing
Did some basic testing and nothing seems broken. Going to do thorough testing on shop computer since Docker is lagging with all the drawing and stuff


Squid5678:
@rishiso what's the status of this PR?

rishiso:
Blocked thinking about some design issues @Squid5678. Going to try to finish during today's meeting

src/rj_planning/src/trajectory_utils.cpp
```diff
@@ -2,7 +2,7 @@
 
 namespace planning {
 
-bool trajectory_hits_static(const Trajectory& trajectory, const rj_geometry::ShapeSet& obstacles,
+bool trajectory_hits_static(const Trajectory& trajectory, const ObstacleSet& obstacles,
```
jvogt23: Can this be renamed to trajectory_hits_obstacle or are we still differentiating between obstacle types?

src/rj_planning/include/rj_planning/obstacle.hpp
```diff
@@ -1,29 +1,53 @@
...
+    virtual bool obstacle_near(rj_geometry::Point pt, float thresh) {
+        return obstacle->near_point(pt, thresh);
+    }
+
+    virtual bool padding_near(rj_geometry::Point pt, float thresh) {
+        return padding->near_point(pt, thresh);
+    }
+
+    // Convenience methods: hit() delegates to padding_hit() for compatibility
+    virtual bool hit(rj_geometry::Point pt) { return padding_hit(pt); }
```
jvogt23, line 40: If we consider the "obstacle" shape to be the obstacle itself and the padding to be an unsafe zone/an area where you are likely to hit the obstacle, would it be better to have hit() delegate to obstacle_hit()?

Or are we considering a "hit" to the obstacle to be a collision with either the obstacle or the unsafe zone?

src/rj_planning/include/rj_planning/obstacle.hpp
```diff
@@ -1,29 +1,53 @@
...
+    virtual bool hit(const rj_geometry::Segment& seg) { return padding_hit(seg); }
+
+    virtual std::shared_ptr<rj_geometry::Shape> get_obstacle() { return obstacle; }
+
+    virtual std::shared_ptr<rj_geometry::Shape> get_padding() { return padding; }
 
 protected:
     std::shared_ptr<rj_geometry::Shape> obstacle;
     std::shared_ptr<rj_geometry::Shape> padding;
-    std::shared_ptr<rj_geometry::Point> velocity;
```
jvogt23, line 25: Is an obstacle that scales by velocity something we plan to implement in a subclass, or are we getting rid of that idea altogether?

If neither of those are the case, I would think we would still need line 25.

src/rj_planning/include/rj_planning/plan_request.hpp
```diff
@@ -30,7 +29,6 @@ namespace planning {
 struct PlanRequest {
     PlanRequest(RobotInstant start, MotionCommand command,  // NOLINT
```
jvogt23, line 30: Do we still want PlanRequest to expect a ShapeSet for field_obstacles instead of an ObstacleSet?

EDIT: Is this a ShapeSet for compatibility with ROS messages?

src/rj_planning/include/rj_planning/trajectory_utils.hpp
```diff
@@ -21,25 +18,7 @@ namespace planning {
...
- * @param out_hit_obstacle A bounding circle for the obstacle that was hit.
- * @param out_hit_time     The time of the collision (output parameter).
- * @return                 Whether or not there is a collision.
- */
-bool trajectory_hits_dynamic(const Trajectory& trajectory,
-                           const std::vector<DynamicObstacle>& obstacles,
-                           RJ::Time start_time,
-                           rj_geometry::Circle* out_hit_obstacle,
-                           RJ::Time* out_hit_time);
+bool trajectory_hits_static(const Trajectory& trajectory, const ObstacleSet& obstacles,
```
jvogt23: May make sense to rename to "trajectory_hits_obstacle"

jvogt23 (changes requested):
Looks pretty good, just a few questions/comments

src/rj_planning/src/trajectory_utils.cpp
rishiso: renamed, we can keep it the same

src/rj_planning/include/rj_planning/obstacle.hpp
rishiso: I decided to make `hit()` delegate to `padding_hit()` since this is the typical use case for obstacle

src/rj_planning/include/rj_planning/obstacle.hpp
rishiso: We recreate obstacles every tick based on their current velocity, so velocity is only needed at obstacle creation time. Thus, I decided to remove it from the members. If we do need to store velocity for some reason, we can create a subclass later.

src/rj_planning/include/rj_planning/plan_request.hpp
rishiso: Yes it is mainly for ROS compatibility

src/rj_planning/include/rj_planning/trajectory_utils.hpp
rishiso: Fixed

jvogt23 (approved):
LGTM

shourikb (approved):
LGTM

PR #2445, opened 2025-11-02
Double Wall Spacing

karsh-s:
doubled wall spacing 
image


karsh-s:
/robocup-software/src/rj_msgs/msg/Trajectory.msg

Squid5678:
marking as done since we went over tutorial sol

PR #2448, opened 2025-11-03
RoboCup Tutorial - Runner + Soccer Mom

LightningBoltz21:
Screenshot 2025-11-02 at 8 14 15 PM
Screenshot 2025-11-02 at 8 14 46 PM


install/setup.bash
Squid5678, line 1: In general, it's always a good practice to try and avoid committing files that aren't necessary to the commit. Even though this looks like the file is unchanged, it's best to remove it from being committed.

src/rj_radio/src/soccer_mom.cpp
```diff
@@ -0,0 +1,32 @@
+#include "rj_radio/soccer_mom.hpp"
+
+namespace tutorial {
+
+SoccerMom::SoccerMom() : Node("soccer_mom") {
+    team_color_sub_ = create_subscription<rj_msgs::msg::TeamColor>(
+        referee::topics::kTeamColorTopic, rclcpp::QoS(1).transient_local(),
+        [this](rj_msgs::msg::TeamColor::SharedPtr msg) { team_color_callback(msg); });
```
Squid5678, line 8: Since this callback is fairly simple, you may consider just keeping all the logic for the callback as you make this subscription.

src/rj_strategy/include/rj_strategy/agent/position/runner.hpp
```diff
@@ -0,0 +1,28 @@
...
+
+private:
+    std::optional<RobotIntent> derived_get_task(RobotIntent intent) override;
+
+    enum State { RUNNING };
+
+    State update_state();
+    std::optional<RobotIntent> state_to_task(RobotIntent intent);
+
+    State current_state_ = State::RUNNING;
```
Squid5678, line 24: Let's try to rethink having a generalized state for running.

Within this Runner position, we know the Runner is going to do 1 of 4 things:
- Moving from top left to top right
- Moving from top right to bottom right
- Moving from bottom right to bottom left
- Moving from bottom left to top left

Within this position, it is useful to know the current state of the runner. You'll notice with the code for some of our other positions that our states are typically defined as enums. You might consider defining your states as follows:
- TOP_LEFT
- TOP_RIGHT
- BOTTOM_RIGHT
- BOTTOM_LEFT

Then, we can accordingly update these states in `update_state` and use these state definitions in order to determine which motion command to run.

src/rj_strategy/include/rj_strategy/agent/position/runner.hpp
```diff
@@ -0,0 +1,28 @@
...
+    int current_vertex_index_ = 0;
```
Squid5678, line 25: The above would also eliminate the need to have these additional variables.

src/rj_strategy/src/agent/position/runner.cpp
```diff
@@ -0,0 +1,51 @@
...
+            return RUNNING;
+        }
+    }
+    return RUNNING;
+}
+
+std::optional<RobotIntent> Runner::state_to_task(RobotIntent intent) {
+    rj_geometry::Point center = field_dimensions_.center_field_loc();
+
+    rj_geometry::Point vertices[kNumVertices] = {
```
Squid5678, line 32: You can also explicitly define the points based on the location of your cursor on the field simulation. This would allow you to avoid needing to reference the field dimensions.

src/rj_strategy/src/agent/position/runner.cpp
```diff
@@ -0,0 +1,51 @@
...
+        rj_geometry::Point{center.x() + 2.0, center.y() + 1.5},
+        rj_geometry::Point{center.x() - 2.0, center.y() + 1.5},
+        rj_geometry::Point{center.x() - 2.0, center.y() - 1.5},
+        rj_geometry::Point{center.x() + 2.0, center.y() - 1.5}};
+
+    switch (current_state_) {
```
Squid5678, line 38: Since your position will also be in this RUNNING state, a switch statement would be redundant in this case (unless, of course, you define additional states)

Squid5678 (commented):
Nice work getting this done! Definitely be sure to go over the comments regarding code structure, as they will be helpful going forward when you are working on tasks.

PR #2450, opened 2025-11-05
Improved free kick routine

eziCode:
## Description
Updated free kick logic to shoot at position farthest from the goalie

### Test Case 1
1. Open the simulator
2. Set up a free kick
3. Watch ball go in corner instead of straight down the middle

## Key Files to Review
Group 1
 * /robocup-software/src/rj_strategy/src/agent/position/free_kicker.cpp


FarkasJoseph:
@eziCode May I ask what prompted the move from using dot products / tan calculations and to using cross products? Based on my quick glance the outcome is the same, so I'm just curious if that was a design choice or if there was a specific bug you were targetting.

eziCode:
I tried what was there, and it didn't seem to work correctly. It kept
shooting into the first position every time, and the loop was not updating
the shot correctly. So I just changed it to use atan and cross product.

FarkasJoseph:
image

Looks great!

FarkasJoseph:
image


wait i found an edge case

FarkasJoseph:
looking to add goal post/edge detection into the model. looks like this is missing from offense as well, might add it there if this is a success

FarkasJoseph:
i've been told we just don't have the physical feasibility to launch from such tight angles, so it doesn't matter

FarkasJoseph:
i'm going to slightly increase our offsets from the goal sides since we are leaving essentially zero tolerance for error

install/setup.bash
Squid5678, line 1: in the future, check you stash to see the commits that you're going to push. something like this does not need to be committed.

src/rj_strategy/src/agent/position/offense.cpp
```diff
@@ -15,7 +15,7 @@ std::optional<RobotIntent> Offense::derived_get_task(RobotIntent intent) {
     if (current_state_ != new_state) {
         reset_timeout();
 
-        SPDLOG_INFO("Robot {}: now {}", robot_id_, state_to_name(current_state_));
+        // SPDLOG_INFO("Robot {}: now {}", robot_id_, state_to_name(current_state_));
```
Squid5678: let's get rid of this comment

src/rj_strategy/src/agent/position/free_kicker.cpp
```diff
@@ -11,11 +11,51 @@ std::optional<RobotIntent> FreeKicker::derived_get_task(RobotIntent intent) {
...
-    planning::LinearMotionInstant target{goal_corner};
+    for (const RobotState& enemy : their_robots) {
+        // Get position of their goalie
+        if (this->field_dimensions_.their_defense_area().hit(enemy.pose.position())) {
+            enemy_goalie_location = enemy.pose.position();
+            break;
+        }
+    }
+
+    constexpr double BALL_WIDTH = 0.025;
```
Squid5678: is this a measured quantity? wondering if there's a constant that already exists for this

src/rj_strategy/src/agent/position/free_kicker.cpp
```diff
@@ -11,11 +11,51 @@ std::optional<RobotIntent> FreeKicker::derived_get_task(RobotIntent intent) {
...
+    constexpr double BALL_WIDTH_OFFSET = BALL_WIDTH * 3;
```
Squid5678: why times 3?

src/rj_strategy/src/agent/position/free_kicker.cpp
```diff
@@ -11,11 +11,51 @@ std::optional<RobotIntent> FreeKicker::derived_get_task(RobotIntent intent) {
...
+        rj_geometry::Point((this->field_dimensions_.goal_width() / 2.0) - BALL_WIDTH_OFFSET, 0.0);
+
+    rj_geometry::Point const left_goal_post =
+        this->field_dimensions_.their_goal_loc() -
+        rj_geometry::Point((this->field_dimensions_.goal_width() / 2.0) - BALL_WIDTH_OFFSET, 0.0);
+    rj_geometry::Point best_shot = right_goal_post;
+    double best_distance = -1.0;
+    rj_geometry::Point ball_position = this->last_world_state_->ball.position;
+
+    int num_samples = 20;
```
Squid5678: is there a way to do this that doesn't require ``num_samples``? Makes this computation a bit inefficient.

Squid5678 (changes requested):

src/rj_strategy/src/agent/position/offense.cpp
FarkasJoseph: Yup! Sorry about that

src/rj_strategy/src/agent/position/free_kicker.cpp
FarkasJoseph: I am not sure exactly why / from what this was declared, however, I see kBallRadius is 0.0215 in rj_constants. I will replace with this value

FarkasJoseph:
Updates:

Replaced the sampling method with a simple check to shoot at either the left or right corner of the goal. I expanded the padding on the sides of the goal to 10*ball width, which makes straight shots (or really anything +/- 45deg of a straight shot) work quite well. However, it does exceptionally poorly with tight angles (i.e. corner kicks). Sanat recommends we look at passing logic next (if angle too extreme, pass to a partner instead), but that's likely outside the scope of this specific PR.

PR #2461, opened 2025-11-17
Feature/latency benchmarking

yuvrajdhadwal:
## Description
Created a pluggable RAII Timer object that starts timer at construction and ends timer at object destruction. Will show up in session csv file for future latency analysis.

## Steps to Test

1. `#include <rj_benchmarking/timer.hpp>`
2. `Timer t("timer_label", robot_id);`


Squid5678:
hmmmmmm

launch/soccer.launch.py
```diff
@@ -238,5 +245,10 @@ def generate_launch_description():
                 parameters=[param_config_filepath],
                 on_exit=Shutdown(),
             )
+            # Node(
+            #     package="rj_benchmarking",
+            #     executable="benchmarking_node",
```
rishiso: Remove comments

src/rj_benchmarking/include/rj_benchmarking/registry.hpp
```diff
@@ -0,0 +1,35 @@
...
+private:
+    void topic_callback(const rj_msgs::msg::Latency &msg);
+    void dump();
+
+    std::string get_curr_datetime();
+    
+    // registry[robot_id][label] -> latency sampling
+    std::array<std::unordered_map<std::string, std::vector<uint64_t>>, 6> registry_{};
+    rclcpp::Subscription<rj_msgs::msg::Latency>::SharedPtr subscription_{};
+    int max_rows_{};
```
rishiso:
```suggestion
    size_t max_rows_{};
```

src/rj_benchmarking/src/registry.cpp
```diff
@@ -0,0 +1,94 @@
+#include "rj_benchmarking/registry.hpp"
+
+Registry::Registry() : rclcpp::Node{"rj_benchmarking"}
+{
+    // SPDLOG_INFO("TESTING: Registry Built");
```
Squid5678: delete out commented logs

src/rj_benchmarking/src/registry.cpp
```diff
@@ -0,0 +1,94 @@
...
+    subscription_ = this->create_subscription<rj_msgs::msg::Latency>(
+            "/registry", 100, std::bind(&Registry::topic_callback, this,
+                std::placeholders::_1));
+}
+
+Registry::~Registry()
+{
+    dump();
+    // SPDLOG_INFO("TESTING: Registry Destroyed");
```
Squid5678: same as above

src/rj_benchmarking/src/registry.cpp
```diff
@@ -0,0 +1,94 @@
...
+void Registry::topic_callback(const rj_msgs::msg::Latency &msg)
+{
+    registry_.at(msg.robot_id)[msg.label].push_back(msg.duration_ns);
+    max_rows_ = std::max(max_rows_, static_cast<int>(registry_.at(msg.robot_id)[msg.label].size()));
+}
+
+
+void Registry::dump()
+{
+    // SPDLOG_INFO("TESTING: Dump Called");
```
Squid5678: same as above

src/rj_benchmarking/src/registry.cpp
```diff
@@ -0,0 +1,94 @@
...
+        //     for (uint64_t e : p.second)
+        //     {
+        //         output_file << e << ", ";
+        //     }
+        //     output_file << "AVG: " << std::accumulate(p.second.begin(), p.second.end(), 0)
+        //                                                                         / p.second.size();
+        //     output_file << '\n';
+        // }
+    }
+    // output_file.close();
```
Squid5678: remove these

src/rj_benchmarking/include/rj_benchmarking/registry_publisher.hpp
```diff
@@ -0,0 +1,38 @@
...
+    {
+        if (instance == nullptr)
+        {
+            instance = new RegistryPublisher();
+        }
+
+        return instance;
+    }
+
+    void publish(std::string label, std::int8_t robot_id, uint64_t time);
```
rishiso: `robot_id` should be `uint8_t` to be consistent with rest of codebase

src/rj_benchmarking/src/registry.cpp
```diff
@@ -0,0 +1,94 @@
...
+
+    // std::ofstream output_file;
+    // output_file.open(path_);
+    for (int i = 0; i < 6; i++)
+    {
+        std::stringstream ss;
+        ss << "/robot_" << i << ".csv";
+        std::ofstream robot_csv{ base_path + ss.str() };
+
+        for (int row = -1; row < max_rows_; row++)
```
Squid5678: can we separate the logic for the headers of the rows and the actual data? the row = -1 here is a bit confusing and might be best handled separately.

src/rj_benchmarking/src/registry_publisher.cpp
```diff
@@ -0,0 +1,20 @@
+#include "rj_benchmarking/registry_publisher.hpp"
+
+RegistryPublisher* RegistryPublisher::instance = nullptr;
+
+RegistryPublisher::RegistryPublisher()
+{
+    // printf("Testing");
```
Squid5678: remove this comment

src/rj_benchmarking/src/registry.cpp
rishiso: Delete comment

src/rj_benchmarking/include/rj_benchmarking/registry_publisher.hpp
```diff
@@ -0,0 +1,38 @@
...
+
+#include <rj_msgs/msg/latency.hpp>
+
+#include <rclcpp/rclcpp.hpp>
+
+class RegistryPublisher
+{
+public:
+    // Singleton pattern
+    static RegistryPublisher* getInstance()
```
Squid5678: Not sure if this is something you want to try to add in, but I believe there is a more modern way of using the Singleton pattern in C++:

```
static RegistryPublisher& getInstance() {
    static RegistryPublisher instance;
    return instance;
}
```

src/rj_benchmarking/src/registry.cpp
```diff
@@ -0,0 +1,94 @@
...
+            robot_csv << '\n';
+        }
+
+        
+        // output_file << "Robot " << i << '\n';
+        // for (auto& p : registry_[i])
+        // {
+        //     output_file << p.first << "     ";
+        //     for (uint64_t e : p.second)
+        //     {
```
rishiso: Remove

src/rj_benchmarking/src/timer.cpp
```diff
@@ -0,0 +1,17 @@
+#include "rj_benchmarking/timer.hpp"
+
+Timer::Timer(std::string label, int8_t robot_id) : label_(label),
+    robot_id_(robot_id), start_(std::chrono::steady_clock::now())
+{
+    // SPDLOG_INFO("Testing: Timer Created " + label);
```
Squid5678: remove this

src/rj_benchmarking/src/registry_publisher.cpp
```diff
@@ -0,0 +1,20 @@
...
+}
+
+void RegistryPublisher::publish(std::string label, std::int8_t robot_id, uint64_t time)
+{
+    // idk what type this should be im just following tutorial code
```
Squid5678: probably don't want to add this in as a comment

src/rj_benchmarking/src/registry_publisher.cpp
rishiso: set constructor to default

src/rj_planning/include/rj_planning/planners/rotate_path_planner.hpp
```diff
@@ -63,4 +63,4 @@ class RotatePathPlanner : public PathPlanner {
 
     double isDoneAngleChangeThresh{1.0};
 };
-}  // namespace planning
```
Squid5678, line 66: not sure why these are getting pushed, but in the future just double check what you're committing

Squid5678 (changes requested):
couple of cleanup things, otherwise looks good

src/rj_ui/package.xml
```diff
@@ -30,6 +30,7 @@
   <depend>rj_referee</depend>
   <depend>rj_config_client</depend>
   <depend>rj_topic_utils</depend>
+  <!-- <depend>rj_benchmarking</depend> -->
```
rishiso: remove

src/rj_benchmarking/src/timer.cpp
rishiso: remove

src/rj_geometry/CMakeLists.txt
```diff
@@ -20,6 +20,7 @@ find_package(rj_convert REQUIRED)
 find_package(rosidl_default_generators REQUIRED)
 find_package(std_msgs REQUIRED)
 find_package(rj_geometry_msgs REQUIRED)
+find_package(rj_benchmarking REQUIRED)
```
rishiso: I think you should remove these links to rj_benchmarking from CMakeLists where the package is not used. We can add these add needed when we use benchmarking

src/rj_benchmarking/include/rj_benchmarking/registry_publisher.hpp
rishiso: make this a shared ptr

src/rj_benchmarking/src/registry.cpp
```diff
@@ -0,0 +1,94 @@
...
+        //                                                                         / p.second.size();
+        //     output_file << '\n';
+        // }
+    }
+    // output_file.close();
+}
+
+std::string Registry::get_curr_datetime()
+{
+    std::time_t time = std::time({});
```
rishiso: Use steady clock and `std::format()`

src/rj_benchmarking/include/rj_benchmarking/registry.hpp
```diff
@@ -0,0 +1,35 @@
...
+    ~Registry();
+
+private:
+    void topic_callback(const rj_msgs::msg::Latency &msg);
+    void dump();
+
+    std::string get_curr_datetime();
+    
+    // registry[robot_id][label] -> latency sampling
+    std::array<std::unordered_map<std::string, std::vector<uint64_t>>, 6> registry_{};
```
rishiso: This should be `kNumShells` instead of 6

src/rj_benchmarking/src/registry.cpp
```diff
@@ -0,0 +1,94 @@
...
+}
+
+void Registry::topic_callback(const rj_msgs::msg::Latency &msg)
+{
+    registry_.at(msg.robot_id)[msg.label].push_back(msg.duration_ns);
+    max_rows_ = std::max(max_rows_, static_cast<int>(registry_.at(msg.robot_id)[msg.label].size()));
+}
+
+
+void Registry::dump()
```
rishiso: Remove `dump()` and move this into destructor. We don't use `dump()` elsewhere so no need to make a method

src/rj_benchmarking/src/registry.cpp
```diff
@@ -0,0 +1,94 @@
...
+    // std::ofstream output_file;
+    // output_file.open(path_);
+    for (int i = 0; i < 6; i++)
+    {
+        std::stringstream ss;
+        ss << "/robot_" << i << ".csv";
+        std::ofstream robot_csv{ base_path + ss.str() };
+
+        for (int row = -1; row < max_rows_; row++)
+        {
```
rishiso: Should we skip writing any rows if `max_rows_ == 0`?

src/rj_benchmarking/include/rj_benchmarking/registry_publisher.hpp
rishiso:
```suggestion
    void publish(const std::string& label, std::int8_t robot_id, uint64_t time);
```

src/rj_benchmarking/src/registry.cpp
```diff
@@ -0,0 +1,94 @@
...
+
+Registry::~Registry()
+{
+    dump();
+    // SPDLOG_INFO("TESTING: Registry Destroyed");
+}
+
+void Registry::topic_callback(const rj_msgs::msg::Latency &msg)
+{
+    registry_.at(msg.robot_id)[msg.label].push_back(msg.duration_ns);
```
rishiso: Use [] operator instead of `.at()`

src/rj_benchmarking/src/registry.cpp
```diff
@@ -0,0 +1,94 @@
...
+        ss << "/robot_" << i << ".csv";
+        std::ofstream robot_csv{ base_path + ss.str() };
+
+        for (int row = -1; row < max_rows_; row++)
+        {
+            if (row == -1)
+            {
+                for (std::pair<std::string, std::vector<uint64_t>> labels : registry_.at(i))
+                {
+                    robot_csv << labels.first << ',';
```
rishiso: Nit: Use structured bindings for the pair

src/rj_benchmarking/src/registry.cpp
```diff
@@ -0,0 +1,94 @@
...
+    base_path += "/session_";
+    base_path += get_curr_datetime();
+    std::filesystem::create_directories(base_path);
+
+    // std::ofstream output_file;
+    // output_file.open(path_);
+    for (int i = 0; i < 6; i++)
+    {
+        std::stringstream ss;
+        ss << "/robot_" << i << ".csv";
```
rishiso: Just use a string here, no need for stringstream

src/rj_benchmarking/src/registry.cpp
rishiso: Nit:
```suggestion
        for (int row = -1; row < max_rows_; ++row)
```

rishiso (changes requested):
Some initial comments

src/rj_benchmarking/include/rj_benchmarking/registry_publisher.hpp
```diff
@@ -0,0 +1,31 @@
+#pragma once
+
+#include <rclcpp/rclcpp.hpp>
+
+#include <rj_msgs/msg/latency.hpp>
+
+class RegistryPublisher {
+public:
+    // Singleton pattern
```
Squid5678: nit: since this is not following the singleton pattern, let's see if we can remove this comment and rename this to getRegistry instead of getInstance? that way it's more clear

Squid5678 (approved):
small nit otherwise looks good

src/rj_benchmarking/include/rj_benchmarking/registry_publisher.hpp
```diff
@@ -0,0 +1,31 @@
...
+    static std::shared_ptr<RegistryPublisher> getInstance() {
+        static std::shared_ptr<RegistryPublisher> instance{new RegistryPublisher()};
```
rishiso: I thought we were switching this using smart pointers?

src/rj_benchmarking/src/registry.cpp
```diff
@@ -0,0 +1,68 @@
...
+    subscription_ = this->create_subscription<rj_msgs::msg::Latency>(
+        "/registry", 100, std::bind(&Registry::topic_callback, this, std::placeholders::_1));
+}
+
+Registry::~Registry() {
+    /*
+        1. Make LatencyLogs Directory if not already there
+        2. Make latency_curr-date_curr-time folder
+        3. Make csv for Robot 1
+            3a. first row is labels
```
rishiso: Is this comment up to date?

src/rj_benchmarking/src/registry.cpp
```diff
@@ -0,0 +1,68 @@
...
+            std::ofstream robot_csv{base_path + ss};
+
+            // Print out labels
+            for (const auto& [label, timestamps] : registry_[i]) {
+                robot_csv << label << ',';
+            }
+            robot_csv << '\n';
+
+            // Print out data row by row
+            for (size_t row = 0; row < max_rows_; ++row) {
```
rishiso, line 41: Are we guaranteed that each column (label) will be measured the same number of times for a given robot? Based on the `Timer` API, I don't think we enforce this behavior. We should place a filler value like -1 or nothing (not sure if csv allows that) when a column doesn't have a valid value for a given row

rishiso (changes requested):
Couple more comments

src/rj_benchmarking/src/registry.cpp
yuvrajdhadwal: good catch, added -1 sentinal value

src/rj_benchmarking/include/rj_benchmarking/registry_publisher.hpp
yuvrajdhadwal: i am a little confused on this, isnt this shared ptr a smart pointer? i switched this to make_shared instead of new in the latest commit

src/rj_benchmarking/include/rj_benchmarking/registry_publisher.hpp
```diff
@@ -0,0 +1,31 @@
...
+            std::make_shared<RegistryPublisher>();
+        return registry_pub;
+    }
+
+    void publish(const std::string& label, uint8_t robot_id, uint64_t time);
+    RegistryPublisher() = default;
+
+private:
+    // Private Constructor
+    // RegistryPublisher() = default;
```
rishiso: Delete this comment

rishiso (approved):
Approved with one comment

src/rj_benchmarking/include/rj_benchmarking/registry_publisher.hpp
rishiso: I think I was seeing an old version of the code maybe

rishiso:
Wait something I just thought about, won't we want to benchmark things that are unrelated to a specific robot?

PR #2463, opened 2025-11-17
standardized cmakes

yuvrajdhadwal:
## Description
Cleans up dependencies in cmake files

### DIFF
--- a/src/rj_common/CMakeLists.txt
+++ b/src/rj_common/CMakeLists.txt
@@ -23,6 +23,19 @@ find_package(rj_param_utils REQUIRED)
 find_package(rj_utils REQUIRED)
 find_package(rj_drawing_msgs REQUIRED)
 
+set(RJ_COMMON_DEPS
+  rj_msgs
+  Eigen3
+  rclcpp
+  rclcpp_components
+  rj_convert
+  rj_protos
+  rj_geometry
+  rj_param_utils
+  rj_utils
+  rj_drawing_msgs
+)
+
 add_library(${PROJECT_NAME} SHARED
   src/debug_drawer.cpp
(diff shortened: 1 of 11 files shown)


src/rj_common/CMakeLists.txt
```diff
@@ -23,6 +23,19 @@ find_package(rj_param_utils REQUIRED)
...
+  Eigen3
+  rclcpp
+  rclcpp_components
+  rj_convert
+  rj_protos
+  rj_geometry
+  rj_param_utils
+  rj_utils
+  rj_drawing_msgs
+)
```
N8BWert, line 37: IIRC rj_common doesn't use these dependencies so they should not be listed in CMakeLists.txt.  Additionally, if they were used here they would need to be declared in find_package and the package.xml file for rj_common

N8BWert (dismissed):

src/rj_common/CMakeLists.txt
yuvrajdhadwal: I double checked and they are being used in rj_common and they all are being declared in find_package and the package.xml file for rj_common

(also funny enough looking at the git blame you were the one who did that two months ago in "Build Stack with Colcon)

Squid5678: Seems to be that these are valid dependencies are they're being rewritten to this "variable" in CMake (from my understanding), so this lgtm.

Squid5678 (approved):
lgtm

N8BWert (approved):
LGTM.  Last time when I was looking this through I thought you standardized the dependencies which would have broken things.  I see now you just created a variable for the dependencies so the changes look fine to me

PR #2467, opened 2025-11-20
Waller Coordinator

sanatd33:
Changes the waller from communication/Role Interface to a coordinator
The logic for wallers previously defined in `position/waller.cpp` has not been changed.

Overview:
Defense sends join group request through Client to Coordinator
Coordinator maintains ordered list of walling robots, sends updated copy to clients
Clients maintain their own copy of walling list
Defense asks Client for walling position, Client computes
Robot switches off from defense, sends leave group request


src/rj_strategy/include/rj_strategy/coordinator/waller_client.hpp
```diff
@@ -0,0 +1,75 @@
...
+private:
+    rclcpp::Node::SharedPtr node_;
+    const uint8_t robot_id_;  // NOLINT(cppcoreguidelines-avoid-const-or-ref-data-members) -- class
+                              // isn't move/copy-able anyway
+    rclcpp::Client<rj_msgs::srv::Waller>::SharedPtr client_;
+    rclcpp::Subscription<rj_msgs::msg::Waller>::SharedPtr subscription_;
+
+    bool am_i_member_ = false;
+    bool request_pending_ = false;
+    std::array<u_int8_t, kNumShells> walling_robots_;
```
rishiso:
```suggestion
    std::array<uint8_t, kNumShells> walling_robots_;
```

src/rj_strategy/include/rj_strategy/coordinator/waller.hpp
```diff
@@ -0,0 +1,39 @@
...
+    Waller(const Waller&) = delete;
+    Waller& operator=(const Waller&) = delete;
+    Waller(Waller&&) = delete;
+    Waller& operator=(Waller&&) = delete;
+
+    void service_callback(RequestPtr request, ResponsePtr response);
+
+private:
+    std::array<u_int8_t, kNumShells> walling_robots_;
+    u_int8_t num_wallers_ = 0;
```
rishiso:
```suggestion
    std::array<uint8_t, kNumShells> walling_robots_;
    uint8_t num_wallers_ = 0;
```

src/rj_strategy/include/rj_strategy/coordinator/waller_client.hpp
```diff
@@ -0,0 +1,75 @@
...
+ *
+ * Manages membership in the waller group and tracks the currently waller list.
+ */
+class WallerClient {
+public:
+    struct Result {
+        bool success{false};
+    };
+
+    using StatusCallback = std::function<void(Result)>;
```
rishiso, line 27: Do we need to create a struct here? Why not just use a boolean?

src/rj_strategy/src/coordinator/waller.cpp
```diff
@@ -0,0 +1,48 @@
+#include "rj_strategy/coordinator/waller.hpp"
+
+namespace strategy {
+
+Waller::Waller() : Coordinator("waller_srv", "waller_data", "waller_node") {
+    walling_robots_.fill(-1);
```
rishiso:
```suggestion
    walling_robots_.fill(std::numeric_limits<uint8_t>::max());
```
I would prefer not to have overflow and make this explicit

src/rj_strategy/src/coordinator/waller_client.cpp
```diff
@@ -0,0 +1,168 @@
...
+            }
+
+            am_i_member_ = false;
+
+            subscription_.reset();
+            walling_robots_.fill(-1);
+            num_wallers_ = 0;
+
+            if (callback) {
+                callback(Result{false});
```
rishiso: Shouldn't this be `success=true`?

src/rj_strategy/src/coordinator/waller_client.cpp
```diff
@@ -0,0 +1,168 @@
...
+
+bool WallerClient::am_i_member() const { return am_i_member_; }
+
+std::optional<rj_geometry::Point> WallerClient::get_walling_point(
+    const WorldState* world_state, FieldDimensions field_dimensions) const {
+    if (!am_i_member_) return std::nullopt;
+
+    // Creates Minimum wall radius is slightly greater than  box bounds
+    // Dimension accessors should be edited when we figure out how we are doing dimensions realtime
+    // from vision
```
rishiso, line 124: Can you break this up with some helpers? Pretty hard to go through

Some more comments would also be helpful

src/rj_strategy/src/coordinator/waller_client.cpp
```diff
@@ -0,0 +1,168 @@
...
+    auto angle = (mid_point - goal_pos).angle();
+    auto delta_angle = (wall_spacing * (waller_pos - num_wallers_ / 2. - 0.5)) / min_wall_rad;
+    auto target_angle = angle - delta_angle;
+
+    target_point =
+        (goal_pos + rj_geometry::Point{1, 0}).normalized(min_wall_rad).rotated(target_angle);
+
+    if (abs(robot_pos.dist_to(goal_pos) - min_wall_rad) < kRobotRadius &&
+        robot_pos.dist_to(target_point) > kRobotRadius) {
+        uint8_t parent_id =
```
rishiso: Why are we using a lambda here?

rishiso (changes requested):
Overall looks good, a few comments

src/rj_strategy/include/rj_strategy/coordinator/waller_client.hpp
sanatd33: I had a struct here to 1. keep the standard coordinator usage across all coordinators of callbacks taking in some Result struct, instead of having to figure out what the callback takes in for each coordinator, 2. leave room for other things to be used in the callback for future improvements, i.e. if we want waller coordinator to pick a waller to clear the ball that could be in the struct, 3. structs dont have any additional overhead so it doesn't cost anything

rishiso: Makes sense

rishiso (approved):
There are a few places we are still overflowing -1 but besides that LGTM

src/rj_strategy/include/rj_strategy/coordinator/waller.hpp
```diff
@@ -0,0 +1,39 @@
...
+    Waller& operator=(const Waller&) = delete;
+    Waller(Waller&&) = delete;
+    Waller& operator=(Waller&&) = delete;
+
+    void service_callback(RequestPtr request, ResponsePtr response);
+
+private:
+    std::array<uint8_t, kNumShells> walling_robots_;
+    uint8_t num_wallers_ = 0;
+    static constexpr int kMaxWallers = 4;
```
shourikb, line 34: The value here is different from the one in defense.hpp. Can we remove one of them?

src/rj_strategy/src/agent/position/defense.cpp
```diff
@@ -29,31 +26,19 @@ Defense::State Defense::update_state() {
...
 
-    if (current_state_ != WALLING && current_state_ != JOINING_WALL && waller_id_ != -1) {
-        send_leave_wall_request();
-        walling_robots_ = {(u_int8_t)robot_id_};
-        waller_id_ = -1;
-    }
-
     switch (current_state_) {
         case IDLING:
+            next_state = JOINING_WALL;
```
petergarud, line 32: Do we want to have a comment explaining why this is default behavior. Perhaps, this is something to be added later but it might be nice to have a brief strategy comment

src/rj_strategy/src/agent/position/defense.cpp
```diff
@@ -29,31 +26,19 @@ Defense::State Defense::update_state() {
...
-            walling_robots_ = {(u_int8_t)robot_id_};
+            client_handles_->waller->join_group([this](WallerClient::Result result) {
+                if (result.success)
+                    current_state_ = WALLING;
+                else
+                    current_state_ = ENTERING_MARKING;
+            });
             break;
         case WALLING:
-            // If a wall is already full,
```
shourikb, line 48: Shouldn't we have a check if we're in the group? If we are not in the group, then leave Walling and go back to IDLING

src/rj_strategy/src/agent/position/defense.cpp
```diff
@@ -29,31 +26,19 @@ Defense::State Defense::update_state() {
...
+            next_state = JOINING_WALL;
             break;
         case JOINING_WALL:
-            send_join_wall_request();
-            // SPDLOG_INFO("join wall {}", robot_id_);
-            next_state = WALLING;
-            walling_robots_ = {(u_int8_t)robot_id_};
+            client_handles_->waller->join_group([this](WallerClient::Result result) {
+                if (result.success)
+                    current_state_ = WALLING;
```
petergarud, line 37: Is this call idempotent? Otherwise could run into race conditions.

src/rj_strategy/src/agent/position/defense.cpp
```diff
@@ -29,31 +26,19 @@ Defense::State Defense::update_state() {
...
+                else
+                    current_state_ = ENTERING_MARKING;
+            });
             break;
         case WALLING:
-            // If a wall is already full,
-            // Remove the robot with the highest ID from a wall
-            // and make them a marker instead.
```
petergarud, line 50: Don't we want logic to determine whether we have been kicked out of the wall group or other (maybe this is later stuff to do), have logic to move on to a better state.

src/rj_strategy/src/coordinator/waller.cpp
```diff
@@ -0,0 +1,48 @@
...
+            response->success = true;
+        }
+    } else {
+        auto it = std::find(walling_robots_.begin(), walling_robots_.end(), request->robot_id);
+        if (it != walling_robots_.end()) {
+            num_wallers_--;
+            *it = std::numeric_limits<uint8_t>::max();
+            update_wallers();
+        }
+        response->success = true;
```
petergarud, line 29: if the waller isn't in the list, shouldn't we make it a success = false.

src/rj_strategy/src/coordinator/waller.cpp
```diff
@@ -0,0 +1,48 @@
...
+    }
+}
+
+void Waller::update_wallers() {
+    auto prev_wallers = walling_robots_;
```
petergarud: why do we have this value. it doesn't seem like it is used?

src/rj_strategy/src/coordinator/waller.cpp
```diff
@@ -0,0 +1,48 @@
...
+    std::sort(walling_robots_.begin(), walling_robots_.end());
```
petergarud, line 34: Shouldn't we have smarter logic on ordering the list based on positions of robots.

src/rj_strategy/include/rj_strategy/coordinator/waller_client.hpp
```diff
@@ -0,0 +1,92 @@
...
+
+/**
+ * @brief Client for interacting with the Waller coordinator.
+ *
+ * Manages membership in the waller group and tracks the currently waller list.
+ */
+class WallerClient {
+public:
+    struct Result {
+        bool success{false};
```
shourikb, line 24: Why are we using a struct with only one boolean? Can we not just use a boolean instead?

src/rj_strategy/src/coordinator/waller.cpp
```diff
@@ -0,0 +1,48 @@
...
+        if (it != walling_robots_.end()) {
+            num_wallers_--;
+            *it = std::numeric_limits<uint8_t>::max();
+            update_wallers();
+        }
+        response->success = true;
+    }
+}
+
+void Waller::update_wallers() {
```
petergarud, line 33: I understand this runs when waller joins or leaves wall. However, if this implements more complicated logic based on position of wallers, etc, it would probably be better to have this run on a timer.

src/rj_strategy/src/coordinator/waller_client.cpp
```diff
@@ -0,0 +1,203 @@
+#include "rj_strategy/coordinator/waller_client.hpp"
+
+namespace strategy {
+
+WallerClient::WallerClient(rclcpp::Node::SharedPtr node, uint8_t robot_id)
+    : node_{std::move(node)}, robot_id_{robot_id} {
+    client_ = node_->create_client<rj_msgs::srv::Waller>("waller_srv");
+    walling_robots_.fill(-1);
```
petergarud: magic number?

src/rj_strategy/src/coordinator/waller_client.cpp
```diff
@@ -0,0 +1,203 @@
...
+}
+
+void WallerClient::join_group(StatusCallback callback) {
+    if (am_i_member_ || request_pending_) {
```
petergarud, line 12: I worry not idempotent because even though we have the check for request pending this service wait for potentially  one second and don't set the request_pending_ boolean yet. There probably isn't a worry but can you just explain why not?

shourikb (changes requested):

src/rj_strategy/include/rj_strategy/coordinator/waller_client.hpp
sanatd33: see https://github.com/RoboJackets/robocup-software/pull/2467/files#r2547314163

src/rj_strategy/include/rj_strategy/coordinator/waller.hpp
sanatd33: done

src/rj_strategy/src/agent/position/defense.cpp
sanatd33: Done

src/rj_strategy/src/agent/position/defense.cpp
sanatd33: discussed this tdy, we reached conclusion that it doesnt matter that much, any race conditions will be resolved in the next tick

src/rj_strategy/src/agent/position/defense.cpp
sanatd33: true, resolved

src/rj_strategy/src/coordinator/waller.cpp
sanatd33: the precedent from kicker_picker client is "success" when leaving a group should be true if we successfully left the group

src/rj_strategy/src/coordinator/waller.cpp
sanatd33: Yes, if that is added, i think it should run on timer, but until then it's just extra computation

src/rj_strategy/src/coordinator/waller.cpp
sanatd33: resolved

src/rj_strategy/src/coordinator/waller.cpp
sanatd33: i agree it might not be the best, but its probably out of scope for now. we've been just doing numerical order since before, so if we want to change it we can always just do it later

src/rj_strategy/src/coordinator/waller_client.cpp
sanatd33: resolved

src/rj_strategy/src/coordinator/waller_client.cpp
sanatd33: i dont think this is an actual issue, the wait for service only is an actual thing at the startup of the stack, if we're ever running defense we can safely assume the service is operational, if it isnt, then none of this will work anyway, so i think its fine

Squid5678 (approved):
lgtm

shourikb (approved):
Looks good

PR #2473, opened 2025-11-24
Seeker coordinator

jvogt23:
## Description
Added Seeker coordinator to handle seekers and seeker points without requiring several robots to keep track of each other.
Single source of truth for current seeker points.

Resolves Clickup Issue: 86b6ubv6c
## Design Documents
[Link](link-to-design-doc)

### Test Case 1
1. Start the sim
2. Hit play
3. Seekers go brrrr

**Expected result:** Seekers go brrr and no weird Seeker requests and responses needed (except with the coordinator ofc)

## Key Files to Review
Group 1
 * File 1: seeker_coordinator.hpp and .cpp
 * File 2: seeker_client.cpp and .hpp

Group 2
 * File 3: offense.cpp and .hpp
 * File 4: Seeker coordinator .msg files

## Review Checklist

- [ ] Step 1
- [ ] Step 2


launch/soccer.launch.py
```diff
@@ -237,6 +237,13 @@ def generate_launch_description():
                 output="screen",
                 parameters=[param_config_filepath],
                 on_exit=Shutdown(),
+            ),
+            Node(
+                package="rj_strategy",
+                executable="seeker_coordinator_node",
```
petergarud: Make sure using same naming convention as the other coordinator nodes.

src/rj_strategy/src/agent/position/offense.cpp
```diff
@@ -162,17 +164,24 @@ std::optional<RobotIntent> Offense::state_to_task(RobotIntent intent) {
...
-            std::optional<RobotIntent> actual_intent =
-                seeker_.get_task(std::move(intent), last_world_state_, field_dimensions_);
-            broadcast_seeker_request(seeker_.get_target_point(), true);
-            return actual_intent;
+            client_handles_->seekerClient->poll_for_target();
+            intent.motion_command = planning::MotionCommand{};
+            return intent;
+        }
+
+        case SEEKING_PROBE: {
```
petergarud: I think we can delete seeking probe and just make it a part of seeking. Also, I feel like we should make sure that if seeking points change when we are in the seeking state we continue to go to that point. Therefore, we should be continusouly looking for our seeking_target by calling selected_target() all the time in the seeking state.

src/rj_strategy/src/agent/position/offense.cpp
```diff
@@ -162,17 +164,24 @@ std::optional<RobotIntent> Offense::state_to_task(RobotIntent intent) {
...
 
         case SEEKING_START: {
-            // Calculate a new seeking point
-            seeker_.reset_target();
-            seeker_.set_seeker_points(seeker_points_);
-            std::optional<RobotIntent> actual_intent =
-                seeker_.get_task(std::move(intent), last_world_state_, field_dimensions_);
-            broadcast_seeker_request(seeker_.get_target_point(), true);
-            return actual_intent;
+            client_handles_->seekerClient->poll_for_target();
```
petergarud: if you look at marking/walling coordaintors, we use join_group instead of a polling name. Why did you choose this name?

src/rj_strategy/src/coordinator/seeker_client.cpp
```diff
@@ -0,0 +1,89 @@
+#include "rj_strategy/coordinator/seeker_client.hpp"
+
+namespace strategy {
+
+SeekerClient::SeekerClient(rclcpp::Node::SharedPtr node, uint8_t robot_id)
+    : node_{std::move(node)}, robot_id_{robot_id}, selected_target_{-1, -1} {
+    client_ = node_->create_client<rj_msgs::srv::SeekerCoordinator>("seeker_coordinator_srv");
+}
+
+void SeekerClient::poll_for_target(StatusCallback callback) {
```
petergarud: make sure idempotent

src/rj_strategy/src/coordinator/seeker_coordinator.cpp
```diff
@@ -1,47 +1,58 @@
-#include "rj_strategy/agent/position/seeker.hpp"
+#include "rj_strategy/coordinator/seeker_coordinator.hpp"
```
petergarud: make sure using same naming convention for file as in the walling and marking coordiantors.

src/rj_strategy/src/coordinator/seeking_coordinator.cpp
```diff
@@ -1,47 +1,58 @@
...
+            last_world_state_ = rj_convert::convert_from_ros(*world_state);
+        });
+}
 
-std::optional<RobotIntent> Seeker::get_task(RobotIntent intent, const WorldState* last_world_state,
-                                            FieldDimensions field_dimensions) {
-    // Determine target position for seeking
-    rj_geometry::Point current_loc = last_world_state->get_robot(true, robot_id_).pose.position();
+void SeekerCoordinator::service_callback(RequestPtr request, ResponsePtr response) {
+    is_seeking_[request->robot_id] = request->wants_to_seek;
```
petergarud: make sure idempotent

src/rj_strategy/src/coordinator/seeker_coordinator.cpp
```diff
@@ -1,47 +1,58 @@
...
-void Seeker::reset_target() { target_valid_ = false; }
-
-rj_geometry::Point Seeker::get_target_point() { return target_pt_; }
+void SeekerCoordinator::publish_seeker_points() {
+    std::array<rj_geometry_msgs::msg::Point, kNumShells> msg_points;
+    for (size_t i = 0; i < kNumShells; i++) {
+        msg_points[i].x = seeker_points_[i].x();
+        msg_points[i].y = seeker_points_[i].y();
+    }
+    this->publisher_->publish(rj_msgs::msg::SeekerCoordinator().set__positions(msg_points));
```
petergarud: is there a reason we need to copy the array first using a for loop. Can we not directly pass it in. If copying by reference then just use a function to copy the array if possible.

src/rj_strategy/src/coordinator/seeker_coordinator.cpp
```diff
@@ -1,47 +1,58 @@
...
+}
 
-void Seeker::set_seeker_points(const std::unordered_map<int, rj_geometry::Point>& seeker_points) {
-    seeker_points_ = seeker_points;
+void SeekerCoordinator::update_target(int robot_id) {
+    rj_geometry::Point robot_pos = last_world_state_.our_robots.at(robot_id).pose.position();
+    seeker_points_[robot_id] = get_open_point(last_world_state_, robot_pos, field_dimensions_);
+    SPDLOG_INFO("Open Point Found for Robot {}: {}, {}", robot_id, seeker_points_[robot_id].x(),
```
petergarud: remove, comment out, logging statement

src/rj_strategy/src/coordinator/seeker_coordinator.cpp
```diff
@@ -1,47 +1,58 @@
...
 }
 
-rj_geometry::Point Seeker::get_open_point(const WorldState* world_state,
-                                          rj_geometry::Point current_position,
-                                          const FieldDimensions& field_dimensions) const {
-    return Seeker::calculate_open_point(3.0, .2, current_position, world_state, field_dimensions);
+rj_geometry::Point SeekerCoordinator::get_open_point(
+    const WorldState world_state, rj_geometry::Point current_position,
+    const FieldDimensions& field_dimensions) const {
+    return SeekerCoordinator::calculate_open_point(3.0, .2, current_position, world_state,
```
petergarud: magic numbers

src/rj_strategy/src/coordinator/seeking_coordinator.cpp
petergarud: i'm worried not currently idempotent because can call again and recalculate a new point. Feels like wasted computation. Potentially have a boolean variable for a pending request and check that. This could potentially be solved at the client level.

src/rj_strategy/src/coordinator/seeking_coordinator.cpp
```diff
@@ -205,3 +218,11 @@ double Seeker::eval_point(rj_geometry::Point ball_pos, rj_geometry::Point curren
 }
 
 }  // namespace strategy
+
+int main(int argc, char** argv) {
+    rclcpp::init(argc, argv);
```
petergarud, line 261: Overall comment. What I would like to see is instead of the coordinator being reactive to new seeking requests, it is instead running continusouly to evaluate the best positions to travel to. This could help with situations where robots shift and there is a better position to seek to if enemy robots change. This would involve a continuously running get open point function that gets open points for all seeking robots. It would also involve ensuring that robots are mapped to the right points and that they only switch to new points if that new point is signifnicantly better.

src/rj_strategy/src/agent/position/offense.cpp
jvogt23: I originally set up the seeker coordinator to constantly send updated points, and it caused pretty immense slowdowns and some very strange movement. The seeker point calculation is pretty expensive which is why I do probing at the beginning of every seek cycle.

src/rj_strategy/src/coordinator/seeking_coordinator.cpp
jvogt23: Issue here is that to achieve this with any efficiency the entire seeking algorithm would need to change quite a bit. I initially had it run the seeker target point calculation algorithm for each robot at every world state update, and it caused everything to slow down. I can keep trying this, but the solution I came up with preserves the original functionality while removing the spaghetti the original seeker setup required.

src/rj_strategy/src/coordinator/seeker_coordinator.cpp
jvogt23: The messages sent through ROS2 use a different data type; passing in the original array directly crashes.

src/rj_strategy/src/agent/position/offense.cpp
jvogt23: Chose this name because the function is called both when a robot joins the group of seekers and when it wants to change its seeking target

src/rj_strategy/src/coordinator/seeker_client.cpp
jvogt23: What does idempotent mean in context?

src/rj_strategy/src/coordinator/seeking_coordinator.cpp
jvogt23: Set up regular interval of 500ms where it evaluates all current seeker points and calculates new points if the current points receive an eval score of >= 2.0 (higher scores are worse)

src/rj_strategy/include/rj_strategy/coordinator/seeking_client.hpp
```diff
@@ -0,0 +1,70 @@
...
+
+private:
+    rclcpp::Node::SharedPtr node_;
+    const uint8_t robot_id_;  // NOLINT(cppcoreguidelines-avoid-const-or-ref-data-members) -- class
+                              // isn't move/copy-able anyway
+    rclcpp::Client<rj_msgs::srv::SeekingCoordinator>::SharedPtr client_;
+    rclcpp::Subscription<rj_msgs::msg::SeekingCoordinator>::SharedPtr subscription_;
+
+    bool am_i_member_{false};
+    rj_geometry::Point selected_target_{-1, -1};
```
Squid5678: why is this initialized with -1?

src/rj_strategy/include/rj_strategy/agent/position/offense.hpp
```diff
@@ -41,6 +41,8 @@ class Offense : public Position {
 
     std::string get_current_state() override;
 
+    void die() override;
```
Squid5678, line 44: nit: better name

src/rj_strategy/include/rj_strategy/coordinator/seeking_coordinator.hpp
```diff
@@ -0,0 +1,130 @@
...
+public:
+    SeekingCoordinator();
+    ~SeekingCoordinator() override = default;
+    SeekingCoordinator(const SeekingCoordinator&) = delete;
+    SeekingCoordinator& operator=(const SeekingCoordinator&) = delete;
+    SeekingCoordinator(SeekingCoordinator&&) = delete;
+    SeekingCoordinator& operator=(SeekingCoordinator&&) = delete;
+
+    void service_callback(RequestPtr request, ResponsePtr response);
+    static rj_geometry::Point invalidPoint() { return rj_geometry::Point{-1, -1}; }
```
Squid5678: can we replace this invalidPoint() with just a null value once a robot leaves seeking?

src/rj_strategy/src/coordinator/seeking_client.cpp
```diff
@@ -0,0 +1,88 @@
+#include "rj_strategy/coordinator/seeking_client.hpp"
+
+namespace strategy {
+
+SeekingClient::SeekingClient(rclcpp::Node::SharedPtr node, uint8_t robot_id)
+    : node_{std::move(node)}, robot_id_{robot_id}, selected_target_{-1, -1} {
+    client_ = node_->create_client<rj_msgs::srv::SeekingCoordinator>("seeking_coordinator_srv");
+}
+
+void SeekingClient::join_group(StatusCallback callback) {
```
Squid5678, line 10: can we add a check for `am_i_member_` prior to the callback to ensure that we aren't making repeated `join_group()` calls?

src/rj_strategy/src/coordinator/seeking_coordinator.cpp
Squid5678: +1 to this, I think this is better than the 500ms approach

src/rj_strategy/src/coordinator/seeking_coordinator.cpp
```diff
@@ -1,49 +1,81 @@
...
+            get_open_point(robot_id, last_world_state_, current_pos, field_dimensions_);
+    }
 }
 
-rj_geometry::Point Seeker::get_open_point(const WorldState* world_state,
-                                          rj_geometry::Point current_position,
-                                          const FieldDimensions& field_dimensions) const {
-    return Seeker::calculate_open_point(3.0, .2, current_position, world_state, field_dimensions);
+rj_geometry::Point SeekingCoordinator::get_open_point(
+    int robot_id, const WorldState world_state, rj_geometry::Point current_position,
```
Squid5678: should this be `const WorldState& world_state` to avoid copying in the entire world state?

Squid5678 (changes requested):
some changes

src/rj_strategy/include/rj_strategy/agent/position/offense.hpp
jvogt23: die() is inherited from Position

src/rj_strategy/include/rj_strategy/coordinator/seeking_client.hpp
jvogt23: { -1, -1} is a known bad point that can't be reached on the field

src/rj_strategy/src/coordinator/seeking_coordinator.cpp
jvogt23: I think I did this because subscribing to the World State topic gets you the entire world state rather than a reference, but passing a reference to the resulting methods would probably be more memory efficient

PR #2481, opened 2026-01-12
Adding Feature for Global Latency Benchmarking (Independent of a Robot)

yuvrajdhadwal:
In a comment right before break, @rishiso mentioned that we should be able to profile events independent of the robots. I have implemented that as a feature of the benchmarking module. You would set robot_id to -1 to do this.


yuvrajdhadwal:
oops i made the branch off a branch instead of off main let me fix that

yuvrajdhadwal:
closing

PR #2482, opened 2026-01-12
Adding Feature for Global Latency Benchmarking (Independent of a Robot)

yuvrajdhadwal:
In a comment right before break, @rishiso  mentioned that we should be able to profile events independent of the robots (#2461). I have implemented that as a feature of the benchmarking module. You would set robot_id to -1 to do this.

### DIFF
--- a/src/rj_benchmarking/src/registry.cpp
+++ b/src/rj_benchmarking/src/registry.cpp
@@ -29,14 +29,13 @@ Registry::~Registry() {
             ss += "/robot_";
             ss += std::to_string(i);
             ss += ".csv";
-            std::ofstream robot_csv {base_path + ss};
+            std::ofstream robot_csv{base_path + ss};
             print_data(robot_csv, i);
         }
 
-        if (!registry_[kNumShells].empty())
-        {
-            std::string file_name {"/global.csv"};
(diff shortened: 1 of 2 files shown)


src/rj_benchmarking/src/registry.cpp
```diff
@@ -30,33 +30,28 @@ Registry::~Registry() {
...
+            std::ofstream global_csv{base_path + file_name};
+            print_data(global_csv, kNumShells);
         }
     }
 }
 
 void Registry::topic_callback(const rj_msgs::msg::Latency& msg) {
-    registry_[msg.robot_id][msg.label].push_back(msg.duration_ns);
-    max_rows_ = std::max(max_rows_, static_cast<size_t>(registry_[msg.robot_id][msg.label].size()));
+    if (msg.robot_id == -1)  // robot independent; will go at end
```
Squid5678, line 45: small nit but could be nice for future users to understand that robot id = -1 leads to independent robot benchmarking

Squid5678 (changes requested):

rishiso (approved):
LGTM

Squid5678 (approved):
ok

PR #2503, opened 2026-02-24
Grey Field Fix

CameronLyon:
This bug has been in the stack for years, but I believe this will fix it.

The bug in question:
- About 10-15% of the time, when running the stack, the field will fail and become grey
- All bots, balls, and game items are non-existent
- The user is forced to re-run (or rebuild if consistent issue) 

The ConfigServer's parse_game_settings function builds a ROS message where all bool fields default to false, and it never sets use_our_half or use_their_half -- so both are published as false, telling the VisionReceiver to discard every robot and ball and telling the UI to grey out both halves of the field. The MainWindow tries to correct this by sending the right values via a service call, but ~40% of the time it fires before DDS has discovered the ConfigServer's service, so the call is silently lost and the bad defaults persist for the entire session. Adding game_settings.use_our_half = true and game_settings.use_their_half = true in parse_game_settings fixes the defaults at the source, eliminating the race condition.

Additionally, RoboNav inserted some things into our repo (somehow) when making this PR, so jetson-inference and jetson-utils was moved/modified. If there are any issues related to that, please let me know.


rishiso (approved):
Good to merge after resolving merge conflict

Squid5678 (changes requested):
can we get rid of these weird robonav changes? otherwise lgtm

sanatd33 (approved):
lgtm after you remove the jetson stuff
should just be able to `rm -rf jetson-*`

PR #2513, opened 2026-03-09
Unique PID Parameters for each Robot

shourikb:
## Description
Changed the sim and real params .yamls to use per robot PID values for rotation and translation, kp, kd, ki for robots 0-5. Modified the motion_control cpp and hpp files for this change, removed the constant DEFINES.

https://app.clickup.com/t/86b6w4xcu

## Steps to Test
1. Source, make, and make run-sim
2. In another terminal window, open up RQT using the rqt command after sourcing
3. Run the simulator and note which robots are moving
4. Change the kp value for one of the moving robots to 100000

**Expected result:**
Should see the robot that had it's value change start moving weirdly, while the other robots remain fine.

## Key Files to Review
 * sim_params.yaml
 * real_params.yaml
 * motion_control.cpp
 * motion_control.hpp

### DIFF
--- a/src/rj_control/include/rj_control/motion_control.hpp
+++ b/src/rj_control/include/rj_control/motion_control.hpp
@@ -34,43 +34,6 @@ namespace control {
 // DECLARE_FLOAT64(params::kMotionControlParamModule, translation_kd);
 // DECLARE_INT64(params::kMotionControlParamModule, translation_windup);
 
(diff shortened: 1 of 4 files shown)


CameronLyon (approved):
Worked out of sim so that's a success on many more levels 😅

Squid5678 (commented):
works

Squid5678 (approved):

PR #2516, opened 2026-03-09
Passing improvements

CameronLyon:
This pull request tackles the issues with passing by basically revamping the entire code/behavior behind the communication between bots, shooting calculations, and ball interpretation fixes. Here's the current issue:

Issues:
- **No wait on pass.** When the bot kicks the ball towards a teammate, the bot immediately enters seeking/stealing and proceeds to chase after the ball it just passed to its friend.
- **Passing repeatedly.** After passing once, the receiver will try to pass again back to the previous bot (who is clearly not open since it passed first), causing a chain reaction of non-stop passing until they are basically next to each other.
- **No checks on whether the bot can shoot.** The reason behind the repeated passing is that the bots do not check if they are able to actually shoot uninterrupted from where they are standing, only if there are any bots blocking them.
- **Ball disappearing.** Whenever a bot gets possession of the ball, the simulator freaks out and teleports (or remove) the ball from where it was. This happens 70% of the time

This pull request responds to these issues with the following changes:
- **Additional communication request/acknowledgment on the pass.** When the bot passes the ball, there is one more communication request that waits for the receiver to send a signal that it received it. Until then, the bot waits (unless it fails to kick the ball or if it times out). This is the reason behind the creation of pass_received_request() file.
- **can_i_shoot() function implemented.** Similar to check_is_open(), but focuses more on the the bots shooting trajectory rather than the bots receiving trajectory. If the bot can shoot from where it is, it is considered open.
- **broadcast_direct_pass() function repeated throughout POSSESSION**, constantly calling out for an open receiver if it can't shoot directly and defaults to SHOOTING if it can't find someone withing 5 seconds.
- **Prevention of passing over the enemy's goalie box**, avoiding the silly scenario where we pass into the goalie box.
- **Commented out "portion of ball_ = WorldBall()."** This is one factor towards the ball vanishing and removing this line in world.cpp helps fix the issue (but not completely). If this line is for a more important purpose and mess up other features, please revert it.

If there are any logical mistakes, vague areas that need commenting, or anything else, lmk on this PR. Thx 😀

### DIFF
--- a/src/rj_strategy/src/agent/position/offense.cpp
+++ b/src/rj_strategy/src/agent/position/offense.cpp
@@ -309,7 +309,7 @@ bool Offense::check_if_open(int target_robot_shell) {
     // are no other robots
     // in the passing line, process the request Currently, max_receive_distance is used to
     // determine when we are open, but this may need to change
-    return (min_robot_dist > max_receive_distance && min_path_dist > max_receive_distance/2);
+    return (min_robot_dist > max_receive_distance && min_path_dist > max_receive_distance / 2);
 }
 
 communication::PosAgentResponseWrapper Offense::receive_communication_request(
@@ -325,18 +325,16 @@ communication::PosAgentResponseWrapper Offense::receive_communication_request(
         auto response = Position::receive_pass_request(*pass_request);
 
         rj_geometry::Point passer_pos =
-        last_world_state_->get_robot(true, pass_request->from_robot_id).pose.position();
+            last_world_state_->get_robot(true, pass_request->from_robot_id).pose.position();
         rj_geometry::Point receiver_pos =
             last_world_state_->get_robot(true, robot_id_).pose.position();
         rj_geometry::Segment pass_trajectory{passer_pos, receiver_pos};
 
         // If the pass is over any defense area, do not pass
-        bool crosses_penalty_area =
-            field_dimensions_.our_defense_area().hit(pass_trajectory) ||
-            field_dimensions_.their_defense_area().hit(pass_trajectory);
+        bool crosses_penalty_area = field_dimensions_.our_defense_area().hit(pass_trajectory) ||
+                                    field_dimensions_.their_defense_area().hit(pass_trajectory);
 
         bool too_close = passer_pos.dist_to(receiver_pos) < kMinPassDistance;
-            
 
         if (check_if_open(pass_request->from_robot_id) && can_i_shoot()) {
             response.direct_open = true;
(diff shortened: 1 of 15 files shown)


src/rj_strategy/src/agent/position/offense.cpp
```diff
@@ -118,15 +136,18 @@ Offense::State Offense::next_state() {
         case RECEIVING: {
             // If we got it, cool, we have it!
             if (check_is_done() && distance_to_ball() < kOwnBallRadius) {
+                send_pass_received_to_passer(face_robot_id);
                 return POSSESSION_START;
             }
 
             if (ball_in_red()) {
```
shourikb: NIT: we can combine the two if statements into a if (ball_in_red() || timed_out()) {

src/rj_strategy/src/agent/position/offense.cpp
```diff
@@ -276,11 +303,13 @@ bool Offense::check_if_open(int target_robot_shell) {
...
     }
 
+    min_robot_dist = std::min(min_robot_dist, robot_position.dist_to(from_robot_position));
+
     // If the current robot is far enough away from other robots and there
     // are no other robots
     // in the passing line, process the request Currently, max_receive_distance is used to
     // determine when we are open, but this may need to change
-    return (min_robot_dist > max_receive_distance && min_path_dist > max_receive_distance);
+    return (min_robot_dist > max_receive_distance && min_path_dist > max_receive_distance/2);
```
shourikb: Add comment for why it's max_receive_distance/2

src/rj_strategy/src/agent/position/offense.cpp
```diff
@@ -308,6 +351,15 @@ communication::PosAgentResponseWrapper Offense::receive_communication_request(
         } else {
             seeker_points_.erase(seeker_request->robot_id);
         }
+    } else if (std::get_if<communication::PassReceivedRequest>(&request.request)) {
+        // Receiver has controlled the ball; we can leave PASSING_FINISHED.
+        if (current_state_ == PASSING_FINISHED) {
```
shourikb, line 350: Also can merge these if statements

shourikb (approved):
Some minor things, other than that logically looks good

Squid5678 (approved):
checked IRL seems good

src/rj_strategy/src/agent/position/offense.cpp
CameronLyon: While I do agree that merging these if statements reduces the # of lines, I believe keeping them separate maintains simplicity and allows readers to understand what's going on instead of looking at: "std::get_if<communication::PassReceivedRequest>(&request.request) && (current_state_ == PASSING_FINISHED || (current_state_ == RECEIVING || current_state_ == RECEIVING_START))"
If you mean something else or if you have a different way of showcasing the logic in a condensed yet simplified way, just lmk. I did the rest of your other changes.

shourikb (approved):

PR #2519, opened 2026-03-11
debug_drawer expansion into strategy/position files

CameronLyon:
**This addition to the stack is not necessary, but I believe will help tremendously towards debugging and adding additional visual features to the simulator**

Implementing debug_drawer among all instances of position/strategy for the sim:
- Position classes like offense.cpp, defence.cpp, seeker.cpp, solo_offense.cpp, and many more will now be capable of incorporating debug_drawer features (drawing circles/squares, inputting text, etc)
- This feature is linked to a flag variable (located in position.hpp called "debug_draw_enabled_") which is defaulted to false, but can be selectively enabled for specific position roles
- Example of enabling and incorporating this debugging feature is shown in offense.cpp, where offense bots will now showcase their current state when moving, avoiding the messy need to utilize SPDLOG_INFO and flood the terminal with print statements. If you want to remove this, just let me know with a comment and I can comment out the currently enabled example

### DIFF
--- a/src/rj_strategy/CMakeLists.txt
+++ b/src/rj_strategy/CMakeLists.txt
@@ -19,6 +19,7 @@ find_package(rj_msgs REQUIRED)
 find_package(rj_constants REQUIRED)
 find_package(rj_convert REQUIRED)
 find_package(rj_common REQUIRED)
+find_package(rj_drawing_msgs REQUIRED)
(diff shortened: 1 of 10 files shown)


rhkoulen (approved):

Squid5678 (approved):
lgtm

PR #2525, opened 2026-04-08
Creating a Shared Utility File for rj_strategy Positions

rhkoulen:
## Description
This PR introduces `src/rj_strategy/include/rj_strategy/agent/position_utils.hpp`. I'm hoping to address something that has really bothered me about strategy for a while.

In strategy files, we often rewrite the same function in so many different places. Not only that, but we also make different meta-decisions in different places. What counts as an open shot? What shots should be evaluated in shot selection? What field geometry is accessible? How hard should we kick the ball?

With a single source of truth, hopefully this will be more simple. If we change our approach to shot acquisition, every class that chooses to use the central shot planning function `calculate_best_shot` will all share that improved approach. Moreover, it attempts to tackle the huge amount of duplicated code. This file should contain functions about meta-strategy, common calculations, etc.

I've already adapted Offense and SoloOffense to use this utils file. I'm going to need to crusade to get people to use this, but I hope this becomes the most carefully maintained file.

## Testing
You can simply build and try out Offense or SoloOffense. Their behavior should not have changed.

### DIFF
--- a/src/rj_strategy/include/rj_strategy/agent/position_utils.hpp
+++ b/src/rj_strategy/include/rj_strategy/agent/position_utils.hpp
@@ -285,11 +285,16 @@ inline bool we_have_ball(const WorldState* world_state, double possession_radius
  * @return an int to shove in the motion command
  */
 inline int calculate_kick_speed(double distance_to_target, double intended_velo_at_target) {
-    // TODO: these numbers are imaginary; based on estimates, NEED to measure
+    // Without measurement of anything, we cannot make a cogent estimate of kick speed.
+    return 4;
+
+    // TODO: these numbers are imaginary; based on estimates, we NEED to measure
+    // This is the approach we should take to kick speed, do not delete this.
     // This could be a fun collab with the hardware subteam! They could build a light gate!
-    // This maps "kick power" to the initial velocity of the ball.
-    const double kick_speed_map[16] = {3.0, 3.5, 4.0, 4.5, 5.0, 5.5, 6.0,  6.5,
-                                       7.0, 7.5, 8.0, 8.5, 9.0, 9.5, 10.0, 10.5};
+
+    // This maps "kick power" to the initial velocity of the ball (m/s).
+    constexpr std::array<double, 16> kick_speed_map = 
+        {3.0, 3.5, 4.0, 4.5, 5.0, 5.5, 6.0, 6.5, 7.0, 7.5, 8.0, 8.5, 9.0, 9.5, 10.0, 10.5};
     // This notes the deceleration of the ball on the pitch (m/s^2).
     const double ball_deceleration = 1.75;  // ideally, this is parametrized in rqt
 
--- a/src/rj_strategy/include/rj_strategy/agent/position_utils.hpp
+++ b/src/rj_strategy/include/rj_strategy/agent/position_utils.hpp
@@ -1,7 +1,7 @@
 #pragma once
 
-#include <cstdlib>
 #include <cmath>
+#include <cstdlib>
 #include <limits>
 #include <string>
 #include <unordered_map>
@@ -23,15 +23,17 @@
 /**
  * This file is just to collect a bunch of common utilities.
  * Longterm TODO:
- *  - a lot functions in here do pairwise distance between various objects. can we have a coordinator that does a grand pairwise distance we can just reference?
+ *  - a lot functions in here do pairwise distance between various objects. can we have a
+ *    coordinator that does a grand pairwise distance we can just reference?
  * Rules:
  *  - these functions have to be stateless and context-free
  *  - try to keep big object args as const references (to minimize overhead)
  *  - try to keep the execution time minimal (to minimize overhead)
  *  - try to keep function names unnecessarily verbose and accurate (to minimize headache)
- *  - remember that if you change something, other people may be using that; don't fundamentally change invariants
- *      if you're doing something a little different, do it in your own file, or add a new util, avoid revising where possible
- *  - if you love me, do not use auto. figure out what type you want
+ *  - remember that if you change something, other people may be using that; don't fundamentally
+ *    change invariants if you're doing something a little different, do it in your own file,
+ *    or add a new util, avoid revising where possible
+ *  - please don't use auto. figure out what type you want :)
  */
 
 // namespace strategy {
@@ -39,7 +41,7 @@
 // Example
 /**
  * @brief Euclidean distance between two points.
- * 
+ *
  * @param a The first point.
  * @param b The second point.
  * @return the Euclidean distance between a and b
@@ -48,12 +50,10 @@ inline double distance(const rj_geometry::Point& a, const rj_geometry::Point& b)
     return (a - b).mag();
 }
 
-
-
 // Geometry confirmation
 /**
  * @brief Determines whether the ball is in bounds (whole field rectangle).
- * 
+ *
  * @param world_state (often named last_world_state_ in Position subclasses)
  * @param field_dimensions (often named field_dimensions_ in Position subclasses)
  * @return if ball.position is in bounds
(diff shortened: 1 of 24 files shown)


rhkoulen:
One thing I'm not sure about in C++ syntax is namespaces. All of our position files are wrapped in a namespace strategy block. I initially did that in my file as well, but commenting it out didn't break anything, so I figured I would leave it out, to prevent my functions from polluting the namespace.

Basically, it's an hpp with a bunch of inline functions. Should those inline functions be defined inside namespace strategy? I don't see a reason, since any Position subclass that wants to use these can just include the hpp.

See lines 39 and 328. Should I uncomment those so all of the funcs are in the namespace? I don't know precisely what that does, so idk

CameronLyon (commented):
I love it 👍, this is needed badly. 
1. Comments are extremely helpful and point out future TODOs on this part of this implementation
2. Some small syntax issues (unnecessary elses, combining if statements, etc), but not enough to change them tbh. I thinking leaving them makes the code easier to read
3. Definitely need to make another PR request implementing this utility in other strategy positions, expanding the position_utils a little more (defense, free_kicker, penalty_player, goalie, etc). That can be for later though since this PR already tackles one of the most bloated files (offense).

I plan on testing this code soon and I'll approve the PR if there are no issues, good stuff.

CameronLyon:
Replying to your comment, namespace does two things from my experience:
1. Prevents name collisions with similarly named classes elsewhere in the project or in dependencies.
2. Keeps strategy-related code logically grouped, so Position, Offense, Defense, Goalie, all live under strategy.

This is more preference than a requirement on whether or not you want to include it, since removing it doesn't brick the codebase. I think including it would cause all of your functions to require something like "position_utils::FUNCTION_NAME" rather than simply saying "FUNCTION_NAME" to reference the position_utils. So do you want to:
- Leave it how it is and simplify how we reference position_utils? Might cause more confusion on where certain functions are being called but will avoid using "position_utils::" everytime we want to call it
- Bring namespace in and require all references to it to use "position_utils::FUNCTION_NAME"? While slightly annoying, does make it more understandable of where the function came from

I'd say leave it how it is, but if I'm wrong about how I see this and Aalind has a better reason to switch, then we can go that path. I don't mind either tbh

rhkoulen:
I mean, it builds and runs with and without the namespace. Like, when I had namespace strategy { ... } before I commented it out, I didn't need to do position_utils::FUNCTION_NAME, I just did FUNCTION_NAME. And it built just the same after commenting it out. I really don't know how that's possible, because that's what I thought it was doing as well. Is it something to do with the fact that I defined them all as inline functions?

It may also be inappropriate for these to be inline functions. To my understanding, the compiler deals with all that, but inline is supposed to be faster because there's no linking, it just shoves in the code. But some of these functions loop a lot so idk

Need someone who knows C++ semantics to look at this.

CameronLyon (approved):
Tested it, everything works as intended.

src/rj_strategy/include/rj_strategy/agent/position_utils.hpp
```diff
@@ -0,0 +1,328 @@
...
+
+// Example
+/**
+ * @brief Euclidean distance between two points.
+ *
+ * @param a The first point.
+ * @param b The second point.
+ * @return the Euclidean distance between a and b
+ */
+inline double distance(const rj_geometry::Point& a, const rj_geometry::Point& b) {
```
rishiso: The `Point` class has a built-in `dist_to(point)` function. Can we use this instead to avoid adding a redundant function?

src/rj_strategy/include/rj_strategy/agent/position_utils.hpp
```diff
@@ -0,0 +1,328 @@
...
+/**
+ * @brief Gets a good shot worth taking, given our pratical realities.
+ * Use this function and rest assured it will go in on an undefended goal.
+ *
+ * @param world_state (often named last_world_state_ in Position subclasses)
+ * @param field_dimensions (often named field_dimensions_ in Position subclasses)
+ * @return a point to aim at for the good shot
+ *   * currently just returns the center of the goal
+ */
+inline rj_geometry::Point calculate_a_shot([[maybe_unused]] const WorldState* world_state,
```
rishiso, line 158: Doesn't look like this function is used anywhere. Can we remove it?

src/rj_strategy/include/rj_strategy/agent/position_utils.hpp
```diff
@@ -0,0 +1,328 @@
...
+ * @param ignore_posts whether to count the goalposts as candidate shots (you should use True,
+ * aiming for the goalposts is stupid)
+ * @return a point to aim at for the best shot
+ */
+inline rj_geometry::Point calculate_best_shot(const WorldState* world_state,
+                                              const FieldDimensions& field_dimensions,
+                                              double granularity, bool ignore_posts) {
+    // Geometry
+    rj_geometry::Point their_goal_pos =
+        field_dimensions.their_goal_loc();  // returns center of goal
```
rishiso:
```suggestion
    const rj_geometry::Point& their_goal_pos =
        field_dimensions.their_goal_loc();  // returns center of goal
```

src/rj_strategy/include/rj_strategy/agent/position_utils.hpp
```diff
@@ -0,0 +1,328 @@
...
+    if (granularity <= 0.0) {               // protection
+        SPDLOG_ERROR(
+            "Invalid granularity value passed into calculate_best_shot, must use a positive "
+            "float.");
+        return their_goal_pos;
+    }
+    double goal_width = field_dimensions.goal_width();
```
rishiso:
```suggestion
    const double goal_width = field_dimensions.goal_width();
```

src/rj_strategy/include/rj_strategy/agent/position_utils.hpp
```diff
@@ -0,0 +1,328 @@
...
+    rj_geometry::Point negXmost_shot = their_goal_pos - rj_geometry::Point(goal_width / 2.0, 0.0);
```
rishiso: Can we call these variables `left_bound` and `right_bound` maybe?

src/rj_strategy/include/rj_strategy/agent/position_utils.hpp
```diff
@@ -0,0 +1,328 @@
...
+// TODO: overload these functions so that we can have some default possession_radius, such that
+// people using it don't need to turn on they brain
+/**
+ * @brief Determines whether the enemy has possession of the ball.
+ *
+ * @param world_state (often named last_world_state_ in Position subclasses)
+ * @param possession_radius the distance at which a robot is defined to "have" the ball (m)
+ * @return do they have ball
+ */
+inline bool they_have_ball(const WorldState* world_state, double possession_radius) {
```
rishiso:
```suggestion
inline bool enemy_has_ball(const WorldState* world_state, double possession_radius) {
```
Making it a bit more clear who "they" is

src/rj_strategy/include/rj_strategy/agent/position_utils.hpp
```diff
@@ -0,0 +1,328 @@
...
+}
+
+/**
+ * @brief Determines whether a specific robot has possession of the ball.
+ *
+ * @param world_state (often named last_world_state_ in Position subclasses)
+ * @param possession_radius the distance at which a robot is defined to "have" the ball (m)
+ * @return does it have ball
+ */
+inline bool robot_has_ball(const WorldState* world_state, const RobotState& robot,
```
rishiso: Maybe we could use `robot_has_ball()` in `they_have_ball()` and `we_have_ball()`?

src/rj_strategy/include/rj_strategy/agent/position_utils.hpp
```diff
@@ -0,0 +1,328 @@
...
+                           double possession_radius) {
+    rj_geometry::Point ball_pos = world_state->ball.position;
+    rj_geometry::Point robot_pos = robot.pose.position();
+    if (distance(ball_pos, robot_pos) < possession_radius) {
+        return true;
+    } else {
+        return false;
+    }
```
rishiso:
```suggestion
    return distance(ball_pos, robot_pos) < possession_radius);
```

src/rj_strategy/include/rj_strategy/agent/position_utils.hpp
```diff
@@ -0,0 +1,328 @@
...
+ * @brief Provides a good suggestion for kick speed, designed for passing.
+ *
+ * Motion intents take an integer from [0,15]. This provides that.
+ *
+ * @param distance_to_target distance from kicker to target (m)
+ * @param intended_velo_at_target the velocity you want the ball to be going when it reaches the
+ * target (m/s)
+ * @return an int to shove in the motion command
+ */
+inline int calculate_kick_speed(double distance_to_target,
```
rishiso: I think we should remove the unused argument for now and then add it back in once it is actually used. It makes things a little confusing

src/rj_strategy/src/agent/position/offense.cpp
```diff
@@ -63,7 +63,7 @@ Offense::State Offense::next_state() {
                 return SEEKING_START;
             }
             if (has_open_shot() || timed_out()) {
-                target_ = calculate_best_shot();
+                target_ = calculate_best_shot(last_world_state_, field_dimensions_, 0.04, true);
```
rishiso: Can we make 0.04 a constant?

src/rj_strategy/src/agent/position/solo_offense.cpp
```diff
@@ -8,175 +8,128 @@ SoloOffense::SoloOffense(Position&& other) : Position{std::move(other)} {
...
             auto kick_cmd =
-                planning::MotionCommand{"line_kick", target, planning::FaceTarget{}, true};
+                planning::MotionCommand{"line_kick", kick_target_, planning::FaceTarget{}, true};
             intent.motion_command = kick_cmd;
             intent.shoot_mode = RobotIntent::ShootMode::KICK;
             intent.trigger_mode = RobotIntent::TriggerMode::ON_BREAK_BEAM;
-            intent.kick_speed = 4.0;
-            counter_++;
-            if (counter_ > 15) {
-                kick_ = false;
```
rishiso, line 125: Nice rewrite of state machine

rishiso (changes requested):

src/rj_strategy/include/rj_strategy/agent/position_utils.hpp
rhkoulen: I'm anticipating this being used in comp. Everyone thinks we should shoot for the most optimal shot, then the on-field realities present themselves. In Brazil comp, we were seeing crazy deviation, almost a quarter of the goal width every meter away.

src/rj_strategy/include/rj_strategy/agent/position_utils.hpp
rhkoulen: Yeah I made this function first as a placeholder to show someone and forgot to get rid of it

src/rj_strategy/include/rj_strategy/agent/position_utils.hpp
rhkoulen: Left and right are abstracted here, since we may be playing +y or -y side (which would flip which way negative X is), but I agree, I'll rename stuff for clarity. Perhaps `lower_bound` and `upper_bound`.

src/rj_strategy/include/rj_strategy/agent/position_utils.hpp
rhkoulen: I'll get all of the they and theirs in here

rhkoulen: I mean, the field dimension class uses their and they so idk actually

src/rj_strategy/include/rj_strategy/agent/position_utils.hpp
rhkoulen: Yeah good point, in case we ever redefine what "has" means, only need to change it there

src/rj_strategy/include/rj_strategy/agent/position_utils.hpp
rishiso: oh wait that is true, feel free to leave as is

src/rj_strategy/src/agent/position/offense.cpp
rhkoulen: Good idea, I've also applied it as a default value, so caller doesn't need to specify if they don't really care

src/rj_strategy/src/agent/position/solo_offense.cpp
rhkoulen: Thanks :-)

src/rj_strategy/include/rj_strategy/agent/position_utils.hpp
rhkoulen: I'll do a boilerplate implementation with values that match our estimates (and some numbers stolen from tigers), but we need to measure these things if we ever intend to do good passing.

rhkoulen: Does that look alright?

src/rj_strategy/include/rj_strategy/agent/position_utils.hpp
rhkoulen: I guess the comp software folks are perfectly capable of calling their_goal_loc(), so this is kinda useless as is, but I think it would be cool to have a function which generates a practical shot with some chance of being blocked, taking into account our measured rotational uncertainty, rather than generating the sim-optimal shot. i.e., I mostly want to keep this for the TODO message.

src/rj_strategy/include/rj_strategy/agent/position_utils.hpp
```diff
@@ -0,0 +1,321 @@
...
+ *
+ * @param distance_to_target distance from kicker to target (m)
+ * @param intended_velo_at_target ideal ball velocity at the target (m/s)
+ * @return an int to shove in the motion command
+ */
+inline int calculate_kick_speed(double distance_to_target, double intended_velo_at_target) {
+    // TODO: these numbers are imaginary; based on estimates, NEED to measure
+    // This could be a fun collab with the hardware subteam! They could build a light gate!
+    // This maps "kick power" to the initial velocity of the ball.
+    const double kick_speed_map[16] = {3.0, 3.5, 4.0, 4.5, 5.0, 5.5, 6.0,  6.5,
```
rishiso: Can we use `std::array` instead C-style arrays?

src/rj_strategy/include/rj_strategy/agent/position_utils.hpp
```diff
@@ -0,0 +1,321 @@
...
+    const double ball_deceleration = 1.75;  // ideally, this is parametrized in rqt
+
+    // Compute ideal velo (v_final^2 = v_initial^2 - 2ad).
+    double target_velocity = std::sqrt(std::pow(intended_velo_at_target, 2.0) +
+                                       2 * ball_deceleration * distance_to_target);
+    if (target_velocity > 6.5) {  // rules tolerance
+        target_velocity = 6.5;
+    }
+
+    // Return the first candidate that exceeds the needed velocity (err on faster passes).
```
rishiso, line 308: This implementation makes sense. For now, since we haven't tested it and may not have time to test it at comp, could you put this into another function and call it `calculate_kick_speed_smart()` or something? Either at comp or sometime next semester, we can test it and make it the new `calculate_kick_speed()` function.

rishiso (changes requested):
looks good but few last things

Squid5678:
LGTM, make sure to address Rishi's changes. I would make sure everyone going to comp knows about this shared utility to streamline debugging.

Squid5678 (approved):

src/rj_strategy/include/rj_strategy/agent/position_utils.hpp
rhkoulen: For sure, good catch
I'm not good with healthy C++ code

src/rj_strategy/include/rj_strategy/agent/position_utils.hpp
rhkoulen: I'm worried about hiding it or commenting it, lest it never be seen again, but I agree we shouldn't let untested stuff play. I just made it unreachable below a `return 4;`

rishiso (approved):
LGTM

PR #2529, opened 2026-04-14
Better build time

Squid5678:
## Problem
Build times take too long, and I don't think we can use Aiden's computer for our builds. Ideally, we can build the stack without it taking 1/4 of our time in the shop.

Currently, multiple packages are building the same source files multiple times. In particular, packages were built once into a shared library, and another time for each executable. This leads to absurdly slow build times.

## Changes Made
Instead of having each executable rebuild all the relevant packages, we now only compile each executable's entry point (think ``main.cpp``) and then link things against the package's library. This means that any shared sources are just compiled once.

One example to think of (this helped me understand things a bit more):
For something like ``rj_vision_filter``, we used to have something like this:

```cmake
add_executable(rj_vision_filter_node
    src/main.cpp
    src/ball/ball_bounce.cpp  
    src/camera/camera.cpp
    ...many other files...
  )
```

AND we would also have something like this:

```cmake
add_library(rj_vision_filter
    src/ball/ball_bounce.cpp   
    src/camera/camera.cpp
    ...again, many other files...
  )
```

The issue here is that the same files just get compiled twice. This sucks. This PR proposes the following:

```cmake
add_executable(rj_vision_filter_node
    src/main.cpp               
  )
```

Notice how we just compile the entry point. And we leave the ``add_library`` part the same:

```cmake
add_library(rj_vision_filter
    src/ball/ball_bounce.cpp 
    src/camera/camera.cpp
    ..the other files..
  )
```

Basically every package was updated, but thanks to Cameron for pointing out these specific issues:
- ``rj_radio`` and ``rj_referee`` use the same shared logic (``radio.cpp``, ``referee_base.cpp``, etc), which is duplicated. This is now just compiled once into the library.
- ``rj_param_utils`` and ``global_param_server_node`` was recompiling a bunch of library source files. These just compile their own entry point now.

Another thing I found after looking at some other CMakeLists (for other ROS packages), they use this thing called Unity builds, which is specified with ``UNITY_BUILD_ON``. From my understanding, this merges ``.cpp`` files into 1 compilation unit, so the compiler doesn't have to work with things on a per-file basis.

Another change overall is just removing a bunch of redundant ``target_include_directories`` and ``ament_target_dependencies``. Everything is just inherited with the library now being PUBLIC.

## Result
Using our current Docker container, running ``make perf_docker`` on a fresh build took about 1.5 minutes.

### DIFF
--- a/src/rj_benchmarking/CMakeLists.txt
+++ b/src/rj_benchmarking/CMakeLists.txt
@@ -41,6 +41,7 @@ set(BENCHMARKING_SRC
 add_library(${PROJECT_NAME} SHARED
     ${BENCHMARKING_SRC}
 )
+set_target_properties(${PROJECT_NAME} PROPERTIES UNITY_BUILD ON)
 
 target_include_directories(${PROJECT_NAME}
     PUBLIC
--- a/src/rj_benchmarking/CMakeLists.txt
+++ b/src/rj_benchmarking/CMakeLists.txt
@@ -60,24 +60,12 @@ ament_target_dependencies(${PROJECT_NAME}
 
 # Benchmarking Node
 add_executable(${PROJECT_NAME}_node
-    ${BENCHMARKING_SRC}
     src/main.cpp
 )
 
-target_include_directories(${PROJECT_NAME}_node
-    PUBLIC
-    $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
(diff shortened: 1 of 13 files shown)


shourikb (approved):

CameronLyon (approved):
Everything looks good at first glance, and bots move after compilation. tbh I'm not the best at CMake and we need this desperately.

PR #2530, opened 2026-06-22
Goalie sweeping from left to right

rishiso:
## Description
Changed goalie to sweep from left to right instead of tracking the ball. We observed several other teams doing this during the Peachtree Comp.

## Testing
Build and see that goalie is sweeping when in idle state.


src/rj_planning/src/planners/goalie_idle_path_planner.cpp
```diff
@@ -43,17 +43,14 @@ Trajectory GoalieIdlePathPlanner::plan(const PlanRequest& plan_request) {
...
-    double goalie_dist = 0.5;
-    rj_geometry::Point idle_pt = (ball_pos - goal_pt).norm();
-    idle_pt *= goalie_dist;
-    // TODO(Kevin): clamp y to 0 so goalie doesn't go backwards
+    const double sweep_amplitude = kSweepFraction * field_dimensions->goal_width() / 2.0;
+    const double seconds = RJ::Seconds(RJ::now().time_since_epoch()).count();
+    const double x = sweep_amplitude * std::sin(seconds * kSweepRate);
 
-    return idle_pt;
+    return rj_geometry::Point{x, kGoalLineOffset};
```
rhkoulen, line 53: Every time I try to swap teams in sim, it crashes, so I can't test this, but don't we use a fixed coordinate system? Wouldn't this Y value be problematic? What if we're defending the y=9 goal instead of the y=0 goal?

You could make a vector from our_goal_loc() to center_field_loc() and use that to offset.

But maybe the coordinates flip? Idk I literally can't test it

rhkoulen: What if this was causing the thing at Peachtree where we were attacking our own goal and defending theirs? Do you know what determines where X and Y are defined? Most importantly, does that differ from sim to real?

rishiso: I believe the planner operates entirely in team-relative coordinates. Notice how `our_goal_loc_` is hardcoded as 0, 0 in field dimensions. In comp, that might have happened because we didn't set +x properly maybe? I am not too sure since I wasn't there

rhkoulen: Ah yeah, that's fine then

Still no clue what was happening at comp then

rhkoulen (approved):

Squid5678 (approved):

PR #2534, opened 2026-06-26
Rules compliance and other fixes

rishiso:
## Description
Some fixes

- **Defender distance (opponent restart):** Added a `Ready` case in `planner_for_robot.cpp` so all robots keep `min_dist_from_ball = 0.5` during an opponent free kick / kick-off

- **Stop speed cap:** Lowered the STOP `max_robot_speed` from `1.5 → 1.4` (`planner_for_robot.cpp`) for headroom under the stop-speed foul threshold.

- **Double touch:** `RobotFactoryPosition` now locks the free-kick/kick-off kicker into `SmartIdle` after it kicks, releasing when play stops or another robot touches the ball (`another_robot_touched_ball()`).

- **Kick-off positioning:** During a kick-off, non-kicker robots get the opponent half + center circle as `local_obstacles`, keeping them in our half and out of the circle; the selected kicker is exempt.

- **Ball speed:** Goalie clear changed from full power `15.0` to `max_kick_speed()` in `goalie.cpp`

- **Offense stealing during STOP** `handle_stop()` swaps any `Offense` robot to `SmartIdle`, stopping the drive-into-padding / escape-obstacle oscillation around the ball during stop.

- **Position Labels:** All positions have labels in the form of `POSITION/STATE` now

## Testing
Verified each of the changes using game controller testing. Noticed a few other small issues (marking) but will fix these on comp branch

### DIFF
--- a/src/rj_strategy/include/rj_strategy/agent/position.hpp
+++ b/src/rj_strategy/include/rj_strategy/agent/position.hpp
@@ -99,6 +99,8 @@ public:
     // returns the current state of the robot
     virtual std::string get_current_state() = 0;
 
+    virtual std::string get_state_name() const { return "None"; }
+
     /**
      * @brief setter for time_left_
      */
@@ -351,7 +353,7 @@ protected:
 
     // Debug drawing support: shared across position swaps via move semantics
     std::shared_ptr<rj_drawing::RosDebugDrawer> debug_drawer_;
-    bool debug_draw_enabled_ = false;
+    bool debug_draw_enabled_ = true;
 
 private:
     /**
--- a/src/rj_strategy/src/agent/position/robot_factory_position.cpp
+++ b/src/rj_strategy/src/agent/position/robot_factory_position.cpp
@@ -42,7 +42,18 @@ std::optional<RobotIntent> RobotFactoryPosition::derived_get_task([
     // Every tick, update position based on PlayState
     update_position();
 
-    return current_position_->get_task(*last_world_state_, field_dimensions_, current_play_state_);
+    auto task =
+        current_position_->get_task(*last_world_state_, field_dimensions_, current_play_state_);
+
+    if (task.has_value() && current_play_state_.is_kickoff() &&
+        !client_handles_->kicker_picker->is_selected()) {
+        task->local_obstacles.add(
+            std::make_shared<rj_geometry::Rect>(field_dimensions_.their_half()));
+        task->local_obstacles.add(std::make_shared<rj_geometry::Circle>(
+            field_dimensions_.center_field_loc(), field_dimensions_.center_radius()));
+    }
+
+    return task;
 }
 
 void RobotFactoryPosition::process_play_state() {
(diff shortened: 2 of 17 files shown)


sanatd33 (approved):
brief lgtm

src/rj_strategy/src/agent/position/robot_factory_position.cpp
```diff
@@ -203,6 +237,24 @@ void RobotFactoryPosition::update_position() {
     }
 }
 
+bool RobotFactoryPosition::another_robot_touched_ball() const {
```
CameronLyon, line 240: I like this "locking" mechanism to prevent double-touching, but what happens if the vision doesn't recognize a robot touching the ball and the ball is loose/moving? This scenario might cause the robots to stay locked despite the ball moving. I'd say this is unlikely, but something to point out, maybe also account for the ball moving after a certain distance?

src/rj_strategy/include/rj_strategy/agent/position.hpp
```diff
@@ -351,7 +353,7 @@ class Position {
 
     // Debug drawing support: shared across position swaps via move semantics
     std::shared_ptr<rj_drawing::RosDebugDrawer> debug_drawer_;
-    bool debug_draw_enabled_ = false;
+    bool debug_draw_enabled_ = true;
```
CameronLyon, line 356: Making debug_drawer enabled on all positions could slow down the software. I checked with my computer and it doesn't seem to be a big issue, but maybe turn it off if the comp computer is having trouble during matches (or specify what positions you want to see).

src/rj_strategy/include/rj_strategy/agent/position.hpp
```diff
@@ -99,6 +99,8 @@ class Position {
     // returns the current state of the robot
     virtual std::string get_current_state() = 0;
 
+    virtual std::string get_state_name() const { return "None"; }
```
CameronLyon, line 102: You got this overrided with Offense, Goalie, and Defense properly, but there is nothing for other positions like FreeKicker, PenaltyPlayer, or SmartIdle. Less important roles yes, but still beneficial to know.

CameronLyon (approved):
Made a few comments, but more recommendations rather than outright necessities. I'll approve it if you want to get it on the main branch asap for comp, just look over what I noticed :)

src/rj_strategy/include/rj_strategy/agent/position.hpp
rishiso: Yes I agree, will make this more granular if needed

src/rj_strategy/src/agent/position/robot_factory_position.cpp
rishiso: Based on the rules, even if the ball is loose, the kicking robot still can't touch it. The only condition to unlock it is if another robot has touched the ball or a stop is called.

rishiso: Worst case, if vision is not working, I think stops are called frequently enough such that it shouldn't be too much of an issue

src/rj_strategy/include/rj_strategy/agent/position.hpp
rishiso: Many of these positions are stateless, so I didn't include them for now. I added for PenaltyPlayer in the comp branch

PR #2548, opened 2026-09-09
propagated geometry_msg README to every package

CameronLyon:
This should rewrite every README doc (except for geometry_msg) to include only CURRENT features of the package instead of comparing the features from ros2 branch. Hopefully this should be a quick merge so long as I did this right.


N8BWert (approved):
Changes look good to me.

shourikb (approved):
LGTM

PR #2549, opened 2026-09-14
Migration of `rj_constants`

rhkoulen:
It seems a bit silly to have a package rj_constants providing some constants when rj_common is already providing other constants to every package. I want to remove several of these items, but I'll deal with that refactoring mission in the next package. Basically, if a constant has a narrow scope, it should be defined only in the package for which it is relevant—should that ever create a circular inclusion, we can just export the constant back up to rj_common. If a constant has a broad scope, it should be clearly filed in a relevantly-named folder in rj_common.

Most of everything is being moved to rj_common. However, I'm already making one move to rj_geometry. Since rj_geometry depends on kRobotRadius for Shape::hit, and rj_common depends on rj_geometry for Shape, kRobotRadius cannot go into rj_common. I don't want to move too much out of rj_common, but this is necessary (and it makes sense anyways). rj_geometry is higher on the topography so no issues there.

Files Moved:
 - <rj_constants/topic_names.hpp> : Moved over to <rj_common/topic_names.hpp>
 - <rj_constants/constants.hpp> : Any constants that were geometry related were moved to <rj_geometry/ball.hpp> and <rj_geometry/bot.hpp>, whereas the list of other miscellaneous constants that were tossed into this file over the years were moved to <rj_constants/game_constants.hpp>. That's probably a bad fit, but they can be refiled during the rj_common refactor.
 
 Files Changed:
  - Anything that previously included rj_constants now either includes rj_common or rj_geometry, where appropriate. I simply removed all references to rj_constants, then went 1-by-1 resolving build issues. If this is a bad approach insofar as manifesting latent inclusion chains, let me know and I'll take a more granular approach putting an #include statement wherever there's a relevant constant used (the stack had previously not done that everywhere, so I didn't bother).


rhkoulen:
To any peepers, this is still a draft PR—no need for review yet. I kinda just wanted to see the diff in the Github UI. It's basically done, but I have a few things I still need to check (and style of course).

src/rj_strategy/include/rj_strategy/agent/position/defense.hpp
rhkoulen, line 1: @ reviewer, why did all of the position files each include rj_constants/constants.hpp separately? I get that many of them needed kRobotRadius, but why not just include it in position.hpp? All of the subsequent position implementations include position.hpp, so they'll get kRobotRadius from the inevitable inclusion chain.

Is it just for stability's sake?
Should I put `#include <rj_geometry/bot.hpp>` in each of these?

src/rj_common/include/rj_common/game_constants.hpp
```diff
@@ -0,0 +1,33 @@
...
+
+/** constants for dot patterns */
+inline constexpr float kDotsSmallOffset = 0.035;
+inline constexpr float kDotsLargeOffset = 0.054772;
+inline constexpr float kDotsRadius = 0.02;
+
+/** constants for planning */
+inline constexpr double kAvoidBallDistance = 0.01;
+
+inline const std::string kTeamNameLower = "robojackets";
```
sanatd33, line 32: not sure the purpose of this? but i can see this as a "check its usage in the `rj_common` rewrite" thing

src/rj_common/include/rj_common/topic_names.hpp
```diff
@@ -4,26 +4,27 @@
 
 /**
  * \file This file contains all the topic name strings to avoid typos.
+ * 2026 GRAND REFACTOR QUESTIONS
```
sanatd33, line 7: not sure the purpose of these comments in the codebase?

src/rj_strategy/include/rj_strategy/agent/position/defense.hpp
sanatd33: i'm assuming someone manually added it by mistake once and the AI writing all our code (or the human copy pasting) cloned it into every file i think

src/rj_common/include/rj_common/game_constants.hpp
sanatd33: nvm didnt read your PR body

> I want to remove several of these items, but I'll deal with that refactoring mission in the next package

src/rj_common/include/rj_common/topic_names.hpp
rhkoulen: Note for me not to forget when I do rj_common

sanatd33: i think i prefer not having these types of comments committed, copying what i commented in pooja's pr
> generally i prefer only having functionality-explaining comments, so if this is a todo would pref making this a task on clickup/if we want to us github issues/something else? @CameronLyon what's your opinion

src/rj_common/include/rj_common/game_constants.hpp
N8BWert, line 1: I like these changes here

src/rj_common/include/rj_common/topic_names.hpp
N8BWert, line 1: Can you add a `/` before each of the topic names.  I don't think this will break anything, but right now if we namespace any of our nodes things will break

src/rj_geometry/include/rj_geometry/ssl_ball_constants.hpp
N8BWert, line 1: can we name this file ssl_ball_constants.hpp

src/rj_geometry/include/rj_geometry/ssl_robot_constants.hpp
N8BWert, line 1: can we name this file ssl_robot_constants.hpp

N8BWert (changes requested):
Overall changes look good to me.  I left a few small semantic comments, but the more important change in my opinion is to start topic names with a slash so they can't get namespaced in any weird ways.

N8BWert (approved):
Looks good to me!

sanatd33 (approved):

PR #2551, opened 2026-09-14
Remove Logger

sanatd33:
## Description
Removes the Logger functionality and associated protobuf files. Uses a `LiveFrame` struct to store UI info.

Resolves [ClickUp card](https://app.clickup.com/t/2209401/wdvf3aa7az)

## Steps to Test
1. Run Sim

**Expected result:** The UI is functional

## Key Files to Review
 * `live_view.hpp`
 * `context.hpp`
 * `main_window.hpp`
 * `debug_drawer.hpp`
 * `field_view.cpp`

## Review Checklist


src/rj_common/include/rj_common/context.hpp
```diff
@@ -98,14 +55,74 @@ struct Context {
...
 
-    Logs logs;
     std::string behavior_tree;
+
+    std::vector<std::shared_ptr<rj_common::LiveFrame>> frames;
+
+    RJ::Time start_time;
 };
+
+static std::shared_ptr<rj_common::LiveFrame> create_log_frame(const Context& context) {
```
sanatd33, line 72: oops this should probably be `create_live_frame`

src/rj_common/include/rj_common/context.hpp
```diff
@@ -98,14 +55,74 @@ struct Context {
...
+    // Add everything to the log frame.
```
sanatd33, line 73: these comments probably arent necessary?

src/rj_geometry/include/rj_geometry/point.hpp
```diff
@@ -69,12 +64,6 @@ class Point {
      */
     [[nodiscard]] QPointF to_q_point_f() const { return QPointF(x(), y()); }
```
N8BWert, line 65: This isn't for this PR, but this method probably doesn't need to exist.  It introduces a weird dependency to rj_geometry and its just a single line.

src/rj_common/include/rj_common/live_frame.hpp
```diff
@@ -0,0 +1,119 @@
...
+};
+
+struct LiveBall {
+    const rj_geometry::Point& pos() const { return position; }
+    const rj_geometry::Point& vel() const { return velocity; }
+    rj_geometry::Point position;
+    rj_geometry::Point velocity;
+};
+
+struct LiveFrame {
```
N8BWert, line 54: I think `LiveFrame` is a bit of an odd name.  It doesn't really describe what the frame does.  I think something like `UIFrame` with a comment describing its usage

src/rj_common/include/rj_common/live_frame.hpp
```diff
@@ -0,0 +1,119 @@
...
+    rj_geometry::Point position;
+    double heading = 0;
+    bool has_ball = false;
+    bool kicker_ok = true;
+    float battery = 0;
+    std::vector<MotorStatus> motors;
+    std::vector<DebugText> texts;
+};
+
+struct LiveBall {
```
N8BWert, line 47: I think if we're going to change the name of `LiveFrame` we might as well change this to `UIBall` (not happy with that name)

src/rj_common/include/rj_common/live_frame.hpp
N8BWert, line 1: For structs in C++, all fields are inherently constant so we don't need all of the const getters.  I think having the getters blurs the line between a struct and a class so if we want to conform to a struct being a simple data storage we probably shouldn't have trivial getter or setter methods

src/rj_common/include/rj_common/live_frame.hpp
```diff
@@ -0,0 +1,119 @@
...
+#include "rj_common/control/motion_setpoint.hpp"
+#include "rj_common/debug_drawings.hpp"
+#include "rj_common/radio/robot_status.hpp"
+#include "rj_common/robot_intent.hpp"
+#include "rj_common/time.hpp"
+#include "rj_common/world_state.hpp"
+
+namespace rj_common {
+
+enum BallSenseStatus { NoBall, HasBall, Dazzled, Failed };
```
N8BWert, line 18:
```suggestion
enum BallSenseStatus {
    kNoBall = 0,
    kHasBall,
    kDazzled,
    kFailed
};
```
According to the google cpp style guide enums should be named like constants and should have a default value

src/rj_common/include/rj_common/live_frame.hpp
```diff
@@ -0,0 +1,119 @@
...
+enum MotorStatus { Good, Fault };
```
N8BWert, line 19:
```suggestion
enum MotorStatus {
    kGood = 0,
    kFault
};
```
See comments on R18

src/rj_common/include/rj_common/live_frame.hpp
```diff
@@ -0,0 +1,119 @@
...
+
+struct LiveRobot {
```
N8BWert, line 21: Same comment on the `LiveRobot` name

src/rj_common/include/rj_common/live_frame.hpp
```diff
@@ -0,0 +1,119 @@
...
+    bool blue = true;
+    bool defend_plus = false;
+    bool use_our = true;
+    bool use_opponent = true;
+    int manual = -1;
+    std::string blue_name;
+    std::string yellow_name;
+    DebugDrawFrame debug_draw_frame;
+
+    static void fill_robot(LiveRobot* out, int shell_id, RobotState const& state,
```
N8BWert, line 99: Can we have a docstring on this method.  Its not entirely obvious what this method does from the name and signature

N8BWert, line 99:
```suggestion
    static void fill_robot(LiveRobot* out, int shell_id, const RobotState& state,
```
Our codebase typically uses const T& name

src/rj_common/include/rj_common/live_frame.hpp
```diff
@@ -0,0 +1,119 @@
...
+                           RobotStatus const* status) {
```
N8BWert, line 100: Is there a way we can do this without raw pointers.  I'd prefer smart pointers or references if possible

src/rj_common/include/rj_common/live_frame.hpp
N8BWert: As another note, I think `LiveFrame` should probably be a class.  I think it does more than group relevant data so it should probably be a class.

src/rj_common/include/rj_common/debug_drawings.hpp
```diff
@@ -0,0 +1,100 @@
...
+    bool center() const { return center_; }
+
+    int layer_ = -1;
+    uint32_t color_ = 0;
+    rj_geometry::Point pos_;
+    std::string text_;
+    bool center_ = true;
+};
+
+struct DebugDrawFrame {
```
N8BWert, line 82: This should definitely be a class

src/rj_common/include/rj_common/debug_drawings.hpp
N8BWert, line 1: I think with this file.  You can either get rid of the getters or leave struct members as public.  Right now it kind of blurs the line between a struct and class.  `DebugDrawFrame` should definitely be a class though

src/rj_common/src/debug_drawer.cpp
N8BWert, line 1: I like these changes.  If you can update the datatypes this code looks good to me

src/rj_common/include/rj_common/context.hpp
```diff
@@ -98,14 +55,74 @@ struct Context {
...
-    std::vector<SSL_WrapperPacket> raw_vision_packets;
 
     WorldState world_state;
 
     FieldDimensions field_dimensions;
 
     GameSettings game_settings;
 
-    Logs logs;
     std::string behavior_tree;
```
N8BWert, line 65: Do we use this property

src/rj_common/include/rj_common/context.hpp
```diff
@@ -98,14 +55,74 @@ struct Context {
...
+    frame->debug_draw_frame = context.debug_drawer.published_frame();
+
+    frame->blue = context.blue_team;
+
+    // Our robots
+    for (size_t shell = 0; shell < kNumShells; shell++) {
+        const auto& state = context.world_state.our_robots.at(shell);
+        const auto& status = context.robot_status.at(shell);
+
+        if (RJ::now() - status.timestamp < RJ::Seconds(0.5)) {
```
shourikb: Where does 0.5 come from

src/rj_ui/include/rj_ui/field_view.hpp
```diff
@@ -115,7 +111,7 @@ class FieldView : public QWidget {
     // How many degrees to rotate text so it shows up the right way on screen
     int _textRotation{};
 
-    const std::vector<std::shared_ptr<Packet::LogFrame> >* _history{};
+    const std::vector<std::shared_ptr<rj_common::LiveFrame> >* _history{};
```
N8BWert, line 114: Is there a reason we need a const pointer here.  Is there any way the field view can just own the history vector?

src/rj_ui/src/field_view.cpp
```diff
@@ -187,51 +183,16 @@ void FieldView::paintEvent(QPaintEvent* /*e*/) {
 
 void FieldView::drawWorldSpace(QPainter& p) {
     // Get the latest LogFrame
-    const LogFrame* frame = _history->back().get();
+    const rj_common::LiveFrame* frame = _history->back().get();
 
     // Draw the field
     drawField(p, frame);
```
N8BWert, line 189: Is there any way this method can take a const reference?  If not, can you place `_history->back().get()` in the place of frame.  Raw pointers are worrisome

N8BWert (changes requested):
Left a few comments.  Anywhere possible can you see if we can use const references instead of const pointers (if not it works the way it is).  Using raw pointers in our stack always worries me (we have done some weird stuff in the past).

PR #2553, opened 2026-09-16
Rj topic utils cleaning

pd2025:
## Description
- removed unused message queue templates in message_queue.hpp and async_message_queue.hpp
- replaced instances of message queues in rj_vision_filter and rj_ui with direct subscriptions to message topics
- filled out the documentation template

Resolves # or https://app.clickup.com/t/2209401/wdvf3a9383

## Design Documents
[Link](link-to-design-doc)

## Steps to Test
Testing the behavior of the sim:
1. run the UI
2. observe behavior:
     a. robots should take the same amount of time to transition between stealing and shooting states as they did before changes (6 - 8 seconds)
     b. path planning behavior should be the same as it was before changes were made

**Expected result:**
Everything should behave as normal after these changes

## Key Files to Review
rj_vision_filter
 * vision_filter.cpp
 * vision_filter.hpp

rj_ui
 * processor.hpp
 * processor.cpp

rj_topic_utils
 * message_queue.hpp
 * async_messge_queue.hpp

## Review Checklist

- [ ] Step 1
- [ ] Step 2


src/rj_control/include/rj_control/motion_control_node.hpp
```diff
@@ -9,7 +9,7 @@
 #include <rj_msgs/msg/world_state.hpp>
 #include <rj_param_utils/global_params.hpp>
 #include <rj_param_utils/ros2_local_param_provider.hpp>
-#include <rj_topic_utils/async_message_queue.hpp>
+// #include <rj_topic_utils/async_message_queue.hpp>
```
sanatd33: pref delete instead of comment out

src/rj_vision_filter/src/vision_filter.cpp
```diff
@@ -17,18 +16,23 @@ DEFINE_FLOAT64(kVisionFilterParamModule, publish_hz, 60.0,
...
     : rclcpp::Node{"vision_filter", options},
       config_client_{this},
-      team_color_queue_{this, referee::topics::kTeamColorTopic},
       param_provider_{this, kVisionFilterParamModule} {
     // Create a timer that calls predict on all of the Kalman filters.
     const std::chrono::duration<double> predict_timer_period(PARAM_vision_loop_dt);
     auto publish_callback = [this]() { publish_state(); };
     publish_timer_ = create_wall_timer(predict_timer_period, publish_callback);
 
+    // Create a subscripter for TeamColorMsg
```
sanatd33: dont think this comment is necessary

src/rj_ui/include/rj_ui/processor.hpp
```diff
@@ -152,11 +154,22 @@ class Processor {
 
     // ROS2 temporary modules
     using WorldStateMsg = rj_msgs::msg::WorldState;
+
     using AsyncWorldStateMsgQueue = rj_topic_utils::AsyncMessageQueue<
         WorldStateMsg, rj_topic_utils::MessagePolicy::kQueue, 1>;
 
+    // replace this w a sub and callback
```
sanatd33: not sure what these comments are saying? what are we replacing with a sub and callback? 
generally i prefer only having functionality-explaining comments, so if this is a todo would pref making this a task on clickup/if we want to us github issues/something else?

sanatd33: @CameronLyon what's your opinion

pd2025: I wrote that while I was working and I meant to delete it but forgot, I'll delete it now.

src/rj_control/include/rj_control/motion_control_node.hpp
pd2025: That was another thing that I meant to delete and forgot about. Will delete now.

src/rj_ui/include/rj_ui/processor.hpp
```diff
@@ -10,6 +10,8 @@
 
 #include <QMutexLocker>
 #include <rclcpp/executors/single_threaded_executor.hpp>
+#include <rclcpp/rclcpp.hpp>
+#include <rclcpp_action/rclcpp_action.hpp>
```
CameronLyon, line 14: Is this dependency used anywhere? rclcpp_action is the library for ROS 2 actions, like the request/feedback/result pattern for long-running goals like "drive to this point and report progress." This PR simply uses ROS2 subscription based updates, avoiding the need of using this from the get-go. Additionally, I don't see any dependency declarations within the package.xml for rj_ui. I believe this is a delete line if its truly unused.

src/rj_ui/include/rj_ui/processor.hpp
CameronLyon: I agree. Its completely fine if you want to leave comments for yourself, but when it comes to making a PR with those comments, you have to make sure that those comments are understandable or deleted to simplify everything. Thankfully its an easy fix so np honestly.

Additionally, UniquePtr is a std::unique_ptr, so this is a pointer that starts out null. Before your PR, Processor's constructor filled it in with std::make_unique<AsyncWorldStateMsgQueue>(...) and run() called world_state_queue_->get() each loop. 

What's left now is a pointer that is never assigned and never read. It sits there as a null pointer for the lifetime of every Processor. I think it should just be deleted.

src/rj_vision_filter/include/rj_vision_filter/vision_filter.hpp
```diff
@@ -92,15 +92,10 @@ class VisionFilter : public rclcpp::Node {
...
-        rj_topic_utils::MessageQueue<TeamColorMsg, rj_topic_utils::MessagePolicy::kLatest>;
-
     config_client::ConfigClient config_client_;
 
-    /**
-     * @brief Message Queue for TeamColor that takes the latest message.
-     */
-    TeamColorMsgQueue team_color_queue_;
+    rclcpp::Subscription<TeamColorMsg>::SharedPtr team_color_sub_;
+    std::atomic<bool> us_blue_{true};
```
CameronLyon, line 98: I believe this might cause issues that causes the publisher to default towards blue whenever the game controller / referee is disconnected. 

My understanding is:
Before the first team-color message lands, VisionFilter publishes a WorldState asserting we're blue. If we're actually yellow, "our robots" and "their robots" are swapped for every downstream consumer. This should be quickly fixed once referee is connected and config_client is loaded. If either one is never connected though, we are constantly publishing fake info onto the UI and everything else that is reliant on this subscription. The original solution was to refuse to post any info until referee/config_client were there, which I think should be the play. 

Let me know if I'm misunderstanding this though. Could be overthinking and I'll try to test this later.

src/rj_topic_utils/README.md
CameronLyon, line 1: Thank you for filling this out :)

src/rj_ui/include/rj_ui/processor.hpp
```diff
@@ -152,11 +154,17 @@ class Processor {
 
     // ROS2 temporary modules
     using WorldStateMsg = rj_msgs::msg::WorldState;
+
     using AsyncWorldStateMsgQueue = rj_topic_utils::AsyncMessageQueue<
         WorldStateMsg, rj_topic_utils::MessagePolicy::kQueue, 1>;
```
CameronLyon, line 159: This might be an issue. You are making an AsyncMessageQueue that brings (WorldStateMsg, rj_topic_utils::MessagePolicy::kQueue, 1), basically trying to make a 1-sized queue for WorldState messages. [https://github.com/RoboJackets/robocup-software/pull/2553/changes#diff-9082a4c18adc31db2d1888bfa8af988dc6d5644954b01d1bb811e66a8a4ffbf9L49](url) is where the issue lies as the size given to this queue does not actually get passed, making this parameter useless and creating I think a boundless queue. This can be fixed by simply deleting it! This is being used nowhere now that we are using ROS2 subscriptions, but maybe look at fixing the queue struct from the link.

src/rj_topic_utils/include/rj_topic_utils/async_message_queue.hpp
CameronLyon, line 49: Can we simply add the size parameter? Or revert the deletion of the more specific AsyncMessageQueue? Or is this being used at all and we can just delete this struct all together?

CameronLyon (commented):
Made comments. Overall the new structure is there, but there are some lingering comments/queue systems left with nothing to do, just dead code. Unless there's anything important, I'd say delete world_state_queue_, rclcpp_action/rclcpp_action.hpp, AsyncWorldStateMsgQueue. The main two issues I see are us_blue_{true}, and the struct formatting of the queue system you've simplified earlier. I plan on testing this on a simulated environment soon. I like it! 👍

src/rj_ui/include/rj_ui/processor.hpp
pd2025: I just double-checked and you're right, they're not used anywhere. I'll go ahead and delete those.

pd2025: While I'm at it I'll check and see if there's anything else that can be deleted too.

PR #2555, opened 2026-09-16
Removing RJ param utils from rj_joystick and rj_vision_filter

shourikb:
## Description
Uses ros parameters instead of param_utils for rj_joystick and rj_vision_filter

Part of rj_param_utils cleanup

## Steps to Test
Build and run sim, rj_vision_filter has unit tests that pass


src/rj_joystick/src/manual_control_node.cpp
```diff
@@ -66,6 +58,62 @@ ManualControlNode::ManualControlNode()
     });
 }
 
+void ManualControlNode::declare_params() {
```
yuvrajdhadwal, line 61: ??? Just put this in the node and remove this lambda and just declare_parameter the way it is defined in `radio.cpp` here

[example](https://github.com/RoboJackets/robocup-software/commit/4f63f3a45499eebb8e5eea4bcd2fb18cb4511bde)

src/rj_joystick/src/manual_control_node.cpp
```diff
@@ -66,6 +58,62 @@ ManualControlNode::ManualControlNode()
...
+    get_parameter("max_rotation_speed", params_.max_rotation_speed);
+    get_parameter("max_damped_rotation_speed", params_.max_damped_rotation_speed);
+    get_parameter("max_translation_speed", params_.max_translation_speed);
+    get_parameter("max_damped_translation_speed", params_.max_damped_translation_speed);
+    get_parameter("kick_power_increment", params_.kick_power_increment);
+    get_parameter("dribble_power_increment", params_.dribble_power_increment);
+    
+}
+
+rcl_interfaces::msg::SetParametersResult ManualControlNode::on_param_change(
```
yuvrajdhadwal, line 98: is any of this actually necessary? i am pretty sure none of this stuff actually materially changes mid match

src/rj_joystick/src/manual_control_node.cpp
```diff
@@ -66,6 +58,62 @@ ManualControlNode::ManualControlNode()
...
 }
 
+void ManualControlNode::declare_params() {
+    auto declare = [this](const std::string& name, auto default_value, const std::string& desc) {
+        rcl_interfaces::msg::ParameterDescriptor d;
+        d.description = desc;
+        declare_parameter(name, default_value, d);
+    };
+
+    declare("use_field_oriented_drive", params_.use_field_oriented_drive,
```
jvogt23, line 68: May not need to declare the parameter here if the Node has 
- automatically_declare_parameters_from_overrides
- allow_undeclared_parameters
These settings would allow you to remove all of these declare statements
Then you would use get_parameter_or() to get the params and set them to variables within the class. 

You don't have to do any of this, though. It's just a suggestion.

src/rj_joystick/src/manual_control_node.cpp
```diff
@@ -66,6 +58,62 @@ ManualControlNode::ManualControlNode()
...
+    
+}
+
+rcl_interfaces::msg::SetParametersResult ManualControlNode::on_param_change(
+    const std::vector<rclcpp::Parameter>& parameters) {
+    for (const auto& p : parameters) {
+        if (p.get_name() == "use_field_oriented_drive") params_.use_field_oriented_drive = p.as_bool();
+        else if (p.get_name() == "damped_translation") params_.damped_translation = p.as_bool();
+        else if (p.get_name() == "damped_rotation") params_.damped_rotation = p.as_bool();
+        else if (p.get_name() == "max_rotation_speed") params_.max_rotation_speed = p.as_double();
```
jvogt23, line 104: Make certain that all of these are gated on the correct ranges - Example, fail if something that should always be positive is negative

src/rj_joystick/src/manual_control_node.cpp
```diff
@@ -66,6 +58,62 @@ ManualControlNode::ManualControlNode()
     });
 }
 
+void ManualControlNode::declare_params() {
+    auto declare = [this](const std::string& name, auto default_value, const std::string& desc) {
+        rcl_interfaces::msg::ParameterDescriptor d;
```
jvogt23, line 63: This lambda isn't really needed and seems to have been added for brevity. Are you sure you want to keep this?

src/rj_joystick/src/manual_control_node.cpp
```diff
@@ -105,7 +153,9 @@ void ManualControlNode::publish(int robot_id, const ControllerCommand& command)
                                                     .velocity_x_mps(command.translation.x())
                                                     .velocity_y_mps(command.translation.y())
                                                     .velocity_z_radps(command.rotation));
-    uint8_t trigger_mode = PARAM_kick_on_break_beam
+    bool kick_on_break_beam = false;
```
jvogt23, line 156: Params can be saved to instance variables, with the param change callback updating those instance variables. That way, the variable will change when the param does, but you don't have to call get_parameter() constantly.

src/rj_joystick/src/manual_control_node.cpp
```diff
@@ -117,8 +167,12 @@ void ManualControlNode::publish(int robot_id, const ControllerCommand& command)
         trigger_mode = rj_msgs::msg::ManipulatorSetpoint::TRIGGER_MODE_STAND_DOWN;
     }
 
-    double kick_speed = lerp(soccer::robot::PARAM_min_kick_speed,
-                             soccer::robot::PARAM_max_kick_speed, command.kick_power);
+    double min_kick_speed = 0.0;
+    double max_kick_speed = 15.0;
+    get_parameter("min_kick_speed", min_kick_speed);
```
jvogt23, line 172: Can make instance variable with callback

src/rj_param_utils/config/real_params.yaml
```diff
@@ -80,6 +80,12 @@
     team_name: RoboJackets
     use_sim_time: false
 
+/manual_control:
```
jvogt23, line 83: Are these all the parameters on manual_control?

src/rj_param_utils/config/sim_params.yaml
```diff
@@ -80,6 +80,12 @@
     team_name: RoboJackets
     use_sim_time: false
 
+/manual_control:
+  ros__parameters:
+    use_sim_time: false
```
jvogt23, line 85: Do we even do manual control in sim?

src/rj_vision_filter/src/camera/camera.cpp
```diff
@@ -25,10 +15,12 @@ Camera::Camera(int camera_id)
 bool Camera::get_is_valid() const { return is_valid_; }
 
 void Camera::process_ball_bounce(const std::vector<WorldRobot>& yellow_robots,
-                                 const std::vector<WorldRobot>& blue_robots) {
+                                 const std::vector<WorldRobot>& blue_robots,
+                                 const VisionFilterParams& params_) {
```
jvogt23, line 19: I'm noticing a lot of passing around of VisionFilterParams - Would it make sense to make a node that publishes this stuff or have an existing node publish it? 

This is just a question, ignore if not a good idea.

src/rj_vision_filter/src/vision_filter.cpp
```diff
@@ -6,21 +6,246 @@
...
+      params_{load_vision_filter_params(this)},
+      world_{params_},
       config_client_{this},
       team_color_queue_{this, referee::topics::kTeamColorTopic},
-      param_provider_{this, kVisionFilterParamModule} {
+      param_setters_{build_param_setters(params_)} {
+    // Keep params_ live-updated whenever a parameter is changed (e.g. via
+    // `ros2 param set`).
+    param_callback_handle_ = add_on_set_parameters_callback(
+        [this](const std::vector<rclcpp::Parameter>& changed_params) {
```
jvogt23, line 234: Again, we may need some validation steps in this callback.

jvogt23 (changes requested):
Seems mostly fine. There's a few nitpicks here, and also some suggestions/questions. I am also wondering about what can/should be done about the amount of times I saw params being passed between objects and into functions. Obviously, the previous architecture encouraged a lot of coupling, and we are trying to remove it pretty much for that reason, so I am wondering if we may need to place some things that are in separate classes/structs together with each other or change the way the params are handled to make all this variable passing less necessary. Not necessarily saying that needs to be done, just wondering about it.

src/rj_joystick/src/manual_control_node.cpp
```diff
@@ -127,4 +181,4 @@ void ManualControlNode::publish(int robot_id, const ControllerCommand& command)
             .dribbler_speed(static_cast<float>(command.dribble_power)));
 }
 
-}  // namespace joystick
+}  // namespace joystick
```
yuvrajdhadwal, line 184: ?

src/rj_vision_filter/include/rj_vision_filter/params.hpp
```diff
@@ -0,0 +1,85 @@
...
+ * parameters once at startup (and kept live-updated) by VisionFilter, and
+ * passed by const reference down to the plain (non-Node) classes that use
+ * them.
+ */
+struct VisionFilterParams {
+    double vision_loop_dt = 1.0 / 60.0;
+    int64_t max_num_cameras = 12;
+    double publish_hz = 60.0;
+
+    struct {
```
yuvrajdhadwal, line 18: why are these C structs ? within a cpp struct?

src/rj_vision_filter/src/camera/camera.cpp
```diff
@@ -302,10 +305,10 @@ void Camera::update_robots_mhkf(RJ::Time calc_time, const std::list<CameraRobot>
     for (const CameraRobot& camera_robot : single_robot_list) {
         bool was_used = used_camera_robot.at(camera_robot_idx);
 
-        if (!was_used &&
-            single_kalman_robot_list.size() < (unsigned long)PARAM_max_num_kalman_robots) {
+        if (!was_used && single_kalman_robot_list.size() <
+                             (unsigned long)params_.camera.max_num_kalman_robots) {
```
yuvrajdhadwal, line 309: c styled casting is unsafe

yuvrajdhadwal (changes requested):
this was a slog

